Issue #6, opened 2016-04-25
Docker Image


thedude459:
Is there a docker file for this?

mstenta:
Nope - but there can be... want to make one?

thedude459:
Never have made one before but I will try. If I can get a working one I will let you know

mstenta:
I'm going to close this because there hasn't been activity. Feel free to reopen if someone wants to work on it!

mstenta:
Reopening this... @rjsteinert got the ball rolling: https://github.com/farmOS/farmOS/pull/8

It would be great to include this in the 1.0 release. Let's use this issue to coordinate what still needs to be done...

mstenta:
Things to figure out:
- Setting up a MySQL container for the database.
- How to manage file storage?
- Document setup procedure.

thedude459:
This is the database docker that I used https://hub.docker.com/r/linuxserver/mariadb/

This is the apache docker that I used https://hub.docker.com/r/linuxserver/apache/

I just dropped the farmOS into the apache docker and pointed it to the mariadb docker for database.

mstenta:
Thanks @thedude459 - how is it working for you?

PS: If you're using farmOS actively, please consider adding your farm to the list on farmOS.org (http://farmos.org/community/farms)!

thedude459:
I only have a small garden right now but I hope to have a farm in the next 2-3 years.  I just wanted to check this out for when I get the farm.  Do you have a user manual for this yet?

mstenta:
Awesome! Documentation is available on farmOS.org (always more to do, of course) - I plan to add some screen captures this week

thedude459:
Thanks, I will keep you updated and I will log anything that I find.

mstenta:
I merged the pull request that adds a Dockerfile. Let's turn attention to writing some documentation for how to get up and running. Here is the open issue for that: https://github.com/farmOS/farmOS.org/issues/14

PR #8, opened 2016-08-05
Add a Dockerfile

rjcorwin:
So far this seems to be working for getting going with building a Docker image of farmos. Need to test linking in a mysql database to make sure though, not to mention documentation on how to do that.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,24 @@
+FROM drupal:7 
+
+ENV DEBIAN_FRONTEND noninteractive
+RUN echo 'debconf debconf/frontend select Noninteractive' | debconf-set-selections
+
+RUN apt-get update && \
+    apt-get install -y git unzip 
+
+RUN php -r "readfile('http://files.drush.org/drush.phar');" > drush && \
+    chmod +x drush && \
+    mv drush /usr/local/bin
+
+ADD build-farm.make /farmos/build-farm.make
+ADD drupal-org-core.make /farmos/drupal-org-core.make  
+ADD drupal-org.make /farmos/drupal-org.make  
+ADD farm.info /farmos/farm.info  
+ADD farm.install /farmos/farm.install  
+ADD farm.profile  /farmos/farm.profile
+
+WORKDIR /farmos
+RUN cd /farmos && drush make build-farm.make farm 
+RUN rm -rf /var/www/html && \
+    ln -s /farmos/farm /var/www/html && \
+    chown -R www-data:www-data /farmos/farm/sites


mstenta:
Awesome!!!!!!! This will be great for development purposes. And it might also make it easier for folks to host their own farmOS instances.

I'm testing this out with the built-in SQLite database option (so that a separate MySQL container isn't necessary). But I'm running into some issues, which might be SQLite's fault... (I've never used Drupal with SQLite before, so I'm not familiar with any hurdles there might be).

Here are the steps I went through to get up and running:
1. Install Docker: https://docs.docker.com/engine/installation/
2. Clone the farmOS repo, and checkout this pull request branch. Open the directory in a terminal.
3. Build the farmOS image: <code>sudo docker build -t farmos .</code>
4. Deploy a new farmOS container: <code>sudo docker run --name my_farmos_instance -i -t farmos</code>
5. Find the IP address of the container: <code>sudo docker inspect my_farmos_instance | grep "IPAddress"</code>
6. Open the IP address up in a browser and go through the farmOS installation - select SQLite during the "Database configuration" phase.

Something went wrong with Apache the first time I tried it, but then I removed the container and built a new one and it worked without any issues.  It happened on step 4... it seems that Apache crashed, leaving me with a non-running container and the following message:

<code>
AH00558: apache2: Could not reliably determine the server's fully qualified domain name, using 172.17.0.2. Set the 'ServerName' directive globally to suppress this message
AH00558: apache2: Could not reliably determine the server's fully qualified domain name, using 172.17.0.2. Set the 'ServerName' directive globally to suppress this message
[Fri Aug 05 13:49:59.135599 2016] [mpm_prefork:notice] [pid 1] AH00163: Apache/2.4.10 (Debian) PHP/7.0.9 configured -- resuming normal operations
[Fri Aug 05 13:49:59.135646 2016] [core:notice] [pid 1] AH00094: Command line: 'apache2 -D FOREGROUND'
[Fri Aug 05 13:49:59.169904 2016] [mpm_prefork:notice] [pid 1] AH00170: caught SIGWINCH, shutting down gracefully
</code>

However, starting the container up again seemed to get it running:

<code>sudo docker start my_farmos_instance</code>

I tried removing the container and starting fresh, and I did NOT run into the same Apache issue. So it may have been a fluke.

During the Drupal/farmOS installation, however, it gets halted by the following error:

<code>
An AJAX HTTP error occurred. HTTP Result Code: 500 Debugging information follows. Path: http://172.17.0.2/install.php?profile=farm&locale=en&id=1&op=do StatusText: Service unavailable (with message) ResponseText: PDOException: SQLSTATE[HY000]: General error: 1 no such table: farm_asset_type: SELECT base.id AS id, base.type AS type, base.label AS label, base.status AS status, base.module AS module FROM {farm_asset_type} base; Array ( ) in EntityAPIController->query() (line 187 of /farmos/farm/profiles/farm/modules/contrib/entity/includes/entity.controller.inc).
</code>

Clicking through to the error page shows a few more errors:

<code>
Deprecated function: Methods with the same name as their class will not be constructors in a future version of PHP; ctools_context has a deprecated constructor in require_once() (line 127 of /farmos/farm/profiles/farm/modules/contrib/ctools/ctools.module).
Deprecated function: Methods with the same name as their class will not be constructors in a future version of PHP; ctools_context_required has a deprecated constructor in require_once() (line 127 of /farmos/farm/profiles/farm/modules/contrib/ctools/ctools.module).
Deprecated function: Methods with the same name as their class will not be constructors in a future version of PHP; ctools_context_optional has a deprecated constructor in require_once() (line 127 of /farmos/farm/profiles/farm/modules/contrib/ctools/ctools.module).
</code>

But it also shows the following:

<code>SQLSTATE[HY000]: General error: 1 no such table: farm_asset_type</code>

So I think it's probably the SQLite database.  I've never seen this issues with MySQL before.

I'll have to learn how to set up a separate MySQL container...

But in general this is great! A very exciting step! Thanks @rjsteinert !

mstenta:
There was a feature request for this posted previously on Github: https://github.com/farmOS/farmOS/issues/6

Let's use that to discuss next steps and try to get this included in the 1.0 release of farmOS.

rjcorwin:
@mstenta I had the same error when trying sqlite. I actually tried to spin up mysql also using the Dockerfile with help from @chrowe and @dogi but couldn't get the recipe quite right. I'm inclined to go with jus "linking" a mysql container into the farmOS container. This allows us to lean on the folks working on the MySQL Docker image to deal with the complexity of keeping mysql in a Docker container and just have us concerned with using the API they define for spinning up MySQL containers. Links is the Docker way of mapping a port in one container into another container. I did this a ways back with other Drupal sites (farm hack) but I need to document how to do this. Docker docs on links [here](https://docs.docker.com/engine/userguide/networking/default_network/dockerlinks/) yet that doc is now called "Legacy container links" so I wonder what the "modern" way of doing this is.

mstenta:
@rjsteinert Sounds good to me! I did a little reading on linked containers myself, and that seems like the right approach.

I added a new issue to the farmOS.org repo for documenting how to get set up: https://github.com/farmOS/farmOS.org/issues/14

I think we can go ahead and merge this pull request and then work on the documentation as the next step.

rjcorwin:
+1

PR #10, opened 2016-08-11
Add docker-compose.yml and a gitignore for ignoring the data folder

rjcorwin:
This adds a `docker-compose.yml` file which allows us to get the farmOS container up and running with MySQL using Docker Compose (install instructions for Docker Compose [here](https://docs.docker.com/compose/install/)).

```
git clone git@github.com:farmos/farmos
cd farmos
docker-compose up -d
```

When you visit your server in the browser, you will be prompted to install and enter database information. Select mysql and set database name, username, and password to `farmos`.  Under  Andvanced Options, set Database host to `db`. 

This PR still needs some work though. Currently it will pull the Docker image for farm os from https://hub.docker.com/r/rjsteinert/farmos/ when ideally it will point to yet to be created https://hub.docker.com/r/rjsteinert/farmos/. @mstenta I created a farmos org on Docker Hub, go ahead and create an account for yourself and I'll add you as an owner. Then you will be able to set up an automated build for the `farmos/farmos` image because you have ownership over the github `farmos/farmos` repository.  At that point, we modify the `docker-compose.yml` file to point to that new image on Docker Hub.

PR #11, opened 2016-08-16
Update Dockerfile

thedude459:
The docker file currently there was not working for me.  Please test this one.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,15 +10,9 @@ RUN php -r "readfile('http://files.drush.org/drush.phar');" > drush && \
     chmod +x drush && \
     mv drush /usr/local/bin
 
-ADD build-farm.make /farmos/build-farm.make
-ADD drupal-org-core.make /farmos/drupal-org-core.make  
-ADD drupal-org.make /farmos/drupal-org.make  
-ADD farm.info /farmos/farm.info  
-ADD farm.install /farmos/farm.install  
-ADD farm.profile  /farmos/farm.profile
+RUN git clone https://github.com/farmOS/farmOS.git /farmOS 
 
-WORKDIR /farmos
-RUN cd /farmos && drush make build-farm.make farm 
+RUN cd /farmOS && drush make build-farm.make farm 
 RUN rm -rf /var/www/html && \
-    ln -s /farmos/farm /var/www/html && \
-    chown -R www-data:www-data /farmos/farm/sites
+    ln -s /farmOS/farm /var/www/html && \
+    chown -R www-data:www-data /farmOS/farm/sites


mstenta:
Thanks @thedude459 - could you paste the errors you saw?  I wonder what their cause was...

The Dockerfile worked for me on Ubuntu.

I'm just starting to learn Docker, but one thing I did find is that COPY seems to be recommended over ADD in most cases: https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/

I'm not sure if that would make any difference in your case... could you test it to see?

In general, I do like the brevity of your suggested changes. I also like the OS in farmOS capitalized, where possible.

I'm not sure I like using "git clone", however. It seems redundant since the Dockerfile is in the repository, which implies you've already cloned the repository.

It seems like we should try to get to the bottom of the errors you're seeing with the ADD commands... and maybe try with COPY instead?

With regards to brevity, is it possible to do this?

<code>ADD . /farmOS</code>

Does that add everything in the repo to /farmOS in the container? Still pretty new to Docker...

mstenta:
I just tested with the following in place of the "git clone" line, and it seemed to work:

<code>COPY . /farmOS</code>

@thedude459 : Can you give that a try and see if it works for you?  If not, please paste the errors you see.  On a side note, would you mind updating your pull request branch to split the "farmos" to "farmOS" changes into a separate commit?

thedude459:
So that will work if the person just does a copy of this repo but not if they take the file from the docker hub.  There needs to be something in the docker file that gets the build files from this repo.

thedude459:
I was doing the git clone in case anything in the repo changed in the future but I supposed if anything changed we could just re-build the docker image.

mstenta:
Ah ok - so just so that I understand: does Docker Hub store Dockerfiles? Or just container images? If it does store Dockerfiles, then I see why it might make sense to include a "git clone". But if it just stores images, and the Dockerfile is always in the Github repo, then it sounds like rebuilding the image in Docker Hub is the best approach. I think that's what @rjsteinert is working on setting up in #10 (setting up a farmOS organization on Docker Hub and automating the image build process)... if I understand it all correctly. :-)

I would still be interested in pulling in some of the other changes you made in this pull request, like replacing "farmos" with "farmOS". And if it makes sense to replace all the ADD lines with the single COPY line I described in my comment above: https://github.com/farmOS/farmOS/pull/11#issuecomment-240303960

Do you want to tackle those changes? Maybe in a new pull request? Or feel free to reopen this one too. Ideally they would be in two separate commits, to keep them separated conceptually in the Git history.

thedude459:
I am working on another pull request now.

rjsteinert:
Hi guys - There are a lot of ways to do Docker, there is however a pattern that the official Docker images use which is the same pattern for all the Docker projects I've been working on. I'm suggesting we go with an "automated build repository" on Docker Hub. On change to a Github repository being watched, Hub clones the repository (Dockerfile included) and does a `docker build`, then stores the resulting image on Docker Hub. Images built on Docker Hub are considered trusted over ones that are built on a person's laptop and then pushed to Docker Hub.

rjcorwin:
@thedude459 This is not a big deal but adding or copying the entire directory will add the `.git` folder which will likely trip up the Docker Cache every time you build.  Read up on Docker Cache [here](http://thenewstack.io/understanding-the-docker-cache-for-faster-builds/). Also, the use of capital letters in system level name spacing is [against linux convention](http://unix.stackexchange.com/questions/239295/is-it-considered-a-best-practice-to-not-use-capital-letters-in-file-naming). Again, not a big deal but it hurts old sysadmins like @dogi.

thedude459:
@rjsteinert can't we just add a .dockerignore file to the repo?

mstenta:
Makes sense about copying the .git folder... didn't think of that. Sounds somewhat related to the thoughts around using Docker for local development too: https://github.com/farmOS/farmOS.org/issues/14

I am "pro" capitalization (in this context anyway). The stackexchange thread says that spaces are a no-no, but the consensus seems to be that uppercase is fine, and is used already in lots of places. I would prefer to keep the "OS" capitalized - if only for aesthetic reasons... and consistency.

PR #14, opened 2016-08-20
Add dockerignore file

thedude459:
This is fix the issue with the docker cache that @rjsteinert was talking about.


mstenta:
Hi @thedude459 - can you help me understand what the dockerignore file does? I'm still trying to wrap my head around it.

I think @rjsteinert's comment (https://github.com/farmOS/farmOS/pull/11#issuecomment-240909250) was in reference to using ADD to add the entire farmOS repo contents to the container. But I'm not sure if dockerignore is used in that context? Can't seem to find a definitive answer about that...

On a related note... I realized that we ONLY need to ADD build-farm.make to the container, since we are running Drush Make using that file alone.  The Drush Make script will clone the full farmOS repo itself, so there's no need for us to ADD the other files. I will push a new commit to the docker branch with that change...

thedude459:
https://docs.docker.com/engine/reference/builder/#/dockerignore-file

mstenta:
Ha! It was the second sentence! :-)

> This helps to avoid unnecessarily sending large or sensitive files and directories to the daemon and potentially adding them to images using ADD or COPY.

I guess I skimmed too quickly and missed that.

I just pushed a commit that removes all but two of the ADD commands: https://github.com/farmOS/farmOS/commit/36460dd14bf00a834c29845cc9d2805048eb7abd

We only need those two files in order to build farmOS with drush make.

Since we are not ADDing the whole repo, does that mean we don't need a .dockerignore?

PR #22, opened 2016-10-31
Set the container names to DNS safe values.

kadaan:
Change `network_mode` to `bridge` to support running on Os X under dlite: https://github.com/nlf/dlite
NOTE: To run farmOS under dlite:

1. Download the latest release from https://github.com/nlf/dlite/releases
2. Copy `dlite` to `/usr/local/bin/dlite`
3. Run `chmod +x /usr/local/bin/dlite`
4. Initialize dlite by running: `dlite init`
5. Start dlite host vm by running `dlite start`
6. Start farmOS with docker-compose
6. Profit!

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,6 +2,8 @@ version: '2'
 services:
   db:
     image: mysql:latest
+    container_name: farmos-db
+    network_mode: bridge
     volumes:
       - './.data/db:/var/lib/mysql'
     restart: always
@@ -15,6 +17,8 @@ services:
     depends_on:
       - db
     image: farmos/farmos:7.x-1.x
+    container_name: farmos
+    network_mode: bridge
     volumes:
       - './.data/www:/var/www/html'
     links:


mstenta:
Thanks @kadaan ! It will be great to get this working better on OS X.

Unfortunately, I'm not sure we can add the container_name parameter...

One of the things that I have been doing is building multiple local development sites using Docker Compose. I do this by cloning the repository into folders with different names. For example: `farmOS1`, `farmOS2`, etc.

When `docker-compose up` is run, it generates containers with names based on the parent folder. For example: `farmos1_www_1`, `farmos2_www_`1, etc.

But if `container_name` is defined in docker-compose.yml, then that doesn't work anymore. I get the following error when I try it:

```
Recreating farmostest2_db_1

ERROR: for db  Cannot create container for service db: Conflict. The name "/farmos-db" is already in use by container ce0f223c3080a13074356703f5b0d662225874228d9bc728c0f7a5936a238a7d. You have to remove (or rename) that container to be able to reuse that name.
ERROR: Encountered errors while bringing up the project.
```

I think the same is true of the `bridge` parameter. It seems that Docker creates a single network called "bridge". Whereas without it, it creates a network called "farmos_default" (using the name of the parent folder followed by `_default`).

I wonder if there is another way to solve the issues on OS X without sacrificing the ability to run multiple local dev sites?

Do you have a sense of exactly what problems the `container_name` and `network` parameters were fixing? Maybe we can find another fix.

kadaan:
The reason for setting the `container_name` is that `_` is not a safe DNS character.  That said, it works in most browsers.  Setting the `network_mode` was working around [#208](https://github.com/nlf/dlite/issues/208) in [dlite](https://github.com/nlf/dlite).  It think for now we can alter the Os X  steps to:
1.  Download the latest release from https://github.com/nlf/dlite/releases
2.  Copy dlite to /usr/local/bin/dlite
3.  Run: `chmod +x /usr/local/bin/dlite`
4.  Initialize dlite by running: `dlite init`
5.  Start dlite host vm by running: `dlite start`
6.  Start farmOS by running:  `docker-compose up`
7.  Fix networking by running `docker network connect bridge farmos_www_1`
8.  Open Safari and navigate to: http://farmos_www_1.docker
9.  Profit!
10. Stop farmOS: `docker-compose stop`
11.  Remove networking: `docker network disconnect bridge farmos_www_1`
12. Start farmOS: `docker-compose up`
13. Add networking: `docker network connect bridge farmos_www_1`
14. Open Safari and navigate to: http://farmos_www_1.docker
15. Profit!

mstenta:
OK cool. Should we add those steps to https://github.com/farmOS/farmOS.org under a new section for OS X users?

Still not sure I understand this:

> _ is not a safe DNS character. That said, it works in most browsers.

Is there an underscore somewhere that is causing a problem? I'm not sure what you mean by `it works in most browsers`.

Is Docker setting up DNS of some kind?  If so, I wasn't aware of that - could you fill me in?

PR #38, opened 2017-08-17
added LICENSE file to repo base path

cavebeat:
makes it clear and directly visible in GitHub
GPL according to the homepage and drupal page.

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,0 +1,339 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 2, June 1991
+
+ Copyright (C) 1989, 1991 Free Software Foundation, Inc.,
+ 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+                            Preamble
+
+  The licenses for most software are designed to take away your
+freedom to share and change it.  By contrast, the GNU General Public
+License is intended to guarantee your freedom to share and change free
+software--to make sure the software is free for all its users.  This
+General Public License applies to most of the Free Software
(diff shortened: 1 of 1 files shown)


mstenta:
Thanks @cavebeat ! I think this is a good idea.

Just two small changes are needed to your pull request...

1) We need to stick with GPL version 2+ for the time being. I'm in favor of version 3 personally, but technically all code automatically packaged on drupal.org is packaged with the GPL v2+ license. The Drupal Licensing Working Group has an issue open to enable distro maintainers to use GPL v3, but it is still a work in progress: https://www.drupal.org/node/1449452

2) We should call the file LICENSE.txt instead of just LICENSE. I know GitHub recommends LICENSE or LICENSE.md, but since we are using drupal.org for packaging, we need to use their convention - otherwise the packaged code on drupal.org will include BOTH a LICENSE and LICENSE.txt (because the packaging script automatically adds it (and it will overwrite any existing LICENSE.txt file in the repo: https://www.drupal.org/node/1663716)).

So for now, the easiest thing to do would be to download the packaged tarball from http://drupal.org/project/farm, copy and paste the LICENSE.txt file that is included with it, and then commit that to the farmOS repo.  Then we can be sure that it will be exactly the same.

It's annoying that we can't use GPL v3 yet, but I think it will be possible in the future.  It basically comes down to the fact that we're still relying on drupal.org's packaging.  This too might change in the future - especially with the growing use of Composer in the Drupal world - but for now we're on GPL v2+.

Would you mind updating your pull request to use the LICENSE.txt file that drupal.org packages?

Appreciate the contributions! Thanks again! :-)

cavebeat:
Done 
the pull request contains now the GPLv2
copied LICENSE.txt from the Drupal release 7.x-1.0-beta14

mstenta:
Thanks @cavebeat ! Merged.

PR #127, opened 2019-01-08
Planting quick form asset name override

paul121:
[Drupal issue 3023723]( https://www.drupal.org/project/farm/issues/3023723)

Creates a field for the planting asset name, autofills the default_value as the season and crops are input to the form. The user can change the value before submitting.

### DIFF
--- a/modules/farm/farm_crop/farm_crop.farm_quick.planting.inc
+++ b/modules/farm/farm_crop/farm_crop.farm_quick.planting.inc
@@ -29,6 +29,10 @@ function farm_crop_planting_quick_form($form, &$form_state) {
     '#description' => t('What season will this planting be part of? This is used for organizing plantings for future reference, and can be something like "@year" or "@year Summer". This will be prepended to the planting asset name.', array('@year' => date('Y'))),
     '#autocomplete_path' => 'taxonomy/autocomplete/field_farm_season',
     '#default_value' => $season,
+    '#ajax' => array(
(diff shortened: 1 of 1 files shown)


modules/farm/farm_crop/farm_crop.farm_quick.planting.inc
```diff
@@ -237,13 +237,17 @@ function farm_crop_planting_quick_form($form, &$form_state) {
     implode(', ', $crop_tags),
   );
 
+  // Cannot change the default value without unsetting form_state input
```
paul121, line 240: This solution was found [here](https://drupal.stackexchange.com/questions/251056/set-default-form-values-after-ajax-callback)

mstenta:
Thanks @paul121 - this looks great! Well done!

I merged this in and will close https://www.drupal.org/project/farm/issues/3023723.

I also made two small follow-up commits:

https://github.com/farmOS/farmOS/commit/f6f3c73ade826d02aaf1c0094b131826a23c4e8a

https://github.com/farmOS/farmOS/commit/33937bc2f748118493aab24b17a0ec2e44eb97ed

Thanks again! :-D

modules/farm/farm_crop/farm_crop.farm_quick.planting.inc
mstenta: Good find!  And yea I've had to use the same "solution" myself in the soil amendment and soil disturbance quick forms. Drupal 7 #ajax has some quirks to it, for sure.

PR #128, opened 2019-01-08
Autofill location into the asset name

paul121:
Drupal issue [3023724](https://www.drupal.org/project/farm/issues/3023724)

Uses the transplanting location, then defaults to the seeding location.
Also adds a description to the name field.


mstenta:
Great! Thanks @paul121 !

Issue #136, opened 2019-02-12
This is unecessary in docker entrypoint


sokoow:
This is actually an antipattern:

https://github.com/farmOS/farmOS/blob/ffef8543de8ae9e6b670be2e1cd277f94400992b/docker-entrypoint.sh#L104-L116

you can't be reinstalling the whole product and modules every time the pod/container starts :) It takes too much time, overwrites whole volume on /var/www/html and so on. Make this a part of Dockerfile build please.

mstenta:
Thanks for making an issue for this, @sokoow - it's actually something I've been giving a lot of thought to myself recently.

I am planning on changing the way this works in the 2.x version of farmOS. I was thinking I would leave it as-is for the 1.x branch, just so we don't disrupt anyone currently using it - but we can consider it if it needs to be done in 1.x too. I would just want to think through how it might affect existing deployments first.

For historical context, there is some reasoning behind the decision to build the farmOS codebase in `docker-entrypoint.sh` rather than in the `Dockerfile` itself.

The simplified reason is that the farmOS Dockerfile was originally designed for local development purposes, and in order to use standard development tooling (eg: PHPStorm IDE) you need to have the whole `/var/www/html` available in the local filesystem (mounted as a volume into the container). As you know, mounting a volume into a Docker container REPLACES that directory inside the container - so if we build farmOS in `/var/www/html` via the Dockerfile, and then mount `/var/www/html` as a volume in the container, that directory is then empty when the container starts. So I decided to move the build process to the `docker-entrypoint.sh` script as a compromise.

I admit that when I did this I was still learning Docker best practices - and in hindsight I think it's a bad solution for a number of reasons. So I am eager to change this.

This issue might also provide some good thoughts and context for you (and others): https://github.com/farmOS/farmOS/issues/36

I've been giving thought to redesigning this in 2.x - and perhaps having two different approaches to farmOS+Docker: one for production hosting (which builds the codebase in the image), and one for development (which uses the same image, but mounts over the directory and has a second step to build farmOS's codebase in the volume).

I have some ideas already, but they aren't documented anywhere. Maybe we can use this issue to sketch them out, and then decide if it makes sense to apply them to the 1.x branch as well - or just leave that alone so it doesn't risk breaking anyone's active deployments.

sokoow:
[...]

Issue #132, opened 2019-01-29
trying to get this working on kubernetes


sokoow:
Hey All,

Trying to get farmos working on kubernetes, here's some minimal manifests:

```
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  name: farmos
spec:
  replicas: 1
  template:
    metadata:
      labels:
        app: farmos
    spec:
      containers:
      - name: farmos
        image: farmos/farmos:7.x-1.x
        imagePullPolicy: Always
        ports:
        - containerPort: 80
        env:
          - name: "POSTGRES_USER"
            value: "farmos"
          - name: "POSTGRES_PASSWORD"
            value: "farmos"
          - name: "POSTGRES_DB"
            value: "farmos"
          - name: "FARMOS_DEV"
            value: "true"
---
apiVersion: v1
kind: Service
metadata:
  name: farmos-svc
spec:
  selector:
    app: farmos
  ports:
  - name: web
    port: 80
---
---
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: farmos
  annotations:
   kubernetes.io/ingress.class: traefik
spec:
  rules:
  - host: farm.XXX.net
    http:
      paths:
      - path: /
        backend:
          serviceName: farmos-svc
          servicePort: 80
```

Here's some challenges I've come across, hopefully you can help here:

1. The docker container seems to be reinstalling all modules again and again - is it possible to build a docker image which is a step after that (so has modules pulled in and is starting fast) ?
2. What would be a required persistence volume for drupal configs ? There probably is a file with db settings and so on, where is it ?
3. How would one set proper base url and where ? I'm getting lots of mixed content warnings when on tls... 

Sorry if these questions are silly - nice product and keep it going! :D Greetings

mstenta:
Hi @sokoow thanks for starting this!

I don't have time to go into a ton of detail right now, but I'll try to point you to some things that might help inform your exploration.

First, take a look at the `docker-compose.yml` file that comes with farmOS - it is used to quickly spin up a farmOS instance with the necessary data persistence volumes: https://github.com/farmOS/farmOS/blob/7.x-1.x/docker-compose.yml

Also, read the Docker page on farmOS.org to understand the recommended use: https://farmos.org/development/docker/

Also highly recommend reading this issue - it is rich in information about how the docker volume and codebase build process work, and why: https://github.com/farmOS/farmOS/issues/36

[...]

Issue #36, opened 2017-05-25
Change Docker volume mount from /var/www/html to /var/www/html/sites (?)


mstenta:
**Question:** I'm going to put this up for consideration (and to help myself think it through by writing it out): should we change the default Docker image behavior to mount `/var/www/html/sites` as a volume, instead of the entire `/var/www/html` directory?

**Background:** Currently, the Dockerfile mounts the entire `/var/www/html` directory as a volume, which means that the farmOS codebase itself needs to be built AFTER the container starts (we do this via `docker-entrypoint.sh`). This is because when a directory is volume mounted into a Docker container from the host, it replaces the directory in the container, and therefore starts out as empty. So if we build the farmOS codebase in `/var/www/html` within the container, and then mounted that directory as a volume, it would be empty and the build would have been pointless.

However, this also means that the farmOS codebase cannot be updated simply by pulling a new version of the Docker image. It also needs to be rebuilt in the volume. We have some clever logic (aka extra complexity) in `docker-entrypoint.sh` that takes care of this automatically by detecting the installed farmOS version using Drush, and automatically rebuilding if the version changes.

Additionally, the `docker-entrypoint.sh` has two pathways to build the farmOS codebase: 1) download a packaged release for a specific version, or 2) build the latest development version with Drush. The approach it takes is toggled with the `FARMOS_DEV` boolean environment variable.

I don't love all of the complexity required for these things, but they got the job done for an initial Docker integration (and yes, I learned a lot about Docker through the process). For more historical perspective on these details, see the original issue thread that added Docker integration for farmOS: https://github.com/farmOS/farmOS/issues/15

**Proposal:** So now I am thinking about ways to simplify. If we did build the farmOS directly in the container, and mounted `/var/www/html/sites` as a volume instead of `/var/www/html`, then we could remove a lot of the complex logic in `docker-entrypoint.sh`.

This alone would not solve the tagged release vs development version use-case, however. For that, maybe it would make sense to offer two different Dockerfiles (and two different docker-compose.yml files).

Something to think about...

mstenta:
Of course, if the volume mount point does change, it will affect anyone who is running farmOS with Docker currently. So care needs to be taken to ensure that it does not break anything.

mstenta:
[...]

PR #137, opened 2019-02-13
Modified docker build process to be more containers friendly

sokoow:
This is not rebuilding every time container starts anymore. TODO: still have issues understanding why modules install on new installs, maybe related to database setup


mstenta:
For anyone looking at this for the first time, please read https://github.com/farmOS/farmOS/issues/36 in full to understand the context and considerations.

mstenta:
Also read @sokoow's other two recent issues that led to this pull request:

https://github.com/farmOS/farmOS/issues/132

https://github.com/farmOS/farmOS/issues/136

mstenta:
As I said in https://github.com/farmOS/farmOS/issues/36 and https://github.com/farmOS/farmOS/issues/136, I am interested in taking an approach like this in the 8.x-2.x branch of farmOS, but I'm still not sure about merging it into 7.x-1.x - because it might be considered a BC breaking change (has the potential to break/disrupt existing deployments - but we should think this through in more detail to be sure). Let's discuss all the considerations here and decide how to proceed.

mstenta:
Here is an outline of the main considerations (mainly copied from https://github.com/farmOS/farmOS/issues/36):

**Consideration 1: Backwards compatibility** - If the volume mount point does change, it will affect anyone who is running farmOS with Docker currently. See related comments here: https://github.com/farmOS/farmOS/issues/36#issuecomment-304022146 and https://github.com/farmOS/farmOS/issues/36#issuecomment-304023553

**Consideration 2: new installations** - Drupal comes with a `default.settings.php` file in `sites/default` that it automatically copies to `sites/default/setting.php` during the installation process (AFTER the container and codebase are built and running). If we build the codebase directly inside the image, and then mount `sites/default` as a volume, `default.settings.php` will not exist. This makes initial installation more difficult because it requires an additional step to create `settings.php` (this pull request is not handling this currently). See related comment here: https://github.com/farmOS/farmOS/issues/36#issuecomment-304024561

**Consideration 3: sites vs sites/default** - We may want to mount `/var/www/html/sites` into the container, rather than `/var/www/html/sites/default`. The `sites/all` directory is often used to put extra modules, themes, and libraries. This can also be done in `sites/default` though - so maybe defaulting to a `sites/default` volume is enough, and we can leave it to downstream users to decide if they want to mount `sites` instead, for more flexibility. (This is what https://github.com/farmOS/farmOS/issues/36 in general proposed.)

**Consideration 4: Drush, Drupal Console, Composer, etc** - If the `/var/www/html` directory is not mounted in a volume as a whole, then we no longer have the ability to mount it into other containers as well. Some examples where doing this is useful include command line build tools like Drush, Drupal Console, and PHP's Composer. If we want to use those tools with the codebase, they would also need to be installed directly inside the image. As it is now, we don't need to install those tools in the image, and they can be used separately via their own Docker images (eg: https://hub.docker.com/r/drush/drush, https://hub.docker.com/_/composer). Maybe this is a moot point, though - since it may make sense to include those in the container anyways (especially in farmOS 8.x-2.x). But are there other examples where we might need to be able to mount the whole `/var/www/html` directory into another container? Or should we just leave that up to the downstream user to configure their volumes differently than our repository default?

**Consideration 5: IDE development** - Similar to consideration 4 above, if the `/var/www/html` directory is not mounted in a volume as a whole, then it's not possible to use an IDE like PHPStorm for development of the codebase, because the IDE requires all of that code be accessible to it on the host (outside of the container).

**Consideration 6: production vs development** - The current way we have the Dockerfile, docker-compose.yml, and docker-entrypoint.sh scripts set up allows us to have two different approaches to deployment: one for production hosting using a tagged release, and one that checks out the git repo directly for development purposes. This pull request does look like it handles both of these still (similar to how it worked previously), but I'd like to give some thought to simplifying even further. Perhaps having two different Dockerfiles and/or docker-compose.yml files would be worth considering.

jasperboyd (approved):

mstenta:
I've merged these changes in (https://github.com/farmOS/farmOS/commit/9234094ecb3960b529c9a36486eef140b8b11bcc), as well as a number of other changes (https://github.com/farmOS/farmOS/commit/38196e698e0cb3db62550e468423db56ee6115a3), to meet the goals of this pull request, and to address all the considerations I've outlined above.

The documentation on farmOS.org has been updated accordingly: https://farmos.org/development/docker/

Thanks for getting this started @sokoow !

PR #164, opened 2019-04-24
First pass at a loraserver module

proffalken:


proffalken:
Closed because I don't want to spoil the main codebase - I've got https://github.com/mockingbirdconsulting/farmOS/pull/1 open on my fork instead.

mstenta:
@proffalken Thanks for starting this!

I agree - this probably makes more sense as an add-on module. I've actually thought about splitting the Sensor module itself out of farmOS core... might still do that when we move to Drupal 8, but for now it will stay.

Let me know when you have something working, and we can add a link to https://farmos.org/development/projects/ and/or https://farmos.org/guide/contrib/

mstenta:
If you want to see what a super-simple add-on module looks like, here is an example that adds an "Atmospi" sensor type: https://github.com/mstenta/farm_sensor_atmospi

(For use with https://github.com/mstenta/atmospi - note that I haven't used the farm_sensor_atmospi module for a while, so it may be broken.)

PR #184, opened 2019-06-19
Gitignore docker-compose.yml

viktorsmari:
When using docker on a server or locally when developing, this file always pops up in `git status`

Can we ignore it, until maybe someone decides it should be in the root (and not in the /docker folder)?

I was also thinking if we should gitignore the `db/` and `www/`? But that can be a new PR

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,4 @@ libraries
 modules/contrib
 modules/dev
 themes/bootstrap
-
+docker-compose.yml


mstenta:
Thanks for opening a pull request @viktorsmari !

However, I don't think this is an issue anymore, if you follow the new instructions here: https://farmos.org/development/docker/#create-containers

Although, maybe we should add a little more to https://farmos.org/hosting/docker/ to make that clear as well.

You don't actually need to check out the farmOS Github repository to run farmOS in Docker. You can simply copy one of the example `docker-compose.yml` files from the repo and modify it for your needs. But it can stand alone in a directory by itself, and that's the recommended way of deploying it now. The `db` and `www` directories will then be in that same directory, and none of it will be in source control (unless you want it to, for your own version control - but then it's a different repo, not farmOS).

Hope that all makes sense! If you think https://farmos.org/hosting/docker/ could benefit from some clarification, and are interested in creating a pull request for that, please feel free! The farmOS.org repo is here: https://github.com/farmOS/farmOS.org

PR #188, opened 2019-07-18
Fixed problems when BC math is missing

tangix:
Changed this manually in my 1.1 installation, but thought it was time to contribute the fix.
Not running BC math on my php 7.2 installation.

### DIFF
--- a/modules/farm/farm_area/farm_area.module
+++ b/modules/farm/farm_area/farm_area.module
@@ -475,7 +475,7 @@ function farm_area_calculate_area_multiple($area_ids, $format = FALSE) {
       if (empty($total_area)) {
         $total_area = 0;
       }
-      $total_area += $area_area;
+      $total_area += floatval($area_area);
     }
   }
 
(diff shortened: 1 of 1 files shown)


mstenta:
Merged! Thanks @tangix !

mstenta:
@tangix We've had a bug report that seems to be related to this change: https://github.com/farmOS/farmOS/issues/192

Have you come across that warning?  Is that the warning that was supposed to be fixed by this change? Or something new?

What version of PHP are you running?

tangix:
The pull request addressed the issues I saw on my installation with my configuration of areas. I think I remember seeing other issues with some areas that I fixed (hacked) in my own install of 1.1. When 1.2 was released I created the pull request.

I'm running 7.2.18 without bc math on Debian 9.9. When I tried installing bcmath I completely messed up the php installation on the system and haven't retried this since then.

mstenta:
Thanks @tangix - the other issue went away, so I think it's all good!

Issue #37, opened 2017-07-14
FarmOS Offline Mode


DomtronVox:
Hello! First, thank you for creating and maintaining this project.

I was wondering if it would be possible to have an "offline mode" for FarmOS. That is, allow FarmOS to work without an internet connection.

For example my family lives out where there is no real internet connection to speak of. However, I have a local network that covers the house, which I plan to extend out to the main fields as well, with a personal web server. I have FarmOS running and can access it with all local computers and tablets, but each page load is very slow and takes up the precious little bandwidth I do have because it has to fetch external fonts, stylesheets, and scripts.

I would like to use FarmOS to track information, but the requirement for an internet connection makes FarmOS almost useless to me. :(  I'm mostly looking for instructions on how to remove**, or better yet host locally**, all the parts requiring an internet connection. Though it would be really nice if it was an option/step in the installation process. 

Thank You!

mstenta:
Hi @DomtronVox - great question! This is something I've thought a lot about, and we actually discussed it on the last monthly call this past Wednesday.

The main things that depend on an internet connection in farmOS (off the top of my head) are:

1. Google Maps - farmOS uses Google Maps layers for it's default maps. However, it is possible to turn off the maps entirely in farmOS. The setting for that is at: /admin/config/farm/map
2. Openlayers library - By default, farmOS is set up to load the Openlayers javascript library via a CDN. This only happens if the maps are turned on, so the setting described above is also a workaround for this. Alternatively, it is possible to set up the Openlayers Drupal module to load the library from local files, rather than from a CDN.
3. Bootstrap library - The Drupal theme for farmOS (farm_theme) is based on the Bootstrap theme (http://drupal.org/project/bootstrap), and similarly it also loads the Bootstrap library JS and CSS from a CDN by default. It is possible to host those files locally, as well.

So ultimately, right now it is mainly the maps that need an internet connection - for the Google layers. If you turn off the maps, then that covers points 1 and 2 above. If you set up the Bootstrap CSS and JS files locally, then that will cover point 3.

(You also mentioned fonts... any custom font requirements must be coming from the Bootstrap library - so I would assume they would also be included with a local install of the Bootstrap library.)

I wish I could provide more detailed instructions for setting up Openlayers and/or Bootstrap locally, but I haven't done it in a while. I think both projects provide documentation for that, though:

https://www.drupal.org/project/bootstrap

https://www.drupal.org/project/openlayers

[...]

PR #189, opened 2019-07-20
Add farm_report module, simple CSV export of animal weights, simple graph of animal weights.

paul121:
- farm_livestock_weight adds a simple graph of animal weights under the "Weight" tab
- farm_report adds a "Reports" tab to the Dashboard menu
- farm_livestock_weight adds an "Animal Weights" report, exports CSV of animal weights from selected groups


paul121:
The "Reports" tab - 
csv_export_tab

Simple CSV Export report form -
csv_export_form

Sample animal report CSV file -
csv_export_file

mstenta:
FYI I rebased and renamed this branch to `weight_reports` in my fork. Check that branch out locally, then reset your `7.x-1.x` branch to it, then force push to your fork - and that should update this PR automatically so it knows where the new work is.

paul121:
Thanks @mstenta ! I made the two changes we discussed.
- Changed the format of the CSV
- Unset the start date in the form
- I also modified the default date_year_range to be -10 years to +1 years. Is that sufficient?

mstenta:
Great!  Testing it out now - the new columns look good in the CSV! And unsetting the start date works. -10/+1 years sound good to me.

One issue I'm seeing in my test site: I have an animal that has two weight logs. I see both sets of columns in my CSV, but only the first one has data in it. See attached, and screenshot of my animal's weight logs.

[animal_weights.txt](https://github.com/farmOS/farmOS/files/3461969/animal_weights.txt)

Screenshot from 2019-08-02 11-42-56

The only thing I can think that might be weird about it is that the second weight log (July 24th) references TWO assets (Child test 1, Test animal 1). Could that be causing it?

mstenta:
Oh actually... now that I look closer... the first weight in the CSV is wrong too. It should be 150 lbs, but it says 100 lbs.

paul121:
Hmmm. I didn't do any testing for logs with multiple assets. I guess I'm
not fully sure how that works - is a separate log not created for each
asset? Right now I loop through each asset, then collect all weight logs
associated with each asset. That is what populates the dates in CSV
header... Interesting that the 7/24 date is there, but the values aren't.

Also it looks like log #36 was 100 pounds, #37 was 150. I probably have it
taking the the first log recorded on a date, not the last log recorded.
Should I switch that?

mstenta:
> I didn't do any testing for logs with multiple assets. I guess I'm
not fully sure how that works - is a separate log not created for each
asset?

It's possible to create a single weight log that references multiple assets. But if you are just iterating through assets and loading all weight logs for each, then it should pick it up. It seems like it did... at least partly... because it created the column, albeit didn't fill in the values.

> Also it looks like log #36 was 100 pounds, #37 was 150. I probably have it
taking the the first log recorded on a date, not the last log recorded.
Should I switch that?

Yea generally in farmOS we go by the latest log in cases like that... and if two logs have the same timestamp, we go by the one with the higher log ID.

paul121:
OK - I believe these issues are fixed. Both stemmed from how I was "collecting" the dates. I was using a global array of dates instead of one unique to each animal. Attempted duplicates were dropped for animals if another had already claimed a date.

Now, being unique to each animal, only duplicate log dates for the same animal are dropped. And because `farm_quantity_log_data` loads logs in descending order, the most recent log comes first followed by older ones on the same date, which get dropped.

mstenta:
Finally got around to merging this.

https://github.com/farmOS/farmOS/commit/f13327f584735991d8a0dd0b689648c927db7fda

Thanks again @paul121 !

PR #214, opened 2019-11-21
Add configuration UI for administering the farm_access module

### LINKED ISSUE #187: Using farmOS.app with self-hosted farmOS
@jgaehring and I have been working on getting the farmOS Field Kit (https://github.com/farmOS/farmOS-client) working as a PWA (progressive web app) on https://farmOS.app. So far it's working well!

One remaining piece is CORS configuration. Since the app runs in the browser and is served from a different domain than farmOS, the browser prevents cross-domain requests, so the client doesn't work.

I set up Farmier-hosted instances to work with farmOS.app by adding the following headers to page responses:

```
Access-Control-Allow-Origin: https://farmos.app
Access-Control-Allow-Credentials: true
Access-Control-Allow-Headers: Content-Type,Authorization,X-CSRF-Token
Access-Control-Allow-Methods: GET,POST,PUT,DELETE,HEAD,OPTIONS
```

This is currently being added by a custom module, so it isn't included in farmOS itself. This means that farmOS.app will not work if you are self-hosting farmOS.

There is a spectrum of approaches we could take:

1. Document how to install and configure the Drupal CORS module (https://www.drupal.org/project/cors) in your own self-hosted farmOS. This would be the most burdensome on end-users as it would require both filesystem and configuration changes to set up. But it means we aren't adding any additional code or dependencies to farmOS core (a good thing).
2. Include the Drupal CORS module in farmOS, and provide some default configuration that allows it to work automatically with https://farmOS.app. This would mean it works with farmOS.app out of the box, and you could go into the CORS configuration and modify it if you wanted to... for example if you wanted to self-host the client instead of using farmOS.app.
3. Provide our own simplified CORS configuration module in farmOS that is tailored for farmOS's API, and defaults to working with https://farmOS.app. This would mean farmOS.app works out of the box, and we could also add the ability for the user to add additional domains, while keeping all the other access control headers fixed. This would provide the most user-friendly UI for configuring CORS if you needed to add your own domains, and wouldn't add a new dependency on the CORS module.

It's also worth noting that in the long run I'd like to include the farmOS-client UI directly in farmOS, which means CORS would not be an issue in the first place, and https://farmOS.app would no longer be necessary. With that in mind, approach (1) above would be the best because it doesn't involve adding anything to farmOS, and would be considered a stop-gap for users who are interested in self-hosting and using farmOS.app in the meantime.

(Also note that using the Android/iOS apps do not need CORS configuration - only https://farmOS.app because it is in the browser.)

jgaehring:
This was needed to give users (particularly self-hosted users) a means of configuring which third-party client sites can access the farmOS API. Prior to this we had farmos.app hardcoded in, but in the future, other clients may want to connect as well. It also alleviates a pain point in development of Field Kit, where we had to go into the module and modify the hardcoded value in order to connect via http:localhost:8080 instead of farmos.app.

This resolves #187 and https://www.drupal.org/project/farm/issues/3083205

Thanks @mstenta for all the help!

### DIFF
--- a/modules/farm/farm_access/farm_access.install
+++ b/modules/farm/farm_access/farm_access.install
@@ -32,6 +32,9 @@ function farm_access_uninstall() {
   foreach ($roles as $name) {
     user_role_delete($name);
   }
+
+  // Clean up variables.
+  variable_del('farm_access_allow_origin');
 }
 
 /**
--- a/modules/farm/farm_access/farm_access.module
+++ b/modules/farm/farm_access/farm_access.module
@@ -33,6 +33,69 @@ function farm_access_hook_info() {
 }
 
 /**
+ * Implements hook_permission().
+ */
+function farm_access_permission() {
+  $perms = array(
+    'administer farm_access module' => array(
+      'title' => t('Administer farm access module'),
+    ),
+  );
+  return $perms;
+}
+
+/**
+ * Implements hook_farm_access_perms().
+ */
+function farm_access_farm_access_perms($role) {
+  $perms = array();
+
+  // Load the list of farm roles.
+  $roles = farm_access_roles();
+
+  // If this role has 'config' access, grant access to farm_access configuration.
+  if (!empty($roles[$role]['access']['config'])) {
+    $perms[] = 'administer farm_access module';
+  }
+
+  return $perms;
+}
+
+/**
+ * Implements hook_menu().
+ */
+function farm_access_menu() {
+
+  // Access configuration form.
+  $items['admin/config/farm/access'] = array(
+    'title' => 'Access',
+    'description' => 'Access configuration settings.',
+    'page callback' => 'drupal_get_form',
+    'page arguments' => array('farm_access_settings_form'),
+    'access arguments' => array('administer farm_access module'),
+  );
+
+  return $items;
+}
+
+/**
+ * Access settings form.
+ */
+function farm_access_settings_form($form, &$form_state) {
+
+  // Metric or US/Imperial.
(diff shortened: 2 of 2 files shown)


modules/farm/farm_access/farm_access.module
```diff
@@ -32,6 +32,69 @@ function farm_access_hook_info() {
...
+/**
+ * Implements hook_farm_access_perms().
+ */
+function farm_access_farm_access_perms($role) {
+  $perms = array();
+
+  // Load the list of farm roles.
+  $roles = farm_access_roles();
+
+  // If this role has 'config' access, grant access to quantity configuration.
```
mstenta: "grant access to **quantity** configuration."

modules/farm/farm_access/farm_access.install
```diff
@@ -43,3 +43,8 @@ function farm_access_update_7000(&$sandbox) {
     module_enable(array($module));
   }
 }
+
+/**
+ * Clean up variables.
+ */
+variable_del('farm_access_allow_origin');
```
mstenta: This needs to be inside `hook_uninstall()` so it only runs when the module is uninstalled.

mstenta (commented):
Thanks @jgaehring - looks great!

Just two minor things I see...

jgaehring:
Thanks for catching those mistakes, @mstenta! Just amended my commits. Let me know if there's anything else.

mstenta:
Merged! Thanks @jgaehring !

Closed https://www.drupal.org/project/farm/issues/3083205 as well.

PR #226, opened 2020-01-06
Fix location input in Planting Quick Form

### LINKED ISSUE #209: Improve date selectors in Planting Quick Form
The date selector for `Seeding`, `Transplanting` and `Harvest` logs in the Planting Quick Form is inconsistent with those supplied in the UI for creating each individual `Log` type.

When creating an individual `log` of these types you can select a date -10/+3 years from the current year. In the Planting Quick Form each date selector provides -3/+10 years.

1. I think the Quick Form selectors should be changed to provide -10/+3 years, as the UI for individual logs provides.
2. It would be great if you could optionally input *any* year... I was going to enter `2006` but can't. The only way around this (that I can think of) is doing it via API. I'm sure there's a barrier to adding this ability, can we overcome it?

### LINKED ISSUE #210: Add description to input fields in Planting Quick Forms
There are descriptions under these fields in the `New Planting` form - we should add the input field descriptions to the `Crop` and `Location` fields in the Planting Quick Form. (`Location`s are included under `seeding`, `transplanting`, and `harvest` options)

Consistent UI! Specifically, its good to know that you can supply multiple locations in the `location` input field.

New Planting form:
Screen Shot 2019-10-26 at 10 19 21 AM

Planting Quick Form:
Screen Shot 2019-10-26 at 10 19 41 AM

### LINKED ISSUE #211: location input field character limit is too small in Planting Quick Form
The `location` field in the Planting Quick Form only allows 128 characters. The `location` field in the New Planting form has no limit. When entering multiple locations with a long name, you can hit the character limit rather quickly, which could prevent users from adding multiple locations.

paul121:
This fixes #211.

- Updated the `location` texfield to have unlimited length to allow inputting many locations (previously was limited to html/css default of 128 characters)
- Trim the `location` string to ensure the length of the planting asset `name` field is < 255 characters. This also takes into account the `season` and `crop tags` to make sure the TOTAL length (with spaces) is less than 255.
- Set the planting asset `name` field's `maxlength` property to 255 characters. Previous was 128 character default.
- Add description fields for Crops and Locations. Closes #210 
- Update the `date_year_range` to `-10:+3` for consistency with other date selectors. Closes #209 

@mstenta I saw your commit https://github.com/farmOS/farmOS/commit/94291e4b92cadae6cb6ebc40c9c4b35e1fc2f985 that added a `farm_movement_asset_location_validate` function to validate the length of each location name to < 255 characters. I thought about copy/pasting the same validator here, but was hesitant for a couple reasons: 
1. Is that code reusable? Not sure if I could simply link to `farm_movement_asset_location_validate` in this file. Also, I don't think the `form_state['values']` are structured the same way, so it would probably fail if I could. Can `farm_asset` provide a callback to validate an `asset_name` ?
2. I also omitted that for sake of simplicity. Kinda unlikely that location names > 255 characters will be input. It would be best if we had a reusable validation function so we aren't copy/pasting wherever we have `asset name` in a form.

### DIFF
--- a/modules/farm/farm_crop/farm_crop.farm_quick.planting.inc
+++ b/modules/farm/farm_crop/farm_crop.farm_quick.planting.inc
@@ -177,6 +177,7 @@ function farm_crop_planting_quick_form($form, &$form_state) {
         '#type' => 'textfield',
         '#title' => t( 'Location'),
         '#autocomplete_path' => 'taxonomy/autocomplete/field_farm_area',
+        '#maxlength' => NULL,
         '#ajax' => array(
           'callback' => 'farm_crop_planting_quick_form_name_ajax',
           'wrapper' => 'planting-name',
--- a/modules/farm/farm_crop/farm_crop.farm_quick.planting.inc
+++ b/modules/farm/farm_crop/farm_crop.farm_quick.planting.inc
@@ -249,10 +249,13 @@ function farm_crop_planting_quick_form($form, &$form_state) {
   }
 
   // The planting will be named based on the season, location and crop.
+  // The location string may be trimmed to 253 char (accounting for spaces)
+  // to ensure that the asset name is less than 255 characters.
+  $crop_string = implode(',', $crop_tags);
   $planting_name_parts = array(
     $season,
-    $location,
-    implode(', ', $crop_tags),
+    substr($location, 0, (253 - strlen($season) - strlen($crop_string)) ),
+    $crop_string,
   );
   $planting_name = implode(' ', $planting_name_parts);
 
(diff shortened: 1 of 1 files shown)


paul121:
Ah, I was curious if we could do a simple regex validation for the location text field? Check that each comma separated value is < 255 characters?

mstenta:
Thanks @paul121! Finally got around to merging this.  :-)

> Ah, I was curious if we could do a simple regex validation for the location text field? Check that each comma separated value is < 255 characters?

This is a good thought. We can think about it in a new issue if we want, but it doesn't need to hold up these improvements. The chances of someone entering an area name that is over 255 characters long is pretty slim, I think. But I agree it would be good to have a check in place to ensure it doesn't cause an error.

PR #230, opened 2020-01-15
Add a farm metrics module to collect high level metrics

paul121:
This module does a few things:

-  Refactors `hook_farm_dashboard_metrics` into `hook_farm_metrics`
-  Implements `hook_farm_dashboard_pane` to display metrics on the dashboard
-  Creates an endpoint at `/farm_metrics.json` to expose metrics over API

Currently this module does not check for any OAuth Scope. I'll include this feature in the PR for the `farm_api_oauth` module.


mstenta:
Awesome thanks @paul121 ! I rebased+merged this with a few changes (discussed with @paul121 in chat):

- Allow all farm roles to access metrics (not just managers).
- Check the new permission when displaying the metrics dashboard pane (only show pane if user has perm).
- Install farm_metrics on new and existing sites via update.php.
- Added a `farm_metrics()` helper function that is used by `farm_metrics_info()` and `farm_metrics_dashboard_pane()`.
- Moved metrics to `/farm.json` instead of `/farm_metrics.json`.

PR #251, opened 2020-02-21
Add farm setup wizard

### LINKED ISSUE #248: Allow modules to be configured in the setup wizard
One main feature of the setup wizard will allow `default` and `optional` `farm` modules to be enabled. General requirements:
- Once enabled, modules should not be disabled/uninstalled
  - Explain this to users
  - In the future, perhaps some "simple" modules could be uninstalled?
- Display enabled modules as "checked"
- Display checkboxes in a fieldset

### LINKED ISSUE #249: Enable contrib modules via the setup wizard
It would be great to also be able to enable contrib modules via the setup wizard.

Some considerations:
- It likely will not be possible to _download_ contrib modules via the setup wizard. What this means is, it will be up to the _host_ to download contrib modules to the `sites/all/modules` directory.
  - Some hosts, such as Farmier, could download pre-approved contrib modules to this directory and allow users to enable them at a later time.
- We can use `system_rebuild_module_data()` to check if they are present when building the form, and then show them as optional checkboxes in the setup wizard.
- @mstenta Perhaps rather than a pre-approved list of contrib modules, could this just look for modules in the `farmOS Contrib` package? In this way, the "pre-approved list" is really just the host deciding which modules are downloaded

paul121:


paul121:
@mstenta I think this PR is nearly done for a first pass on `farm_setup`

What I've done is provide "start" and "finished" steps in `farm_setup` that are used in the setup wizard.
- I think we could add more info to the "start" to briefly explain the setup process & welcome users.
- Similarly the "finished" step needs a bit more polishing. This is also displayed as the "finished" tab in the setup menu, where the list of taxonomies is provided.

It's probably easier to get a feel for those once you test and have it running locally. I don't think we'll need more than simple visual changes there.

mstenta:
Great thanks @paul121 ! I have a few things in the queue ahead of this, but looking forward to reviewing soon!

mstenta:
Despite the fact that this is mostly done, @paul121 and I chatted about this and decided to close this PR for the 7.x-1.x branch as "won't fix". It is already implemented in 2.x, mostly the same but with some differences, and we are just going to focus on that branch instead of worrying about getting this into 7.x-1.x.

There are some features in this that are not in 2.x, so @paul121 will leave his branch up for future reference.

PR #266, opened 2020-03-27
Fix minor bug when enabling OAuth Clients via admin form.

paul121:
The `client_secret` is an optional attribute for OAuth2 Clients provided via `hook_farm_api_oauth2_client`.

This adds a quick to check to see if `client_secret` was provided. Avoids error saying `client_secret` is not set.


mstenta:
Good catch! Thanks @paul121 !

One small thing: the PHP null constant is `NULL` not `null`. But actually... it should probably be an empty string, because that's what the `farm_api_enable_oauth_client()` function expects.

paul121:
Oh oops! Yea, I though about making it `''` - let's go with that. I'll fix that.

PR #275, opened 2020-04-16
Fix minor UI bug with Glyphicons being duplicated

symbioquine:
Fixes

### DIFF
--- a/themes/farm_theme/js/glyphicons.js
+++ b/themes/farm_theme/js/glyphicons.js
@@ -15,7 +15,7 @@
       if (glyphicons_text[link_text] !== undefined) {
         icon = glyphicons_text[link_text]
       }
-      if (icon) {
+      if (icon && !$(element).find('span.glyphicon').length) {
         $(element).prepend('<span class="icon glyphicon glyphicon-' + icon + '"></span> ');
       }
     }


mstenta:
Good find @symbioquine !

This is happening because the `farm_theme_glyphicons` JS behavior is running more than it should.

https://github.com/farmOS/farmOS/blob/7.x-1.x/themes/farm_theme/js/glyphicons.js

Drupal will re-run JS behaviors any time that a Drupal AJAX action takes place - to ensure that JS behaviors are re-applied to HTML that may have been replaced by AJAX.

The proper solution to this issue is to use jQuery's `.once()`. This should have been done originally in that JS (my bad).

Here is more info in the Drupal docs: https://www.drupal.org/node/756722#jquery-once

So I think what we basically want to do is wrap all the logic in `attach()` with `$('#navbar ul', context).once('glyphicons', function () { ... }`

So maybe something like this:

```
    attach: function(context, settings) {
      $('#navbar ul', context).once('glyphicons', function () {
        var glyphicons_text = settings['farm_theme']['glyphicons_text'];
        $('#navbar ul li a', context).each(function(index) {
          Drupal.behaviors.farm_theme_glyphicons.glyphicon(this, glyphicons_text);
        });
        $('ul.tabs--primary li a', context).each(function(index) {
          Drupal.behaviors.farm_theme_glyphicons.glyphicon(this, glyphicons_text);
        });
      });
    },
```

What that does is add a `glyphicons-processed` class to the `#navbar ul`, which `once()` will look for again next time... and won't run the code again if it exists.

symbioquine:
Makes sense! Closing in favor of https://github.com/farmOS/farmOS/pull/276

PR #276, opened 2020-04-16
Fix minor UI bug with Glyphicons being duplicated

symbioquine:
Updated fix based on https://github.com/farmOS/farmOS/pull/275#issuecomment-614697939

Fixes


mstenta:
Thanks @symbioquine !

PR #283, opened 2020-04-24
Add farmOS Default OAuth Client

paul121:


paul121:
@mstenta this should be pretty straightforward. I'm just passing the `settings` array that the `oauth2_server` module expects when customizing clients. This is much simpler than trying add another API layer only allowing some of the settings to be configured.

One question: enabling this default client on existing farmOS installs? Will we need to do manually enable the client in a `hook_update` ?

mstenta:
Thanks @paul121! I made two small changes before merging:

1. I updated `farm_api_modules_enabled()` to ensure that it handles the new settings array on behalf of modules that provide it.
2. I added `farm_api_update_7002()` to enable the new default client on existing installations.

PR #285, opened 2020-04-29
Implement a farm_trait module

symbioquine:
Provide a field collection `field_farm_trait` which can be
used for tracking arbitrary traits associated with crops/
varieties, livestock animal types, or specific assets.

Traits are modeled as a key-value pair where the key comes
from a new taxonomy `farm_trait_types` and the value is an
unconstrained text field. The trait types taxonomy includes
a description - which could be used for documenting how the
trait is measured - and a reference ID which could be used
for linking to a canonical ontology.
e.g http://purl.obolibrary.org/obo/TO_0002734


symbioquine:
As an example, if the new field collection were associated with the animal asset (not part of this PR), one could use a trait to track the presence of the [darkened agouti coat color trait](http://purl.obolibrary.org/obo/MP_0005409) in their rabbits;

image

mstenta:
Super interesting idea @symbioquine ! Perhaps we can take a look at this together on the next monthly call. I'd love to see it in more detail, but unfortunately might not have a chance to until then anyway.

One thing I will say, though (and I hate to be the bearer of bad news): I am trying to move away from Field Collections in farmOS.  See https://www.drupal.org/project/farm/issues/2991374

So I don't think we'll be adopting any new Field Collections in the 7.x-1.x branch.  We are already in the process of rebuilding them as custom fields for farmOS 2.x.

That said, this could still be a useful add-on module for farmOS 7.x-1.x if you want to break it out to its own repository! And maybe we can think about creating an upgrade path for farmOS 2.x when the time comes. Perhaps we should include this idea of "traits" in farmOS 2.x right from the start!

mstenta:
I'd also like to think about how this relates to the existing `{farm_asset_properties}` module, which provides a similar key/value storage... but without using taxonomy, and without much in the way of displaying it anywhere automatically. It was added to help with the Grazing module development, but was never a very "complete" solution.

Seems like this would be a great discussion for farmOS 2.x... how to merge these ideas together into something generally useful.  I love the traits idea, and how it is linked to taxonomy.  There are cases where we may want to be more strict, too, and now allow editing them via terms.... things to think about...

symbioquine:
@mstenta Thanks for all the feedback!

> One thing I will say, though (and I hate to be the bearer of bad news): I am trying to move away from Field Collections in farmOS. See https://www.drupal.org/project/farm/issues/2991374
> 
> So I don't think we'll be adopting any new Field Collections in the 7.x-1.x branch. We are already in the process of rebuilding them as custom fields for farmOS 2.x.

Makes sense. I wonder if there might be a way to capture some of these larger development currents so others don't waste time similarly?

> That said, this could still be a useful add-on module for farmOS 7.x-1.x if you want to break it out to its own repository! And maybe we can think about creating an upgrade path for farmOS 2.x when the time comes.

I might do that, but I'd rather invest the energy in finding a path forward for a core concept of traits in farmOS... :)

> I'd also like to think about how this relates to the existing {farm_asset_properties} module, which provides a similar key/value storage... but without using taxonomy, and without much in the way of displaying it anywhere automatically. It was added to help with the Grazing module development, but was never a very "complete" solution.

I had originally investigated both the `farm_asset_property` and `farm_fields_dynamic` modules as dependencies of this functionality here were my findings;

#### `farm_asset_property`:

1. As you just corroborated, I didn't see any UI integration for properties added via `farm_asset_property`. Adding such UI integration would be more complex and would require namespacing/defensive code to ensure that it wouldn't apply to existing properties - such as those from the Grazing module which already has its own UI integration.
2. It felt trickier to tie the keys back to a vocabulary which could include metadata about the trait - ontology ID and description.
3. It should also be possible to add traits to non-assets. For example, crop/variety vocabulary items should also have traits. We might even imagine that the existing `Days to transplant` and `Days to maturity` fields should be superceded by traits;  image

#### `farm_fields_dynamic`:

The main value this provides is allowing modules to inject fields into their dependencies instead of consuming fields from their dependencies. That is important if one is writing a module that both provides some fields and needs to depend on other aspects of the module(s) which use the fields. However, for the `farm_trait` module it seems perfectly reasonable for it to be a normal dependency of the modules defining the entities that have traits.

There are also some costs the the current `farm_fields_dynamic` pattern which we discussed and which helped steer me away from it;
* Development workflows via the Features module are complicated since the dynamic fields are treated as part of the wrong module.
* Dynamic fields are not cleanly removed on uninstallation of the module which defines them.
* `farm_fields_dynamic` would need to be improved to include an API for declaring the field grouping behavior for the dynamic fields. Naively, I didn't see a way to make that API very simple since field grouping is defined in a pretty complex structure.

> Perhaps we should include this idea of "traits" in farmOS 2.x right from the start!

> Seems like this would be a great discussion for farmOS 2.x... how to merge these ideas together into something generally useful. I love the traits idea, and how it is linked to taxonomy. There are cases where we may want to be more strict, too, and no[t] allow editing them via terms.... things to think about...

I like this idea a lot. My ideal outcome here would be to find a path forward for the 7.x which which translates directly to the same functionality on the D8 branch. Let me know if that's unrealistic...

Also, is there a big-picture timeline for the D8 (farmOS 2.x) development?

mstenta:
> I wonder if there might be a way to capture some of these larger development currents so others don't waste time similarly?

Sorry I do feel bad about that.  I'm open to ideas.  Part of the problem is that a lot of the less-recent discussions are in the drupal.org/project/farm issue queue, not in GitHub - so it's necessary to search both to find things. AND now we have the forum too... :-/

Great outline of the considerations with `farm_asset_property` and `farm_fields_dynamic` too! I agree neither really fits the needs.

Another related thought process that I want to bring to your attention is: https://www.drupal.org/project/farm/issues/2630218

It may be a totally different set of requirements, but it feels like there are some common thoughts involved... the idea of having the ability to add more arbitrary key/value data to records in farmOS... above and beyond the more strictly defined "Fields" (Drupal Field API).

My thought is that the "asset lifespan" idea and "asset properties" probably need to merge together. So maybe that's also a good time to consider this "traits" idea more generally too.

mstenta:
> That said, this could still be a useful add-on module for farmOS 7.x-1.x if you want to break it out to its own repository! And maybe we can think about creating an upgrade path for farmOS 2.x when the time comes.

Just to circle back to this idea...

If we DO create a generalized solution for this in farmOS 2.0.x (Drupal 8/9), then Drupal's Migrate API could be used to automatically pull data from this D7 add-on module and put it in the new place in D8/9 during the migration. That's just to say: don't throw this out! And use it with that potential upgrade path in mind.

mstenta:
Ah forgot to answer this most important question... :-)

> Also, is there a big-picture timeline for the D8 (farmOS 2.x) development?

Yes and no! I am actively working on it right now, and we have funding to make it happen, but it's slow going because everyone still needs stuff to happen in 7.x-1.x in the meantime. So it's a lot of juggling.

I had really hoped that 2020 would see a big chunk of the project get done. And maybe it will! I'm still optimistic. But there are always speed bumps (like a global pandemic). :-/

mstenta:
Chatted with @symbioquine today - closing this in the 7.x-1.x branch. Perhaps worth considering implementing as a contrib module for 2.x!

Maybe this would help: https://www.drupal.org/project/key_value_field

PR #291, opened 2020-05-21
Updated postgis image to version managed directly by postgis team; ad…

buckaroogeek:
…ded restart: unless-stopped direction

As requested. Updated docker-compose override that uses the postgis container managed by the postgis team

### DIFF
--- a/docker/docker-compose.override.postgis.yml
+++ b/docker/docker-compose.override.postgis.yml
@@ -1,7 +1,7 @@
 version: '2'
 services:
   db:
-    image: mdillon/postgis:latest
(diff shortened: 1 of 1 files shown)


mstenta:
Thanks @buckaroogeek ! I merged this in. https://github.com/farmOS/farmOS/commit/e576ec2f0d1261d2f18a0273eafd80253d5beaae

I did not include the `restart: unless-stopped` bit. We don't currently include that parameter in any of the example Dockerfiles. That should be a separate thread to think about adding that, but I don't think we need to include it in the examples.

PR #295, opened 2020-05-22
Autocreate term by name if field is not a list

paul121:


mstenta:
Merged! Thanks @paul121 !

PR #296, opened 2020-05-22
Display archived animals in livestock weight report

paul121:
This adds a simple checkbox to include archived animal assets in the weight report. Animals are displayed with `Animal Name (archived)` in the result table.

I also modified the Animal Groups to display archived groups, but with a label such as `Group Name (archived)`. I'm curious if we should *only* include groups that have animal assets - was there a reason we didn't do that? Simplicity?

Screenshot-2020-5-22 Animal Weights weight-report


paul121:
Also, here is the script I use to generate semi-realistic animal weights: https://gist.github.com/paul121/afdf1575f2d9400a2668fc36180c8d06

mstenta:
Merged - thanks @paul121 !

PR #297, opened 2020-05-23
Fix multiple seasons bug in Planting Quick Form

paul121:
I believe allowing multiple seasons on Planting assets broke this functionality. Simple fix. I'll see if I can quickly allow multiple seasons to be added via the form too.

### DIFF
--- a/modules/farm/farm_crop/farm_crop.farm_quick.planting.inc
+++ b/modules/farm/farm_crop/farm_crop.farm_quick.planting.inc
@@ -32,7 +32,7 @@ function farm_crop_planting_quick_form($form, &$form_state) {
   $form['planting']['season'] = array(
     '#type' => 'textfield',
     '#title' => t('Season'),
-    '#description' => t('What season will this planting be part of? This is used for organizing plantings for future reference, and can be something like "@year" or "@year Summer". This will be prepended to the planting asset name.', array('@year' => date('Y'))),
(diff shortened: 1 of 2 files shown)


paul121:
Okay, added the ability to reference multiple seasons (comma separated) in the Planting Quick form. Seems like we should include that functionality there if we're making it more broadly available?

Also I'm curious if there is a slight naming inconsistency in the Taxonomy Vocabularies. Right now `farm_season` is the only vocabulary that isn't plural. Obviously this is very minor, and I'm not sure what all it entails to change that (assuming more than it's worth) but an observation nonetheless!

mstenta:
Thanks @paul121 - merged!

I'm going to take this one step further, too, and add a general `farm_term_parse_names()` which both `farm_area_parse_names()` and `farm_season_parse_names()` delegate to. And I might actually remove `farm_season_parse_names()` in a followup commit. Need to keep `farm_area_parse_names()` because it's in the API and might be used by others, but we can deprecate it in favor of this more general one.

Good point regarding the plurality of vocabulary names. It is a bit tricky to change, as you guessed, and this is actually something I'm thinking about in 2.0.x already. So we can migrate. And it may actually make more sense to make all of them singular, instead of having some singular and some plural, and then being faced with this same issue again if they change.

paul121:
@mstenta all makes good sense! 👍

PR #316, opened 2020-06-18
Display active plans on the dashboard

paul121:
This creates a simple view of the 5 most recent `active` farm plans.

I chose to only display the plan `name` and plan `type` fields in the table. I considered adding the `season` field, but not all plans have this. Also considered the `active` field but felt redundant?

The plans are sorted by the `changed` timestamp, idea being to display the most recently updated plan on top.

### DIFF
--- a/modules/farm/farm_plan/farm_plan.farm_dashboard.inc
+++ b/modules/farm/farm_plan/farm_plan.farm_dashboard.inc
@@ -1,0 +1,18 @@
+<?php
+/**
+ * @file
+ * Farm dashboard hooks implemented by farm plan module.
+ */
+
+/**
+ * Implements hook_farm_dashboard_panes().
+ */
+function farm_plan_farm_dashboard_panes() {
+  return array(
+    'farm_plan_active_plans' => array(
+      'view' => 'farm_plan',
+      'view_display_id' => 'block_active_plans',
+      'group' => 'plans',
+    ),
+  );
+}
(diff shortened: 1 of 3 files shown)


mstenta:
Awesome! Thanks @paul121 !

> The plans are sorted by the changed timestamp, idea being to display the most recently updated plan on top.

This may not work how we want it to... `changed` will only be updated when the plan entity itself is changed. We are not updating that automatically when plantings within the plan are changed. It might make sense to do that, but that would require extra code in any module that adds a plan type to update that timestamp. So I wonder if it's worth it. Maybe just sorting alphabetically would make sense? What do you think?

paul121:
@mstenta ahh makes sense. yeah, alphabetically is probably fine. Do you think the limit of 5 plans is OK?

mstenta:
> Do you think the limit of 5 plans is OK?

Hmm maybe we don't need to limit? Or, Views also has an option to auto-create a "View more" link which could send you to the "All Plans" page.

Oh... that actually makes me think: we should probably just add this as a display to the existing "All Plans" View, rather than making a whole new View. That would allow you to create the "View more" link that points to the "All Plans" display, if that makes sense?

paul121:
Yeah, that makes sense. Almost did that from beginning anyways. I'll force push with the new display

paul121:
I actually disabled the sort by `name`, but it seems like click sorting happens regardless?

Sorting probably isn't a big priority, but I don't like that some may get hidden in the process of adding this "shortcut" to access plans :confused: 

Here is a preview:
Screenshot_2020-06-18 Dashboard crop-plan-dev

mstenta:
That looks great!

Regarding the limit: yea, should we just remove it and show ALL active plans? That's fine with me! Maybe it will encourage archiving old plans to keep the list focused.

paul121:
@mstenta done. Suppose we shouldn't worry about it until its a problem

mstenta:
Merged - thanks @paul121 !

Issue #317, opened 2020-06-25
Mapbox studio classic styles are deprecated


paul121:
Just got an email from mapbox:

> Mapbox deprecated support for Mapbox Studio Classic styles on June 1st, 2020 and you’ve made 833 requests using the following Classic styles in the last 2 weeks: mapbox.outdoors.

It looks like there is an upgrade path available for both the `mapbox.outdoor` and `mapbox.satellite` classic style layers that farmOS is using (although `mapbox.satellite` isn't deprecated yet).

A cool feature might be to have the core `farm_map_mapbox` module support users providing their own custom styles. It looks like they can be provided in the same "styles url" pattern that the core mapbox styles are defined. In addition to custom styles, this would allow customization of which mapbox layers are added to farmOS-map.

Migration docs: https://docs.mapbox.com/help/troubleshooting/migrate-legacy-static-images-api/
Styles available in new "style url" format: https://docs.mapbox.com/api/maps/#styles

paul121:
More specific instructions for OpenLayers: https://docs.mapbox.com/help/tutorials/mapbox-with-openlayers/#openlayers-3x

mstenta:
https://github.com/farmOS/farmOS/pull/318 was merged - closing this.

PR #318, opened 2020-06-25
Update mapbox layers to use new studio style format.

paul121:
See #317. It's a simple change to use the new style format. I kept the default 256x256 `tilesize` that mapbox studio classic was using. But it seems like 512x512 might be recommended? Thoughts? From the [Static Tiles API docs](https://docs.mapbox.com/api/maps/#static-tiles):

> Default is 512x512 pixels. (512x512 image tiles are offset by one zoom level compared to 256x256 tiles from Mapbox Studio Classic styles. For example, 512x512 tiles at zoom level 4 are equivalent to Mapbox Studio Classic styles tiles at zoom level 5.) 256x256 tiles from the endpoint are one quarter of the size of 512x512 tiles. Therefore, they require 4 times as many API requests and accumulate 4 times as many map views to render the same area.

An important note from the email I received:

> Note: the Static Tiles API can return 512x512 map tiles, which cover 4x the area of the 256x256 tiles associated with Classic styles. Be sure you include the tileSize: 512 and zoomOffset: -1 options so you don't incur 4x as many tile requests as necessary and so that your labels, icons, and other map features don't appear smaller than expected. 

Note this does change from using the Mapbox Raster Tiles API to using the Static Tiles API. As per the [pricing](https://www.mapbox.com/pricing/#maps) page, this Static Tiles API **costs twice as much and has a smaller free-usage tier**. This might be a significant change... but seems like the only way to continue using the `mapbox.outdoor` layer (see mapbox [docs for OL 3.x](https://docs.mapbox.com/help/tutorials/mapbox-with-openlayers/#openlayers-3x)) But, with the larger 512x512 tiles 4x less requests would be made??

### DIFF
--- a/modules/farm/farm_map/farm_map_mapbox/js/farmOS.map.behaviors.mapbox.js
+++ b/modules/farm/farm_map/farm_map_mapbox/js/farmOS.map.behaviors.mapbox.js
@@ -2,8 +2,8 @@
   farmOS.map.behaviors.mapbox = {
     attach: function (instance) {
       var key = Drupal.settings.farm_map.behaviors.mapbox.api_key;
-      this.addMapboxLayer(instance, 'Mapbox Outdoors', 'mapbox.outdoors', key);
-      this.addMapboxLayer(instance, 'Mapbox Satellite', 'mapbox.satellite', key, true);
+      this.addMapboxLayer(instance, 'Mapbox Outdoors', 'mapbox/outdoors-v11', key);
+      this.addMapboxLayer(instance, 'Mapbox Satellite', 'mapbox/satellite-v9', key, true);
     },
     addMapboxLayer: function (instance, title, tileset, key, visible = false) {
 
@@ -17,7 +17,7 @@
       // Add the layer.
       var opts = {
         title: title,
-        url: 'https://api.mapbox.com/v4/' + tileset + '/{z}/{x}/{y}.png?access_token=' + key,
+        url: 'https://api.mapbox.com/styles/v1/' + tileset + '/tiles/256/{z}/{x}/{y}?access_token=' + key,
         attribution: attribution,
         group: 'Base layers',
         base: true,


paul121:
It doesn't appear that the mapbox `satellite` layer imagery has changed with the new version.

But notably mapbox `outdoor` *has changed* (I kinda like the old style better, but supposedly it hasn't been updated since 2018)

Old: 
mapbox-old
mapbox-zoom-old

New:
mapbox-new
mapbox-zoom-new

There are also mapbox `light` and `dark` layers:
mapbox-light
mapbox-dark

mstenta:
512x512 sounds like it would make sense! Want to test it out and make a second commit for that change before we merge?

paul121:
@mstenta hmm testing 512x512 right now. I think this may require a change in farmOS-map to support the XYZ `tileSize` parameter. The default `tileSize` is 256 - [docs](https://openlayers.org/en/latest/apidoc/module-ol_source_XYZ-XYZ.html)

It's hard to notice the issue in farmOS because `zoomToLayer` hides the fact that things are rendered smaller than usual. This is related to the `zoomOffset` that I mentioned above (mapbox docs [here](https://docs.mapbox.com/help/troubleshooting/migrate-legacy-static-tiles-api/)). Problem is, Openlayers doesn't have a `zoomOffset` parameter like they document with Leaflet. From what I can tell, changing the `tileSize` on the XYZ source solves this.

mstenta:
I merged https://github.com/farmOS/farmOS-map/pull/83, released [farmOS-map v1.2.0](https://github.com/farmOS/farmOS-map/releases/tag/v1.2.0), updated `drupal-org.make` (https://github.com/farmOS/farmOS/commit/76748f1e1afebc20860056cf1ccf8fc9becf3685), merged this PR (#318), and updated mapbox layers to use 512x512 tile size (https://github.com/farmOS/farmOS/commit/571ffbe7334bfb091c52127a168189bc2e10fd91).

Thanks again @paul121 !

PR #323, opened 2020-07-16
Add additional filters to livestock weight report

paul121:
- Allow more selective filtering of archived assets as `No`, `Yes`, or `Any`
- Add filtering assets by flag(s)
- Add filtering by asset date of birth within a range

### DIFF
--- a/modules/farm/farm_livestock/farm_livestock_weight/farm_livestock_weight.module
+++ b/modules/farm/farm_livestock/farm_livestock_weight/farm_livestock_weight.module
@@ -212,6 +212,31 @@ function farm_livestock_weight_group_report_form($form, &$form_state) {
   $format = 'Y-m-d';
   $current_date = date($format, REQUEST_TIME);
 
+  $form['input']['birth'] = array(
+    '#type' => 'fieldset',
+    '#title' => t('Birth Date'),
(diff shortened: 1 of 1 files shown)


paul121:
@mstenta I was curious about integrating some of these filters into the actual query that retrieves assets but decided not to. At first I though it would be easy enough to use `farm_group_members_query`and add to the query object instead of loading assets with the `farm_group_members` function but couple issues...

- Not sure if we can override how the previous query handles `archived` assets (we need true, false, or any, and the previous query builds it as true or false?)
- Adding a condition for the asset's `field_farm_date` to handle the birth date range seems complicated... since the previous query is more complex than I initially though it would be :D (querying logs to find members of groups before returning asset ids)

Whatcha think? Having it in the form logic rather than query might make it easier to reuse these filters later on?

mstenta:
Thanks @paul121 - this looks great!

Yea I think doing some of the filtering after the query is fine for now. It might improve performance to integrate some of it into the query, but that can be a follow-up. Doesn't need to block merging this. This LGTM!

mstenta:
Merged: https://github.com/farmOS/farmOS/commit/65190926a062405b1946ee9246bc467f52e5c9e1

PR #325, opened 2020-07-16
Allow creating measurements with a value of 0

paul121:
When creating quantities via `farm_quantity_log_add_measurements` if the measurement value is `0`, the value is not included. Changing the check within that function from `!empty()` to `isset()` should fix this.


mstenta:
Good improvement, thanks @paul121 !

PR #328, opened 2020-07-22
Use hook_sensor_view to add graphs to listener sensor asset views

paul121:
I noticed there was a `hook_sensor_view` provided by `farm_sensor`. A simple change to simplify how `farm_sensor_listener` adds graphs to sensor asset pages. This required doing a recursive merge on the build arrays returned from `hook_sensor_views`, but that shouldn't change any current implementations (if any?)


mstenta:
Merged - thanks @paul121 !

PR #329, opened 2020-07-23
Expose sensor_type and sensor_settings fields to sensor assets REST API

paul121:
Exposes both the `sensor_type` and `sensor_settings` fields to `sensor` assets over the API. These fields are returned via the API, but they can also be created and updated.

This adds support for a few major use cases:
- Any "custom" values saved in `sensor_settings` for any Sensor type can be accessed via the API.
- The `public_key` and `private_key` saved in the `sensor_settings` of `listener` sensors can be accessed via the API.
- Any sensor asset can be created _and be configured with custom settings_ via the API

### DIFF
--- a/drupal-org.make
+++ b/drupal-org.make
@@ -118,6 +118,8 @@ projects[restws][patch][] = "http://www.drupal.org/files/issues/2019-06-17/restw
 projects[restws][patch][] = "http://www.drupal.org/files/issues/2020-03-29/restws-auth-check-2490416-12.patch"
 ; Fix Issue #2301237: Allow creating nodes with multi-value fields
 projects[restws][patch][] = "https://www.drupal.org/files/issues/2020-05-31/2301237-34.patch"
+; Fix Issue #3161113: Support fields with unknown data type.
+projects[restws][patch][] = "http://www.drupal.org/files/issues/2020-07-23/3161113-6.patch"
 
 projects[restws_field_collection][subdir] = "contrib"
 projects[restws_field_collection][version] = "1.1"
--- a/modules/farm/farm_sensor/farm_sensor.module
+++ b/modules/farm/farm_sensor/farm_sensor.module
@@ -34,6 +34,28 @@ function farm_sensor_farm_ui_entity_view_groups() {
 }
 
 /**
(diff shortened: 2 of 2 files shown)


paul121:
Simply returning the `sensor_settings` field via the API (read access) was pretty simple. But making it writable was more complicated. This required a patch to `restws` to disable validation for fields with an `unknown` data type. Normally this would have used the `struct` data type, but since we can't determine what every custom field within the settings might be using the `unknown` type is easier for now.

Drupal issue: https://www.drupal.org/project/restws/issues/3161113#comment-13761538

Some notes from chat that helped us arrive to this solution...

@mstenta saw that a previous patch to `restws` was interfering with this feature: 2318839676c937c493e60ab0e063c78236ea24a4

My thoughts... TLDR; ultimately adding a check for the `unknown` data type was the easiest solution.

> maybe restws should just check if the data type is unknown at the top of getPropertyValue - if True, then just continue, don't perform any validation....

> seems like an easy way out but I don't see anywhere where restws has explicit check for unknown types, where it shouldn't be performing validation anyways

> our use case of sensor_settings being an "unstructured" array seems fairly unique since it doesn't match the other data types

> it most-resembles a struct, but the struct type requires that the structure is defined with another level of property info definitions (which I'm assuming could continue on and on and be recursive)

> perhaps one option.... when defining a sensor type, modules could provide a property_info array describing the structure of sensor_settings that they expect. (but feels like we'd be reinventing the wheel doing this)

> > maybe restws should just check if the data type is unknown at the top of getPropertyValue - if True, then just continue, don't perform any validation....
>
> this was pretty easy & doesn't conflict with the other patch. but not sure if there are any other unknown types floating around in our Drupal instance that might be affected by this... (a search found no conflicts)

paul121:
@mstenta are there any other considerations for this?

One thought that comes to mind is the optional `access_callback` property that can be added to a field. But it seems like the defaults will be OK even if a custom sensor type had special `view` or `edit` permissions defined for it.

mstenta:
This all looks good and makes sense - thanks for the detailed notes and explanations @paul121 - merged!

PR #334, opened 2020-07-30
Make plotly graphs responsive to page size.

paul121:
Simple change to make the plotly graphs provided by `farm_sensor_listener` responsive to the page size. Currently graphs will overlap in almost the entire range of the bootstrap "medium" window size. The graphs break into full width at the "small" window size.

I should note that graphs don't resize super fluidly while changing the window size, but I think this is OK. It also depends on the number of graphs & data points being graphed.

Plotly responsive example: https://plotly.com/javascript/responsive-fluid-layout/

See before and after photos below:


paul121:
Screenshots are taken at the same browser width.

Before:
graphs_before

After:
graphs_after

PR #338, opened 2020-08-06
Include sensor type column in sensor asset view

paul121:
I will soon be adding many sensors to a farmOS instance and it would be useful to include a `sensor type` column. I believe I've done this correctly using `farm_sensor_views_data()`. Note I didn't include the `sensor_settings` column from the DB because I don't imagine that would be useful in Views. (Curious if views could handle a serialized array though?)

I also exposed a filter for `sensor type` but its just a simple string "contains" operator. To make the filter a selectable list I think we would need to implement our own [`views_handler_filter_in_operator`](https://api.drupal.org/api/views/handlers%21views_handler_filter_in_operator.inc/class/views_handler_filter_in_operator/7.x-3.x) ?? (related article: https://atendesigngroup.com/articles/drupal-7-views-drop-down-exposed-filter-text-field)

### DIFF
--- a/modules/farm/farm_sensor/farm_sensor.views.inc
+++ b/modules/farm/farm_sensor/farm_sensor.views.inc
@@ -1,0 +1,45 @@
+<?php
+
+/**
+ * @file
+ * Farm Sensor Views data.
+ */
+
+/**
+ * Implements hook_views_data().
+ */
+function farm_sensor_views_data() {
+
+  // Describe the {farm_sensor} table to Views.
+  $data['farm_sensor']['table']['group'] = t('Sensor');
+
+  // Create an implicit relationship to the farm_asset table, so that when the
+  // base table is farm_asset, sensor fields are automatically available.
+  $data['farm_sensor']['table']['join'] = array(
+    'farm_asset' => array(
+      'left_field' => 'id',
+      'field' => 'id',
+    ),
+  );
+
+  // Sensor reading name.
+  $data['farm_sensor']['type'] = array(
+    'title' => t('Sensor type'),
+    'help' => t('Type of sensor.'),
+    'field' => array(
+      'handler' => 'views_handler_field',
+      'click sortable' => TRUE,
+    ),
+    'sort' => array(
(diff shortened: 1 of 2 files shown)


paul121:
sensor_types

mstenta:
This is great @paul121 !

> Note I didn't include the sensor_settings column from the DB because I don't imagine that would be useful in Views. (Curious if views could handle a serialized array though?)

Yea might as well leave it out for now. Views does not have built-in support for de-serializing, as far as I know, but it would be possible with a custom field handler. I don't think that's worth the effort/complexity right now though. De-serializing adds to the processing because it requires passing each result through the PHP `unserialize()` function, rather than just printing the results of the query. So yea, I say we avoid adding that unless absolutely necessary.

> I also exposed a filter for sensor type but its just a simple string "contains" operator. To make the filter a selectable list I think we would need to implement our own `views_handler_filter_in_operator` ??

Yea that makes sense - string filter is probably enough for now. Let's leave the dropdown select for a future feature request.

I have a couple of SUPER small nitpicks:

- Delete lines 351-353 where it adds the `remember_roles` property to the default View code. This is something that gets added automatically by Views/Features, but it isn't required, and it references raw role IDs, which may be different from site to site. I'm not really sure what it's for, but I've made a habit out of excluding it from farmOS code.
- Let's move the "Sensor type" exposed filter so that it appears right after the "Name" exposed filter (before the "Location"). This will be consistent with the way I've been sorting exposed filters in other Views (roughly ordered from specific to general).
- Can we call the column "Sensor type" instead of "Sensor Type"? I'm trying to get better about consistently only capitalizing the first word in strings like this.

Want to take a quick pass at those? Otherwise this looks great and is ready to go I think!

paul121:
@mstenta all things I was curious about! Sounds good. Just made those changes.

mstenta:
Merged: https://github.com/farmOS/farmOS/commit/8dc567145ace578703a80474a1ed2bb1c8d65bcb

Thanks again @paul121 !

PR #351, opened 2020-09-16
Add helper functions for loading Crop and Crop Family terms.

paul121:
Adds two functions (not sure on the names - could they be better?)
- `farm_crop_families()`: Simple function to help load `farm_crop_families` terms.
- `farm_crop_families_crops()`: Load `farm_crop` terms that are apart of specific Crop Families. No parameter loads all Crop Terms.
  - Went with the `entity->load()` style where `ids = FALSE` specifies that all should be loaded. This requires a bit more logic, but seems cleaner than adding another parameter or needing a separate function to load all crop terms IMO. But open to alternatives!

Helps for building forms that use Crops + Varieties (named however you like haha :smile:):
crop_varieties

### DIFF
--- a/modules/farm/farm_crop/farm_crop.module
+++ b/modules/farm/farm_crop/farm_crop.module
@@ -217,3 +217,61 @@ function farm_crop_families() {
   // Load all terms in the farm_crop_families vocabulary.
   return taxonomy_get_tree($vocabulary->vid);
 }
+
+/**
+ * Load Farm Crop terms. Can optionally limit crops by Crop Family.
+ *
+ * @param array|int|bool $crop_family_tids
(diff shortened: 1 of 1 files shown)


mstenta:
Great thanks @paul121 !

> Adds two functions (not sure on the names - could they be better?)

Minor but: how about `farm_crop_family_crops()` instead of `farm_crop_families_crops()`?

paul121:
That sounds better 👍 Any other changes I can make?

mstenta:
Nope I think it looks good!

paul121:
Force-pushed that change. Should be ready

mstenta:
Done! Thanks again!

PR #355, opened 2020-09-28
Alias oauth2/* endpoints to oauth/*

paul121:
https://www.drupal.org/project/farm/issues/3172818

Adds aliases using `path_save()` via `farm_access_update()`. This seems better than adding them via `hook_menu` and redirecting. I'm afraid a redirect could break different oauth client libraries if they aren't expecting a redirect response.

I've tested the aliased endpoints in farmOS.py and all is working as expected :+1:

### DIFF
--- a/modules/farm/farm_access/farm_access.install
+++ b/modules/farm/farm_access/farm_access.install
@@ -46,3 +46,20 @@ function farm_access_update_7000(&$sandbox) {
     module_enable(array($module));
   }
 }
+
+/**
+ * Add aliases for OAuth2 endpoints.
+ *
+ * https://www.drupal.org/project/farm/issues/3172818
+ */
+function farm_access_update_7001(&$sandbox) {
+  $path_aliases = array(
+    'oauth2/authorize' => 'oauth/authorize',
+    'oauth2/revoke' => 'oauth/revoke',
+    'oauth2/token' => 'oauth/token',
+  );
+  foreach ($path_aliases as $source => $alias) {
+    $path = array('source' => $source, 'alias' => $alias);
+    path_save($path);
+  }
+}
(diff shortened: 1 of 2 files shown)


mstenta:
> Adds aliases using `path_save()` via `farm_access_update()`. This seems better than adding them via `hook_menu` and redirecting. I'm afraid a redirect could break different oauth client libraries if they aren't expecting a redirect response.

This would also require a `hook_install()` so that the alias gets created on new farmOS installs. `hook_update_N()` only runs on existing installs during `update.php`.

But... I think there's a better way (that still avoids a redirect)...

```php
/**
 * Implements hook_menu_alter().
 */
function farm_api_menu_alter(&$items) {

  // Make OAuth2 token endpoints available at both oauth2/* and oauth/*.
  // See https://www.drupal.org/project/farm/issues/3172818
  $oauth_aliases = array(
    'oauth2/authorize' => 'oauth/authorize',
    'oauth2/revoke' => 'oauth/revoke',
    'oauth2/token' => 'oauth/token',
  );
  foreach ($oauth_aliases as $source => $alias) {
    if (!empty($items[$source])) {
      $items[$alias] = $items[$source];
    }
  }
}
```

That essentially just clones the records in the `{menu_router}` table so they are first-class routes in their own right.

Just a sketch - I haven't tested that... but I think it should work... and doesn't require dealing with new installs vs updates.

Also note that I put that in `farm_api.module`, not `farm_access.module`. Do you think that makes more sense, because that's where we declare the dependency on `oauth2_server`?

paul121:
> Also note that I put that in farm_api.module, not farm_access.module

Oh yes, definitely!

> Just a sketch - I haven't tested that... but I think it should work... and doesn't require dealing with new installs vs updates.

This is working great via `hook_menu_alter()`. And very nice we don't need to worry about the update. Good idea! :+1: 

Pushing another commit with this implementation.

mstenta:
Merged - thanks @paul121 !

PR #361, opened 2020-10-27
Fixes for livestock quick movement form

paul121:
- Fix bug where post-grazing observation logs might reference the same area multiple times if different assets were located in the same area before being moved.
- Fix any previous post-grazing observation logs via a batch `hook_update_n` in `farm_livestock`
- Get the asset's current location relative to the movement time specified in the form. This allows for "past" movements to be recorded using the quick form.
- Add ajax to update the current location geometry when changing the date

@mstenta everything seems to be working well, there just seems to be a UI bug with the AJAX on the "Movement Date" field. Once one of the date input fields is changed it becomes full width. This wasn't happening before I added the AJAX callback.

Looking closer, it seems like the AJAX adds a new element within the input select `<span class="input-group-addon"></span>` -  which doesn't exist before modifying the field. This confuses me, because the AJAX command I'm returning explicitly updates the `#current_location` wrapper, not the date.

Any ideas? Having AJAX on the date input is nice because when you change the date, the current location geometry updates quite fluidly!

After changing the "month" and "day" input:
quick_move_bug

Span element that is added:
quick_move_html

### DIFF
--- a/modules/farm/farm_livestock/farm_livestock.install
+++ b/modules/farm/farm_livestock/farm_livestock.install
@@ -508,3 +508,54 @@ function farm_livestock_update_7007(&$sandbox) {
     module_enable(array($module));
   }
 }
+
+/**
+ * Remove duplicate area references from movement quickform logs.
+ */
+function farm_livestock_update_7008(&$sandbox) {
+
+  // Setup. Find observation logs associated with the farm_livestock_move_form
+  // that need to be updated.
+  if (!isset($sandbox['progress'])) {
+    $sandbox['progress'] = 0;
+
+    // Query observation logs linked to this quick form.
+    $query = db_select('farm_quick_entity', 'fqe');
+    $query->addField('fqe', 'entity_id');
+    $query->condition('fqe.entity_type', 'log');
+    $query->condition('fqe.quick_form_id', 'farm_livestock_move_form');
+    $log_alias = $query->join('log', 'l', 'fqe.entity_id = l.id');
+    $query->condition($log_alias . '.type', 'farm_observation');
+    $log_ids = $query->execute()->fetchCol();
+
+    // Finish the update if there are no logs to update.
+    if (empty($log_ids)) {
+      $sandbox['#finished'] = 1;
+      return;
+    }
+
+    $sandbox['log_ids'] = $log_ids;
+    $sandbox['total'] = count($log_ids);
+  }
+
+  // Load the Nth log we need to process.
+  $log = log_load($sandbox['log_ids'][$sandbox['progress']]);
+
+  // Load areas that each log references.
+  $log_wrapper = entity_metadata_wrapper('log', $log);
+  $areas = $log_wrapper->field_farm_area->value();
+
+  // Only update areas if some exist.
+  if (!empty($areas)) {
+
+    // Get the unique areas.
+    $new_areas = array_unique($areas, SORT_REGULAR);
+
+    // Update the log.
+    $log_wrapper->field_farm_area->set($new_areas);
+    $log_wrapper->save();
+  }
+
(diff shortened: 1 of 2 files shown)


modules/farm/farm_livestock/farm_livestock.install
```diff
@@ -508,3 +508,54 @@ function farm_livestock_update_7007(&$sandbox) {
...
+
+  // Load areas that each log references.
+  $log_wrapper = entity_metadata_wrapper('log', $log);
+  $areas = $log_wrapper->field_farm_area->value();
+
+  // Only update areas if some exist.
+  if (!empty($areas)) {
+
+    // Get the unique areas.
+    $new_areas = array_unique($areas, SORT_REGULAR);
```
mstenta: I wonder if we should add another condition here to ensure that we are only saving the log if it indeed has/had duplicates. As it stands now, the `if (!empty($areas))` condition alone will most likely match all logs.

mstenta:
> @mstenta everything seems to be working well, there just seems to be a UI bug with the AJAX on the "Movement Date" field. Once one of the date input fields is changed it becomes full width. This wasn't happening before I added the AJAX callback.

I wonder if switching to `'#type' => 'date_popup'` would avoid this? Worth a try! I've been reluctant to change to date popup without changing it everywhere, but if this fixes this issue we can make an exception.

modules/farm/farm_livestock/farm_livestock.install
paul121: Right now this matches all logs *with areas*. But yes, it will save any log that had at least one area. It seems pretty safe to me, but we could add another condition. Maybe:

```php
if (!empty($areas) && count($areas) > 1) {

  // Get the unique areas.
  $new_areas = array_unique($areas, SORT_REGULAR);

  if (count($areas) > count($new_areas)) {
    // update & save log
  }
}
```

mstenta: Yea! I was thinking something similar. That would be simple enough...

It's a minor optimization, I agree. But easy enough to include. The extra benefit we get (still very minor) is that the `changed` timestamp will only be updated on logs that needed to be changed... so if there's an issue it's another metric we can look at to try to fix things.

mstenta:
Quick updates: @paul121 and I looked at the date issue together and what our options are.

Summary:

- Drupal is wrapping each date select dropdown field with a wrapper DIV so that it can show the spinner icon when AJAX is happening. Bootstrap theme is then applying a 100% width to the items within that DIV.
- We tested date_popup as a quick alternative. Two issues with that:
    - Due to a CSS issue the popup renders under the text fields below it. This can be resolved with this: https://github.com/mstenta/farm_crop_plan/blob/a4114bd80a4bfd3fab784726726550c1e8a00eee/farm_crop_plan.css#L39
    - We also need to be able to specify TIME - but for some reason, no popup appears when you click the time field, and the field does not allow you to type anything into it. We did not dig much deeper into this.
- We discussed some alternatives:
    - We could add a button under the date/time field that the user must manually press to trigger the AJAX to reload. I don't love this because the user shouldn't have to do this or understand why they need to.
    - We could change the date/time field to a plain text field (neither select fields nor popup). This would allow AJAX to work simply, but it would mean the user needs to type out the whole date and time in the correct format to pass validation.
    - We could just leave this all as-is, and NOT refresh the map when the date changes. Under most normal conditions, a user will be entering a date and THEN entering the animal, which itself will trigger the same map refresh AJAX... so assuming they follow that typical flow, then the map will show the correct shape. The only time it won't is if they set the animal first and then set the date.

mstenta:
I've merged all but the last commit of this. I created a separate follow-up issue for it here: https://github.com/farmOS/farmOS/issues/370

PR #364, opened 2020-10-29
Cannot query quantity labels that have spaces

paul121:
I wasn't getting any data back when querying quantity logs with the `Pasture Height` label (trying to retrieve data created by the livestock movement quick form).

Looks like this is because the `farm_quantity_log_query_add_filters()` method escapes the `label` using `db_escape_field()`. This is to strict; using `db_like()` should allow the label to have spaces (and other characters) that are valid for a quantity label.


mstenta:
Good catch - thanks @paul121 !

PR #377, opened 2020-11-23
Run phpunit without XDebug to speed up tests

symbioquine:
**Why:** Tests take about 8 - 9 minutes with XDebug, but closer to
3 minutes without it.

**How:** Implement strategy that mstenta@ described in https://www.drupal.org/project/farm/issues/3183687
using the dev docker image to build up the /opt/drupal directory,
but the prod image to actually run it since that image doesn't have
XDebug.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -20,14 +20,14 @@ jobs:
       # but notably it does NOT override the default PROJECT_VERSION, so the
       # farmOS Composer project 2.x branch is always used.
       - name: Build farmOS 2.x-dev Docker image
-        run: docker build --build-arg FARMOS_REPO=https://github.com/${GITHUB_REPOSITORY} --build-arg FARMOS_VERSION=${FARMOS_VERSION} -t farmos/farmos:2.x-dev docker/dev
+        run: docker build --build-arg FARMOS_REPO=https://github.com/${GITHUB_REPOSITORY} --build-arg FARMOS_VERSION=${FARMOS_VERSION} --build-arg WWW_DATA_ID=33 -t farmos/farmos:2.x-dev docker/dev
       - name: Create docker-compose.yml
-        run: curl https://raw.githubusercontent.com/farmOS/farmOS/2.x/docker/docker-compose.development.yml -o docker-compose.yml
+        run: cp docker/docker-compose.testing.yml docker-compose.yml
       - name: Start containers
-        run: sudo docker-compose up -d
-      - name: Wait 10 seconds
-        run: sleep 10
+        run: docker-compose up -d
+      - name: Wait 15 seconds
+        run: sleep 15
       - name: Run PHPUnit tests
-        run: sudo docker exec -i -u www-data farmos_www_1 phpunit --verbose --debug --group farm
+        run: docker-compose exec -u www-data -T test-runner phpunit --verbose --debug --group farm
       - name: Run PHP CodeSniffer
-        run: sudo docker exec -i -u www-data farmos_www_1 phpcs /opt/drupal/web/profiles/farm
+        run: docker-compose exec -u www-data -T www phpcs /opt/drupal/web/profiles/farm
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - '2.x'
+      - '2.x-**'
 
 jobs:
   build:
--- a/docker/docker-compose.testing.yml
+++ b/docker/docker-compose.testing.yml
@@ -1,0 +1,28 @@
+version: '3'
+services:
+  db:
+    image: postgres:12
+    volumes:
+      - './db:/var/lib/postgresql/data'
+    ports:
+      - '5432:5432'
+    environment:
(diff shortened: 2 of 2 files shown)


symbioquine:
https://github.com/symbioquine/farmOS/runs/1443750112?check_suite_focus=true

.github/workflows/run-tests.yml
```diff
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - '2.x'
+      - '2.x-**'
```
symbioquine, line 6: This is worth checking in since it makes testing via github actions much easier for future changes. See https://irc.farmos.org/bot/log/farmOS/2020-11-21#T57794 for background. I can make a follow-up change to the development documentation to explain how to take advantage of this.

.github/workflows/run-tests.yml
```diff
@@ -19,14 +20,14 @@ jobs:
...
       - name: Create docker-compose.yml
-        run: curl https://raw.githubusercontent.com/farmOS/farmOS/2.x/docker/docker-compose.development.yml -o docker-compose.yml
+        run: cp docker/docker-compose.testing.yml docker-compose.yml
       - name: Start containers
-        run: sudo docker-compose up -d
-      - name: Wait 10 seconds
-        run: sleep 10
+        run: docker-compose up -d
+      - name: Wait 15 seconds
+        run: sleep 15
```
symbioquine, line 29: Increased this to 15 seconds to allow for the previous 10 seconds plus startup of the test-runner container.

.github/workflows/run-tests.yml
```diff
@@ -19,14 +20,14 @@ jobs:
...
       - name: Run PHPUnit tests
-        run: sudo docker exec -i -u www-data farmos_www_1 phpunit --verbose --debug --group farm
+        run: docker-compose exec -u www-data -T test-runner phpunit --verbose --debug --group farm
       - name: Run PHP CodeSniffer
-        run: sudo docker exec -i -u www-data farmos_www_1 phpcs /opt/drupal/web/profiles/farm
+        run: docker-compose exec -u www-data -T www phpcs /opt/drupal/web/profiles/farm
```
symbioquine, line 33: Cleaned this up by using `docker-compose exec` and removing unnecessary `sudo` usage.

docker/docker-compose.testing.yml
```diff
@@ -0,0 +1,28 @@
...
+      - './www:/opt/drupal'
+
+  test-runner:
+    depends_on:
+      - www
+    image: farmos/farmos:2.x
+    entrypoint: /bin/sh
+    command: -c 'sleep 10 && exec docker-entrypoint.sh apache2-foreground'
+    volumes:
+      - './www:/opt/drupal'
```
symbioquine, line 28: Having a dedicated "testing" docker compose file will provide a place to start up additional database types to be tested or other testing-specific customizations in the future - which may deviate from the recommended "development" containers.

.github/workflows/run-tests.yml
mstenta: I like this. Can we give it its own commit, though? It is not directly related to "Run phpunit without XDebug to speed up tests". (Keeping it in this PR as a second commit is fine.)

symbioquine: Sure

symbioquine: Made a separate commit and rebased this one on top of it, then force-pushed it to this PR branch.

docker/docker-compose.testing.yml
mstenta: Great! I like this approach a lot.

Question: what is the `sleep 10` for? And how does it relate to the `sleep 15` in the workflow itself? Do we need this one too?

Is it related to the `depends_on: www` statement?

Are they both there to ensure the `www` has a chance to populate the the codebase with a dev version before the `test-runner` container does? That makes sense, but I wonder if there's a small chance that it is inconsistent. We can cross that bridge if/when we come to it... not necessary to change anything now, I don't think.

mstenta:
This all looks great to me!

One final thought: do we want to give thought to recommendations for running tests locally without Xdebug too? Or handle that in a follow up?

docker/docker-compose.testing.yml
symbioquine: > Are they both there to ensure the www has a chance to populate the the codebase with a dev version before the test-runner container does?

Yes. I toyed with some sort of loop to wait until the directory exists, but that didn't seem very conclusive since we actually want to wait until the full copy has occurred. That could be signalled via a sentinel file, however I didn't want to get into that in this change since it would add a fair bit of complexity.

symbioquine: > Question: what is the `sleep 10` for? And how does it relate to the `sleep 15` in the workflow itself? Do we need this one too?

The two sleeps occur in parallel. The relationship between them is that the one in the workflow must be larger than the one here.

symbioquine:
I was planning on making a follow-up change with the documentation updates - both for running tests faster locally and pushing to Github following that branch naming pattern so tests will be run there.

mstenta:
Awesome - sounds good!

mstenta:
Merged - thanks again @symbioquine !!

PR #378, opened 2020-11-24
2.x parallel w paratest v7

symbioquine:
The second commit 13bb207 shouldn't be merged. It is only needed until https://github.com/farmOS/composer-project/pull/1 is merged to show that the build passes - and is faster.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -30,6 +30,6 @@ jobs:
       - name: Run PHPUnit tests
       # The overridden result printer is needed until Drupal 9.2 when the updated version of HtmlOutputPrinter will be available
       # https://github.com/drupal/drupal/blob/74fbb0aabdee3e1a5da7b8e489a725afdc5824fd/core/tests/Drupal/Tests/Listeners/HtmlOutputPrinter.php
-        run: docker-compose exec -u www-data -T test-runner paratest --passthru='--printer=PHPUnit\\TextUI\\DefaultResultPrinter' --verbose=1 --processes=16 --group farm /opt/drupal/web/profiles/farm
+        run: docker-compose exec -u www-data -T test-runner paratest --functional --passthru='--printer=PHPUnit\\TextUI\\DefaultResultPrinter' --passthru='--process-isolation' --verbose=1 --processes=16 --group farm /opt/drupal/web/profiles/farm
       - name: Run PHP CodeSniffer
         run: docker-compose exec -u www-data -T www phpcs /opt/drupal/web/profiles/farm
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -28,6 +28,8 @@ jobs:
       - name: Wait 15 seconds
         run: sleep 15
       - name: Run PHPUnit tests
-        run: docker-compose exec -u www-data -T test-runner phpunit --verbose --debug --group farm
+      # The overridden result printer is needed until Drupal 9.2 when the updated version of HtmlOutputPrinter will be available
+      # https://github.com/drupal/drupal/blob/74fbb0aabdee3e1a5da7b8e489a725afdc5824fd/core/tests/Drupal/Tests/Listeners/HtmlOutputPrinter.php
+        run: docker-compose exec -u www-data -T test-runner paratest --passthru='--printer=PHPUnit\\TextUI\\DefaultResultPrinter' --verbose=1 --processes=16 --group farm /opt/drupal/web/profiles/farm
       - name: Run PHP CodeSniffer
         run: docker-compose exec -u www-data -T www phpcs /opt/drupal/web/profiles/farm
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -20,7 +20,7 @@ jobs:
       # but notably it does NOT override the default PROJECT_VERSION, so the
       # farmOS Composer project 2.x branch is always used.
       - name: Build farmOS 2.x-dev Docker image
-        run: docker build --build-arg FARMOS_REPO=https://github.com/${GITHUB_REPOSITORY} --build-arg FARMOS_VERSION=${FARMOS_VERSION} --build-arg WWW_DATA_ID=33 -t farmos/farmos:2.x-dev docker/dev
+        run: docker build --build-arg FARMOS_REPO=https://github.com/${GITHUB_REPOSITORY} --build-arg FARMOS_VERSION=${FARMOS_VERSION} --build-arg PROJECT_REPO="https://github.com/symbioquine/farmOS-composer-project.git" --build-arg WWW_DATA_ID=33 -t farmos/farmos:2.x-dev docker/dev
       - name: Create docker-compose.yml
         run: cp docker/docker-compose.testing.yml docker-compose.yml
       - name: Start containers
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -30,6 +30,6 @@ jobs:
       - name: Run PHPUnit tests
       # The overridden result printer is needed until Drupal 9.2 when the updated version of HtmlOutputPrinter will be available
       # https://github.com/drupal/drupal/blob/74fbb0aabdee3e1a5da7b8e489a725afdc5824fd/core/tests/Drupal/Tests/Listeners/HtmlOutputPrinter.php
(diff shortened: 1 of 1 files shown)


symbioquine:
Waiting on https://github.com/symbioquine/farmOS/runs/1445968147?check_suite_focus=true to show the tests pass (like my previous testing has shown) and are faster as expected...

symbioquine:
A [previous build](https://github.com/symbioquine/farmOS/runs/1445944486?check_suite_focus=true) with almost identical code passed in 1 minute 29 seconds;

image

symbioquine:
The tests failed with what I believe is an intermittent error;

```
1) Drupal\Tests\farm_api\Functional\ConsumerClientIdTest::testValidClientId
mkdir(): File exists

/opt/drupal/web/core/lib/Drupal/Core/File/FileSystem.php:247
/opt/drupal/web/core/lib/Drupal/Core/File/FileSystem.php:232
/opt/drupal/web/core/lib/Drupal/Core/StreamWrapper/LocalStream.php:453
/opt/drupal/web/core/lib/Drupal/Core/Test/TestRunnerKernel.php:91
/opt/drupal/web/core/lib/Drupal/Core/Test/FunctionalTestSetupTrait.php:620
/opt/drupal/web/core/tests/Drupal/Tests/BrowserTestBase.php:399
/opt/drupal/web/modules/contrib/simple_oauth/tests/src/Functional/TokenBearerFunctionalTestBase.php:95
/opt/drupal/web/profiles/farm/modules/core/api/tests/src/Functional/OauthTestBase.php:37
/opt/drupal/vendor/phpunit/phpunit/src/Framework/TestResult.php:730
```

image

Retrying the build...

symbioquine:
Just to provide an update here, I'm troubleshooting the cause of those failures - it appears to be a race-condition similar to the following two issues;

* https://www.drupal.org/project/drupal/issues/2685957
* https://www.drupal.org/project/drupal/issues/3040166

symbioquine:
I put together [a patch](https://raw.githubusercontent.com/symbioquine/farmOS/2.x-parallel-w-paratest-v8/drupal_core_avoid_mkdir_e_warning.patch) that could be a work-around for the race-condition which is causing the tests to fail;

```patch
diff --git a/core/lib/Drupal/Core/File/FileSystem.php b/core/lib/Drupal/Core/File/FileSystem.php
index ab21b365cb..66d2170288 100644
--- a/core/lib/Drupal/Core/File/FileSystem.php
+++ b/core/lib/Drupal/Core/File/FileSystem.php
@@ -244,11 +244,12 @@ public function mkdir($uri, $mode = NULL, $recursive = FALSE, $context = NULL) {
    */
   protected function mkdirCall($uri, $mode, $recursive, $context) {
     if (is_null($context)) {
-      return mkdir($uri, $mode, $recursive);
+      $success = @mkdir($uri, $mode, $recursive);
     }
     else {
-      return mkdir($uri, $mode, $recursive, $context);
+      $success = @mkdir($uri, $mode, $recursive, $context);
     }
+    return $success || (file_exists($uri) && $mode === (fileperms($uri) & 0777));
   }
 
   /**

```

The two test runs I've kicked off with that patch have succeeded, though that's by no means conclusive given the illusive nature of race-conditions;

* https://github.com/symbioquine/farmOS/runs/1451004921?check_suite_focus=true
* https://github.com/symbioquine/farmOS/runs/1451051605?check_suite_focus=true

Short of running every test/test-case against a completely separate file structure, it seems like there needs to be some sort of fix along the lines of the patch above to avoid the E_WARNING from this race-condition where [TestRunnerKernel.php](https://github.com/drupal/drupal/blob/9.0.x/core/lib/Drupal/Core/Test/TestRunnerKernel.php#L90-L92) does `is_dir` followed by `mkdir`;

```php
    if (!is_dir('public://simpletest')) {
      mkdir('public://simpletest', 0777, TRUE);
    }
```

Looking forward to any feedback as to how best to proceed...

mstenta:
That's a big improvement - thanks for digging into this @symbioquine !!

What does the `--process-isolation` flag add?

Regarding the core bug, it should probably be opened as an issue in the Drupal queue upstream. If we do decide to include a core patch, I would prefer it to have some forward-looking plan (ideally it is merged upstream eventually).

Out of curiosity I did a quick search for `paratest` in the Drupal core queue: https://www.drupal.org/project/issues/drupal?text=paratest&status=All

Found this: https://www.drupal.org/project/drupal/issues/2781123 - looks like it stalled, but might be some more relevant info in there worth reviewing.

This one also talks about concurrency and `paratest`: https://www.drupal.org/project/drupal/issues/2867818

symbioquine:
> What does the `--process-isolation` flag add?

It causes phpunit to run each test in its own process.

I was originally seeing whether that would resolve the race condition before I understood it more fully. I can get rid of that, but I was focused on getting to the bottom of the race condition at that point so I haven't cleaned up yet.

> Regarding the core bug, it should probably be opened as an issue in the Drupal queue upstream. If we do decide to include a core patch, I would prefer it to have some forward-looking plan (ideally it is merged upstream eventually).

Happy to open one and propose the patch. I might need help figuring out where to open the issue though...

> Out of curiosity I did a quick search for `paratest` in the Drupal core queue...

I think those two issues are stalled for different reasons that wouldn't necessarily affect whether paratest is a reasonable path forward for running the farmOS tests;

* https://www.drupal.org/project/drupal/issues/2781123 - Needs a lot of dependency wrangling
* https://www.drupal.org/project/drupal/issues/2867818 - Paratest isn't a 1:1 replacement for the existing CI script and they're hesitant to support a fourth "test runner" mechanism

Fortunately, the dependencies didn't turn out to be too complex and all the tests are under a single directory so it's easy to tell Paratest how to find them.

mstenta:
> Happy to open one and propose the patch. I might need help figuring out where to open the issue though...

Sure! So the Drupal core issue queue would be the place for it: https://www.drupal.org/node/add/project-issue/drupal

I think 9.2.x-dev is the current HEAD, so select that under "Version", and there's a "phpunit" component.

> I think those two issues are stalled for different reasons that wouldn't necessarily affect whether paratest is a reasonable path forward for running the farmOS tests

+1 to using paratest in farmOS regardless of those. Just posted them in case they had any other useful information - I haven't read them thoroughly.

mstenta:
> Happy to open one and propose the patch. I might need help figuring out where to open the issue though...

Drupal.org has instructions for creating a patch too... if you need them: https://www.drupal.org/docs/develop/git/using-git-to-contribute-to-drupal/making-a-patch

Generally patches are named `[issue-number]-[comment-number].patch`

mstenta:
Regarding patching Drupal core - I like to avoid it whenever possible, because it introduces another thing that can break our build process if/when upstream changes occur.

The thought occurred to me, though: we would only need this particular patch in the dev environment, not in production. So I did some digging to see what might be possible...

We currently use https://github.com/cweagans/composer-patches to apply patches to Drupal core and contrib packages via our `composer.json`. For example: https://github.com/farmOS/farmOS/blob/267ee074198082fa4f324df98b5d59d3a037c9d0/composer.json#L43

There appears to be an open feature request for adding a `patches-dev` option, which would allow patches to be applied only on dev builds: https://github.com/cweagans/composer-patches/issues/237

Doesn't look like there's been much activity on that. However, it led me to find this: https://github.com/nathandentzau/composer-patches-dev - which "extends the cweagans/composer-patches plugin to allow patches to be applied to packages when composer is in dev mode".

So, one idea is to add that package under the dev dependencies, and add our patch via `patches-dev`. Then we know it's only applied in dev environment builds - which is the only place we need it.

symbioquine:
> So, one idea is to add that package under the dev dependencies, and add our patch via `patches-dev`. Then we know it's only applied in dev environment builds - which is the only place we need it.

I like that idea! I'll try to get an issue/patch sent to Drupal/Core then try that...

symbioquine:
I've opened https://www.drupal.org/project/drupal/issues/3192365 with Drupal Core regarding the `mkdir` race condition.

mstenta:
Very nice @symbioquine ! Nice simple patch. :-)

I'm looking forward to using paratest - our test time has been growing!

symbioquine:
The most recent 10 builds for farmOS 2.x took an average of 307 seconds for the phpunit step;

```
311s
284s
314s
250s
329s
308s
272s
325s
333s
345s
```

I've executed two test runs with the current version of these changes. (new PR coming soon)

* `189s` https://github.com/symbioquine/farmOS/runs/1690088541?check_suite_focus=true
* `171s` https://github.com/symbioquine/farmOS/runs/1690400928?check_suite_focus=true

symbioquine:
Closing in favor of https://github.com/farmOS/farmOS/pull/399

PR #379, opened 2020-11-26
Wait only as long as necessary for Postgres and the farmOS file copying to be done

symbioquine:
Cuts about 15s of unnecessary waiting out of the build because it turns out the test-runner and Postgres containers were usually already done starting up before that `sleep 15` step of `run-tests.yml`.

This is tested via [this Github action run](https://github.com/symbioquine/farmOS/runs/1460645605?check_suite_focus=true) which includes;

* https://github.com/paul121/farmOS/commit/61c129fd4c594d6311d79780168a60ce0718f3df

The branch of this PR won't pass because it doesn't include that commit.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -25,8 +25,10 @@ jobs:
         run: cp docker/docker-compose.testing.yml docker-compose.yml
       - name: Start containers
         run: docker-compose up -d
-      - name: Wait 15 seconds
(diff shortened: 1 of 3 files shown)


mstenta:
Beautiful! Thanks @symbioquine !!

Merged: https://github.com/farmOS/farmOS/commit/4f88cdda92dc5bc0c184b067a502d1656ae6bb6b

PR #383, opened 2020-12-04
Fix google maps api version parameter.

paul121:
Based on these [docs](https://developers.google.com/maps/documentation/javascript/versions), I think we should be using the `weekly` version of the Google Maps API.

> For most applications, we recommend the weekly channel. This is the most current and up-to-date version and contains the latest bug fixes and performance improvements.

Version 3 was [deleted](https://developers.google.com/maps/documentation/javascript/versions#documentation-for-the-api-versions), so I believe we have been using the `weekly` version anyways.

> If you do not explicitly specify a channel or version, you will receive the weekly channel by default.

### DIFF
--- a/modules/farm/farm_map/farm_map_google/farm_map_google.farm_map.inc
+++ b/modules/farm/farm_map/farm_map_google/farm_map_google.farm_map.inc
@@ -25,7 +25,8 @@ function farm_map_google_farm_map_view($name, $element) {
   if (!empty($google_api_key)) {
 
     // Add remote Google Maps JavaScript library, with the API key.
-    $url = 'https://maps.googleapis.com/maps/api/js?v=3&key=' . $google_api_key;
+    // See https://developers.google.com/maps/documentation/javascript/versions
+    $url = 'https://maps.googleapis.com/maps/api/js?v=weekly&key=' . $google_api_key;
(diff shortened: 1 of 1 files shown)


paul121:
> Specify a version by number only if you regularly test and update your application. This is because numbered versions are eventually deleted as new versions are created every quarter.

https://developers.google.com/maps/documentation/javascript/versions#documentation-for-the-api-versions

mstenta:
This makes sense. And yet I'm still always nervous to make any change to the Google integration because it feels brittle to me. :-)

I'll try to test this out soon. If anyone else has a chance to as well, add your thumbs up here!

Having read the doc linked above, should we consider using `quarterly` instead of `weekly`?

> Some applications may benefit from fewer but larger updates, because it gives greater predictability. These applications should use the quarterly channel. Use this approach if you want bug fixes and performance improvements to arrive at an expected time

I am honestly more and more tempted to split all farmOS Google Map integration out of farmOS core and into a contrib module, because I don't want us to have to be tied to something as flaky as Google.

mstenta:
Oops this closed because I was cleaning up branches on origin, and this was based on origin. @paul121 would you mind reopening this from your fork at some point?

If you think it's still worth fixing in 7.x-1.x, that is...

paul121:
Sorry yea, meant to close this awhile ago. Let's just drop it...

PR #384, opened 2020-12-06
Backport inclusion of WWW_DATA_ID docker devel image build arg to 7.x-1.x

symbioquine:
Make default devel image uid/gid=1000 and allow for it to be easily customized - motivation for those things is already well described in the comment that was also copied from the 2.x branch.

*Note: This could be considered a breaking change since it changes the default id from 33 to 1000, but I'm hoping its still a go because it only affects the devel image.*

### DIFF
--- a/docker/dev/Dockerfile
+++ b/docker/dev/Dockerfile
@@ -7,6 +7,16 @@ ARG FARMOS_REPO=https://git.drupal.org/project/farm.git
 # Set the farmOS branch to git clone.
 ARG FARMOS_BRANCH=7.x-1.x
 
+# Change the user/group IDs of www-data inside the image to match the ID of the
+# developer's user on the host machine. This allows Composer to create files
+# owned by www-data inside the container, while keeping those files editable by
+# the developer outside of the container.
+# This defaults to 1000, based on the assumption that the developer is running
+# as UID 1000 on the host machine. It can be overridden at image build time with:
+# --build-arg WWW_DATA_ID=$(id -u)
+ARG WWW_DATA_ID=1000
+RUN usermod -u ${WWW_DATA_ID} www-data && groupmod -g ${WWW_DATA_ID} www-data
+
 # Install Xdebug.
 RUN yes | pecl install xdebug \
 	  && echo "zend_extension=$(find /usr/local/lib/php/extensions/ -name xdebug.so)" > /usr/local/etc/php/conf.d/xdebug.ini \


mstenta:
This sounds good to me! We struggled with this same issue in 7.x-1.x and have just been working around it.

> Note: This could be considered a breaking change since it changes the default id from 33 to 1000, but I'm hoping its still a go because it only affects the devel image.

Curious what the actual impact will be with this. In theory, anyone who followed the instructions at https://farmos.org/development/docker/#file-ownership will have a codebase that is owned by `1000`, but with a group of `33` in many files/folders. I think that's OK though? Everything will be owned by `1000`, which will be the new Apache ID inside the container - so it should still be able to read/write everything it needs to, I would think.

Ideally you would run `sudo chown -R ${USER}:${USER} www` to set the group to `1000` on everything as well.

Just curious if there's anything I'm overlooking, which would cause things to mysteriously break.

mstenta:
We will have to update the Docker instructions alongside this... because it will no longer be necessary to run the `chown` command on the `www` directory.

It might be best to prepare these PRs, but wait to merge them until 7.x-1.7 is ready to be tagged. Then we can roll it out at the same time and add a blurb to the release notes. So if anyone does experience problems, the release notes explain why.

symbioquine:
> Curious what the actual impact will be with this. In theory, anyone who followed the instructions at https://farmos.org/development/docker/#file-ownership will have a codebase that is owned by `1000`, but with a group of `33` in many files/folders. I think that's OK though? Everything will be owned by `1000`, which will be the new Apache ID inside the container - so it should still be able to read/write everything it needs to, I would think.
> 
> Ideally you would run `sudo chown -R ${USER}:${USER} www` to set the group to `1000` on everything as well.
> 
> Just curious if there's anything I'm overlooking, which would cause things to mysteriously break.

Well, the container wouldn't know anything about uid/gid 33 any more so it will just be running as uid/gid 1000, but with a contents of `/var/www` which has many files/folders owned by the now unknown uid/gid 33.

The most likely thing to break would be file-uploads and module installation through the UI/Drush. Drupal should complain in the status report that it's not seeing the file permissions it expects too I believe.

However, it will only affect those using the `farmos/farmos:dev` docker image and as you say the fix should be a simple `sudo chown...` which isn't any more complex than what the documentation already describes for fixing the file ownership.

> We will have to update the Docker instructions alongside this... because it will no longer be necessary to run the `chown` command on the `www` directory.
> 
> It might be best to prepare these PRs, but wait to merge them until 7.x-1.7 is ready to be tagged. Then we can roll it out at the same time and add a blurb to the release notes. So if anyone does experience problems, the release notes explain why.

Yep, makes sense. I'll get a PR ready for that too.

symbioquine:
@mstenta I've opened https://github.com/farmOS/farmOS.org/pull/103 for the aforementioned documentation updates.

mstenta:
Merged - thanks again @symbioquine !

PR #385, opened 2020-12-06
Add convenience env vars to devel docker entrypoint

symbioquine:
Adds;

* `FARMOS_INSTALL_CMD` - A command to be executed just after copying the
  farmOS source from /tmp/www
* `FARMOS_DEV_MODULES_DIRECTORY` - A directory containing modules under
  development. These will be symlinked into the Drupal modules directory
  and enabled using drush.

Together these allow for a docker-compose.yml configuration which
automatically sets up farmOS and installs contrib modules which are
under development - since these are bind-mounted they can be edited
live in the source tree on the host system.

```yaml
  www:
    depends_on:
      - db
    image: farmos/farmos:dev
    volumes:
      - './www:/var/www/html'
      - './php-custom.ini:/usr/local/etc/php/conf.d/php-custom.ini'
      - '../farmos_wfs:/farmos_dev_modules/farmos_wfs'
    ports:
      - '80:80'
    environment:
      XDEBUG_CONFIG: remote_host=172.17.0.1
      FARMOS_INSTALL_CMD: while { ! exec 3<>/dev/tcp/db/3306; } > /dev/null 2>&1; do sleep 0.1; done && drush site-install farm --debug --yes --locale=us --db-url=mysql://farm:farm@db/farm --site-name=Test0 --account-name=root --account-pass=test install_configure_form.update_status_module='array(FALSE,FALSE)'
      FARMOS_DEV_MODULES_DIRECTORY: /farmos_dev_modules
```

Though not integral to this pattern, the php-custom.ini file contains;

```ini
opcache.enable=0
```

Which ensures php includes are executed every page load - making
develpment testing easier.

### DIFF
--- a/docker/dev/docker-entrypoint.sh
+++ b/docker/dev/docker-entrypoint.sh
@@ -5,6 +5,22 @@ set -e
 if ! [ "$(ls -A /var/www/html/)" ]; then
   echo "farmOS webroot not detected. Copying from pre-built codebase in the Docker image."
   cp -rp /tmp/www/. /var/www/html
+
+  # If the FARMOS_INSTALL_CMD variable is defined run the expression it contains. This allows
+  # for a drush site-install command to be specified as part of a docker-compose.yml file.
+  if [ -n "$FARMOS_INSTALL_CMD" ]; then
+    echo "Running '$FARMOS_INSTALL_CMD'"
+    eval "$FARMOS_INSTALL_CMD"
+  fi
+fi
+
+# If the FARMOS_DEV_MODULES_DIRECTORY variable is defined, symlink and use drush to enable
+# each module contained therein. This allows modules to be bind-mounted for live development.
+if [ -n "$FARMOS_DEV_MODULES_DIRECTORY" ]; then
+  find "$FARMOS_DEV_MODULES_DIRECTORY" -maxdepth 1 -mindepth 1 -execdir sh -c '\
+      echo Symlinking and drush enabling $(basename {}) && \
+      ln -sTf "$FARMOS_DEV_MODULES_DIRECTORY"/$(basename {}) /var/www/html/sites/all/modules/$(basename {}) && \
+      drush --root=/var/www/html pm-enable --yes $(basename {})' \;
 fi
 
 # Execute the arguments passed into this script.


mstenta:
> `FARMOS_INSTALL_CMD` - A command to be executed just after copying the
farmOS source from /tmp/www

This is a neat idea, but it doesn't feel like something we should include in farmOS itself. This can be accomplished in a downstream/custom `Dockerfile` + `docker-entrypoint.sh` that inherits from `farmos/farmos:7.x-1.x`.

> `FARMOS_DEV_MODULES_DIRECTORY` - A directory containing modules under
development. These will be symlinked into the Drupal modules directory...
> ... contrib modules which are
under development - since these are bind-mounted they can be edited
live in the source tree on the host system.

I'm not sure this is necessary. We are already bind-mounting `'./www:/var/www/html'` - and the modules directory is `/var/www/html/sites/all/modules`, so they are already editable from the host system.

> `opcache.enable=0`
> Which ensures php includes are executed every page load - making
develpment testing easier.

Ah! I took a similar approach in 7.x-1.x development by just manually adding `php_flag opcache.enable Off` to my `.htaccess` file locally. But that's not a great approach for a lot of reasons. For one, it gets overwritten anytime `.htaccess` is rebuilt by Drush. Setting it in an `ini` file is better in that regard.

However, I found a better way to fix the "opcache in dev" issue in the 2.x branch, which could be backported:

```
# Set OPcache's revalidation frequency to 0 seconds for development.
# See https://www.php.net/manual/en/opcache.configuration.php#ini.opcache.revalidate-freq
RUN sed -i 's|opcache.revalidate_freq=60|opcache.revalidate_freq=0|g' /usr/local/etc/php/conf.d/opcache-recommended.ini
```

https://github.com/farmOS/farmOS/blob/8ff0353dd3122776d241c002436939cd777c5a84/docker/dev/Dockerfile#L22

Ultimately the problem is that Opcache only checks for file changes every 60 seconds by default. In the dev environment we can set that to 0, which means it will always check for file updates.

The key benefit here is that it is only *checking* for updates to the files... and only refreshing a particular file in the cache if its timestamp has changed. It is not turning the cache off entirely. So we still get the benefit of the opcode caching. At least that is how I understand it to be working...

Happy to consider merging a backport of that to 7.x-1.x!

With all of the above in mind, I'm going to close this PR - but as always I'm open to discussion! :-)

symbioquine:
> > `FARMOS_INSTALL_CMD` - A command to be executed just after copying the
> > farmOS source from /tmp/www
> 
> This is a neat idea, but it doesn't feel like something we should include in farmOS itself. This can be accomplished in a downstream/custom `Dockerfile` + `docker-entrypoint.sh` that inherits from `farmos/farmos:7.x-1.x`.

Certainly doable, but that would require replacing the whole `docker-entrypoint.sh` (Otherwise there isn't a good way to "hook" into executing only on that first startup.) which could be considered a little fragile if upstream changes occur. (Arguably such breaking changes are probably unlikely in `7.x-1.x`, but I'm aiming to create a pattern for module development which is replicable in `2.x` as well.)

> I'm not sure this is necessary. We are already bind-mounting `'./www:/var/www/html'` - and the modules directory is `/var/www/html/sites/all/modules`, so they are already editable from the host system.

That's true, but the modules directory is only editable after the image populates the `./www` bind-mount. This means that the lack of a mechanism such as this encourages one of two things;

1. One-off dev environments which are harder to document and reproduce
2. Axillary scripting or manual steps which must occur before a functional dev environment is reached

> Ah! I took a similar approach in 7.x-1.x development by just manually adding `php_flag opcache.enable Off` to my `.htaccess` file locally. But that's not a great approach for a lot of reasons. For one, it gets overwritten anytime `.htaccess` is rebuilt by Drush. Setting it in an `ini` file is better in that regard.

Thanks for the background/advice, but I wish I'd left the explanation of the opcache part out of my example in my commit description since it's a bit of a distraction - it's not part of the actual changes in the PR.

----

As you rightly point out all of this ***could*** be achieved in downstream/derivative/custom docker configuration. There's a lot of approaches;

1. Check in a `Dockerfile` + `docker-entrypoint.sh` as part of each module's testing/devel harness
2. Check in a much more complex `docker-compose.yml` file which hacks around these problems directly as part of each module's testing/devel harness
3. Vend a `symbioquine/farmOS-module-dev-harness` Docker Hub image

However, I believe those all have pretty significant maintenance/complexity costs and that the vision represented by these changes is one which ***should*** be supported out-of-the-box by the devel farmOS docker image.

To make that vision a little more explicit;

> It should be possible with only a (relatively) trivial `docker-compose.yml` file to encode a fully functional development environment for a contrib module. That environment should be startable with only a single command (i.e. no clicking or follow-up commands) which yields a ready-for-testing default farmOS installation with the contrib module(s) already installed/enabled. Changes to that module's code should easy to test and check back in.

mstenta:
Thanks for the additional context @symbioquine - all of that is definitely valid, and I see more now what your vision is, I think.

It gives me another idea...

If the goal is to spin up a development environment specifically for just ONE module, then perhaps you don't need to mount the entire codebase at all. Perhaps the `docker-compose.yml` file for module development ONLY needs to mount the module repo directory into `/opt/drupal/web/sites/all/modules`.

Then you wouldn't need to jump through hoops to alter/extend the `docker-entrypoint.sh` script - whose sole purpose is to populate/persist a bind-mounted volume of the farmOS codebase as a whole (or just the `sites` directory in production).

It means that you would be making a choice, however: are you using the environment for BOTH farmOS core development AND farmOS module development? Or are you just trying to automate the setup and testing for a single module?

If the former (both core and contrib dev), then automation is more tricky - but it doesn't sound like that's the goal here. If the latter, then a simple `docker-compose.yml` file included with the module itself, which only mounts the module directory, could work.

That should work for automated testing via GitHub Actions. And for local development, you could also mount the `sites/default` directory to preserve `settings.php`.

Would something like that work?

symbioquine:
> If the goal is to spin up a development environment specifically for just ONE module, then perhaps you don't need to mount the entire codebase at all. Perhaps the `docker-compose.yml` file for module development ONLY needs to mount the module repo directory into `/opt/drupal/web/sites/all/modules`.

Doing that kind of works, but it requires some manual steps since one would have to modify the docker-compose.yml file to create the bind-mount only after the container has started for the first time and populated the `/opt/drupal` directory. (Otherwise, the `docker-entrypoint.sh` discovers that the `/opt/drupal` directory already exists and doesn't copy the code-base at all.) Bind-mounting into `/tmp/www` also kind of works, but doesn't allow for live editing of the code.

> are you using the environment for BOTH farmOS core development AND farmOS module development?

That wasn't my intention, though I do see applications of the `FARMOS_INSTALL_CMD` var to both scenarios.

mstenta:
In 2.x, `/opt/drupal` is populated during image build with the full codebase. So if you don't bind-mount anything, it still works. We basically just have the added layer of `docker-entrypoint.sh` because bind-mounting volumes replaces that directory with an empty directory from the host, so we just check to see if it's empty and then copy from `/var/farmOS` in the container (where we build via Composer during image build).

https://github.com/farmOS/farmOS/blob/2fa3220c82f2051c6a0baa109e6e08fa7ae8f4c7/docker/Dockerfile#L70

So... it should still be possible to bind-mount a module repo directory into `/opt/drupal/web/sites/all/modules` via a simple `docker-compose.yml`. The entrypoint script will see that `/opt/drupal` and `/opt/drupal/web/sites` already exist (from the original image, not from a bind-mount) and won't do anything.

mstenta:
It's sort of buried in this last line, so easy to miss... :-)

https://github.com/farmOS/farmOS/blob/2fa3220c82f2051c6a0baa109e6e08fa7ae8f4c7/docker/Dockerfile#L75

`cp -rp /var/farmOS /opt/drupal`

symbioquine:
> In 2.x, `/opt/drupal` is populated during image build with the full codebase. So if you don't bind-mount anything, it still works. We basically just have the added layer of `docker-entrypoint.sh` because bind-mounting volumes replaces that directory with an empty directory from the host, so we just check to see if it's empty and then copy from `/var/farmOS` in the container (where we build via Composer during image build).
> 
> https://github.com/farmOS/farmOS/blob/2fa3220c82f2051c6a0baa109e6e08fa7ae8f4c7/docker/Dockerfile#L70
> 
> So... it should still be possible to bind-mount a module repo directory into `/opt/drupal/web/sites/all/modules` via a simple `docker-compose.yml`. The entrypoint script will see that `/opt/drupal` and `/opt/drupal/web/sites` already exist (from the original image, not from a bind-mount) and won't do anything.

If I'm understanding correctly that's only applicable for 2.x right now. Are you suggesting changing 1.x work the same way? As far as I can tell, I'm going to be developing/maintaining 1.x modules at least until more of the core for 2.x is done - e.g. areas...

Also, the bind-mount is only part of the module setup. In 1.x there's still `drush pm-enable`. In 2.x there's the composer install step + `drush pm-enable`.

mstenta:
> If I'm understanding correctly that's only applicable for 2.x right now. Are you suggesting changing 1.x work the same way?

Yes that's only in 2.x. The prod image almost does the same thing in 1.x, but in general the baked in assumptions in 1.x are that you'll only be preserving the `/var/www/html/sites` directory in produciton, and the whole `/var/www/html` directory in development. I refactored that in 2.x so that both images work the same, and EITHER `/opt/drupal/` OR `/opt/drupal/web/sites` will be preserved depending on which one you bind-mount.

I'd certainly be open to backporting the way that works to 1.x. It would simplify both the prod and dev images, and they could share the `docker-entrypoint.sh` in the same way.

I wonder if in doing so it would also reveal a better way to extend the image for customizations like this.

> I'm going to be developing/maintaining 1.x modules at least until more of the core for 2.x is done - e.g. areas...

FWIW I have a `2.x-areas` branch that has a LOT of progress - so we are indeed getting close. But I definitely understand that there will be need to stick with 1.x for a little while longer.

> Also, the bind-mount is only part of the module setup. In 1.x there's still `drush pm-enable`. In 2.x there's the composer install step + `drush pm-enable.`

That's true. Not needed for automated testing via GitHub Actions... but convenient if the goal is to set up a local dev environment with a single command.

symbioquine:
I still think the changes in this PR would probably have been worthwhile, but I'm not planning to pursue it further given that I've been able to work-around it;

```yml
version: '2'
services:
  db:
    image: mariadb:latest
    volumes:
      - './db:/var/lib/mysql'
    ports:
      - '3306:3306'
    environment:
      MYSQL_ROOT_PASSWORD: farm
      MYSQL_DATABASE: farm
      MYSQL_USER: farm
      MYSQL_PASSWORD: farm

  www:
    depends_on:
      - db
    image: farmos/farmos:dev
    entrypoint: /bin/bash
    command:
      - -c
      - |
        set -e

        wait_db_ready() {
            while { ! exec 3<>/dev/tcp/db/3306; } > /dev/null 2>&1; do sleep 0.1; done
        }

        if ! [ "$$(ls -A /var/www/html/)" ]; then
          echo "farmOS webroot not detected. Copying from pre-built codebase in the Docker image."
          cp -rp /tmp/www/. /var/www/html
          wait_db_ready
          drush site-install farm --debug --yes --locale=us --db-url=mysql://farm:farm@db/farm --site-name=Test0 --account-name=root --account-pass=test install_configure_form.update_status_module='array(FALSE,FALSE)'
        fi

        wait_db_ready

        find /farmos_dev_modules -maxdepth 1 -mindepth 1 -execdir sh -c '\
            echo Symlinking and drush enabling $$(basename {}) && \
            ln -sTf /farmos_dev_modules/$$(basename {}) /var/www/html/sites/all/modules/$$(basename {}) && \
            drush --root=/var/www/html pm-enable --yes $$(basename {})' \;

        exec docker-entrypoint.sh apache2-foreground
    volumes:
      - './www:/var/www/html'
      - './php-custom.ini:/usr/local/etc/php/conf.d/php-custom.ini'
      - '../farmos_wfs:/farmos_dev_modules/farmos_wfs'
    ports:
      - '80:80'
    environment:
      XDEBUG_CONFIG: remote_host=172.22.0.1

```

PR #388, opened 2020-12-15
Minor fixes after converting config fields to bundle fields.

paul121:
Just a few small fixes/improvements after https://www.drupal.org/project/farm/issues/3183372 was merged.

Biggest issue is that the animal sex values are displaying as `M` and `Male`:
animal_sex_values

I translated some other string labels & descriptions that were missing from various log field definitions. Thought I'd test out the localization and see how things are looking - after enabling Spanish, I'm surprised that not very much is being translated... even "Male" and "Female". Is this because we haven't released a version of farmOS 2.x yet? So technically no translations have been built on Drupal.org for this release of the project?

### DIFF
--- a/modules/asset/animal/src/Plugin/Asset/AssetType/Animal.php
+++ b/modules/asset/animal/src/Plugin/Asset/AssetType/Animal.php
@@ -61,7 +61,7 @@ class Animal extends FarmAssetType {
       ],
     ];
     foreach ($field_info as $name => $info) {
-      $fields[$name] = \Drupal::service('farm_field.factory')->bundleFieldDefinition($info);
+      $fields[$name] = $this->farmFieldFactory->bundleFieldDefinition($info);
     }
     return $fields;
   }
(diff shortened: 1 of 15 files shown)


mstenta:
Nice improvements - thanks @paul121! Merged.

Thanks for catching the missing translations - I got the asset bundles but forgot the log bundles in my pass. :-)

> I translated some other string labels & descriptions that were missing from various log field definitions. Thought I'd test out the localization and see how things are looking - after enabling Spanish, I'm surprised that not very much is being translated... even "Male" and "Female". Is this because we haven't released a version of farmOS 2.x yet? So technically no translations have been built on Drupal.org for this release of the project?

Good question. It does appear that localize.drupal.org translations are tied directly to releases, so that may be why.

I do see "Male" and "Female" have translations in 1.x: https://localize.drupal.org/translate/languages/es/translate?project=farm&status=0&release=all&search=Female&author&context=all&limit=10&sid=0

This would be good to follow-up on in https://www.drupal.org/project/farm/issues/3159886

PR #390, opened 2020-12-17
Add bundle fields to views data

paul121:
Just noticed that after the recent changes bundle fields are not included in the entity's views data. Entities that have bundle plugin fields need to set their `views_data` handler to the contrib `EntityViewsData` class. There are a few ways we could do this.. here is my thought process:

We *could* set this value for asset and plant entities directly in the original entity annotation. The problem is that we can't do this for the log entity since it is contrib. But even then, I'm not sure if this is the best place to do it since the original entity definitions never depend on the entity bundle plugin features. This is added with `farm_entity`.

So... It would be simple enough & make good sense to set the `views_data` handler for all our entity types in `farm_entity_entity_type_build()` right after we set the `bundle_plugin` handler class. The problem with this is that logs have a custom `LogViewsData` class that specifies a couple views plugins to use (for correct sorting of timestamp + id), so logs will need a new class that extends contrib `EntityViewsData`, but also sets the same views plugins.

Rather than creating a `Farm*ViewsData` class for each entity type, I created a single `FarmEntityViewsData` class that can be used for all entity types. This centralizes logic to one place, which might be a downside (?), but also allows us to re-use any views logic that is needed across different entity types which might be needed?


paul121:
Ohh actually, on second thought, it might be better to just change the `log` views data in `hook_views_data_alter()`, which `farm_entity` already implements. This keeps things a bit simpler by not requiring any new classes.

The first two commits should be sufficient, leaving the third as an example of another approach we could take.

mstenta:
Wow great find and sleuthing @paul121 ! I think your analysis and solution sound like the right approach, and I agree we don't need the extra class.

We also may not need to replicate the Log views data stuff. Right now the Log module already provides its own `LogViewsData` class and sets that as its `views_data` handler, so ideally we could leave that untouched (in case anything else changes or is added in the future upstream), and only set the `views_data` handler for Asset and Plan entities.

However, the Log module's `LogViewsData` class extends from `Drupal\views\EntityViewsData`, not `Drupal\entity\EntityViewsData`, so I think it would suffer from the same issue. But that seems like a reasonable change to make upstream. I don't think it would affect anything for the Log module to inherit from `Drupal\entity\EntityViewsData` instead.

I will open an issue upstream and talk to @pcambra about it.

mstenta:
I created an issue with a patch in the Log module upstream: https://www.drupal.org/project/log/issues/3190371

Let's see if those tests pass, and check with @pcambra to see if there are any other concerns.

mstenta:
Patch committed upstream!

I'm going to add a commit that changes this branch so that it only applies the `EntityViewsData` class to Asset and Plan entities. I will keep all three of your commits @paul121 for the historical context they provide.

mstenta:
Oops hmm... it appears that this is breaking some tests related to the new `location` field. Will need to debug this...

https://github.com/mstenta/farmOS/runs/1626497059?check_suite_focus=true

paul121:
Great! Didn't know if that was a reasonable change for the log module, but it's certainly the easiest.

But hmm... that error must be related to computed fields?

mstenta:
Yea appears so:

`Drupal\Core\Entity\Sql\SqlContentEntityStorageException: Column information not available for the 'location' field.`

mstenta:
Ah probably because we are adding the `location` field via `hook_views_data_alter()`... maybe that needs to be updated.

mstenta:
Did some digging but didn't get very far with it - need to move on to some other things today.

I pushed my wip branch to my fork here: https://github.com/mstenta/farmOS/tree/2.x-bundle_field_views_data

paul121:
The issue seems to be related to the contrib `EntityViewsData` handler adding reverse relationships for all `entity_reference` fields. The logic in the `addReverseRelationships` method assumes that the entity reference data will always come from a DB table. Since our `location` field is a computed entity reference, this isn't stored in a DB table.

The `addReverseRelationships` method is called here: https://git.drupalcode.org/project/entity/-/blob/8.x-1.x/src/EntityViewsData.php#L59

What's interesting is that the reverse relationships aren't created for fields defined by the `bundle_plugin` class. The bundle plugin fields are collected after reverse relationships are created. To confirm this, I moved the `location` from a base field, to instead being provided by the `FarmAssetType` bundle plugin, and it worked!

Still, it seems like there is an issue with computed entity reference fields. I'm hesitant to create an issue in the Entity project since it seems like this would have come up before...but I can't find any existing issues that mention this. I'm also curious if there is anything we can add to our field definition that would exclude it from having a reverse relationship created, but looking at the code nothing comes to mind.

One potential fix is to prevent computed entity reference fields from getting a reverse relationship. This fixes the problem, but won't provide the reverse relationship in views (that's probably OK for our use case). One line change:

```php
/**
   * {@inheritdoc}
   */
  public function getViewsData() {
    $data = parent::getViewsData();

    $this->tableMapping = $this->storage->getTableMapping();
    $entity_type_id = $this->entityType->id();
    // Workaround for core issue #3004300.
    if ($this->entityType->isRevisionable()) {
      $revision_table = $this->tableMapping->getRevisionTable();
      $data[$revision_table]['table']['entity revision'] = TRUE;
    }
    // Add missing reverse relationships. Workaround for core issue #2706431.
    $base_fields = $this->getEntityFieldManager()->getBaseFieldDefinitions($entity_type_id);
    $entity_reference_fields = array_filter($base_fields, function (BaseFieldDefinition $field) {
      //return $field->getType() == 'entity_reference';
      // ADD THIS LINE. Exlcude computed fields.
      return !$field->isComputed() && $field->getType() == 'entity_reference';
    });
    $this->addReverseRelationships($data, $entity_reference_fields);
    // Add views integration for bundle plugin fields.
    // Workaround for core issue #2898635.
    if ($this->entityType->hasHandlerClass('bundle_plugin')) {
      $bundles = $this->getEntityTypeBundleInfo()->getBundleInfo($entity_type_id);
      foreach (array_keys($bundles) as $bundle) {
        $field_definitions = $this->getEntityFieldManager()->getFieldDefinitions($entity_type_id, $bundle);
        foreach ($field_definitions as $field_definition) {
          if ($field_definition instanceof BundleFieldDefinition) {
            $this->addBundleFieldData($data, $field_definition);
          }
        }
      }
    }

    return $data;
  }
 ```

Whatcha think @mstenta? If my findings seem correct...

1) Temporary fix is to provide the field as a bundle plugin (not ideal)
2) If this is a bug, we could patch the Entity module?
3) And/or provide our own EntityViewsData class that just overrides the `getViewsData` method?

paul121:
After chatting with @mstenta we decided a patch to the Entity module would be easiest for now. This way we don't have to create our own EntityViewsData handler class, and the contrib Log module can continue using its own. If we do need reverse relationship functionality on the `location` field, we would likely need to create our own views handler to provide that functionality.

I opened this [issue](https://www.drupal.org/project/entity/issues/3190436) and included a patch with the same workaround as described above.

The test for this commit passed on my branch: https://github.com/paul121/farmOS/runs/1628136604?check_suite_focus=true

mstenta:
Thanks @paul121 - this looks good to me! I'll follow that upstream issue you created to see what happens there. But this works for us in the meantime :-)

PR #391, opened 2020-12-20
Pin XDebug to version 2.9.8

symbioquine:
* Why: XDebug >= 3 has many breaking changes which don't
  make sense to deal with in farmOS 7.x-1.x.


mstenta:
Thanks @symbioquine! I agree.

PR #394, opened 2020-12-24
Issue collecting bundle field definitions

paul121:
Just a little bug with the operator that collects all the bundle field definitions from `hook_farm_entity_bundle_field_info()`. Right now only the *last* bundle's fields are being included. I don't think we've noticed this issue because we don't implement this hook in farmOS core (yet??). I added a failing test case that demonstrates this: https://github.com/paul121/farmOS/actions/runs/441518917

The current tests create the *same* fields for both the `test` and `test_override` log bundles which means this bug doesn't cause any problems. It's a bit complicated to replicate in the tests since it requires that we create fields that are only included in some bundles. To simplify this a little, I added a field called `{bundle_name}_specific_field` to each log type, where the field name is unique to the bundle name.

One other change: I expanded `testFarmFields()` to test that `hook_farm_entity_bundle_field_info()` can provide `FieldDefinitions` as well as `FieldStorageDefinitions`. A little confusing, but the hook is invoked in both the `getFieldStorageDefinitions()` and `getFieldDefinitions($bundle)` methods of `FarmEntityBundlePluginHandler`, so this should be possible. This part tripped me up when writing the tests... note that this bug is only affecting the `FieldStorageDefinitions` for logs since the `getFieldDefinitions` method requires the `$bundle` name as an argument, and means the hook is only invoked one time.

### DIFF
--- a/modules/core/entity/tests/src/Kernel/FarmEntityFieldTest.php
+++ b/modules/core/entity/tests/src/Kernel/FarmEntityFieldTest.php
@@ -92,26 +92,36 @@ class FarmEntityFieldTest extends KernelTestBase {
    */
   public function testHookFarmEntityBundleFieldInfo() {
 
-    // Test log field storage definitions.
-    $fields = $this->entityFieldManager->getFieldStorageDefinitions('log');
-    $this->assertArrayHasKey('test_hook_base_field', $fields);
-    $this->assertArrayHasKey('test_hook_bundle_field', $fields);
-    $this->assertArrayHasKey('test_hook_bundle_test_specific_field', $fields);
-    $this->assertArrayHasKey('test_hook_bundle_test_override_specific_field', $fields);
+    // Get the log field storage definitions.
+    $log_storage_definitions = $this->entityFieldManager->getFieldStorageDefinitions('log');
 
-    // Test log field definitions for test logs.
+    // Test that 'test_hook_bundle_field' has a storage definition.
+    $this->assertArrayHasKey('test_hook_bundle_field', $log_storage_definitions);
+
+    // Test fields definitions for the 'test' log type.
     $fields = $this->entityFieldManager->getFieldDefinitions('log', 'test');
-    $this->assertArrayHasKey('test_hook_base_field', $fields);
     $this->assertArrayHasKey('test_hook_bundle_field', $fields);
-    $this->assertArrayHasKey('test_hook_bundle_test_specific_field', $fields);
-    $this->assertArrayNotHasKey('test_hook_bundle_test_override_specific_field', $fields);
 
-    // Test log field definitions for test_override logs.
+    // Test fields definitions for the 'test_override' log type.
     $fields = $this->entityFieldManager->getFieldDefinitions('log', 'test_override');
-    $this->assertArrayHasKey('test_hook_base_field', $fields);
     $this->assertArrayHasKey('test_hook_bundle_field', $fields);
-    $this->assertArrayHasKey('test_hook_bundle_test_override_specific_field', $fields);
-    $this->assertArrayNotHasKey('test_hook_bundle_test_specific_field', $fields);
+
+    // Get all log bundles.
+    /** @var \Drupal\Core\Entity\EntityTypeBundleInfoInterface $entity_type_bundle_info */
+    $entity_type_bundle_info = $this->container->get('entity_type.bundle.info');
+    $bundles = $entity_type_bundle_info->getBundleInfo('log');
+
+    // Test that all log types have a bundle specific field.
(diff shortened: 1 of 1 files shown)


mstenta:
Good find @paul121 !

Was about to merge this but then realized I made some changes to `FarmEntityFieldTest.php` that I haven't pushed to `origin` which conflict with your changes. We'll have to sort those out... probably not too hard - just need to review.

mstenta:
Ok I think I've got this sorted.

I had reorganized the tests in `FarmEntityFieldTest.php` pretty significantly in my `2.x` branch (which hadn't been pushed to `origin`) so it was easier for me to start with that and rebase/refactor your test on top.

I ended up simplifying your test a bit (and by that I mean flattened them down to explicit assertions without any looping) just to help wrap my head around things. (And maybe it's better to keep tests simpler in this way in general for readability and future merge-ability?)

I kept your commit credit. But I want to have you double-check before I merge this, to make sure I didn't miss or misinterpret anything. I've tacked the two new commits onto my `2.x` branch in my fork.

Here is the one that adds the tests: https://github.com/mstenta/farmOS/commit/da3b031f73a56dccef92473ef2f863241c0bf260

Here is the failing test (with the test, but without your fix commit): https://github.com/mstenta/farmOS/runs/1621552700?check_suite_focus=true

And here is the passing test (with your fix commit): https://github.com/mstenta/farmOS/runs/1621545244?check_suite_focus=true

Let me know if that looks good to you and I will push my `2.x` to `origin`.

paul121:
Thanks for doing the dirty work @mstenta!

The test looks good, I think it covers all the bases. Also like that it is in a separate test case. On that note, should we remove the `$this->assertArrayHasKey('test_hook_base_field', $fields);` assertions since that field isn't provided by `hook_farm_entity_bundle_field_info` ? (I know I had included that in the loops/assertions I added)

> I ended up simplifying your test a bit (and by that I mean flattened them down to explicit assertions without any looping) just to help wrap my head around things. (And maybe it's better to keep tests simpler in this way in general for readability and future merge-ability?)

Yea, looking back at the test I wrote the looping is a little complicated. It shouldn't be checking for `test_hook_bundle_field` in the loop over all log types, since that field is only added to specific log types.

But for `test_hook_bundle_{bundle}_specific_field` in particular, I think the looping actually improves the readability and maintainability going forward.  Based on our logic that creates the field, this test should really be testing that *all log types* have a "bundle specific field", which it is not currently doing.

Since we're testing two different pieces of logic we created within the hook, I suggest we use both approaches for this test case. Maybe (untested):

```php
    // Get the log storage definitions.
    $log_storage_definitions = $this->entityFieldManager->getFieldStorageDefinitions('log');

   // 'test_hook_bundle_field' should have a storage definition.
    $this->assertArrayHasKey('test_hook_bundle_field', $log_storage_definitions);

    // The test log type should have the `test_hook_bundle_field'.
    $fields = $this->entityFieldManager->getFieldDefinitions('log', 'test');
    $this->assertArrayHasKey('test_hook_bundle_field', $fields);

    // The test_override log type should have the 'test_hook_bundle_field'.
    $fields = $this->entityFieldManager->getFieldDefinitions('log', 'test_override');
    $this->assertArrayHasKey('test_hook_bundle_field', $fields);

    // Get all log bundles.
    /** @var \Drupal\Core\Entity\EntityTypeBundleInfoInterface $entity_type_bundle_info */
    $entity_type_bundle_info = $this->container->get('entity_type.bundle.info');
    $bundles = $entity_type_bundle_info->getBundleInfo('log');

    // All log types should have a bundle specific field.
    foreach (array_keys($bundles) as $bundle) {
      $fields = $this->entityFieldManager->getFieldDefinitions('log', $bundle);
      $field_name = 'test_hook_bundle_' . $bundle . '_specific_field';

      // Assert field storage definition.
      $this->assertArrayHasKey($field_name, $log_storage_definition);

     // Assert field definition for the bundle.
      $this->assertArrayNotHasKey($field_name, $fields);
    }
```

> Here is the failing test (with the test, but without your fix commit): https://github.com/mstenta/farmOS/runs/1621552700?check_suite_focus=true

It looks like that test failed for a different reason (failed to build the image?), but the passing test LGTM!

mstenta:
Good points @paul121 and thanks for taking a look! Would you mind making the changes you suggest and force push?

mstenta:
Looks good @paul121 - thanks!

I merged this and then added a commit to restore the test for `test_hook_base_field` - but in `testHookEntityBaseFieldInfo()` instead of `testHookFarmEntityBundleFieldInfo()` (where it belongs). These tests are all looking good now IMO. :-)

PR #398, opened 2021-01-07
Standardize method and property visibility to be protected.

paul121:
Found myself double checking if I should be declaring a property as `private` or `protected` a couple times.. seems like something we can standardize in farmOS core. Overall, it seems like `protected` is the best bet for the sake of inheritance and general extendibility.

https://www.drupal.org/project/drupal/issues/2727011
https://www.php.net/manual/en/language.oop5.visibility.php


mstenta:
LGTM - thanks @paul121 !

PR #399, opened 2021-01-12
Make tests much faster with paratest to run tests in parallel

symbioquine:
**Why?** Faster testing leads to faster iteration and faster
overall development.

**How?** Leverage paratest which is - mostly - a drop-in for the
invocation of phpunit.

**Note:** The tests for this PR won't pass until the prerequisite change in https://github.com/farmOS/composer-project/pull/2 is merged.

The most recent 10 builds for farmOS 2.x took an average of 307 seconds for the phpunit step;

```
311s
284s
314s
250s
329s
308s
272s
325s
333s
345s
```

The following two builds demonstrate that it will pass and is faster;

* `189s` https://github.com/symbioquine/farmOS/runs/1690088541?check_suite_focus=true
* `171s` https://github.com/symbioquine/farmOS/runs/1690400928?check_suite_focus=true

Some background on earlier versions of this PR can be found at https://github.com/farmOS/farmOS/pull/378

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -33,7 +33,9 @@ jobs:
       # so waiting for that lets us know it is safe to start the tests
         run: until [ -f ./www/test-runner-container-fs-ready ]; do sleep 0.1; done
       - name: Run PHPUnit tests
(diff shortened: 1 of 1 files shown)


mstenta:
Merged - thanks @symbioquine !

PR #401, opened 2021-01-15
Improvements to GH action

paul121:
- Use the correct HEAD ref for pull requests. On pull requests `GITHUB_REF` = `refs/pull/{number}/`, which sets `FARMOS_VERSION` to an incorrect branch name.
- Specify the SHA that triggered the event so that the tests are run against that specific commit. Currently, tests checkout the HEAD of the target branch, which is problematic if multiple commits are pushed before the test has finished building the docker images. This is useful for pushing a commit that should result in a failing test, followed by a commit that fixes the test.

Some references just because this was a little confusing to figure out:
- Environment variables: https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables#default-environment-variables
- Contexts and expressions: https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#about-contexts-and-expressions

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -15,15 +15,23 @@ jobs:
     steps:
       - name: Checkout the repository
         uses: actions/checkout@v2
-      - name: Set FARMOS_VERSION environment variable
-        run: echo "FARMOS_VERSION=${GITHUB_REF:11}" >> $GITHUB_ENV
+      - name: Set FARMOS_VERSION for push event.
+        if: ${{ !github.event.pull_request }}
+        run: |
+          echo "FARMOS_VERSION=${GITHUB_REF:11}" >> $GITHUB_ENV
+          echo "FARMOS_REPO=${GITHUB_REPOSITORY}" >> $GITHUB_ENV
+      - name: Set FARMOS_VERSION for pull request event.
+        if: ${{ github.event.pull_request }}
+        run: |
+          echo "FARMOS_VERSION=${GITHUB_HEAD_REF}" >> $GITHUB_ENV
+          echo "FARMOS_REPO=${{ github.event.pull_request.head.repo.full_name }}" >> $GITHUB_ENV
       - name: Build farmOS 2.x Docker image
-        run: docker build --build-arg FARMOS_REPO=https://github.com/${GITHUB_REPOSITORY} --build-arg FARMOS_VERSION=${FARMOS_VERSION} -t farmos/farmos:2.x docker
+        run: docker build --build-arg FARMOS_REPO=https://github.com/${FARMOS_REPO} --build-arg FARMOS_VERSION=${FARMOS_VERSION} -t farmos/farmos:2.x docker
       # This builds the dev Docker image using the specified FARMOS_VERSION,
       # but notably it does NOT override the default PROJECT_VERSION, so the
       # farmOS Composer project 2.x branch is always used.
       - name: Build farmOS 2.x-dev Docker image
-        run: docker build --build-arg FARMOS_REPO=https://github.com/${GITHUB_REPOSITORY} --build-arg FARMOS_VERSION=${FARMOS_VERSION} --build-arg WWW_DATA_ID=33 -t farmos/farmos:2.x-dev docker/dev
+        run: docker build --build-arg FARMOS_REPO=https://github.com/${FARMOS_REPO} --build-arg FARMOS_VERSION=${FARMOS_VERSION} --build-arg WWW_DATA_ID=33 -t farmos/farmos:2.x-dev docker/dev
       - name: Create docker-compose.yml
         run: cp docker/docker-compose.testing.yml docker-compose.yml
       - name: Start containers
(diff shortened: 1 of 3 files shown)


MorganBorman:
@paul121 I have a hunch, but have to ask; this fixes the tests for all PR after this (assuming they have no other issues) but isn't expected to pass the checks itself right?

.github/workflows/run-tests.yml
```diff
@@ -15,15 +15,19 @@ jobs:
...
+        run: echo "FARMOS_VERSION=${GITHUB_HEAD_REF}#${GITHUB_SHA}" >> $GITHUB_ENV
       - name: Build farmOS 2.x Docker image
-        run: docker build --build-arg FARMOS_REPO=https://github.com/${GITHUB_REPOSITORY} --build-arg FARMOS_VERSION=${FARMOS_VERSION} -t farmos/farmos:2.x docker
+        run: docker build --build-arg FARMOS_REPO=https://github.com/$GITHUB_REPOSITORY --build-arg FARMOS_VERSION=$FARMOS_VERSION -t farmos/farmos:2.x docker
       # This builds the dev Docker image using the specified FARMOS_VERSION,
       # but notably it does NOT override the default PROJECT_VERSION, so the
       # farmOS Composer project 2.x branch is always used.
       - name: Build farmOS 2.x-dev Docker image
-        run: docker build --build-arg FARMOS_REPO=https://github.com/${GITHUB_REPOSITORY} --build-arg FARMOS_VERSION=${FARMOS_VERSION} --build-arg WWW_DATA_ID=33 -t farmos/farmos:2.x-dev docker/dev
+        run: docker build --build-arg FARMOS_REPO=https://github.com/$GITHUB_REPOSITORY --build-arg FARMOS_VERSION=$FARMOS_VERSION --build-arg WWW_DATA_ID=33 -t farmos/farmos:2.x-dev docker/dev
```
MorganBorman, line 30: Is changing the way these variables are referenced relevant to the change as a whole?

symbioquine: Is changing the way these variables are referenced relevant to the change as a whole?

symbioquine:
@paul121 I have a hunch, but have to ask; this fixes the tests for all PR after this (assuming they have no other issues) but isn't expected to pass the checks itself right?

paul121:
@symbioquine , no I think these changes should pass the tests as well! I thought using `GITHUB_HEAD_REF` would be enough, but apparently not... hard to debug for the PR event, sorry for the noise!

.github/workflows/run-tests.yml
paul121: No, I thought it might just happen to fix it. Referencing these docs: https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables#about-environment-variables

paul121:
Okay, another change required for the PR was to specify the correct repo name. I *think* `GITHUB_REPOSITORY` is returning `farmOS/farmOS` for pull request events, where this should actually be the user's repository eg: `paul121/farmOS`. Using `${{ github.event.pull_request.head.repo.full_name }}` seems to grab the correct repo name.

But then adding the commit SHA (via `GITHUB_SHA`) broke the tests for pull_requests, but IS working for push events. There must be a different SHA to use for the pull_request event..

paul121:
Trying `${{ github.event.pull_request.head.sha }}` for the pull request event, and leaving `${GITHUB_SHA}` for the push event.

paul121:
That worked! Now to confirm that this is all truly using the correct repos + commits... the tests should fail for this last commit.

mstenta:
Excellent - thanks for digging into this @paul121 - merged!

mstenta:
Doh! Tests are failing in the origin repo now. :-(

https://github.com/farmOS/farmOS/runs/1723823552?check_suite_focus=true

paul121:
Hmm. Seems like Composer failed to get/checkout the repo. Previously when the tests broke at this step it was because the SHA was incorrect. In this failing test it tried `2-x#caf36c721443a95745e6edd112a0ab4c6174328c`  but this seems to be the correct SHA for the latest commit on 2.x, yea?

paul121:
Same error pushing to the 2.x branch on my fork. Is anything special about this branch?

https://github.com/paul121/farmOS/actions/runs/494521054

paul121:
Well, didn't realize this but the Composer docs say explicit commit references work, but are not actively supported:

https://getcomposer.org/doc/04-schema.md#package-links

> Note: This feature has severe technical limitations, as the composer.json metadata will still be read from the branch name you specify before the hash. You should therefore only use this as a temporary solution during development to remediate transient issues, until you can switch to tagged releases. The Composer team does not actively support this feature and will not accept bug reports related to it.

Maybe we're just hitting an edge case with the `2.x` branch name? I just pushed another commit here that ignores explicit commit references on the 2.x branch. This works for the 2.x branch in my fork, but doesn't seem like a perfect solution.

Overall it would be a nice "convenience" feature, but maybe we should just revert my commit that added the SHA reference

mstenta:
> Overall it would be a nice "convenience" feature, but maybe we should just revert my commit that added the SHA reference

I've reverted this commit until we can figure it out.

paul121:
I opened #403 to consider adding the commit reference later on.

PR #400, opened 2021-01-13
Use a test matrix to run tests and site-install against MariaDB/SQLite/Postgres in parallel

symbioquine:
**Why?** Detect if things break on specific DBMS' and have ongoing proof
that the top DB options work with farmOS.

PR #402, opened 2021-01-19
Fix DB_URL part of test matrix

symbioquine:
Tested at https://github.com/symbioquine/farmOS/runs/1728732854?check_suite_focus=true


mstenta:
Looks good thanks @symbioquine !

For future reference, here is the previous PR that added these variables: https://github.com/farmOS/farmOS/pull/400

PR #406, opened 2021-01-31
Fix issue with broken links in asset map popup

symbioquine:
**Why?** When assets don't have a URL alias and have an id
greater than 1000, the link to those assets was getting rendered
with a comma in the map popup, which breaks the link - often leading
to the wrong asset like `/farm/asset/1` instead of `/farm/asset/1234`.


mstenta:
Thanks @symbioquine !

PR #408, opened 2021-02-04
Auto create taxonomy term references

paul121:
- Support the `auto_create` and `auto_create_bundle` fields via FarmFieldFactory
- Auto create animal asset animal_type terms
- Auto create plant asset plant_type terms
- Auto create input log material terms

I think these are the only taxonomy term references we want to auto create? Maybe include the plant_type.crop_family?

Log category was a `list_item` field in 1.x, but right now it is an auto complete field in 2.x. This is a little confusing, because it seems like you could auto create a lot category. Shall we change this?

### DIFF
--- a/modules/asset/animal/src/Plugin/Asset/AssetType/Animal.php
+++ b/modules/asset/animal/src/Plugin/Asset/AssetType/Animal.php
@@ -26,6 +26,7 @@ class Animal extends FarmAssetType {
         'description' => $this->t("Enter this animal asset's species/breed."),
         'target_type' => 'taxonomy_term',
         'target_bundle' => 'animal_type',
+        'auto_create_bundle' => 'animal_type',
         'required' => TRUE,
         'weight' => [
           'form' => -90,
--- a/modules/asset/plant/src/Plugin/Asset/AssetType/Plant.php
+++ b/modules/asset/plant/src/Plugin/Asset/AssetType/Plant.php
@@ -26,7 +26,7 @@ class Plant extends FarmAssetType {
         'description' => "Enter this plant asset's crop/variety.",
         'target_type' => 'taxonomy_term',
         'target_bundle' => 'plant_type',
-        'auto_create' => TRUE,
+        'auto_create_bundle' => 'plant_type',
         'required' => TRUE,
         'multiple' => TRUE,
         'weight' => [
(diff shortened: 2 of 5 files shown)


paul121:
I pushed two more commits to display the `log.category` options in an `options_select` form element. This seems like the correct option if we don't want to support auto creation of log categories.

This change also enables the default FieldFarmFactory `form` and `view` display options to be overridden. This seemed like the simplest & most robust way to support a specific `$form_display_options['type']`. In general it allows for more flexibility with future "one-off" edge cases like this. But perhaps this isn't something we want, and FarmFieldFactory should truly remain "opinionated"...

The other options I could think of:
- Add `form_display_type` as another optional variable in the FarmFieldFactory API. But different types will need different options?? So we would need `form_display_options` anyways?
- Add a check so if the taxonomy_term reference is limited to the `log_categories` bundle, use the `options_select` type instead. This way any field that references the `log_categories` terms would use an `options_select`, but right now we only have one field that references log categories, and it seems reasonable that a different `type` may be wanted in other contexts?

For reference, here is the difference between `options_select` and `entity_reference_autocomplete`: 
option_select_vs_autocomplete

mstenta:
Thanks @paul121 - this all looks good!

I just made one small change to the `auto_create` stuff... instead of requiring that the `auto_create_bundle` be declared, now it just requires that `auto_create` be set to `TRUE`, and then it will automatically use the `target_bundle` as the `auto_create_bundle`. There isn't a need to declare the same bundle name in two places IMO. They'll always be the same bundle.

> I pushed two more commits to display the `log.category` options in an `options_select` form element. This seems like the correct option if we don't want to support auto creation of log categories.

This makes sense too! Thanks for taking care of that at the same time!

paul121:
> I just made one small change to the auto_create stuff... instead of requiring that the auto_create_bundle be declared, now it just requires that auto_create be set to TRUE, and then it will automatically use the target_bundle as the auto_create_bundle. There isn't a need to declare the same bundle name in two places IMO. They'll always be the same bundle.

I agree, this does make sense.. with the assumption that only one `target_bundle` will be needed in these field definitions.

But I'm curious if we want to support multiple bundles via `target_bundles` instead. Maybe this isn't needed in core (yet?), but something to consider for contrib modules? Really depends if FarmFieldFactory is meant to be used by contrib. If that is the case, I think supporting multiple `target_bundles` would be a good feature.

We can tackle that later, but curious your thoughts @mstenta

mstenta:
Yea that was exactly my thought! We haven't had a need for it yet, but could expand it in the future if we needed to.

I am hoping that contrib can use `FarmFieldFactory` yes, so this might be worth considering sooner rather than later, if it would otherwise require a change to the API surface we're cobbling together with that class...

Or, we could just keep it simple, and if module needs to do more they can choose NOT to use `FarmFieldFactory`. We certainly don't need to re-invent the wheel (too much). ;-)

paul121:
Sounds good. It's a good point, we're only going to get closer to re-inventing the wheel with every feature we add

PR #409, opened 2021-02-14
Fix quantity queries in migration to avoid extraneous rows

symbioquine:
**Why?** Drupal 7 allows field re-use which means there may
be extra unrelated rows in addition to the ones pertaining
to the core logs.

More context available at https://farmos.discourse.group/t/symbioquines-2-x-migration-testing-log/660/11?u=symbioquine


mstenta:
Good find thanks @symbioquine !

PR #413, opened 2021-02-26
Add support for drush site:install profile arguments

paul121:
Currently all of the default modules are installed when using `drush site:install farm`. This would allow specifying only certain modules to be installed with a `farm_install_modules.modules` profile argument ([drush docs](https://drushcommands.com/drush-9x/site/site:install/))

Example usage:

```bash
drush site:install farm farm_install_modules.modules='["farm_observation", "farm_land", "farm_ui"]'
```

Since Drush only supports strings, numeric and NULL values to be supplied in this way, the `modules` variable is expected to be a JSON array of module names encoded as a string.

I didn't add any error handling here... an invalid string, JSON structure, or invalid module name might cause errors. But since Drupal has already been installed, I'm not sure if we can fail gracefully anyways. If this value is somehow being provided by users, extra care should be taken to ensure it is OK before being used.

### DIFF
--- a/farm.install
+++ b/farm.install
@@ -68,7 +68,7 @@ function farm_install_modules(array &$install_state) {
 
   // Assemble the batch operation for installing modules.
   $operations = [];
-  foreach ($modules as $module => $weight) {
+  foreach ($modules as $module) {
     $operations[] = [
       '_farm_install_module_batch',
       [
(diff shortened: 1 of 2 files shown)


mstenta:
This is handy - thanks @paul121! Merged.

PR #414, opened 2021-02-26
Simplify adding farmOS repository to composer.json.

paul121:
Just learned that composer has a command for updating the repositories in composer.json via the cli: https://getcomposer.org/doc/03-cli.md#modifying-repositories

The `repositories` section of `composer.json` can either be a simple array or an object with key/value pairs for each repository. Using this command changes the repositories to an object and adds a `farmos` key. Using keys makes it easier to modify the composer.json project file in scripts later on.

I was hoping we could modify the `requires: farmos/farmos {FARMOS_VERSION}` in a similar way, but the `composer config` command doesn't work here. We'd have to use `composer require` which ends up installing everything. Composer v2 has a `composer require --no-install` option that JUST builds the `composer.lock` file. That might work in the future.

### DIFF
--- a/docker/build-farmOS.sh
+++ b/docker/build-farmOS.sh
@@ -20,12 +20,14 @@ rm -rf project
 git checkout ${PROJECT_VERSION}
 git reset --hard
 
+# Add the farmOS repository to composer.json.
+composer config repositories.farmos git ${FARMOS_REPO}
+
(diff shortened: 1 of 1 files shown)


mstenta:
Cool! That's nifty! Thanks @paul121!

PR #415, opened 2021-03-02
farmOS 2.x Data Model Docs

mstenta:
This is a first draft of the official "farmOS Data Model" documentation for farmOS 2.x, to be merged into the farmOS 2.x branch and published at https://docs.farmOS.org.

A preview is available here: https://mstenta.github.io/farmOS/model/

(Note that the links in the content itself are broken because the preview is being shown in a subdirectory. These will work once this is merged into docs.farmOS.org and the links become absolute. In the meantime, you can navigate using the menu on the left of the content.)

This pull request contains two commits, but only the first one is intended to be merged. The second is simply a temporary commit to publish the preview.

On the February 2021 [farmOS monthly call](https://farmos.org/community/monthly-call/) we decided to publish this first as a pull request, to get reviews and comments from the wider community before making it "official". Feel free to leave comments, make suggestions, ask questions, etc! I will "fixup" the draft commit for small changes, and we can consider larger changes in follow-up commits (and/or follow-up PRs). We will leave this PR open for a period of time (TBD on the March monthly call perhaps)

**This is just documentation** - it is not a proposal for new features or ideas, just a description of the existing farmOS data model, nearly identical to that of farmOS 1.x, which up until now has only been documented in code and informally in the API docs, user guide, issue queues, forum topics etc. This documentation serves to formalize it and make it available in a way that is easy to share. It is intentionally non-technical, and does not include specific machine names or API implementation details (the API docs will cover that) - although it does make the same distinction between "attributes" and "relationships" that JSON:API does because I think that's helpful.

Note that while this includes *most* of the current data model (including an overview of the entity types, base fields, logic, and an introduction to the idea of conventions), there are still a few things missing, which I plan to add in the near future. These include:

- **Inventory fields and logic** - Although farmOS 1.x had inventory for animals, 2.x will be more generalized. We are working on this now ahead of the 2.0 alpha release.
- **Data streams** - farmOS 1.x had a "Sensor Listener" module, which provided some custom data architecture for storing sensor data streams. This is more formalized in 2.x, and will be merged soon.

In addition to those two (which were technically part of the farmOS 1.x data model), there are two other changes coming soon, which will be unique to farmOS 2.x:

- **Quantity types** - See https://www.drupal.org/project/farm/issues/3196825
- **Time tracking** - See https://www.drupal.org/project/farm/issues/3196844

I have a `2.x-model-wip` (work in progress) branch (on top of this `2.x-model` branch) containing drafts of these changes, if anyone is interested: https://github.com/mstenta/farmOS/compare/2.x-model...mstenta:2.x-model-wip


mstenta:
There are a few other pieces we might want to think about, and I would like to hear thoughts from others...

**Entity metadata**

Assets, logs, plans, and quantities have some additional metadata available on them, which is not currently exposed in the farmOS UI, but is available via the API. This includes `created` and `changed` timestamps (which indicate when a record itself was created and last updated, separate from the timestamp of the event itself in the case of logs), `uid` (which indicates which user created the content, separate from who it is assigned to in the case of logs).

**Revisions**

In farmOS 2.x, many of the entity types support revisions, which means that every change to a record is stored as a separate revision, and you can see what the previous values were. This was not the case in farmOS 1.x. This is an important feature for historical and archival understanding of changes, but it's an open question whether or not this is technically part of the "farmOS data model". I'm curious to hear what others think.

**Bundle-specific fields**

The draft doc only includes "base fields" of assets, logs, plans, and quantities. It does not include bundle-specific fields for each asset/log type yet. I do think we should include these in the docs, but decided to make that a follow-up change. I'm also curious to hear what people think of the overall structure of the docs right now, and where it might make sense to put bundle-specific fields (at the bottom of the entity type pages perhaps?)

**Config entities**

There is some internal discussion about exposing "config entities" over the API, and so perhaps we should think through how they relate to the data model itself as well. This might be a bigger discussion...

docs/model/convention/index.md
```diff
@@ -0,0 +1,44 @@
...
+
+One of the longer term goals of farmOS is to be a platform that supports the
+collaborative development of these conventions over time. As new standards
+are developed and adopted in the community, they can be written into modules
+that provide different levels of "enforcement of" or "compliance to" these
+conventions.
+
+The data model, combined with established conventions, and tools like the
+[farmOS Aggregator](https://github.com/farmOS/farmOS-aggregator) can enable a
+wide variety of data sharing, aggregating, and reporting possibilities.
```
symbioquine: Maybe `%s/possibilities/use-cases/` ?

docs/model/convention/index.md
```diff
@@ -0,0 +1,44 @@
...
+
+## In practice
+
+Within an individual farmOS instance, you may develop your own conventions
+around naming your Assets. Or you may develop a standard operating procedure
+for certain types of common data entry, to ensure that it always goes into
+farmOS in the same manner and format.
+
+Across separate farmOS instances, developing conventions enables data to be
```
symbioquine: Maybe "developing [+shared] conventions"?

docs/model/index.md
```diff
@@ -0,0 +1,36 @@
...
+## Goals
+
+The goal of the farmOS data model is to represent and organize a wide variety
+of agricultural and ecological systems to allow for easy access and analysis.
+This ranges from small-scale garden records to industrial agriculture to
+national forestry management.
+
+farmOS can model data that was recorded via manual data entry, as well as data
+streams from sensors or other applications.
+
```
symbioquine, line 12: This discusses data ingress via manual and automated means. Should it also discuss data egress? e.g. "via a user friendly UX and documented canonical APIs"

docs/model/logic/location.md
```diff
@@ -0,0 +1,72 @@
...
+geometry as well as relation to other Assets.
+
+## Fixed Assets
+
+Some Assets are *fixed* in location. The geometry of these Assets is
+*intrinsic* to the Asset itself.
+
+Land is an example of an Asset with a fixed location. Although geography
+changes over time, farmOS models land in discreet period parcels with a defined
+geometry.
```
symbioquine: * "models land in discreet period parcels with a defined geometry" seems like a pretty confusing way to say this...
* `%s/discreet/discrete/`

docs/model/type/user.md
```diff
@@ -0,0 +1,8 @@
+# Users
```
symbioquine, line 1: Title and path of this document don't match how it is referred to earlier and in the text. i.e. It seems like this should either be consistently "user(s)" or "people", but not both interchangeably...

docs/model/index.md
mstenta: My intention here was to highlight that the farmOS data model itself covers both types of data (eg: logs and data streams). I wasn't really thinking in terms of "input" vs "output", although maybe it should be reworded to make this more clear? The model itself doesn't need to concern itself with input/output methods, IMO - that can be described in the user guide and API docs.

docs/model/logic/location.md
mstenta: Ha! I suppose it is confusing. I was particularly pleased with that phrase when I came up with it, to be honest. ;-)

Open to suggestions!

docs/model/type/user.md
mstenta: Good catch - I actually changed from "People" to "Users" at the last minute, but didn't do a thorough job updating it everywhere. I'll fix this.

docs/model/index.md
```diff
@@ -0,0 +1,36 @@
...
+## Data types
+
+Data is organized into high-level data types called entities. The two primary
```
jgaehring: I believe I've raised this concern in past conversations, but I wonder if it's wise to adopt Drupal terminology into our own model for farmOS. I'm worried this tight coupling with the Drupal implementation could lead to confusion down the road.

Specifically, I'm worried about how the farmOS concept of "entity" could evolve in the future, particularly if it's implemented in other languages and frameworks to mean something different from what a Drupal "entity" entails. Similarly, though probably less likely, Drupal's concept of "entity" could change over time. This could come to a head if and when we have farmOS entities that are not implemented as Drupal entities, and/or Drupal entities in farmOS which are not strictly farmOS entities. This could get confusing real fast.

Two alternatives that briefly come to mind: farm elements, or farm objects.

mstenta:
Thanks for taking the time to review @symbioquine ! I've force pushed the branch with your recommendations.

docs/model/index.md
mstenta: Thanks for pointing that out @jgaehring ! I agree.

For what it's worth, I intentionally named the section "Data types" to avoid using the word "entities", but in searching through it looks like I did use "entity" in a few places (old habits die hard). I'll push a follow-up commit that changes those.

> Two alternatives that briefly come to mind: farm elements, or farm objects.

How about "records"? I've used this as an alternative to entities in presentations I've given in the past. I agree "entity" is too Drupal-specific, and sounds more technical/complicated than it needs to. "Records" is nice also because it fits the domain model. "Record keeping"! :-)

mstenta: Similarly, there are three references to "bundles" that I can find, which are not really necessary. I'll remove those too.

mstenta:
Thanks for the suggestions @jgaehring! I pushed two commits that remove references to "entities" and "bundles", to keep Drupal-specific terminology out of the model docs.

docs/model/logic/location.md
symbioquine: Maybe something like;

> "Land is an example of an Asset with a fixed location. Although geography changes over time, land assets in farmOS should be considered a snapshot of the physical land they represent at the current time. This is useful because it provides a simplified model for referring to a logical piece of land and its geometry."

symbioquine:
@mstenta I think the force-pushes are a bit counter-productive in the context of reviewing such a large change. It makes it a bit hard to track what's already been reviewed vs what has recently changed...

mstenta:
Roger that @symbioquine - I will refrain from force-pushes during the review process. I'll make fixup commits for small changes and squash them before the final merge.

docs/model/logic/location.md
mstenta: I like that! Thanks! :-)

symbioquine:
Thanks @mstenta! To be fair your most recent force-push wasn't as bad since it retained new commits for the new changes. Either way works for me.

I was mostly reacting to;

image

That said, the review experience is way better with pushes like your most recent one;

image

docs/model/index.md
```diff
@@ -0,0 +1,36 @@
...
+**Quantities**, **Terms**, **Plans**, and **Users**.
+
+- [Assets](/model/type/asset)
+- [Logs](/model/type/log)
+- [Quantities](/model/type/quantity)
+- [Terms](/model/type/term)
+- [Plans](/model/type/plan)
+- [Users](/model/type/user)
+
+## Logic
```
jgaehring, line 26: Ok, I know I've caused a lot of fuss over "logic" in the past, but what if we rephrased this as "Operations" instead of "Logic"? So hear me out...

First, logic seems a little too broad to me in these examples. There is a _lot_ of logic in farmOS, but I feel like we're trying to encapsulate something much more specific here, with regards to the actual _data structures_. For reference, here's how Wikipedia defines a [Data structure](https://en.wikipedia.org/wiki/Data_structure):

> More precisely, a data structure is a collection of data values, the relationships among them, and the functions or __operations__ that can be applied to the data. [emphasis added]

A good example here, outside of farmOS, would be to think about the built-in data structures in a language, like numbers, strings and arrays. There are different operators for different data structures, for example:

```php
// Numeric operators
4 + 2;
4 - 2;

// String operator
"Hello" . "World";

// Array operator
array("a" => 42, "b" => "foo") + array("c" => "bar");
```

And these can get proportionally more complex as the data structures get more complex.

For farmOS, we could think of a _movement_ as a specific type of operation, where the operands are an asset and a log, with the result being a given change in the asset's location. Same would go for _group membership_.

I guess this is splitting hairs a bit, and it's not to say there still isn't logic involved in each operation, but I feel like it could help to frame the different types of operations, based on the data structures they operate on. This could eventually help inform future work on inventory, as specific operations on assets and quantities.

And despite the fact that "operation" can have a rather mathematical connotation to it, it has a second connotation to the running of a business that might make it more intuitive to non-programmers.

docs/model/type/term.md
```diff
@@ -0,0 +1,77 @@
...
+## ID
+
+Each Term will be assigned two unique IDs in the database: a universally unique
+identifier (UUID), and an internal numeric ID.
+
+The UUID will be unique **across** farmOS databases. The internal ID will only
+be unique to a **single** farmOS database. Therefore, the farmOS API uses UUIDs
+to ensure that IDs pulled from multiple farmOS databases do not conflict.
+Internally, farmOS modules use the internal IDs to perform CRUD operations.
+
```
jgaehring, line 35: We're missing the term type here. Related to my comments on vocabulary below.

docs/model/type/term.md
```diff
@@ -0,0 +1,77 @@
...
+additional relationships.
+
+### Standard relationships
+
+Relationships that are common to all Term types include:
+
+- Vocabulary
+- Parent
+
+#### Vocabulary
```
jgaehring: This is a little hard to understand. Like the phrase "entity", this seems a little too particular to the Drupal implentation. I wonder if, since the vocabulary is essentially equivalent to the term's type (which is also missing; see above comment), we can omit all references to the vocabulary in these docs, and just say that taxonomy terms require a type. This would be consistent with how logs, assets, etc all require a type, and how JSON:API will require a `type` for any resource.

docs/model/index.md
```diff
@@ -0,0 +1,36 @@
...
+
+The goal of the farmOS data model is to represent and organize a wide variety
+of agricultural and ecological systems to allow for easy access and analysis.
+This ranges from small-scale garden records to industrial agriculture to
+national forestry management.
+
+farmOS can model data that was recorded via manual data entry, as well as data
+streams from sensors or other applications.
+
+## Data types
```
jgaehring: Thinking about taxonomy term types has got me thinking about this heading again. I would suggest we say "Resources" instead. This is coupling us a little closer to JSON:API, but we're already borrowing their concept of `attributes` and `relationships`, and I think "resources" is a more broadly articulated term in web dev.

This would round out our terminology quite nicely, I think:

```
A log is a resource.
An asset is a resouce.
An activity is a resource type, specifically a log type.
The Drupal entity implements the farmOS resource.
The Drupal bundle implements the farmOS resource type.
```

I think that resolves any collisions we might otherwise have in naming.

jgaehring (commented):
Some comments re: vocabularies, taxonomy terms and types.

docs/model/index.md
symbioquine: Interesting... makes sense. Though I'd argue there's a bit of asymmetry here where you're saying a specific way the data model is designed to support recording data, but not saying anything a specific about how the data model is designed to support "easy access and analysis". Not a big deal either way, just pointing it out.

docs/model/index.md
mstenta: This is an interesting way to look at it.

> I feel like we're trying to encapsulate something much more specific here, with regards to the actual data structures.

Yes!

> There are different operators for different data structures

True - but this may be more atomic than what we're going for with this documentation section.

It isn't so much about describing what kind of operations can be performed (in some ways that's what the "Logs" section covers). Instead, it tries to describe "how the granular records themselves, when looked at together, bubble up greater meaning".

> For farmOS, we could think of a movement as a specific type of operation, where the operands are an asset and a log, with the result being a given change in the asset's location. Same would go for group membership.

Maybe these are two ways to say the same thing, but I feel like there is a key distinction here...

I think about farmOS as similar to a ledger. It has accounts (assets) and transactions (logs). The "balance" of the account is calculated based on the transactions, and this "balance" can be determined for any point in time. In farmOS, the "balance" is whatever "pseudo property" it is we're trying to determine, like location, group membership, inventory, etc.

"Moving" an animal can be considered an "operation" - but the act of performing that is accomplished by creating a log. And so the log itself *is* the operation. However, *how* the location is determined based on logs is the logic.

In order for logs and assets to have any meaning, we have to define how a set of logs can be looked at collectively to determine an asset's "state" at *any point in time*. Perhaps we don't emphasize that point enough in the current logic docs. The logic can be used to get an asset's "current" location, but it is the same exact logic used to get the asset's location at a given timestamp in the past or future.

So, I see "logic" as meaning "we've defined how to compute a given asset property at a given timestamp, and this is the logic that is followed to do that".

In some ways, this logic itself is a form of convention - one that has been baked into farmOS code. But ultimately, it is a definition for how to pull meaning out of the lower level records.

The main reason to document this logic is so that *other* implementations of the farmOS data model can replicate the same logic if they need to. If someone were to pull all their farmOS records down to JSON files in a directory, and wanted to compute the location of an asset from them, they would need to perform the same logic.

> it could help to frame the different types of operations, based on the data structures they operate on

Also worth noting that a single log can perform multiple "operations". You can have a log that assigns group membership to an asset, changes its location, and increments its inventory all at once. These are just three separate "balance calculations" that use the same log in their equations.

>  it has a second connotation to the running of a business

I suppose the other hesitation I have is that "operations" is used to mean something more domain specific in farming/agriculture, and is perhaps more analogous to what logs themselves represent (farming operations, events, things that were done on the farm, etc). I worry that a farmer coming to these docs and seeing "Operations" would expect to find information about the type of operations they can record in farmOS, which would actually be found under "log types".

---

All that said, I agree "logic" might be a bit opaque. What is it they say about the 2 hardest problems in programming? 1. cache invalidation, 2. **naming things**, and 3. off-by-one errors. :-)

docs/model/index.md
mstenta: Yea I see what you mean... maybe the word "record" is the problem, because it connotes "ingress". Maybe rephrasing to use the word "represent" is a better choice. Something like:

> farmOS can represent both high-level metadata records about assets and events on a farm, as well as raw time-series data from sensors or other applications.

Still needs some phrasing work... but the key point I'm trying to make is that there are two types of data. And the farmOS data model provides ways to represent both (or *will* once the data streams stuff is merged).

*Typically* the former is user-entered and the latter is automatically collected, but that's not always the case. So this particular sentence isn't so much about ingress/egress as it is about the types of data. Probably still needs some more wordsmithing to make that clear. :-)

mstenta: Another motivation for including this sentence is for newcomers who find this doc and are wondering things like "what can I record with farmOS?" / "does it *only* represent one or the other?" - I feel like this kind of question has come up a lot.

docs/model/index.md
mstenta: I thought about using the word "resources" originally too! I like it! Ultimately I decided to go with "Data types" instead, simply because I thought that would be a bit more approachable to a non-technical audience. I like "resource" personally, but if you aren't an API programmer it could be very confusing. Then layer on the fact that "resources" in farming means "stuff you use to farm" - I worry it would be confusing.

At the same time, I think "attributes" and "relationships" are easy to understand.

What about instead of "data types" we say "record types"?

mstenta: > What about instead of "data types" we say "record types"?

Following your pattern:

```
A log is a record.
An asset is a record.
An activity is a record type, specifically a log type.
The Drupal entity implements the farmOS record.
The Drupal bundle implements the farmOS record type.
```

docs/model/type/term.md
mstenta: Good catch! I will make a fixup commit...

docs/model/type/term.md
mstenta: Hmm yea I see your point. In this specific instance, though, if you open up a term in JSON:API, you will see "vocabulary" as one of the relationships. So it seems like we should keep that here, since this is a list of the relationships.

I agree that this "type" vs "vocabulary" piece in particular is an inconsistency, and does stem from Drupal's own internal legacy naming (the `taxonomy` module dates back to before entities/bundles entirely).

I did try to keep these lists of "Attributes" and "Relationships" identical to what one would see in the API, though. So this might just be a case where we compromise on that consistency a bit. 🤷

mstenta: Ah actually... looking more closely at the actual JSON for logs and terms I realize logs also have `log_type` as a relationship, and it is `vid` in the term. But, we don't include `log_type` in the "relationships" overview in `log.md`.

So maybe we should just omit `vocabulary` form `term.md` as well? That would make them consistent. But it would mean we are intentionally omitting those relationships. But like you said, maybe that's more of a Drupalism than a farmOS data model piece...

docs/model/type/term.md
mstenta: What do you think? https://github.com/farmOS/farmOS/pull/415/commits/9c254ee72e0f1bf4847c9647b9ca71197b256dcd

docs/model/type/term.md
mstenta: Here is a commit that does this: https://github.com/farmOS/farmOS/pull/415/commits/10c1da4883a1b9266257593b9d26821b7fa2e46d - think that makes sense?

docs/model/index.md
mstenta: Whaddya think? https://github.com/farmOS/farmOS/pull/415/commits/d5d8f10defd04bfd5cd4fa5ffa1dc49261e047ff

docs/model/index.md
jgaehring: Taking a step back, I admit that "operation" and "logic" are kind of interchangeable in my usage. But the main thrust of my argument might boil down to: "operation" implies logic that is applied to a particular set of data structures, whereas "logic" alone does not imply the associated data structures, or at least does not have as strong an implication.

> I think about farmOS as similar to a ledger. It has accounts (assets) and transactions (logs).

Right, but I think accounts and transactions are still plain data structures by themselves, which require an operation to determine the new balance (another data structure).

> And so the log itself is the operation.

I don't think it is though. A log can be either a data structure or an operation (again, "operation" here implying "logic that is applied to a particular set of data structures"), not both. I think we've come down pretty squarely on the side that a log is a data structure, insofar as we can send them over the wire as JSON. Perhaps it would help to clarify that operations can have multiple operands, so the movement operation has two operands: an asset to be moved, and a movement log.

> but the act of performing that is accomplished by creating a log

We might be saying the same thing here, but I'm trying to say that the "operation" is the verb in this sentence, whereas the  "log" is the noun (the "asset" is a noun, too, though unstated here). I don't know if that's helpful, so I won't get hung up on that analogy if it doesn't resonate with you.

> The main reason to document this logic is so that other implementations of the farmOS data model can replicate the same logic if they need to. If someone were to pull all their farmOS records down to JSON files in a directory, and wanted to compute the location of an asset from them, they would need to perform the same logic.

I think I'm fully in agreement with you on all of this, except I would just substitute the words "operation" for "logic". Just arguing that it's a more precise usage, which clarifies the meaning.

> Also worth noting that a single log can perform multiple "operations".

Again, I think we have to distinguish that the log is data; it doesn't _do_, it _is_. The "operation" is what's performed on the log. And multiple operations can be performed on the same data. I can _add_ 5 and 6 just as easily as I can _subtract_ 5 and 6.

Maybe it would be helpful to look at it this way. I'm thinking of the operation as the middle term in both of these expressions:

```
5 + 6
log MOVES asset
```

Where both `+` and `MOVES` are merely symbols representing a computational operation. Does that make sense?

So what I'm arguing is that "operation" is a better word for the computation performed by such an infix operator. We could call that logic, too, but the difference, in my mind, is that operation is more specific to the data structures it can be performed on. In other words, it implies that the `MOVES` operation above is only used with a log (on the left) and an asset (on the right). Both of these expressions would then be considered illegal:

```
quantity MOVES asset
asset MOVES log
```

Those expressions don't make sense, because `MOVES` is an operation specific to the data structures it operates on.

I think the biggest sticking point here is whether "logic" has the same implications. In my mind, it doesn't, but that could be a subjective thing. I'd be curious to know what other people think on this.

docs/model/type/term.md
jgaehring: If you send a term to the server with its `type` field set, but no `vid` under `relationships` it will accept it, right? But if the `type` is missing, it will reject it, even if the `vid` is there? I'm just assuming, haven't tried, but for me that would clinch it. `vid` and "vocabulary" as a whole is a detail of the Drupal implementation, which may be necessary for Drupal, but I don't see any reason why it would be necessary in other implementations.

docs/model/index.md
jgaehring: Sounds good to me!

docs/model/type/term.md
jgaehring: Ehhh, this still feels very drupally. :-) 

Why can't we just say

> ## Type
> Each term must have a type. farmOS includes modules that define the following term types, which it also refers to as vocabularies:

docs/model/type/term.md
mstenta: I believe that's true, yes.

mstenta: The `log_type` and `vid` relationships are sort of like "computed" relationships... is my understanding. They provide quick access to the actual resource that is referenced by the `type` string. So it's a helper, but not really "core data model" IMO. I think we can omit them.

docs/model/index.md
mstenta: > We might be saying the same thing here, but I'm trying to say that the "operation" is the verb in this sentence, whereas the "log" is the noun (the "asset" is a noun, too, though unstated here).

Yea I think we're in sync. :-)

"Operation" might just be too loaded of a term. I do appreciate how it tries to relate these ideas to mathematical operations though - I certainly agree there. But does it confuse more than it helps, I wonder?

The goal of having a dedicated "logic/operations" section, and not just adding info to the "record types" themselves, is to explicitly highlight that this "logic/operations" is a separate layer. I think both terms serve that - but I worry that "operations" might be initially confusing.

I'd love to hear what others think of these ideas!

---

> `log MOVES asset`

I like the simplicity of describing things this way - but I'm curious how this can also capture the deeper parameters involved. Specifically, a log only `MOVES` an asset "if it has a status of 'done' and has a timestamp less than or equal to the reference point (which by default is the current timestamp)". It's also possible to ask "where was an asset in the past" and "where will an asset be in the future" (which includes "pending" logs).

I like the idea of describing some of these things as simple "operations" but I think the "logic" *includes* operations, but also includes decisions about how those operations will be interpreted, if that makes sense?

In my mind the "farmOS data model logic" defines not only the allowable operations, but also how we are deciding to interpret the results of these equations. So I think there is a level of meaning that goes beyond just the mathematical analogy? But maybe I'm overthinking it. :-)

---

Perhaps another point I'd like to highlight, though, is that the "logic" section does have some other stuff in it that doesn't fit neatly into the "operations" idea. For example, at the bottom of the Location logic page:

> When a Log is saved without a geometry, and it references locations that have geometries, the locations geometries will be copied to the Log.

Now, maybe that piece in particular is out of place there... but I think little things like that should also be documented, and the "logic" section seemed like the right place for it. But perhaps that's a separate thread. :-)

docs/model/type/term.md
mstenta: Split the difference? https://github.com/farmOS/farmOS/pull/415/commits/ad913b0f9600a456051329f65cfef1f58459334a :-)

I don't think "vocabulary" itself is "Drupally", and in fact helps to convey the fact that these are ontologies that the user can add terms to. So in this case, I think making a tighter conceptual link between "term type" and "vocabulary" is good.

mstenta:
@jgaehring @symbioquine (and others) curious what your thoughts are on this idea:

> **Entity metadata**
> 
> Assets, logs, plans, and quantities have some additional metadata available on them, which is not currently exposed in the farmOS UI, but is available via the API. This includes `created` and `changed` timestamps (which indicate when a record itself was created and last updated, separate from the timestamp of the event itself in the case of logs), `uid` (which indicates which user created the content, separate from who it is assigned to in the case of logs).

vital-agro:
Interesting stuff, guys.

I'm coming at this from a less technical perspective and my first impression is that this information needs to be communicated to a non-technical audience as well. As architects of the system, all of this may make sense to you, but to a newcomer without any programming experience it may seem like a foreign language. What I'm trying to emphasize is that you should be able to "explain this to me like I'm 5 years old" or more precisely, "I don't speak your language". 

For example, what are config entities and bundle-specific fields? how are these useful to a farmer? Is there more intuitive terminology?

Maybe I don't correctly understand the purpose of this documentation, but my sense is that the user interface should communicate this information without the user needing to read a thesis on the data model. 

I also agree with @symbioquine comment: "Though I'd argue there's a bit of asymmetry here where you're saying a specific way the data model is designed to support recording data, but not saying anything a specific about how the data model is designed to support "easy access and analysis". Detailed record keeping is valuable only to the degree that the data can be leveraged with easy access. Perhaps this is more the role of the UI/UX but for example, can I export data streams and activity records in a csv file or pdf? I hear this request from growers on a weekly basis.

mstenta:
Hi @vital-agro - thanks for the feedback! Great to have your perspective on this. :-)

> Maybe I don't correctly understand the purpose of this documentation, but my sense is that the user interface should communicate this information without the user needing to read a thesis on the data model.

This is a great point - and we should be more clear about what the *purpose* of this "data model documentation" is. The intended audience is actually a bit different than "normal users" and "developers" - I would put it in between those. It is for someone who wants to better understand how farmOS "models reality". :-)

This is very useful to understand if you are going to be developing farmOS modules or using its API (and up until now the "API Docs" were pretty much the same as the "data model docs" in 1.x -  we didn't have an organized/dedicated document that just described the abstract ideas behind the model), and it's also helpful for normal users of farmOS, but generally the user interface itself, as well as the "User guide" (a separate document), will aim to communicate "how to use farmOS" for normal day-to-day tasks. But there is a third audience: people who won't necessarily be using farmOS themselves, and won't be developing themselves, but want to understand whether or not the farmOS data model is a good fit for integration with their own systems.

The other purpose of this document is to emphasize that "the farmOS data model is itself independent of the farmOS software implementation"! This is an important concept - because it means that we're trying to think about this beyond our own implementations (eg the farmOS codebase, or the Field Kit codebase). @gbathree is thinking through how to use this model outside of ANY implementation - as a common format that can be used to translate from one platform to another, without any actual farmOS database/server involved. (see https://gitlab.com/groups/our-sci/software/-/epics/8)

So one way to think about it is: the farmOS data model is the "[platonic ideal](https://en.wikipedia.org/wiki/Platonic_idealism)", and the farmOS codebase is the "[reference implementation](https://en.wikipedia.org/wiki/Reference_implementation)" of that model, along with UI/UX/DX features for working with it. And so the purpose of this documentation is to describe the model itself, separate from the implementation.

> Perhaps this is more the role of the UI/UX but for example, can I export data streams and activity records in a csv file or pdf? I hear this request from growers on a weekly basis.

Yes - this is something that should be covered in the User Guide, because it would be specific to the implementation itself (in this case the farmOS server UI - but perhaps we'll have similar features in Field Kit one day as well - or other systems that just push/pull via API!).

> For example, what are config entities and bundle-specific fields? how are these useful to a farmer? Is there more intuitive terminology?

Ha! Yea... I bring these up to highlight other ideas that we might want to include in the model itself. Ultimately, we probably won't call them "config entities" and "bundle-specific fields" - because those are somewhat implementation specific terms ("fields that are on specific log types" would be a better way to describe what "bundle specific fields" are, for instance).

Hope that helps to summarize what the purpose of this is. Appreciate the feedback and ideas! :-)

docs/model/index.md
jgaehring: > But does it confuse more than it helps, I wonder?

Yea, I think that's what really remains to be seen.

> I like the simplicity of describing things this way - but I'm curious how this can also capture the deeper parameters involved. Specifically, a log only `MOVES` an asset "if it has a status of 'done' and has a timestamp less than or equal to the reference point (which by default is the current timestamp)". It's also possible to ask "where was an asset in the past" and "where will an asset be in the future" (which includes "pending" logs).

I think part of what I like about this framing is it forces us to be explicit about the deeper parameters involved. If, instead of just one log we're talking about a collection of logs that must all be taken together to get the current location of an asset, or its location at a given point in time, we have to say so. And we have to distinguish that operation from the operation that only requires one log. Also, we could express the operation on many logs in terms of the operation on one log, which gives a level of composition.

> but I think the "logic" includes operations

I mean, I could argue that operations includes logic (and I am). I also think operations can be composed of other operations. But I'm going to reserve my judgment on that for the moment.

> In my mind the "farmOS data model logic" defines not only the allowable operations, but also how we are deciding to interpret the results of these equations.

I don't think this rules out the idea of operations. I didn't address it in my previous example, but operations have results. And there are lots of ways that can happen, which is why I avoided it, b/c I didn't want to get bogged down in that. But you can have something like an expression which evaluates to the result:

```lisp
(+ 4 5) ;; evaluates to 9
(MOVE log asset) ;; evaluates to an asset w/ new geometry
```

You can have an assignment type of operation which mutates the original variable holding that value:

```go
$x := 4;
$x += 5; // $x is now 9
$asset := create_asset();
$asset.move($log); // mutates the asset's geometry 
```

Anyways, I think those are implementation details, but I still think the core idea of operations requires that those results be explicit in a way that logic doesn't imply. Even if the result is that nothing changed:

```js
x += 0;
```

At least, I struggle to think of a concept of "operation" that does _not_ imply a result, but perhaps that's just me.

> Now, maybe that piece in particular is out of place there... but I think little things like that should also be documented, and the "logic" section seemed like the right place for it.

I agree! And again, I think conceptualizing these as operations only reinforces that requirement. If the documentation fails to specify all the results of an operation on its operands, that's a failure of the documentation, and I think becomes more clear when we think of these as operations instead of logic.

docs/model/type/term.md
jgaehring: This reminds me of the issue of "computed" values, and how to distinguish them... :thinking:

jgaehring:
> curious what your thoughts are on this idea:
>
>> Entity metadata

Generally, @mstenta, I think this metadata should not be part of the data model. Where are you planning to add this?

mstenta:
> Generally, @mstenta, I think this metadata should not be part of the data model. Where are you planning to add this?

No plans to add it - but wanted to see what others thought. farmOS itself does not display this information anywhere itself, but it may be helpful to one day. And perhaps the same would therefore be true in Field Kit or other contexts.

mstenta:
I've merged the `2.x-quantity_types` branch (https://www.drupal.org/project/farm/issues/3196825) into farmOS 2.x, and pushed a simple commit that updates the Quantity data model doc page to mention it. More will be added once some of the new quantity type branches are added (eg: `2.x-price_quantity` https://www.drupal.org/project/farm/issues/3196836).

rburghol:
Appreciating the dialog here, it is giving me an understanding of the aspirations of the abstract implementation and is motivating me.  

FYI the sample docs page has dead links once you click on the sub-topics like "Assets" so reviewing that way is hampered. Any recommendations on how to view the docs for a deep dive -- other than clicking thru the directory tree? (Clicking is fine if that's how it is)

mstenta:
Welcome @rburghol! - Yea there are some broken links in the content itself (because they are "absolute" links to work once this is merged into docs.farmos.org), but you can also navigate using the menu on the left hand side - that should give you access to everything when you're browsing in this preview: https://mstenta.github.io/farmOS/model/

Thanks for asking! I should have mentioned that in the description. I'll add a little note now...

rburghol:
Ahh - thanks! That dropdown works great (viewing on mobile)

mstenta:
Per the discussion on today's [farmOS monthly call](https://farmos.org/community/monthly-call/), we decided to leave this open for comment until Friday of next week. If there aren't any objections before then, we'll merge it.

That said, I expect this to be a living document, so we can always continue iterating on it with future pull requests. :-)

mstenta:
Thanks for all the feedback everyone! I've merged this "first draft" into the farmOS 2.x branch! :tada: 

@jgaehring I'm still very interested in following up on the "logic/operations" thoughts above, as well as some of the other pieces I outlined in my first comment - but these can be considered separately IMO. These documents will undoubtedly evolve into the future. :rocket:

mstenta:
For future reference, the Data Model docs are now public here: https://docs.farmos.org/model/

PR #418, opened 2021-03-21
Simplify composer require farmos/farmos version.

paul121:
A follow up from #414. Composer 2 adds the `composer require --no-install` option which resolves dependencies and generates a lock file, but doesn't perform the install. This allows the `build-farmOS.sh` script to still accept arguments that are passed to `composer install`.

### DIFF
--- a/docker/build-farmOS.sh
+++ b/docker/build-farmOS.sh
@@ -31,7 +31,9 @@ composer config repositories.farmos git ${FARMOS_REPO}
 # it is a branch, and prepend it with "dev-".
 # Otherwise FARMOS_VERSION is a valid semantic versioning string. We assume
 # that it is a tagged version and require that version.
-if [[ ! "${FARMOS_VERSION}" =~ ^(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-((0|[1-9][0-9]*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*)(\.(0|[1-9][0-9]*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*))*))?(\+([0-9a-zA-Z-]+(\.[0-9a-zA-Z-]+)*))?$ ]]; then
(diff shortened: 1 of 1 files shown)


mstenta:
Seems that this causes tests to fail when the branch is `2.x`: https://github.com/mstenta/farmOS/runs/2165458384?check_suite_focus=true

```
  [InvalidArgumentException]                                          
  Could not find package farmos/farmos in a version matching dev-2.x  
```

paul121:
Ah of course! `2.x` is not a valid semantic version. I thought we could get away without another edge case but requesting `2.x` should end up requiring `2.x-dev`.

This is one thing that has confused me with composer... the `dev-` prefix and `-dev` suffix. But investigating this helped clear it up:

> When branch names look like versions, we have to clarify for composer that we're trying to check out a branch and not a tag. In the above example, we have two version branches: v1 and v2. To get Composer to check out one of these branches, you must specify a version constraint that looks like this: v1.x-dev. 

https://getcomposer.org/doc/articles/versions.md#branches

paul121:
Hmm the tests are failing for another reason.. something must be down. But I think this should work.

```
curl: (22) The requested URL returned error: 504 Gateway Time-out
The command '/bin/sh -c apt-get update && apt-get install -y libgeos-dev   && curl -fsSL 'https://git.osgeo.org/gitea/geos/php-geos/archive/1.0.0.tar.gz' -o php-geos.tar.gz   && tar -xzf php-geos.tar.gz   && rm php-geos.tar.gz   && (     cd php-geos     && ./autogen.sh     && ./configure     && make     && make install   )   && rm -r php-geos   && docker-php-ext-enable geos' returned a non-zero code: 22
```

mstenta:
> This is one thing that has confused me with composer... the `dev-` prefix and `-dev` suffix.

Same here... I still need to look it up to remember how that works.

> Hmm the tests are failing for another reason.. something must be down.

I triggered the tests to re-run, and they are passing now. I also merged into my fork's `2.x` branch and those tests passed as well. Good to go! :-)

Thanks @paul121 - merged!

PR #419, opened 2021-04-06
Update farm_livestock_weight.module

Skipper-is:
Added default categories to the farm_livestock_weight_set function, and the ability to add other categories to the logs.

### DIFF
--- a/modules/farm/farm_livestock/farm_livestock_weight/farm_livestock_weight.module
+++ b/modules/farm/farm_livestock/farm_livestock_weight/farm_livestock_weight.module
@@ -191,7 +191,7 @@ function farm_livestock_weight_dlwg($asset) {
  * @return \Log
  *   Returns the log that was created.
  */
-function farm_livestock_weight_set($assets, $weight, $units, $timestamp = REQUEST_TIME, $done = TRUE) {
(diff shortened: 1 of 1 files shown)


mstenta:
Thanks @Skipper-is ! Nice and simple... :-)

One small coding standards nit-pick: needs spaces between the function arguments in `farm_quantity_log_create()`.

My only other hesitation in merging this is that I'm worried by including the "Animals" category as a default we are mixing "conventions" with lower-level API functions, and I think those should be kept separate. Instead, maybe we just add the optional category parameters like you did, but leave the array blank by default, and leave the decision up to higher-level components - like your weight quick form.

We are not going to be including the "default categories" in farmOS 2.x like we did in 1.x (I don't think), and therefore "Animals" won't be a standard term. I added those default terms early on, but in hindsight I don't think they are a good fit for everyone, and tend to clutter the list of categories. (Although glad to hear this one works for you!)

This idea of "conventions" (https://docs.farmos.org/model/convention/) has been helping to formalize how we think about these decisions, and how to separate low-level code from high-level features, with the intention of keeping farmOS itself relatively unopinionated.

Long-winded thoughts for such a simple PR, I know... hope they make sense. :-)

mstenta:
Now that 2.x is the default branch, and seeing how the animal weight module has been split out to contrib (drupal.org/project/farm_animal_weight) I think I'm going to close this. But please feel free to open a new issue in the 2.x version of that module if you want to suggest changes!

PR #420, opened 2021-04-08
Add support for Pi

frakman1:
Used `buildx` option to `docker build` to support following architectures:

linux/amd64
linux/arm/v7
linux/arm64

Base Image can be found here:
https://hub.docker.com/r/frakman1/farmos-base-pi

Dev Image built based on Base Image above can be found here:
https://hub.docker.com/r/frakman1/farmos-dev-pi

Then used that in docker-compose-development-pi.yml

### DIFF
--- a/docker/docker-compose-development-pi.yml
+++ b/docker/docker-compose-development-pi.yml
@@ -1,0 +1,24 @@
+version: '2'
+services:
+  db:
+    image: jsurf/rpi-mariadb
+    volumes:
+      - './db:/var/lib/mysql'
+    ports:
+      - '3306:3306'
+    environment:
+      MYSQL_ROOT_PASSWORD: farm
+      MYSQL_DATABASE: farm
+      MYSQL_USER: farm
+      MYSQL_PASSWORD: farm
+
+  www:
+    depends_on:
+      - db
+    image: frakman1/farmos-dev-pi:latest
+    volumes:
+      - './www:/var/www/html'
+    ports:
+      - '80:80'
+    environment:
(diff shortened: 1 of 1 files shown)


docker/docker-compose-development-pi.yml
```diff
@@ -0,0 +1,24 @@
...
+    environment:
+      MYSQL_ROOT_PASSWORD: farm
+      MYSQL_DATABASE: farm
+      MYSQL_USER: farm
+      MYSQL_PASSWORD: farm
+
+  www:
+    depends_on:
+      - db
+    image: frakman1/farmos-dev-pi:latest
```
symbioquine, line 18: I'm not sure it's a great idea to point out at random other Docker Hub namespaces like this and line 4.

Surely there's an officially supported mariadb image for the pi?

For farmOS itself why not also update the Github action workflows for farmOS to do the buildx build and push to the farmOS Docker Hub namespace automatically?(Obviously you'd need to test it with your own account until the change got merged and could use secrets only the main repo has.)

mstenta:
Thanks for the contribution @frakman1! Appreciate the effort and exploration you put into this!

My preference would be to focus our efforts on this: https://www.drupal.org/project/farm/issues/3203129

That way we can provide an "official" image for ARM architectures in our Docker Hub organization, and a separate Pi-specific template file wouldn't be necessary.

If you have time/interest in sharing some of your lessons in that issue, that would be great! Or if you are interested in starting the balls rolling on the GitHub Actions changes, please feel free! There are some challenges to think through, so please don't hesitate to ping one of us in chat to coordinate if you do. :-)

With all that in mind, I'm going to close this PR.

PR #421, opened 2021-04-23
Entity reference revisions patch for compatibility with IEF

pcambra:
Include patch from https://www.drupal.org/project/entity_reference_revisions/issues/2667748#comment-14070377

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -57,6 +57,9 @@
             "drupal/entity_browser": {
                 "Issue #3160482: Error with UntrustedCallbackException": "https://www.drupal.org/files/issues/2021-03-02/error-with-untrustedcallbackexception-3160482-5.patch"
             },
+            "drupal/entity_reference_revisions": {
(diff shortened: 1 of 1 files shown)


mstenta:
Oops sorry @pcambra this slipped between the cracks!

If I remember correctly, this is specifically affecting the work in https://github.com/mstenta/farm_nfa/tree/2.x, where an Inline Entity Form is being used edit Log entities, along with Quantity entities (which are referenced on Logs via Entity Reference Revisions) - is that correct?

If that's the only place this has come up so far, and isn't affecting anything in farmOS core itself (yet), then I'm inclined to say we just include the patch in the `composer.json` for the `farm_nfa` project, and maybe wait to either a) add the patch to farmOS's `composer.json` we decide to do something similar (I love the form in `farm_nfa` and look forward to exploring how we can take a similar approach in farmOS 2.x plans!), or b) perhaps Entity Reference Revisions will merge the patch upstream itself.

Do you think that makes sense?

pcambra:
Makes sense @mstenta ! I've just pushed https://github.com/mstenta/farm_nfa/commit/0888de4d481cc0e3e95100e76eade8f09e3452cd to farm_nfa instead and we'll see how it plays out :)

Issue #396, opened 2021-01-03
Warning divion by Zero 


maoneid:
**Describe the bug**
Warning Error, division by zero 

**To Reproduce**
1. Go to animal item view and detail 'http://localhost/farm/farm/animal/sapi'
2. See error
or
1. see weight report http://localhost/farm/farm/asset/1/weight

run on PHP 7.4, Ubuntu 20
Warning: Division by zero in farm_livestock_weight_dlwg() (line 137 of /var/www/html/farm/profiles/farm/modules/farm/farm_livestock/farm_livestock_weight/farm_livestock_weight.module).

maoneid:
It is happened when I input weght more than twice a day. so timediff_days = 0 .

`$timediff_days = round($timediff/86400,2);

      // Calculate dlwg.
      $dlwg_value = round($weight_difference/$timediff_days,3);`

mstenta:
Good find @maoneid ! If you think there's an obvious fix based on your investigation please feel free to suggest it or start a pull request.

Otherwise we'll try to remember to fix this when we port the weight module to farmOS 2.x in the next few months.

AugustoHA:
@maoneid could you add more details about this problem, including some input examples? 
I manage to reproduce it, but only when both timestamps were equal in day and minute (EG: 04/10/2021 10:00). Otherwise, it didn't show the warning.

So if that is the problem, a simple but not elegant solution is to verify if the timediff_days equal zero and then set to one.

mstenta:
Closing old issues, see https://github.com/farmOS/farmOS/pull/422#issuecomment-1004316168

PR #422, opened 2021-04-27
Fixes "Warning divion by Zero" : https://github.com/farmOS/farmOS/issues/396

alexliu4:
If there are multiple inputs entered on a single day, $timediff_days will round down to 0. This causes a division-by-0 error when $dlwg_value is calculated. This change makes it so that a $timediff_days value of 0 is rounded up to 1 in order to avoid the previously mentioned issue.

### DIFF
--- a/modules/farm/farm_livestock/farm_livestock_weight/farm_livestock_weight.module
+++ b/modules/farm/farm_livestock/farm_livestock_weight/farm_livestock_weight.module
@@ -132,6 +132,11 @@ function farm_livestock_weight_dlwg($asset) {
       // Calculate time difference.
       $timediff = $latest_log->timestamp - $previous_log->timestamp;
       $timediff_days = round($timediff/86400,2);
+      
+      // When multiple inputs are entered on a single day, round $timediff_days up to 1.
+      if ($timediff_days == 0) {
+        $timediff_days = 1;
+      }
 
       // Calculate dlwg.
       $dlwg_value = round($weight_difference/$timediff_days,3);


modules/farm/farm_livestock/farm_livestock_weight/farm_livestock_weight.module
```diff
@@ -132,6 +132,11 @@ function farm_livestock_weight_dlwg($asset) {
       // Calculate time difference.
       $timediff = $latest_log->timestamp - $previous_log->timestamp;
       $timediff_days = round($timediff/86400,2);
+      
+      // When multiple inputs are entered on a single day, round $timediff_days up to 1.
+      if ($timediff_days == 0) {
+        $timediff_days = 1;
+      }
```
symbioquine, line 139: Kudos for jumping in and fixing bugs you've found!

Take all this with a grain of salt since I don't personally use the weight tracking in farmOS right now and acknowledge there's a lot to be said for ensuring that the page doesn't crash as a first level of bug fix...

---

Besides the division by zero error, it seems like this logic was a little broken to start with and I suspect this change introduces as many issues as it fixes.

For example, integer division by 86400 isn't a great way to bucket timestamps into days because timestamps which are a few seconds apart can get counted as being different days while timestamps which are 23h59s apart can get counted as being the same.

From a different perspective, honoring time zones to bucket the weight logs into days wouldn't necessarily produce the right results either. With that approach, a log at 11 PM would be counted as being a "day later" than a log at 6 AM even though they're only 7 hours apart.

Similarly, this change would produce pretty odd behavior. Suppose you had a measurement for a flock of animals 3 days ago and had just entered new measurements for them today, but accidentally weighed and logged two measurements for one of the animals. All the other animals would show the correct weight gain, but the one with a duplicate log would presumably just show the variance in the scale's measurements and claim that was the "daily weight gain" for that animal.

---

It's more work, and I don't know what the stakeholders care about with this stat, but I would argue a better strategy would be to have a window heuristic where this logic averages datapoints which are close together and looks for two datapoints (or averages thereof) which are a day or more apart, then uses floating point division to calculate the "daily weight gain" from those datapoints.

I'm curious whether @mstenta can comment on who the stakeholders are and if they can be looped in on what the expected behavior is?

AngeloR: Does this actually need to bucket things into "days" or can it just continue grouping in 24hr blocks with an update to the UI to indicate this? 

Days is, as you mention, pretty arbitrary and if we can just say "24hrs" that would match what the code actually does until a better solution is found?

mstenta:
Sorry for the late reply on this... I haven't been closely involved with the development of the weight report, so I'm not super familiar with the logic (and it's been a long time since I've looked at it).

That said, we are going to need to port this feature to farmOS 2.x in the near future - which might be a good opportunity to review in more detail, and potentially take it in new directions. I opened a new issue for that, and added a comment back to this PR for further review: https://www.drupal.org/project/farm/issues/3220601

mstenta:
Hi all - this PR hasn't seen much action, and farmOS 2.x is now the default branch. The weight report module has been split out to contrib (https://drupal.org/project/farm_animal_weight) so I'm going to go ahead and close this. We may want to open a companion issue in the new repo for that report, to make sure the issue gets fixed in the new 2.x branch (ping @paul121).

PR #423, opened 2021-05-10
2.x Gatsby docs

jgaehring:
__WARNING: This should not be merged until https://github.com/jgaehring/farmos.org-2.x has been deployed to docs.farmos.org, because it removes the MkDocs and GitHub Pages configuration files, thus breaking the current build process.__

Just wanted to submit this now for early feedback on how to adjust the farmOS docs in preparation for farmOS.org 2.x. In order to test if Frontmatter headers would work in the markdown files, I needed to create a fork I could source it from anyways.

More to come...


jgaehring:
From the [work proposal](https://farmos.discourse.group/t/work-proposal-scaffold-a-new-gatsby-site-for-farmos-org-2-x-docs/699/10?u=jgaehring) on the forum:

> I wound up using the `mkdocs.yml` file after all, while still retaining the parsing logic that generates the navigation tree based on the file structure alone if there is not a `mkdocs.yml`. Basically I realized there was more information encoded into one of the `nav` YAML property than I first ascertained, and it wasn't worth trying to encode all that into some combination of Frontmatter and `index.yml` files to make that work. At least for now. I think it's still a good idea worth revisiting once we've tried out using this with other source repos for a while.
>  
> For now, I'm gonna just close that PR though. Maybe we won't need the favicon in the source repo any more, and we'll probably replace the GitHub Pages action with something to trigger Netlify builds, but it seems worthwhile just starting fresh for those changes.

PR #424, opened 2021-05-18
Add optional group membership selection to quick birth form

ghost:
First PR here :)  Happy to receive comments on mistakes, coding guidlines, etc.

I initially considered adding a group selection field with the mother's group pre-selected instead of having a checkbox + ajax displaying the groups without any pre-selection. I eventually chose the latter solution because deselecting groups is error-prone when there are many groups.

I've done a few tests, everything seems to be working properly. I've left a TODO in the code though - not sure if the group entries should be validated or if it's taken care of by farm_asset_load_multiple()

### DIFF
--- a/modules/farm/farm_livestock/farm_livestock.farm_quick.birth.inc
+++ b/modules/farm/farm_livestock/farm_livestock.farm_quick.birth.inc
@@ -150,7 +150,6 @@ function farm_livestock_birth_form($form, &$form_state) {
       '#multiple' => TRUE,
       '#options' => farm_group_options(),
       '#required' => FALSE,
-      '#multiple' => TRUE,
     );
   }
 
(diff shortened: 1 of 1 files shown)


mstenta:
Thanks @taradiddles - this looks great!

I like this approach of defaulting to the mother's group, but showing a checkbox that lets you assign groups manually. Simple and flexible. :-)

> I've left a TODO in the code though - not sure if the group entries should be validated or if it's taken care of by farm_asset_load_multiple()

Yea that should be fine - the `'#options' => farm_group_options(),` will ensure that only valid options are available in the `select` widget, and Drupal protects against overriding that client-side, so we should be good. You can go ahead and remove the `@TODO` line.

The only minor coding standard nitpick I see is that `farm_livestock_birth_form_group_ajax()` should have its own docblock comment, rather than sharing with `farm_livestock_birth_form_ajax()` above it.

Other than that, this looks good!

ghost:
Hi Mike
I've removed the `TODO` and fixed the comment block ; I've also renamed `farm_livestock_birth_form_ajax()` to make it clear it applied to the children form/div (that function isn't used anywhere else so that should be fine - but I can revert the change if needed).

mstenta:
Looks good @taradiddles ! Merged! Thanks again. :-)

PR #425, opened 2021-06-04
Upgrade to farmOS-map 2.x

symbioquine:
**Note:** This change must be merged after we tag the first farmOS-map 2.x release, otherwise the built won't be able to find the right version in NPM.

**Why?** farmOS-map 2.x provides a slightly cleaner API and more modular/extensible behavior loading.

PR #426, opened 2021-06-06
2.x Create FarmMapPrototype element to support JS map creation

symbioquine:
**Note:** This is intended to be merged to the `2.x` branch after https://github.com/farmOS/farmOS/pull/425. Unfortunately, the diff includes the changes from that PRs since [stacked pull requests](http://bentrengrove.com/blog/2020/7/8/how-to-stack-prs-in-github) don't work [across forks](https://github.com/google/ground-android/issues/93) on Github.

**Why?** Currently the extension system for `FarmMap` couples the code that determines which behaviors should be loaded
with the code that instantiates a specific instance of farmOS-map.

With these changes, I've refactored out the map-type/settings/event logic into a common trait which is used both by the existing `FarmMap` `RenderElement` and a new `FarmMapPrototype` element.

This new `FarmMapPrototype` element does all the same steps, except it doesn't include the `farm-map` class which is used by the JS to find the maps to instantiate.

Further, I've refactored the code from the `farm_map.js` Drupal behavior such that a factory method is exposed which other JS can leverage to create maps using all the same boiler-plate that is needed to wire map settings exposed from Drupal into new map instances.

I also had to decouple the mechanism by which settings are accessed in JS given a map instance. Prior to these changes, that assumed the id of the map element was always that same as the key by which those settings
are accessed.

All that together allows for map creation like this;

```js
const mapPrototypeElement = document.getElementById('my-tool-map-prototype');

const mapElement = mapPrototypeElement.cloneNode();
mapElement.removeAttribute('id');

myParentContainer.appendChild(mapElement);

const mapInstance = Drupal.behaviors.farm_map.createMapInstance(mapElement, mapElement, 'my-tool-map-prototype', {});
```

### DIFF
--- a/modules/core/map/js/farmOS.map.behaviors.geofield.js
+++ b/modules/core/map/js/farmOS.map.behaviors.geofield.js
@@ -1,9 +1,10 @@
 (function () {
   farmOS.map.behaviors.geofield = {
     attach: function (instance) {
-      instance.edit.wktOn('featurechange', function(wkt) {
-        console.log('here!');
-        document.querySelector('#' + instance.target).parentElement.querySelector('textarea').value = wkt;
+      instance.editAttached.then(() => {
+        instance.edit.wktOn('featurechange', function(wkt) {
+          document.querySelector('#' + instance.target).parentElement.querySelector('textarea').value = wkt;
+        });
       });
     },
 
--- a/modules/core/map/js/farmOS.map.behaviors.geofield.js
+++ b/modules/core/map/js/farmOS.map.behaviors.geofield.js
@@ -3,7 +3,7 @@
     attach: function (instance) {
       instance.editAttached.then(() => {
         instance.edit.wktOn('featurechange', function(wkt) {
-          document.querySelector('#' + instance.target).parentElement.querySelector('textarea').value = wkt;
+          instance.map.getTargetElement().parentElement.querySelector('textarea').value = wkt;
         });
       });
     },
--- a/modules/core/map/js/farm_map.js
+++ b/modules/core/map/js/farm_map.js
@@ -2,62 +2,17 @@
   Drupal.behaviors.farm_map = {
     attach: function (context, settings) {
 
-      // Get the units.
-      let units = 'metric';
-      if (!!drupalSettings.farm_map.units) {
-        units = drupalSettings.farm_map.units;
-      }
-
-      // Build default options.
-      const defaultOptions = {
-        units,
-        interactions: {
-          onFocusOnly: true
-        },
-      };
       context.querySelectorAll('.farm-map').forEach(function (element) {
 
-        // Only create a map once per element.
-        if (element.getAttribute('processed')) return;
-        element.setAttribute('processed', true);
+          // Only create a map once per element.
+          if (element.getAttribute('processed')) return;
+          element.setAttribute('processed', true);
 
-        element.setAttribute('tabIndex', 0);
-        const mapId = element.getAttribute('id');
-        const mapOptions = { ...defaultOptions, ...drupalSettings.farm_map[mapId].instance};
-        const instance = farmOS.map.create(mapId, mapOptions);
-        context.querySelectorAll('.ol-popup-closer').forEach(function (element) {
-          element.onClick = function (element) {
-            element.focus();
-          };
-        });
+          element.setAttribute('tabIndex', 0);
 
-        // If the map is rendered as part of a form field, update the map size
-        // when the field's visible state changes,
-        const formWrapper = element.closest('div.form-wrapper');
-        if (formWrapper != null) {
-          const formWrapperObserver = new MutationObserver((mutations) => {
-
-            // Only update the map size if the wrapper was previously hidden.
-            if (mutations.some((mutation) => { return mutation.oldValue.includes('display: none')})) {
-              instance.map.updateSize();
-            }
-          });
-
-          // Observe the style attribute.
-          formWrapperObserver.observe(formWrapper, {
-            attributeFilter: ["style"],
-            attributeOldValue: true
-          })
-        }
-
-        // If the map is inside a details element, update the map size when
-        // the details element is toggled.
-        const details = element.closest('details');
-        if (details != null) {
-          details.addEventListener('toggle', function() {
-            instance.map.updateSize();
-          });
-        }
+          const mapId = element.getAttribute('id');
+          const mapInstanceOptions = {};
+          Drupal.behaviors.farm_map.createMapInstance(context, element, mapId, mapInstanceOptions);
       });
 
       // Add an event listener to update the map size when the Gin toolbar is toggled.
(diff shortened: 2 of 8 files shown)


modules/core/map/js/farmOS.map.behaviors.geofield.js
```diff
@@ -3,7 +3,7 @@
     attach: function (instance) {
       instance.editAttached.then(() => {
         instance.edit.wktOn('featurechange', function(wkt) {
-          document.querySelector('#' + instance.target).parentElement.querySelector('textarea').value = wkt;
+          instance.map.getTargetElement().parentElement.querySelector('textarea').value = wkt;
```
symbioquine: OpenLayers maps can be constructed by either passing the id of the target element or by passing the element itself. This code was the only place (I've found) where we're assuming `target` is the id. By changing this code to use OpenLayer's `Map::getTargetElement` method, we can support either map creation scenario - and be more efficient to boot.

modules/core/map/js/farm_map.js
```diff
@@ -79,6 +34,69 @@
...
+        },
+      };
+
+      const mapOptions = {
+        ...defaultOptions,
+        ...drupalSettings.farm_map[drupalSettingsKey].instance,
+        ...mapInstanceOptions
+      };
+      const instance = farmOS.map.create(element, mapOptions);
+      instance.drupalSettingsKey = drupalSettingsKey;
```
symbioquine, line 62: Saving the "settings key" on the map instance so various behaviors can dereference settings from `drupalSettings`. Maybe there's an argument for more directly exposing the settings... looking forward to the discussion on this point.

modules/core/map/src/Element/FarmMapPrototype.php
```diff
@@ -0,0 +1,52 @@
...
+   * @param array $element
+   *   A renderable array containing a #map_type property, which will be
+   *   appended to 'farm-map-' as the map element ID.
+   *
+   * @return array
+   *   A renderable array representing the map prototype.
+   */
+  public static function preRenderMap(array $element) {
+
+    // Add the farm-map class.
```
symbioquine, line 45: doh! Will fix comment...

modules/core/map/src/Element/FarmMapPrototype.php
```diff
@@ -0,0 +1,52 @@
...
+    $element['#attributes']['class'][] = 'farm-map-prototype';
```
symbioquine, line 46: I debated whether it really makes sense to assign this class here and what to do about the `farm_map` class which is overloaded both to provide styling and to determine which maps to create. Curious about folks' thoughts on that too.

modules/core/map/src/Element/FarmMapPrototype.php
```diff
@@ -0,0 +1,52 @@
...
+
+    // Return the element.
```
symbioquine, line 48: Will update comment.

modules/core/map/src/Element/FarmMap.php
```diff
@@ -39,37 +40,11 @@ public function getInfo() {
...
-
-    // Add the instance settings under the map id key.
-    $instance_settings = array_merge_recursive($map_settings, $map_options);
-    $element['#attached']['drupalSettings']['farm_map'][$map_id] = $instance_settings;
-
-    // Create and dispatch a MapRenderEvent.
-    $event = new MapRenderEvent($map, $element);
-    \Drupal::service('event_dispatcher')->dispatch(MapRenderEvent::EVENT_NAME, $event);
-
     // Return the element.
```
symbioquine, line 46: Will update comment.

symbioquine:
@paul121 Especially looking forward to your thoughts on this. I think this is the lightest-weight way to take what you already created and make it possible for the JS layer to control the lifecycle of map instances while still honoring any Drupal/PHP-layer extensions/settings.

modules/core/map/js/farmOS.map.behaviors.geofield.js
paul121: Oh definitely! That is a great improvement.

paul121:
Thanks for opening this @symbioquine!

> This new FarmMapPrototype element does all the same steps, except it doesn't include the farm-map class which is used by the JS to find the maps to instantiate.

I guess I don't fully understand why a new `FarmMapPrototype` element is needed. Is the primary reason to render maps without the `farm-map` class, presumably to allow them to be instantiated in a different way? Are there other motivations for this? Maybe for using maps in a vue app?

I do see how refactoring this logic into a trait would make things more reusable if someone created their own render element that included a map, but it seems like a bit of an early design choice since core doesn't need it. If we're just modifying the wrapper css class, maybe that should be considered an "option" for the render element?

We could add a `#map_wrapper_class` key to the `FarmMap` render element that defaults to `farm-map`. If a map is rendered programmatically in a page controller or similar, `#map_wrapper_class` could be overridden with a different value.

Also if the main motivation is to allow the instantiation to be done elsewhere perhaps we should rethink how this is done altogether, so that the element class isn't what is used.

> Further, I've refactored the code from the farm_map.js Drupal behavior such that a factory method is exposed which other JS can leverage to create maps using all the same boiler-plate that is needed to wire map settings exposed from Drupal into new map instances.

Cool. Although not necessary for core I see why this could be useful in other contexts. The factory method seems like a general code improvement anyways!

Would it make sense to put this factory method in a separate JS library altogether? That would allow the factory method to be available separate from the JS library that instantiates maps based on the `farm-map` class.

modules/core/map/js/farm_map.js
```diff
@@ -2,62 +2,17 @@
...
-        if (details != null) {
-          details.addEventListener('toggle', function() {
-            instance.map.updateSize();
-          });
-        }
+          element.setAttribute('tabIndex', 0);
+
+          const mapId = element.getAttribute('id');
+          const mapInstanceOptions = {};
+          Drupal.behaviors.farm_map.createMapInstance(context, element, mapId, mapInstanceOptions);
```
paul121, line 15: Minor, but I think we could change this line to be:

```js
this.createMapInstance(context, element, mapId, mapInstanceOptions);
```

modules/core/map/js/farm_map.js
paul121: Ah interesting. This could be a solution! Could it be abstracted a bit so that `instance.settings()` was a method provided by the `createMapInstance` factory method? That way a behavior wouldn't have to navigate the `drupalSettings` array.

`instance.settings = () => drupalSettings.farm_map[drupalSettingsKey].instance;`

This could maybe be a staging area before something more standardized was adopted in farmOS-map core

modules/core/map/src/Element/FarmMapPrototype.php
paul121: > the farm_map class which is overloaded both to provide styling and to determine which maps to create.

Ah missed this comment, but yeah.. curious if we could use something other than the class to determine which maps to create?

modules/core/map/js/farm_map.js
symbioquine: I think that's true, will test that if we stick with keeping the factory method in this JS library...

symbioquine:
> Thanks for opening this @symbioquine!

Of course! Thanks for taking the time to look through it!

> > This new FarmMapPrototype element does all the same steps, except it doesn't include the farm-map class which is used by the JS to find the maps to instantiate.
> 
> I guess I don't fully understand why a new `FarmMapPrototype` element is needed. Is the primary reason to render maps without the `farm-map` class, presumably to allow them to be instantiated in a different way? Are there other motivations for this? Maybe for using maps in a vue app?

The motivation is to allow flexibility in how/when maps are created/destroyed at the JS layer while still leveraging the extension model provided by map render events/subscribers. Vue is certainly how I am planning to use that flexibility, but I think it will benefit many use-cases.

> I do see how refactoring this logic into a trait would make things more reusable if someone created their own render element that included a map, but it seems like a bit of an early design choice since core doesn't need it. If we're just modifying the wrapper css class, maybe that should be considered an "option" for the render element?
> 
> We could add a `#map_wrapper_class` key to the `FarmMap` render element that defaults to `farm-map`. If a map is rendered programmatically in a page controller or similar, `#map_wrapper_class` could be overridden with a different value.

I considered it and I'm still persuadable in that direction.

I guess my thought process was that a separate render element class makes it clearer semantically why one creates an actual fully instantiated map and the other just does the "pre-work" so that maps can be created at some point later.

Having the controller need to know about and customize a `#map_wrapper_class` "option" would arguably bleed that implementation detail about which class causes maps to get immediately instantiated and any other value requiring extra custom JS code.

> Also if the main motivation is to allow the instantiation to be done elsewhere perhaps we should rethink how this is done altogether, so that the element class isn't what is used.

I'm all ears! I was originally pretty opposed to the "side effects from event subscribers" pattern, but I think it's growing on me.

> > Further, I've refactored the code from the farm_map.js Drupal behavior such that a factory method is exposed which other JS can leverage to create maps using all the same boiler-plate that is needed to wire map settings exposed from Drupal into new map instances.
> 
> Cool. Although not necessary for core I see why this could be useful in other contexts. The factory method seems like a general code improvement anyways!

:)

> Would it make sense to put this factory method in a separate JS library altogether? That would allow the factory method to be available separate from the JS library that instantiates maps based on the `farm-map` class.

Sounds good. My only hesitancy would be around naming and potentially cluttering the cognitive space around which library does what...

paul121:
@symbioquine I guess we didn't discuss too much more about this issue in the dev call today. But on the topic of map behavior settings, I am really interested in the idea we conceived of having the "map settings" be a computed field on the `map_type` config entity.

This would allow for our current map event system to be used for populating the map/behavior settings for external uses like Field Kit. As @symbioquine pointed out a module added to farmOS wouldn't have to thinking about implementing the same logic in a similar Field Kit event system. A specific example is the Mapbox API key which `farm_map_mapbox` adds to all maps, and is something that Field Kit will need.

One complication with the current approach from 1.x is that we don't have a great way to expose "simple config" values via the API in 2.x - we have this meta issue to track this: https://www.drupal.org/project/farm/issues/3176736

Making the map/behavior settings (which would include the Mapbox API key) available via a computed settings field could be a nice solution. But just looking into this a little bit, I'm not sure it's possible in the way I first thought it might be... config entities are not "fieldable" like content entities. Instead, they just have properties defined in their class annotation and configuration schema. I don't see any way to add a "computed property". That said, we might still be able to hook into an `hook_entity_load` or alter the [pre/post-create/load/save](https://git.drupalcode.org/project/drupal/-/blob/9.2.x/core/lib/Drupal/Core/Entity/EntityBase.php#L378) methods of the config entity class. But at this point maybe `map_type` would be better off as a content entity with true computed fields... hmm.

Or maybe we could alter the JSONAPI response for the `map_type` config entity and add settings into the `meta` property... that might be better since the API is our primary motivation for this anyways?

symbioquine:
Closing this in favor of the simplified version in https://github.com/farmOS/farmOS/pull/436

PR #429, opened 2021-06-15
Fix metrics block caching.

paul121:
I noticed that the asset counts in the metrics block were not updating when I created new assets. Just a little bug, simple change.


mstenta:
Merged - thanks!

PR #430, opened 2021-06-16
Fix metrics link to log bundles

paul121:
Each button with log counts is linking to `/assets/{log_type}`. Simple fix!

Also made a minor change to the link's `class` attribute array and changed to use an entity aggregate query for the entity count query.

In this process I noticed that the dashboard metric block is not updated when new bundles of logs or assets are installed... I was hoping there might be a  `log_type_list` cache tag, but doesn't look like it's that easy.

We might be better off with a more general approach, invalidating the `{entity_type}_list` cache tag when new bundles of that entity type are installed. I understand why Drupal core might not want to do this, but I think it would be safe if we did?

### DIFF
--- a/modules/ui/metrics/src/Plugin/Block/FarmMetricsBlock.php
+++ b/modules/ui/metrics/src/Plugin/Block/FarmMetricsBlock.php
@@ -133,7 +133,8 @@ class FarmMetricsBlock extends BlockBase implements ContainerFactoryPluginInterf
     // Count records by type.
     foreach ($bundles as $bundle => $bundle_info) {
       $count = $this->database->query('SELECT COUNT(*) FROM {' . $entity_type . '_field_data} WHERE type=:type', [':type' => $bundle])->fetchField();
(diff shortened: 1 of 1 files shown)


mstenta:
Merged - thanks!

> We might be better off with a more general approach, invalidating the {entity_type}_list cache tag when new bundles of that entity type are installed. I understand why Drupal core might not want to do this, but I think it would be safe if we did?

Hmm interesting - I think that makes sense. Maybe in the `farm_entity` module?

paul121:
Yeah, probably makes sense in `farm_entity`.

paul121:
Oh shoot, I think I introduced a bug with the aggregate query @mstenta - I was testing with only 1 log type and 1 asset type so didn't take multiple bundles into account! As is I believe it would display the same count for all bundles.

The query should add a condition on the `type` just like we do here in the asset log views access check: https://github.com/farmOS/farmOS/blob/91c615c6e1d49ebd4f06c774dfef2b4eae1706cf/modules/ui/views/src/Access/FarmAssetLogViewsAccessCheck.php#L53

I don't think we need to specify the `->aggregate('id', 'COUNT')` though, this is what the `->count()` method does, it turns the query into a count query, so the query result is simply an integer. We could update that access check to be a proper `count` query as well and it would eliminate the need for this logic: `$log_count = (int) $result[0]['id_count'] ?? 0;`

PR #431, opened 2021-06-17
Fix entity count query and metrics caching.

paul121:
A follow up from #429 and #430 

- Fixes the entity bundle count problem I mentioned here: https://github.com/farmOS/farmOS/pull/430#issuecomment-862970395

- And updates the dashboard metrics when bundles are added/removed:

 > In this process I noticed that the dashboard metric block is not updated when new bundles of logs or assets are installed... I was hoping there might be a log_type_list cache tag, but doesn't look like it's that easy.

There actually is a `config:log_type_list` cache tag, so problem solved!!


mstenta:
Merged thanks!

PR #432, opened 2021-06-21
Fix caching of the farm_people view.

paul121:
Currently the farm_people page can be cached and display an incorrect `last_access` value for users. Also, if a user has never logged in the `last_access` value is rendered as 1970.

The first two commits of this PR modify the `last_access` field to rewrite the result as `Never` or the up-to-date HTML date.

The last commit changes the `last_access` value using the `timestamp_ago` formatter. This is the approach the core `user_admin_people` view uses. Due to the nature of "timestamp ago" formatting, it sets the cache max age accordingly (1s, 1m, 1 day, etc depending on the granularity the date is rendered) so that the value is always up to date. This also renders "never" when the user has never logged in. BUT this does not render the actual date of last access.

I see value in both the standard Date as well as "ago" formatting. I think we could accomplish both by rewriting the field to display "{HTML Date format} ({time ago format})". Thoughts?

### DIFF
--- a/modules/ui/views/config/install/views.view.farm_people.yml
+++ b/modules/ui/views/config/install/views.view.farm_people.yml
@@ -22,7 +22,7 @@ display:
         options:
           perm: 'access user profiles'
       cache:
-        type: tag
+        type: none
         options: {  }
       query:
         type: views_query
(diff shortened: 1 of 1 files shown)


mstenta:
Thanks @paul121 - merged!

PR #433, opened 2021-06-22
Remove unused asset and plan page include files.

paul121:
Doing a search for `template_preprocess_asset` I noticed we had two implementations. It seems that these `*.page.inc` files aren't being used at all.


mstenta:
Thanks @paul121!

As I was reviewing this I thought, "huh, I remember these... I thought I removed them already" - and then realized it was in the Log module, but not the Asset and Plan modules (which were both copied/modified from Log): https://github.com/mstenta/log/commit/1e645849e2b6be623324142d3b2896ca421dc34f

Good catch!

PR #434, opened 2021-06-23
Update dependencies

paul121:
`drupal/entity_browser` just made a new release and our patch no longer applies!

Also updated the `drupal/geofield` minimum version to match what we depend on.

The `drupal/views_geojson` patch file was actually incorrect and was creating new files in a `b` directory - this explains why this wasn't conflicting with the correct fix which was included with the 1.1 release we currently depend on.

I also bumped our `drupal/gin` dependency to the latest `alpha35` release. The most significant change here is that menu items are more condensed to take up a bit less space. I think it is a welcome change.


mstenta:
Merged - thanks!

PR #436, opened 2021-07-02
Decouple FarmMap render element and JS instantiation

symbioquine:
Note: This is intended to be merged to the 2.x branch after #425. Unfortunately, the diff includes the changes from that PRs since stacked pull requests don't work across forks on Github.

Decouple FarmMap render element and JS instantiation

**Why?** It should be possible to leverage the farm_map module's extension mechanism and default styling without necessarily giving it full control over when the map instance gets created.

Currently, the FarmMap render element uses the 'farm-map' CSS class to inform the JS code in `farm_map.js` that it should instantiate the map. This is problematic since that CSS class is also used for styling and there's a bunch of code in the FarmMap render element and in `farm_map.js` which should be common to pretty much all farmOS maps.

**What?**

* Introduce a new [data attribute] which is used instead of the CSS class to signal the JS code in `farm_map.js` that it should   actually instantiate the map.
* Refactor most of the logic in `farm_map.js` into a factory method  exposed at `Drupal.behaviors.farm_map.createMapInstance` which can then be leveraged by other JS code as well.
* Expose the map drupal settings as `instance.farmMapSettings` so that behaviors don't need to know about how to dereference   `drupalSettings` and so multiple instance of maps can be created with different ids, but the same settings.
* Fix the places where the JS assumes the target passed into `farmOS.map.create` is an ID string.

All that together allows for a controller to return;

```php
return [
  'map-prototype' => [
    '#type' => 'farm_map',
    '#attributes' => [
        'id' => 'example-tool-map-prototype',
        'data-map-instantiator' => 'example-tool',
    ],
  ],
];
```

Then create map instances from custom JS like this;

```js
const mapPrototypeElement = document.getElementById('example-tool-map-prototype');

const mapElement = mapPrototypeElement.cloneNode();
mapElement.removeAttribute('id');

myParentContainer.appendChild(mapElement);

const mapInstance = Drupal.behaviors.farm_map.createMapInstance(mapElement, mapElement, 'example-tool-map-prototype', {});
```

In doing so, a contrib module would get a canonically styled/extensible map where the contrib module's JS controls the lifecycle of the map element.

[data attribute]: https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes

### DIFF
--- a/modules/core/map/src/Element/FarmMap.php
+++ b/modules/core/map/src/Element/FarmMap.php
@@ -32,16 +32,23 @@ class FarmMap extends RenderElement {
    *
    * @param array $element
    *   A renderable array containing a #map_type property, which will be
-   *   appended to 'farm-map-' as the map element ID.
+   *   appended to 'farm-map-' as the map element ID if one has not already
+   *   been provided.
    *
    * @return array
    *   A renderable array representing the map.
    */
   public static function preRenderMap(array $element) {
 
-    // Set the id to the map name.
-    $map_id = Html::getUniqueId('farm-map-' . $element['#map_type']);
-    $element['#attributes']['id'] = $map_id;
+    if (empty($element['#attributes']['id'])) {
+      // Set the id to the map name.
+      $map_id = Html::getUniqueId('farm-map-' . $element['#map_type']);
+      $element['#attributes']['id'] = $map_id;
+    }
+
+    else {
+      $map_id = $element['#attributes']['id'];
+    }
 
     // Get the map type.
     /** @var \Drupal\farm_map\Entity\MapTypeInterface $map */
--- a/modules/core/map/src/Element/FarmMap.php
+++ b/modules/core/map/src/Element/FarmMap.php
@@ -50,6 +50,11 @@ class FarmMap extends RenderElement {
     // Add the farm-map class.
     $element['#attributes']['class'][] = 'farm-map';
 
+    // By default, inform farm_map.js that it should instantiate the map.
+    if (empty($element['#attributes']['data-map-instantiator'])) {
+      $element['#attributes']['data-map-instantiator'] = 'farm_map';
+    }
+
     // Attach the farmOS-map and farm_map libraries.
     $element['#attached']['library'][] = 'farm_map/farmOS-map';
     $element['#attached']['library'][] = 'farm_map/farm_map';
--- a/modules/core/map/js/farmOS.map.behaviors.geofield.js
+++ b/modules/core/map/js/farmOS.map.behaviors.geofield.js
@@ -3,7 +3,7 @@
     attach: function (instance) {
       instance.editAttached.then(() => {
         instance.edit.wktOn('featurechange', function(wkt) {
-          document.querySelector('#' + instance.target).parentElement.querySelector('textarea').value = wkt;
+          instance.map.getTargetElement().parentElement.querySelector('textarea').value = wkt;
         });
       });
     },
--- a/modules/core/map/js/farmOS.map.behaviors.wkt.js
+++ b/modules/core/map/js/farmOS.map.behaviors.wkt.js
@@ -2,9 +2,11 @@
   farmOS.map.behaviors.wkt = {
     attach: function (instance) {
 
+      const settings = instance.farmMapSettings;
+
       // If WKT was set, create a layer.
-      if (drupalSettings.farm_map[instance.target].wkt) {
-        var wkt = drupalSettings.farm_map[instance.target].wkt;
+      if (settings.wkt) {
+        var wkt = settings.wkt;
         var type = 'vector';
         var opts = {
           title: 'Geometry',
@@ -23,7 +25,7 @@
       var focusLayerPromise = Promise.resolve(layer);
 
       // If edit is true, enable drawing controls.
-      if (drupalSettings.farm_map[instance.target].behaviors.wkt.edit) {
+      if (settings.behaviors && settings.behaviors.wkt && settings.behaviors.wkt.edit) {
         if (layer !== undefined) {
           instance.editAttached = instance.addBehavior('edit', { layer: layer });
         } else {
--- a/modules/core/map/js/farm_map.js
+++ b/modules/core/map/js/farm_map.js
@@ -22,9 +22,13 @@
         element.setAttribute('processed', true);
 
         element.setAttribute('tabIndex', 0);
-        const mapId = element.getAttribute('id');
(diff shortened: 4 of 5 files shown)


symbioquine:
As discussed on the weekly dev call, I'll look through these changes and see whether they can be broken up into finer grained commits.

symbioquine:
@mstenta As requested, I've rebased and broken these changes up into three commits which will hopefully be easier to review. Looking forward to your thoughts/feedback!

modules/core/map/src/Element/FarmMap.php
```diff
@@ -50,6 +57,11 @@ public static function preRenderMap(array $element) {
     // Add the farm-map class.
     $element['#attributes']['class'][] = 'farm-map';
 
+    // By default, inform farm_map.js that it should instantiate the map.
+    if (empty($element['#attributes']['data-map-instantiator'])) {
```
paul121, line 61: I think this could be moved to the `FarmMap::getInfo()` method - that is where "default" values can be defined, but can be overridden in a render array that uses `#type => 'farm_map'`

paul121 (commented):
Just one minor thing

modules/core/map/src/Element/FarmMap.php
paul121: More specifically, we could define `#attributes['data-map-instantiator'] = 'farm_map` in the `getInfo()` method.

symbioquine: Thanks for the review @paul121!

Am I right in thinking you're talking about something like the following?

```patch
diff --git a/modules/core/map/src/Element/FarmMap.php b/modules/core/map/src/Element/FarmMap.php
index 7f2b795f..8c1f7288 100644
--- a/modules/core/map/src/Element/FarmMap.php
+++ b/modules/core/map/src/Element/FarmMap.php
@@ -19,16 +19,19 @@ class FarmMap extends RenderElement {
   public function getInfo() {
     $class = get_class($this);
     return [
       '#pre_render' => [
         [$class, 'preRenderMap'],
       ],
       '#theme' => 'farm_map',
       '#map_type' => 'default',
+      '#attributes' => [
+        'data-map-instantiator' => 'farm_map',
+      ],
     ];
   }
 
   /**
    * Pre-render callback for the map render array.
    *
    * @param array $element
    *   A renderable array containing a #map_type property, which will be
@@ -52,21 +55,16 @@ class FarmMap extends RenderElement {
 
     // Get the map type.
     /** @var \Drupal\farm_map\Entity\MapTypeInterface $map */
     $map = \Drupal::entityTypeManager()->getStorage('map_type')->load($element['#map_type']);
 
     // Add the farm-map class.
     $element['#attributes']['class'][] = 'farm-map';
 
-    // By default, inform farm_map.js that it should instantiate the map.
-    if (empty($element['#attributes']['data-map-instantiator'])) {
-      $element['#attributes']['data-map-instantiator'] = 'farm_map';
-    }
-
     // Attach the farmOS-map and farm_map libraries.
     $element['#attached']['library'][] = 'farm_map/farmOS-map';
     $element['#attached']['library'][] = 'farm_map/farm_map';
 
     // Include map settings.
     $map_settings = !empty($element['#map_settings']) ? $element['#map_settings'] : [];
 
     // Include the map options.
```

I took a glance over some of the [core `ElementInterface::getInfo` implementations](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Render%21Element%21ElementInterface.php/function/overrides/ElementInterface%3A%3AgetInfo/9.0.x) and didn't find any which directly define default attributes as such.

I believe that is because the fields from `ElementInterface::getInfo` [are not deeply merged](https://www.drupal.org/project/drupal/issues/2070131) with values provided in the render array.

Instead, it appears that the convention is something like this;

```patch
diff --git a/modules/core/map/src/Element/FarmMap.php b/modules/core/map/src/Element/FarmMap.php
index 7f2b795f..3029129d 100644
--- a/modules/core/map/src/Element/FarmMap.php
+++ b/modules/core/map/src/Element/FarmMap.php
@@ -19,16 +19,17 @@ class FarmMap extends RenderElement {
   public function getInfo() {
     $class = get_class($this);
     return [
       '#pre_render' => [
         [$class, 'preRenderMap'],
       ],
       '#theme' => 'farm_map',
       '#map_type' => 'default',
+      '#data-map-instantiator' => 'farm_map',
     ];
   }
 
   /**
    * Pre-render callback for the map render array.
    *
    * @param array $element
    *   A renderable array containing a #map_type property, which will be
@@ -52,20 +53,19 @@ class FarmMap extends RenderElement {
 
     // Get the map type.
     /** @var \Drupal\farm_map\Entity\MapTypeInterface $map */
     $map = \Drupal::entityTypeManager()->getStorage('map_type')->load($element['#map_type']);
 
     // Add the farm-map class.
     $element['#attributes']['class'][] = 'farm-map';
 
-    // By default, inform farm_map.js that it should instantiate the map.
-    if (empty($element['#attributes']['data-map-instantiator'])) {
-      $element['#attributes']['data-map-instantiator'] = 'farm_map';
-    }
+    Element::setAttributes($element, [
+      'data-map-instantiator',
+    ]);
 
     // Attach the farmOS-map and farm_map libraries.
     $element['#attached']['library'][] = 'farm_map/farmOS-map';
     $element['#attached']['library'][] = 'farm_map/farm_map';
 
     // Include map settings.
     $map_settings = !empty($element['#map_settings']) ? $element['#map_settings'] : [];
 
```

Naively, that convention doesn't seem like it provides much benefit. It means that each "configurable attribute" now requires two methods to be changed instead of one. Is there some other functional benefit from having properties specified in the `ElementInterface::getInfo` method? The docs seem pretty sparse about the actual purpose/intent of most of this stuff...

paul121: > Am I right in thinking you're talking about something like the following?

Yes, exactly what I was thinking. But shoot, good find! Looks like that won't work because of the deep merge issue.

> Naively, that convention doesn't seem like it provides much benefit.

I agree. It would be confusing to have both the top level `'#data-map-instantiator'` and  `#attributes['#data-map-instantiator']`. Keeping as is with just the latter make sense!

symbioquine:
The latest force-push is just me rebasing these changes now that the farmOS-map 2.x upgrade PR has been pushed.

mstenta:
Thanks @symbioquine (and @paul121) - this looks good to me and makes sense. All seems to be working properly in my local testing. I will go ahead and merge this.

PR #437, opened 2021-08-17
Fix javascript tests

paul121:
Currently our javascript tests are not being executed because they are marked as "incomplete":
https://github.com/farmOS/farmOS/runs/3192325229?check_suite_focus=true#step:12:143

Looking into this, I realized it is because we don't have `chromedriver` available nor configured in our dev container. See drupal docs for running PHPUnit Javascript tests: https://www.drupal.org/docs/automated-testing/phpunit-in-drupal/running-phpunit-javascript-tests

I'm not sure how we missed this (_mostly me because I wrote the javascript tests... and somehow thought they were working... :thinking:_).. but we did! 
There were a few gotchas on all this, and little (up to date) documentation. So some notes on the changes, if not in a commit description:

## Chrome driver

- Rather than install selenium/chrome/chromedriver and other dependencies in our dev image, we can run `chromedriver` in a separate docker container with the [`selenium/standalone-chrome`](https://github.com/SeleniumHQ/docker-selenium#execution-modes) image.
- The drupal docs say to use the `drupalci/webdriver-chromedriver:production` image, but the image hasn't been updated in two years and I could not find a source repo for this image anyways. Using the image provided by Selenium seems more reliable going forward.

- Seemed to find conflicting recommendations on the `MINK_DRIVER_ARGS_WEBDRIVER` PHPUnit config value, specifically which port to use, but I think it depends on the docker container being used. The Drupal `phpunit.xml.dist.yml` example ended up being the ticket, with some additional `chromeOptions` that help speed things up:
```xml
    <env name="MINK_DRIVER_ARGS_WEBDRIVER" value='["chrome", { "chromeOptions": { "w3c": false, "args": ["--disable-gpu","--headless", "--no-sandbox"] } }, "http://chrome:4444/wd/hub"]'/>

```
- The `WebDriver` used for javascript tests does not support some methods like asserting status codes.
- The existing javascript tests needed a few other simple changes to get them passing. Most notably I had to separate JSON:API related location tests into a separate functional test since they were incompatible with the javascript test driver.

These articles helped with the chrome driver config:
- Using `selenium/standalone-chrome`: https://cylab.be/blog/134/automated-web-application-testing-with-php-selenium-and-gitlab)
- Most up-to date: https://drupalize.me/blog/debug-any-drupals-phpunit-tests-phpstorm-ddev-local-environment
- Documents some differences with `Webdriver` (like no support for asserting status codes) https://mglaman.medium.com/running-drupals-functionaljavascript-tests-on-ddev-f9ea74653e86

##  PHPUnit changes

- Add `failOnIncomplete` and `failOnSkipped` so the test action will fail. This is up for debate: many consider it a feature to mark tests as "skipped" or "incomplete". But because we don't depend on this feature in the farmOS code base right now (we want ALL of our tests to execute & succeed), this will alert us if anything like this happens again.
  - At first I thought these options weren't available, but it turns out it was just missing from the documentation (https://github.com/sebastianbergmann/phpunit-documentation-english/pull/247). 

## Docker-compose changes

- I simplified the `docker-compose.testing.*.yml` files so that they can inherit from a single `docker-compose.testing.common.yml`. This is accomplished by [specifying multiple compose files](https://docs.docker.com/compose/reference/#specifying-multiple-compose-files).
- I also added the `chrome` service to our `docker-compose.development.yml` file, but left it commented out by default. It is only needed when writing and executing javascript tests. Most users won't need it so no need to pull the extra docker image & consume additional resources?


paul121:
An example of the new Unit tests result with no "incomplete" tests: https://github.com/farmOS/farmOS/runs/3353751396#step:12:142

mstenta:
Wow! Thanks for identifying this @paul121! Great work!

PR #438, opened 2021-08-17
Add a Seed asset type

mstenta:
See https://www.drupal.org/project/farm/issues/3022158

This is copied from the module that @symbioquine created.

@symbioquine I gave you commit credit, since it was an almost exact copy of your code. Just a few minor tweaks to namespaces, core-consistent text, etc.

### DIFF
--- a/farm.profile
+++ b/farm.profile
@@ -44,6 +44,7 @@ function farm_modules() {
     ],
     'optional' => [
       'farm_inventory' => t('Inventory management'),
+      'farm_seed' => t('Seed assets'),
       'farm_sensor' => t('Sensor assets'),
       'farm_compost' => t('Compost assets'),
       'farm_group' => t('Group assets'),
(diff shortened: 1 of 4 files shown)


symbioquine:
Looks good! I'm really happy to see this brought upstream and agree with the strategy of first checking in the fairly minimal set of fields seen here. My use-cases in fact also require a number of specialized fields, but I left them out of the [farmOS_seed_asset](https://github.com/symbioquine/farmOS_seed_asset) module I published since I wanted that to be more generally applicable to folks.

I'll probably deprecate my [farmOS_seed_asset](https://github.com/symbioquine/farmOS_seed_asset) module in favor of this rather than keep it as a "migration-only" or as place to put additional fields since the naming is sub-optimal for those purposes and it could confuse folks - even with the proposed doc update.

mstenta:
Merged.

PR #441, opened 2021-09-02
Remove test-runner and Disable XDebug with XDEBUG_MODE: off

mstenta:
Back in November of last year, we took a "clever" approach to running tests without XDebug, by spinning up both a `farmos/farmos:2.x-dev` service (to build a bind-mounted codebase in a directory with dev dependencies included) and a `farmos/farmos:2.x` service (for running `phpunit` without XDebug present).

See:

- https://www.drupal.org/project/farm/issues/3183687
- #377 

This worked great!

Then, in December 2020, we upgraded to XDebug 3: https://github.com/farmOS/farmOS/commit/9bb7d4f8557f1be29e3664831c236d5cdf6eeba8

XDebug 3 adds the `XDEBUG_MODE` environment variable, which we set to `debug` in our development environments, but can also be set to `off` to disable XDebug.

XDebug's docs state:

> `off`
> Nothing is enabled. Xdebug does no work besides checking whether functionality is enabled. Use this setting if you want close to 0 overhead.

https://xdebug.org/docs/all_settings#mode

This pull request uses the `XDEBUG_MODE` environment variable to disable XDebug in our testing environment. This removes the need for a separate `test-runner` service, so it also removes that.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -59,15 +59,15 @@ jobs:
           docker-compose -f docker-compose.testing.common.yml -f docker-compose.testing.${{ matrix.dbms }}.yml config > docker-compose.yml
       - name: Start containers
         run: docker-compose up -d
-      - name: Wait until test-runner container is ready
-      # The test-runner-container-fs-ready file is only created once we expect the containers to be online
+      - name: Wait until www container is ready
+      # The www-container-fs-ready file is only created once we expect the containers to be online
       # so waiting for that lets us know it is safe to start the tests
-        run: until [ -f ./www/test-runner-container-fs-ready ]; do sleep 0.1; done
+        run: until [ -f ./www/www-container-fs-ready ]; do sleep 0.1; done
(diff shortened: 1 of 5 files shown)


mstenta:
> we took a "clever" approach to running tests without XDebug

I take the blame for this... ;-)

> Should work... in theory? It's a bit of a "clever" approach, though, which usually means it will come back to bite us in the future...

https://www.drupal.org/project/farm/issues/3183687#comment-13910778

It *did* end up biting me. I spent this afternoon debugging an issue with a custom module test that was due to the fact that `www` and `test-runner` `www-data` users had different IDs. :-P

mstenta:
Ran tests with this twice, and both times it took an average amount of time to run. Not drastically more, as we would expect to see with XDebug.

symbioquine (approved):
Looks great! Glad we can get rid of this kludgy optimization now :)

mstenta:
Thanks for taking a look @symbioquine ! Merged.

PR #442, opened 2021-09-12
Enable the map side panel with layer switcher by default

symbioquine:
**Why?** Provide a place for more settings and information
to be included on farmOS maps without introducing more and more
border clutter.

image

image

image

***Related:** There's a race-condition bug in the farmOS-map `layersSwitcherInSidePanel` behavior which I'm fixing in https://github.com/farmOS/farmOS-map/pull/131. Without that fix, the layer switcher sometimes stays as a separate control - leaving the side panel effectively empty.*

### DIFF
--- a/composer.libraries.json
+++ b/composer.libraries.json
@@ -11,10 +11,10 @@
             "type": "package",
             "package": {
                 "name": "farmos/farmos-map",
-                "version": "2.0.1",
+                "version": "2.0.2",
                 "type": "drupal-library",
                 "dist": {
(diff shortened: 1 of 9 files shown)


mstenta:
This is a great idea @symbioquine! And the commit looks good to me.

So once https://github.com/farmOS/farmOS-map/pull/131 is merged and a new version of farmOS-map is released, we'll probably want to add an additional commit to this PR that includes the new version change. Then we can merge this.

mstenta (approved):

PR #445, opened 2021-09-18
Update testing docs to reflect recent changes

symbioquine:
**Why?** The docs were out of date - not capturing the
need for a Chrome/Selenium container and including info
about an XDebug work-around that isn't needed any more.

image


mstenta:
Thanks @symbioquine - merged!

(With small change: replaced the semicolons with colons at the end of lines, for consistency with the rest of the docs. Maybe semicolons are better grammatically? If so let's make consider that change holistically, but keep this consistent for now.)

PR #446, opened 2021-09-19
Improve asset archive timestamp behavior

symbioquine:
**Why?** Fix some weird behavior and make sure
inconsistent states cannot occur;

* Allow setting the archived timestamp at the same
  time as moving an asset to the archived state.
* Ensure that the archived timestamp cannot be set
  without also transitioning into the archived state.
* Ensure that the arhived timestamp cannot be set to
  NULL without transitioning back to the active state.

### DIFF
--- a/modules/core/asset/src/AssetStorage.php
+++ b/modules/core/asset/src/AssetStorage.php
@@ -29,13 +29,30 @@ class AssetStorage extends SqlContentEntityStorage {
     $new_state = $entity->get('status')->first()->getString();
     $old_state = $entity->original->get('status')->first()->getString();
 
+    $state_unchanged = $new_state == $old_state;
+
+    // If the entity is not archived and this would otherwise not be a state
+    // transition but the archive timestamp is set, then transition to the
+    // archived state.
+    if ($state_unchanged && $old_state != 'archived' && $entity->getArchivedTime() != NULL) {
+      $entity->get('status')->first()->applyTransitionById('archive');
+    }
+
+    // If the entity is archived and this would otherwise not be a state
+    // transition but the archive timestemp is NULL, then transition to the
+    // active state.
+    if ($state_unchanged && $old_state == 'archived' && $entity->getArchivedTime() == NULL) {
+      $entity->get('status')->first()->applyTransitionById('to_active');
+    }
+
     // If the state has not changed, bail.
-    if ($new_state == $old_state) {
+    if ($state_unchanged) {
       return;
     }
 
-    // If the state has changed to archived, save the archived timestamp.
-    if ($new_state == 'archived') {
+    // If the state has changed to archived and no archived timestamp was
+    // specified, set it to the current time.
+    if ($new_state == 'archived' && $entity->getArchivedTime() == NULL) {
       $entity->setArchivedTime(\Drupal::time()->getRequestTime());
     }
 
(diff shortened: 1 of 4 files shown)


mstenta (approved):
Thanks @symbioquine this looks good to me.

I'm going to request @paul121's review too, just to be thorough. :-)

paul121 (commented):
Looks good! But we should probably do the same thing for the `PlanStorage` class, yea?: https://github.com/farmOS/farmOS/blob/cc012931683f5d6c6a9f3795cfeead33c1873b8b/modules/core/plan/src/PlanStorage.php

>  Initial plan module (straight copy+rename of asset module). 

The last commit in that file - it's the same exact code! Would it make sense to refactor this archiving logic into a reusable trait?

mstenta:
Ah great point @paul121!

I love the idea of a Trait, but it raises questions about dependencies. So far the `asset` and `plan` modules have been kept relatively self-contained, without depending on anything in farmOS itself (similar to the `log` module, which is Drupal contrib). I would like to keep them that way, if possible - treating them as if they are contrib modules.

So if we wanted to combine this logic we'd need to do it somewhere in farmOS, and override the storage classes for assets and plans to use it (I think?).

It might be easier to just replicate the code in both places, honestly...

paul121:
Ah yes, agreed. I think that would require overriding the `asset` and `plan` storage classes, so it's probably easier to just replicate the code. Maybe it could be implemented in a `hook_pre_save` or something, but it is nice having this in the storage classes themselves.

symbioquine:
> I love the idea of a Trait, but it raises questions about dependencies. So far the `asset` and `plan` modules have been kept relatively self-contained, without depending on anything in farmOS itself (similar to the `log` module, which is Drupal contrib). I would like to keep them that way, if possible - treating them as if they are contrib modules.
> 
> So if we wanted to combine this logic we'd need to do it somewhere in farmOS, and override the storage classes for assets and plans to use it (I think?).
> 
> It might be easier to just replicate the code in both places, honestly...

Makes sense @mstenta but it is enough identical code that it *does* feel a bit smelly to copy it...

From what I can see the asset/plan modules don't have any farmOS-specific dependencies (yet). I can imagine introducing one - something like "farm_archivable" which provides the trait @paul121 described.

Happy to go either way (new core module or just copying the code), but I don't want to just throw code at the wall. Let me know what you guys prefer...

mstenta:
> Makes sense @mstenta but it is enough identical code that it _does_ feel a bit smelly to copy it...

If it makes you feel any better, the entire `plan` module was copied directly from `asset` so... we wouldn't be adding much smell to what's already there. ;-)

> the asset/plan modules don't have any farmOS-specific dependencies (yet). I can imagine introducing one

Yea I've been trying to avoid that, and I don't think this issue warrants it yet.

If we really want to split things out in the future I think we would remove the `archived` base field from the core `asset` and `plan` modules entirely, and add it via a `farm_*` module. This would avoid the need for a dependency, but would require either a) overriding the storage (as I mentioned above), or doing it via a hook (which would be less code and less invasive perhaps).

Personally I would just copy the code for now. We're already maintaining these two modules (`asset` and `plan`) as nearly identical, so changes to one generally need to be replicated in both. Not great, I agree - but both are relatively simple still, and the considerations for combining things are bigger.

mstenta (approved):

mstenta:
Thanks @symbioquine!! I will rebase and merge this next.

paul121 (approved):

mstenta:
Merged.

PR #447, opened 2021-09-23
Log the result after reverting config.

paul121:
Follow up to #444 

`\Drupal::logger()->notice(...)` will print messages to the console when drush is sufficiently bootstrapped. This works via `drush farm_update:rebuild`:

```sh
$ docker-compose exec -u www-data www drush farm_update:rebuild
 [notice] Reverted config: quantity.type.material
 [notice] Reverted config: views.view.farm_log
```

Unfortunately `drush cr` doesn't get bootstrapped enough to print these messages to the console, but they are saved in the database logs. More info here: https://github.com/drush-ops/drush/issues/3449#issuecomment-490031390

So it doesn't seem like there is any way to print logs during `drush cr` - ah well

### DIFF
--- a/modules/core/update/farm_update.services.yml
+++ b/modules/core/update/farm_update.services.yml
@@ -1,4 +1,7 @@
 services:
+  logger.channel.farm_update:
+    parent: logger.channel_base
+    arguments: [ 'farm_update' ]
   farm.update:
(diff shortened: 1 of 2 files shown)


mstenta:
Merged - thanks @paul121 !

PR #449, opened 2021-09-24
Update to Drupal 9.2

mstenta:
This branch is for testing the update from Drupal 9.1 to 9.2. I had to update one of our core patches to apply, but other than that tests are passing. I have not manually tested yet, but generally minor Drupal core version updates should be pretty stable.

See related `2.x-drupal-9.2` branch in https://github.com/farmOS/composer-project: https://github.com/farmOS/composer-project/compare/2.x-drupal-9.2 - this branch contains a temporary commit to point to that branch, which should not be committed. These should both be merged at the same time.

Moving forward, I wonder if there's a way we can manage these minor version updates without needing to update the `composer-project/composer.json` every time - because this means that anyone managing their farmOS deployments via Composer will need to manually update theirs accordingly.

Can we specify a more general version in the project `composer.json` (eg `^9`) and then constrain the `drupal/core-dev` and `drupal/core-composer-scaffold` versions in `farmOS/composer.json`?

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -66,8 +66,6 @@ jobs:
       - name: Run PHP CodeSniffer
         run: docker-compose exec -u www-data -T www phpcs /opt/drupal/web/profiles/farm
       - name: Run PHPUnit tests
-      # The overridden result printer is needed until Drupal 9.2 when the updated version of HtmlOutputPrinter will be available
-      # https://github.com/drupal/drupal/blob/74fbb0aabdee3e1a5da7b8e489a725afdc5824fd/core/tests/Drupal/Tests/Listeners/HtmlOutputPrinter.php
-        run: docker-compose exec -u www-data -T www paratest --passthru='--printer=PHPUnit\\TextUI\\DefaultResultPrinter' --verbose=1 /opt/drupal/web/profiles/farm
+        run: docker-compose exec -u www-data -T www paratest --verbose=1 /opt/drupal/web/profiles/farm
       - name: Test Drush site install with all modules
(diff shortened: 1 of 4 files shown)


mstenta:
Rebased and added a commit to remove the override we had in our automated test logic. Changed from:

```
- name: Run PHPUnit tests
# The overridden result printer is needed until Drupal 9.2 when the updated version of HtmlOutputPrinter will be available
# https://github.com/drupal/drupal/blob/74fbb0aabdee3e1a5da7b8e489a725afdc5824fd/core/tests/Drupal/Tests/Listeners/HtmlOutputPrinter.php
  run: docker-compose exec -u www-data -T www paratest --passthru='--printer=PHPUnit\\TextUI\\DefaultResultPrinter' --verbose=1 /opt/drupal/web/profiles/farm
```

to:

```
- name: Run PHPUnit tests
  run: docker-compose exec -u www-data -T www paratest --verbose=1 /opt/drupal/web/profiles/farm
```

Running tests now... @symbioquine does that look right to you? I wasn't familiar with the exact reason for this, but I assume just removing that parameter is all we need to do here?

symbioquine:
> Running tests now... @symbioquine does that look right to you? I wasn't familiar with the exact reason for this, but I assume just removing that parameter is all we need to do here?

Yep, looks good! As I recall the tests broke without that parameter - so if they're passing now that should be all the verification that's needed.

mstenta:
Merged.

I spoke with @paul121 about the Drupal version considerations in the project `composer.json` and I'm going to see if we can use `composer-merge-plugin` to handle that... perhaps we'll be able to greatly simplify that `composer.json`!

PR #456, opened 2021-10-12
Add workflow to trigger farmOS.org Netlify builds when docs are updated

jgaehring:
As I mentioned [on the forum](https://farmos.discourse.group/t/work-proposal-scaffold-a-new-gatsby-site-for-farmos-org-2-x-docs/699/12?u=jgaehring), I think it's time to see how we can incorporate this into farmOS 2.x to start triggering builds on the [test deployment](https://gracious-brattain-bdd606.netlify.app/farmos/docs/). This won't impact the current deployment at docs.farmos.org, nor any of the other workflows. And it's a super lightweight workflow as far as GH Actions go; all it does is fire off a `curl` request and exit the process, doesn't even wait for a response. 

Currently I guess it won't even do that, since I have it set to trigger on `push` and `pull_request` events on the `2.x-netlify-build-docs` branch, which doesn't exist in farmOS/farmOS, so we probably want to add another commit to this to change the target branch, either to `2.x` or some other branch if that's preferable.

The only other requirement is to add the Netlify token to GH secrets for this repo. Down the line, we'll want to set up a permanent Netlify deployment that can be controlled by the farmOS GitHub org. That will also make it easier to share the token across all farmOS projects.

One final note, I've also created a special branch in the Gatsby repo, [`preview-farmos`](https://github.com/jgaehring/farmos.org-2.x/tree/preview-farmos), which triggers a separate [branch deployment](https://preview-farmos--gracious-brattain-bdd606.netlify.app/) when there are pull requests. There's nothing there yet, because there hasn't been a PR to trigger the deploy, but should be nice to add similar branch deployments for other repos' docs.

### DIFF
--- a/.github/workflows/netlify-build-docs.yml
+++ b/.github/workflows/netlify-build-docs.yml
@@ -1,0 +1,28 @@
+name: Trigger Netlify to build docs
+on:
+  push:
+    branches:
+      - '2.x-netlify-build-docs'
+  pull_request:
+    branches:
+      - '2.x-netlify-build-docs'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    env:
+      TOKEN: ${{ secrets.NETLIFY_BUILD_HOOK }}
+    steps:
+      - name: Set trigger_title environment variable
+        run: |
+          echo "TRIGGER_TITLE=Triggered+by+farmOS+source+repo" >> $GITHUB_ENV
+      - name: Set trigger_branch env.var for preview build
+        run: |
+          echo "TRIGGER_BRANCH=preview-farmos" >> $GITHUB_ENV
+        if: github.event_name == 'pull_request'
+      - name: Set trigger_branch env.var for main production build
+        run: |
+          echo "TRIGGER_BRANCH=main" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


jgaehring:
Oops, just realized there was a newline missing at the end of the workflow YAML that caused the tests to fail. Force-pushing an amended commit now...

jgaehring:
Grr, stupid automatic tabs on newline. Force-pushing again....

jgaehring:
Blah, whitespace issues resolved, but there were still [2 errors](https://github.com/jgaehring/farmOS/runs/3915382222?check_suite_focus=true#step:12:81) in the sqlite portion of the PHPUnit tests. I'm afraid that's beyond my ken, but hopefully it isn't due to anything wrong with this workflow.

Perhaps we can check up on this sometime next week, @mstenta?

jgaehring:
Ok, I've finished the necessary changes to the workflow (see jgaehring/farmOS.org-2.x#5 and jgaehring/farmOS.org-2.x#6), tested them (hence the forced updates), and everything seems to work!

Lastly, I added one more commit to change the workflow `branches` from `2.x-netlify-build-docs` to `2.x`, so this should be ready to merge!

PR #458, opened 2021-10-21
2.x Link from entities to their referenced terms and show entity views on taxonomy terms

symbioquine:
**Why?** Make it possible for novice users to find data and explore connections between assets/logs without needing to totally grok the farmOS data model first.

i.e.

* Navigate from a plant asset to a list of all the seed/plant assets with the same crop/variety
* Navigate from a plant asset to a list of all the seed/plant assets from the same season
* Navigate from a log in the compost category to a list of all the logs in the compost category
* Navigate from a material asset to a list of all the materials with the same material type

image
image
image
image

### DIFF
--- a/modules/core/ui/views/config/install/views.view.farm_asset.yml
+++ b/modules/core/ui/views/config/install/views.view.farm_asset.yml
@@ -9,6 +9,7 @@ dependencies:
     - csv_serialization
     - entity_browser
     - farm_location
+    - farm_ui_views
     - image
     - options
     - rest
@@ -1584,6 +1585,111 @@ display:
         menu_name: admin
         parent: farm.records
         context: '0'
+    cache_metadata:
+      max-age: 0
+      contexts:
+        - 'languages:language_content'
+        - 'languages:language_interface'
+        - url
+        - url.query_args
+        - user.permissions
+      tags: {  }
+  page_term:
+    display_plugin: page
+    id: page_term
+    display_title: 'By term (page)'
+    position: 4
+    display_options:
+      display_extenders: {  }
+      display_description: ''
+      path: taxonomy/term/%taxonomy_term/assets/%entity_bundle
+      arguments:
+        asset_taxonomy_term_reference:
+          id: asset_taxonomy_term_reference
+          table: asset_field_data
+          field: asset_taxonomy_term_reference
+          relationship: none
+          group_type: group
+          admin_label: ''
+          break_phrase: false
+          default_action: 'not found'
+          exception:
+            value: all
+            title_enable: false
+            title: All
+          title_enable: false
+          title: ''
+          default_argument_type: fixed
+          default_argument_options:
+            argument: ''
+          default_argument_skip_url: false
+          summary_options:
+            base_path: ''
+            count: true
+            items_per_page: 25
+            override: false
+          summary:
+            sort_order: asc
+            number_of_records: 0
+            format: default_summary
+          specify_validation: true
+          validate:
+            type: 'entity:taxonomy_term'
+            fail: 'not found'
+          validate_options:
+            operation: view
+            access: false
+            multiple: 0
+            bundles: {}
+          entity_type: asset
+          plugin_id: entity_taxonomy_term_reference
+        type:
+          id: type
+          table: asset_field_data
+          field: type
+          relationship: none
+          group_type: group
+          admin_label: ''
+          default_action: 'not found'
+          exception:
+            value: all
+            title_enable: false
+            title: All
+          title_enable: false
+          title: ''
+          default_argument_type: fixed
+          default_argument_options:
+            argument: all
+          default_argument_skip_url: false
+          summary_options:
+            base_path: ''
+            count: true
+            items_per_page: 25
+            override: false
+          summary:
+            sort_order: asc
+            number_of_records: 0
+            format: default_summary
+          specify_validation: true
+          validate:
+            type: 'entity:asset_type'
+            fail: 'not found'
+          validate_options:
+            operation: view
+            multiple: 0
+            access: false
+            bundles: {  }
+          glossary: false
+          limit: 0
+          case: none
+          path_case: none
+          transform_dash: false
+          break_phrase: false
+          entity_type: asset
+          entity_field: type
+          plugin_id: string
+      defaults:
+        arguments: false
     cache_metadata:
       max-age: 0
       contexts:
--- a/modules/core/ui/views/config/install/views.view.farm_log.yml
+++ b/modules/core/ui/views/config/install/views.view.farm_log.yml
@@ -2471,7 +2471,7 @@ display:
           entity_type: log
           entity_field: type
           plugin_id: string
-          default_action: ignore
+          default_action: 'not found'
           exception:
             value: all
             title_enable: false
@@ -2511,6 +2511,112 @@ display:
       display_description: ''
       display_extenders: {  }
       path: asset/%asset/logs/%log_type
+    cache_metadata:
+      max-age: 0
+      contexts:
+        - 'languages:language_content'
+        - 'languages:language_interface'
+        - url
+        - url.query_args
+        - user
+        - user.permissions
+      tags: {  }
+  page_term:
+    display_plugin: page
+    id: page_term
+    display_title: 'By term (page)'
+    position: 2
+    display_options:
+      display_extenders: {  }
+      display_description: ''
+      path: taxonomy/term/%taxonomy_term/logs/%entity_bundle
+      arguments:
+        log_taxonomy_term_reference:
+          id: log_taxonomy_term_reference
+          table: log_field_data
+          field: log_taxonomy_term_reference
+          relationship: none
+          group_type: group
+          admin_label: ''
+          break_phrase: false
+          default_action: 'not found'
+          exception:
+            value: all
+            title_enable: false
+            title: All
+          title_enable: false
+          title: ''
+          default_argument_type: fixed
+          default_argument_options:
+            argument: ''
+          default_argument_skip_url: false
+          summary_options:
+            base_path: ''
+            count: true
+            items_per_page: 25
+            override: false
+          summary:
+            sort_order: asc
+            number_of_records: 0
+            format: default_summary
+          specify_validation: true
+          validate:
+            type: 'entity:taxonomy_term'
+            fail: 'not found'
+          validate_options:
+            operation: view
+            access: false
+            multiple: 0
+            bundles: {}
+          entity_type: log
+          plugin_id: entity_taxonomy_term_reference
+        type:
+          id: type
+          table: log_field_data
+          field: type
+          relationship: none
+          group_type: group
+          admin_label: ''
+          default_action: 'not found'
+          exception:
+            value: all
+            title_enable: false
+            title: All
+          title_enable: false
+          title: ''
+          default_argument_type: fixed
+          default_argument_options:
+            argument: all
+          default_argument_skip_url: false
+          summary_options:
+            base_path: ''
+            count: true
+            items_per_page: 25
+            override: false
+          summary:
+            sort_order: asc
+            number_of_records: 0
+            format: default_summary
+          specify_validation: true
+          validate:
+            type: 'entity:log_type'
+            fail: 'not found'
+          validate_options:
+            operation: view
+            multiple: 0
+            access: false
+            bundles: {  }
+          glossary: false
+          limit: 0
+          case: none
+          path_case: none
+          transform_dash: false
+          break_phrase: false
+          entity_type: log
+          entity_field: type
+          plugin_id: string
+      defaults:
+        arguments: false
     cache_metadata:
       max-age: 0
       contexts:
--- a/modules/core/ui/views/farm_ui_views.links.task.yml
+++ b/modules/core/ui/views/farm_ui_views.links.task.yml
@@ -1,3 +1,4 @@
+# Add log view tabs to assets.
 farm.asset.logs:
   title: 'Logs'
   route_name: view.farm_log.page_asset
@@ -13,3 +14,21 @@ farm.asset.logs.all:
   parent_id: farm.asset.logs
 farm.asset.logs.type:
   deriver: Drupal\farm_ui_views\Plugin\Derivative\FarmLogViewsTaskLink
+
+# Add asset/log view tabs to taxonomy terms.
+farm.taxonomy_term.assets:
+  title: 'Assets'
+  route_name: view.farm_asset.page_term
+  route_parameters:
+    entity_bundle: 'all'
+  base_route: entity.taxonomy_term.canonical
+  weight: 50
+farm.taxonomy_term.logs:
+  title: 'Logs'
+  route_name: view.farm_log.page_term
+  route_parameters:
+    entity_bundle: 'all'
+  base_route: entity.taxonomy_term.canonical
+  weight: 50
+farm.taxonomy_term.entities.type:
+  deriver: Drupal\farm_ui_views\Plugin\Derivative\FarmTaxonomyTermViewsTaskLink
--- a/modules/core/ui/views/farm_ui_views.module
+++ b/modules/core/ui/views/farm_ui_views.module
@@ -278,5 +278,8 @@ function farm_ui_views_get_bundle_argument(ViewExecutable $view, string $display
   elseif ($view->id() == 'farm_log' && $display_id == 'page_asset' && !empty($args[1]) && $args[1] != 'all') {
     $bundle = $args[1];
   }
+  elseif (in_array($view->id(), ['farm_asset', 'farm_log']) && $display_id == 'page_term' && !empty($args[1]) && $args[1] != 'all') {
+    $bundle = $args[1];
+  }
   return $bundle;
 }
--- a/modules/core/ui/views/farm_ui_views.services.yml
+++ b/modules/core/ui/views/farm_ui_views.services.yml
@@ -13,6 +13,24 @@ services:
     arguments: [ '@entity_type.manager', '@asset.location' ]
     tags:
       - { name: access_check, applies_to: _asset_children_access }
+  farm_ui_views.asset_term_access:
+    class: Drupal\farm_ui_views\Access\FarmTaxonomyTermEntityViewsAccessCheck
+    arguments:
+     - 'asset'
+     - '@entity_type.manager'
+     - '@entity_type.bundle.info'
+     - '@entity_field.manager'
+    tags:
+      - { name: access_check, applies_to: _asset_term_access }
+  farm_ui_views.log_term_access:
+    class: Drupal\farm_ui_views\Access\FarmTaxonomyTermEntityViewsAccessCheck
+    arguments:
+     - 'log'
+     - '@entity_type.manager'
+     - '@entity_type.bundle.info'
+     - '@entity_field.manager'
+    tags:
+      - { name: access_check, applies_to: _log_term_access }
   farm_ui_views.asset_inventory_access:
     class: Drupal\farm_ui_views\Access\FarmInventoryAssetViewsAccessCheck
     arguments: [ '@entity_type.manager' ]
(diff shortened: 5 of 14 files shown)


modules/core/ui/views/config/install/views.view.farm_asset.yml
```diff
@@ -1720,6 +1720,111 @@ display:
         - url.query_args
         - user.permissions
       tags: {  }
+  page_with_term:
```
mstenta: Change to `page_term` for consistency with other display IDs (eg: `page_children`, `page_type`, `page_location`, etc).

modules/core/ui/views/config/install/views.view.farm_log.yml
```diff
@@ -2472,6 +2472,114 @@ display:
         - user
         - user.permissions
       tags: {  }
+  page_with_term:
```
mstenta: Change to `page_term` for consistency with other display IDs (eg: `page_asset`, `page_user`, `page_type`, etc).

modules/core/ui/views/config/install/views.view.farm_asset.yml
```diff
@@ -1720,6 +1720,111 @@ display:
...
+    id: page_with_term
+    display_title: 'By term (page)'
+    position: 4
+    display_options:
+      display_extenders: {  }
+      display_description: ''
+      path: taxonomy/term/%taxonomy_term/assets/%entity_bundle
+      arguments:
+        asset_taxonomy_term_reference:
+          id: asset_taxonomy_term_reference
```
mstenta: `farm_ui_views` needs to be added as a dependency of this config. Currently the config appears as "changed" in the Configuration updates report even after it is reverted. See diff at /admin/config/development/configuration/report/diff/view/farm_asset

modules/core/ui/views/config/install/views.view.farm_asset.yml
```diff
@@ -1720,6 +1720,111 @@ display:
         - url.query_args
         - user.permissions
       tags: {  }
+  page_with_term:
+    display_plugin: page
+    id: page_with_term
+    display_title: 'By term (page)'
+    position: 4
```
mstenta: Rearrange displays so that this appears before "Children (page)".

modules/core/ui/views/config/install/views.view.farm_log.yml
```diff
@@ -2472,6 +2472,114 @@ display:
...
+    display_plugin: page
+    id: page_with_term
+    display_title: 'By type (page)'
```
mstenta: This should be "By term (page)".

modules/core/ui/views/config/install/views.view.farm_asset.yml
```diff
@@ -1720,6 +1720,111 @@ display:
...
+          entity_type: asset
+          plugin_id: entity_taxonomy_term_reference
+        type:
+          id: type
+          table: asset_field_data
+          field: type
+          relationship: none
+          group_type: group
+          admin_label: ''
+          default_action: ignore
```
mstenta: This `default_action` is different in `farm_asset` vs `farm_log` Views. They should probably be the same?

modules/core/ui/views/config/install/views.view.farm_asset.yml
```diff
@@ -1720,6 +1720,111 @@ display:
...
+          exception:
+            value: all
+            title_enable: false
+            title: All
+          title_enable: false
+          title: ''
+          default_argument_type: fixed
+          default_argument_options:
+            argument: all
```
mstenta: This `argument` is also different between the Views (`all` vs `''`).

modules/core/ui/views/config/install/views.view.farm_asset.yml
mstenta: I think this one is correct ("Show "Page not found" when the `%entity_bundle` argument is not present in the URL). The `farm_asset` View should be changed to this as well.

modules/core/ui/views/farm_ui_views.links.task.yml
```diff
@@ -1,3 +1,4 @@
+# Add log view tabs to assets
```
mstenta: Add period at the end of comment.

modules/core/ui/views/farm_ui_views.links.task.yml
```diff
@@ -13,3 +14,21 @@ farm.asset.logs.all:
   parent_id: farm.asset.logs
 farm.asset.logs.type:
   deriver: Drupal\farm_ui_views\Plugin\Derivative\FarmLogViewsTaskLink
+
+# Add asset/log view tabs to taxonomy terms
```
mstenta: Add period at the end of comment.

modules/core/ui/views/farm_ui_views.links.task.yml
```diff
@@ -13,3 +14,21 @@ farm.asset.logs.all:
...
+farm.taxonomy_term.assets:
+  title: 'Assets'
+  route_name: view.farm_asset.page_with_term
```
mstenta: Change to `page_term` (per comment above).

modules/core/ui/views/farm_ui_views.links.task.yml
```diff
@@ -13,3 +14,21 @@ farm.asset.logs.all:
...
+  route_parameters:
+    entity_bundle: 'all'
+  base_route: entity.taxonomy_term.canonical
+  weight: 50
+farm.taxonomy_term.logs:
+  title: 'Logs'
+  route_name: view.farm_log.page_with_term
```
mstenta: Change to `page_term` (per comment above).

modules/core/ui/views/src/Access/FarmTaxonomyTermEntityViewsAccessCheck.php
```diff
@@ -0,0 +1,105 @@
...
+  /**
+   * A custom access check to filter out irrelevant entity bundles.
+   *
+   * @param \Drupal\Core\Routing\RouteMatchInterface $route_match
+   *   The route match.
+   */
+  public function access(RouteMatchInterface $route_match) {
+
+    // If there is no "asset" or "asset_type" parameter, bail.
+    $term_id = $route_match->getParameter('taxonomy_term');
```
mstenta, line 76: Change "asset" to "taxonomy_term".

modules/core/ui/views/src/Access/FarmTaxonomyTermEntityViewsAccessCheck.php
```diff
@@ -0,0 +1,105 @@
...
+    $this->entityFieldManager = $entity_field_manager;
+  }
+
+  /**
+   * A custom access check to filter out irrelevant entity bundles.
+   *
+   * @param \Drupal\Core\Routing\RouteMatchInterface $route_match
+   *   The route match.
+   */
+  public function access(RouteMatchInterface $route_match) {
```
mstenta, line 73: I'd like to see some more comments in this class, for consistency with the others.

modules/core/ui/views/farm_ui_views.module
```diff
@@ -294,5 +294,8 @@ function farm_ui_views_get_bundle_argument(ViewExecutable $view, string $display
   elseif ($view->id() == 'farm_log' && $display_id == 'page_asset' && !empty($args[1]) && $args[1] != 'all') {
     $bundle = $args[1];
   }
+  elseif (in_array($view->id(), ['farm_asset', 'farm_log']) && $display_id == 'page_with_term' && !empty($args[1]) && $args[1] != 'all') {
```
mstenta: Change to `page_term` (per comment above).

modules/core/ui/views/farm_ui_views.services.yml
```diff
@@ -13,6 +13,24 @@ services:
     arguments: [ '@entity_type.manager', '@asset.location' ]
     tags:
       - { name: access_check, applies_to: _asset_children_access }
+  farm_ui_views.asset_with_term_access:
```
mstenta: Change to `asset_term_access`, for consistency.

modules/core/ui/views/farm_ui_views.services.yml
```diff
@@ -13,6 +13,24 @@ services:
...
+    class: Drupal\farm_ui_views\Access\FarmTaxonomyTermEntityViewsAccessCheck
+    arguments:
+     - 'asset'
+     - '@entity_type.manager'
+     - '@entity_type.bundle.info'
+     - '@entity_field.manager'
+    tags:
+      - { name: access_check, applies_to: _asset_with_term_access }
+  farm_ui_views.log_with_term_access:
```
mstenta: Change to `log_term_access`, for consistency.

modules/core/ui/views/farm_ui_views.services.yml
```diff
@@ -13,6 +13,24 @@ services:
     arguments: [ '@entity_type.manager', '@asset.location' ]
     tags:
       - { name: access_check, applies_to: _asset_children_access }
+  farm_ui_views.asset_with_term_access:
+    class: Drupal\farm_ui_views\Access\FarmTaxonomyTermEntityViewsAccessCheck
+    arguments:
```
mstenta, line 18: Make `arguments` inline array for consistency.

modules/core/ui/views/farm_ui_views.services.yml
```diff
@@ -13,6 +13,24 @@ services:
...
+     - 'asset'
+     - '@entity_type.manager'
+     - '@entity_type.bundle.info'
+     - '@entity_field.manager'
+    tags:
+      - { name: access_check, applies_to: _asset_with_term_access }
+  farm_ui_views.log_with_term_access:
+    class: Drupal\farm_ui_views\Access\FarmTaxonomyTermEntityViewsAccessCheck
+    arguments:
```
mstenta, line 27: Make `arguments` inline array for consistency.

modules/core/ui/views/src/Routing/RouteSubscriber.php
```diff
@@ -28,6 +28,18 @@ public function alterRoutes(RouteCollection $collection) {
       $route->setRequirement('_asset_children_access', 'Drupal\farm_ui_views\Access\FarmAssetChildrenViewsAccessCheck::access');
     }
 
+    // Add our _asset_with_term_access requirement to
```
mstenta: Remove `_with` per comments above.

modules/core/ui/views/farm_ui_views.views_execution.inc
```diff
@@ -107,6 +107,41 @@ function farm_ui_views_views_pre_render(ViewExecutable $view) {
     }
   }
 
+  // If this is the farm_asset/farm_log View and page_with_term display, include
```
mstenta: Remove `_with` per comments above.

modules/core/ui/views/farm_ui_views.views_execution.inc
```diff
@@ -107,6 +107,41 @@ function farm_ui_views_views_pre_render(ViewExecutable $view) {
...
+        $bundles = \Drupal::service('entity_type.bundle.info')->getBundleInfo($view->getBaseEntityType()->id());
+        if (!empty($bundles[$entity_bundle])) {
+          $entity_bundle_label = $bundles[$entity_bundle]['label'] . ' ' . $view->getBaseEntityType()->getPluralLabel();
+        }
+      }
+
+      if (!empty($entity_bundle_label)) {
+        $title = t('%bundle with %vocab term %term', [
+          '%bundle' => $entity_bundle_label,
+          '%vocab' => $vocabulary->get('name'),
```
mstenta: Is there a `$vocabulary->get('label')`?

modules/core/ui/views/farm_ui_views.views_execution.inc
```diff
@@ -107,6 +107,41 @@ function farm_ui_views_views_pre_render(ViewExecutable $view) {
...
+          '%term' => $term->label(),
+        ]);
+      }
+      else {
+        $title = t('%base_type with %vocab term %term', [
+          '%base_type' => ucwords($view->getBaseEntityType()->getPluralLabel()),
```
mstenta: Will `ucwords()` here affect translations?

modules/core/ui/views/farm_ui_views.views_execution.inc
mstenta: (And the one below this as well.)

modules/core/ui/views/farm_ui_views.views_execution.inc
mstenta: @paul121 said we can use the `collection` label for the entity type. Not sure what the exact method name is... `getCollectionLabel()`?

mstenta: https://github.com/farmOS/farmOS/blob/9f7823354c9ea273a8f9c5e04fbb173cb0cc20c5/modules/core/asset/src/Entity/Asset.php#L21

modules/core/ui/views/src/Plugin/views/argument/EntityTaxonomyTermReferenceArgument.php
```diff
@@ -0,0 +1,176 @@
...
+                continue;
+              }
+
+              $column_names = $entity_table_mapping->getColumnNames($field_id);
+
+              $target_id_column_name = $column_names['target_id'];
+
+              // Join the taxonomy reference field table with the entity.
+              /** @var \Drupal\views\Plugin\views\join\JoinPluginBase $join */
+              $join = Views::pluginManager('join')->createInstance('standard', [
```
mstenta, line 154: Talked on the dev call with @symbioquine and @paul121 today - and wondered if these joins could be defined as relationships in `hook_views_data()`...

https://git.drupalcode.org/project/drupal/-/blob/9.3.x/core/modules/views/views.api.php#L176

https://www.flocondetoile.fr/blog/add-relationship-view-programmatically-drupal-8

modules/core/ui/views/farm_ui_views.views_execution.inc
mstenta: https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Entity%21EntityType.php/function/EntityType%3A%3AgetCollectionLabel/9.2.x

modules/core/ui/views/src/Plugin/views/argument/EntityTaxonomyTermReferenceArgument.php
```diff
@@ -0,0 +1,176 @@
...
+      $container->get('entity_type.manager')->getStorage('taxonomy_term'),
+      $container->get('entity_type.manager'),
+      $container->get('entity_type.bundle.info'),
+      $container->get('entity_field.manager'));
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function query($group_by = FALSE) {
```
mstenta, line 89: Would love to see more comments in this code so it can be read/understood more quickly (and for consistency).

mstenta (changes requested):
@symbioquine @paul121 and I reviewed this together on the dev call today.

Most of the comments here are small nit picks. Aside from those there were two followups:

- Try recreating/exporting Views config changes via UI to ensure that the config is "clean" (rather than copying/pasting YML directly).
- Investigate if `hook_views_data()` can be used to describe the relationships instead of explicitly building joins in the argument class. This is not a blocker, but worth at least considering.

Other than that I think this looks great @symbioquine !! Makes sense conceptually and it makes sense how you implemented everything! Well done!

modules/core/ui/views/farm_ui_views.services.yml
symbioquine: It seems like the inline array gets unwieldy above maybe 3 arguments...

**Expanded:**

image

**Inline:**

image

I don't feel strongly about it, but I find it easier to read the expanded version. It also produces cleaner diffs later if arguments are added/removed.

modules/core/ui/views/src/Plugin/views/argument/EntityTaxonomyTermReferenceArgument.php
symbioquine: I think I need a bit more clarification here...

What I thought I understood from the conversation on the call was that there might be some way to leverage the existing Views configuration that defines the entity/field table relationships to avoid needing to specify a join at the `EntityQuery` layer - as I'm doing currently.

However, in looking at both that blog post and the `hook_views_data` documentation, I think I'm seeing that code doing exactly the same thing. i.e. the code I might add in `hook_views_data` (or `hook_views_data_alter`) would still need to know about (or look up) the relevant table/column names.

If that's true, then it would just be a question of where it is cleanest to put the code that adds the join.

In `hook_views_data` I'd presumably need to have some conditional logic based on the view display (e.g. `page_term`) to add the relevant joins and separately I'd still need code in the views argument plugin to add the `WHERE` clauses.

Assuming my understanding above is correct, it seems cleaner to keep all the join/query code in the argument plugin since then the logic is all in one place and views takes care of only invoking the plugin for displays which use the argument.

Am I missing something?

modules/core/ui/views/farm_ui_views.services.yml
mstenta: I agree - I like the non-inline better - we've just been doing them all inline thusfar. But this doesn't need to be a blocker. I don't feel that strongly either - mostly just a consistency thing. And maybe we can do a follow-up commit to change other ones to non-inline.

modules/core/ui/views/src/Plugin/views/argument/EntityTaxonomyTermReferenceArgument.php
mstenta: Yea I think the main question was whether or not we could separate the "definition of the relationship" from the "adding of the JOIN", if that makes sense. But I don't know if it warrants reworking this right now.

`hook_views_data()` (and `hook_views_data_alter()`) are typically where you describe the relationships between tables. These hooks only run when the Views data cache is rebuilt (which happens less frequently than content caches being rebuilt, as would be the case with the logic in the Views argument). So *maybe* there would be a slight performance benefit to describing the relationships in there? It would also mean that those relationships are available for use in other code, potentially.

As for actually *adding* the JOIN, that would still need to happen in the Views argument class regardless, once you know which tables you need. But if the relationship is already described to Views in `hook_views_data()` then adding the join itself would be simpler (just a single method call I believe).

So, the two possible advantages to using `hook_views_data()`:

1. Performance improvement (but maybe only if it avoids running expensive field data lookups in the Views argument handler)
2. Reusable relationships (if someone wanted to build a custom View that used one of these relationships in the future, for instance)

Full disclosure: I'm just describing the high-level thinking here, but haven't fully thought through the specific details of this case, so it could be that those too advantages are not even applicable in this case.

In either case, I'm fine with just leaving the logic in one place for now. We can always consider moving it later, if there were a need. :-)

symbioquine: Wow, that's cool! Thanks for taking the time to explain it.

symbioquine:
In the most recent force push I just rebased against the current `farmOS/2.x` branch.

**Remaining stuff to do overall for this PR:**

* ~~Add updates to `CHANGELOG.md`~~
* ~~Retest changes locally from scratch~~
* Get feedback from @mstenta to see if I've sufficiently addressed existing review comments + any new ones
* Rebase changes into a single commit to prepare for merging

mstenta (approved):

mstenta:
This looks good to me @symbioquine - thanks again!!

> Rebase changes into a single commit to prepare for merging

I'm hoping to tag 2.0.0-beta4 very soon (and hopefully include this) - would you like me to squash these commits and go ahead and merge? Or if you would like to please do!

This is a great addition. :-)

symbioquine:
> I'm hoping to tag 2.0.0-beta4 very soon (and hopefully include this) - would you like me to squash these commits and go ahead and merge? Or if you would like to please do!

I don't feel strongly about who does that part, but I'm happy to...

Coming right up!

symbioquine:
The latest force-push is the same as my previous commits, but squashed and rebased against the main `farmOS:2.x` branch.

mstenta:
Merged! :tada:

PR #462, opened 2021-11-01
Fix quick form development/module link

paul121:
Just a little typo. The current menu link is broken: https://docs.farmos.org/development/modules/quick.md


mstenta:
Good catch - thanks @paul121!

PR #465, opened 2021-12-15
Fix bugs after *_field_allowed_values changes

paul121:
After updating our farm_land_type_field_allowed_values function we forgot to update this usage in the KML importer. I've checked for other usages of these *_field_allowed_values functions but did not see any others that we haven't already updated. Hopefully there aren't too many in 2.x contrib land.

Also, I forgot to update this variable name in the EntityFlagActionForm.


mstenta:
Thanks @paul121 - merged.

PR #467, opened 2022-01-03
Add allow-plugins config.

paul121:
Rather than define this in our composer project, lets define it as a part of our build process. This way downstream users of the `farmos/composer-project` will have to allow the plugins themselves.

Follow up from https://github.com/farmOS/composer-project/pull/6


paul121:
Currently we get the following message: https://github.com/farmOS/farmOS/runs/4694856871?check_suite_focus=true#step:6:1031

> For additional security you should declare the allow-plugins config with a list of packages names that are allowed to run code. See https://getcomposer.org/allow-plugins
You have until July 2022 to add the setting. Composer will then switch the default behavior to disallow all plugins.

The proposed changes remove this message: https://github.com/farmOS/farmOS/runs/4695291089?check_suite_focus=true#step:7:608

mstenta:
Merged - thanks @paul121!

PR #469, opened 2022-01-05
Enforce that the changelog is updated with every pull request

mstenta:
Copied from https://github.com/farmOS/farmOS-map/pull/140


mstenta:
Failed as expected, because I didn't include a `CHANGELOG.md` update... :-)

Now I will force-push with one and it should pass...

PR #470, opened 2022-01-05
Open the Gin toolbar by default 

mstenta:
Currently the Gin toolbar appears collapsed by default. When you hover over *some* of the items, it opens a drawer with a title in the top to let you know what it is. But others (like the "Locations" item at the top) do not. It's not apparent what those items are without clicking on them. And it's not obvious that you can expand the toolbar by clicking the little > button at the bottom.

I propose we open the Gin toolbar by default, so that users can make the choice to close it if they want. This can be accomplished by simply setting the `GinSidebarOpen` local storage item to `true` if it  `=== null` on page load. This gets picked up by Gin's `gin_toolbar.js` behavior, which will then open the toolbar. Thus, this JavaScript needs to be loaded *before* the Gin JS (hence the `weight: -50` in this change).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- [Open the Gin toolbar by default #470](https://github.com/farmOS/farmOS/pull/470)
 - [Enforce that the changelog is updated with every pull request #469](https://github.com/farmOS/farmOS/pull/469)
 
 ### Fixed
--- a/modules/core/ui/theme/farm_ui_theme.libraries.yml
+++ b/modules/core/ui/theme/farm_ui_theme.libraries.yml
@@ -30,3 +30,5 @@ toolbar:
   css:
     theme:
       css/toolbar.css: { }
+  js:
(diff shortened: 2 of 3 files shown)


paul121 (approved):
+1 looks good to me!

I also tested this with the upcoming Gin design changes on drupal/gin 3.x-dev. The toolbar behaves the same as described above, but the local storage variable has changed from `GinSidebarOpen` to [`Drupal.gin.toolbarExpanded`](https://git.drupalcode.org/project/gin/-/blob/8.x-3.x/js/toolbar.js#L14)

I'm unsure when the next release of Gin will be (hopefully before our beta2?) but I don't think that needs to block this. Just something to flag, maybe add a `TODO` here for that, or open a new draft PR.

mstenta:
> the local storage variable has changed from `GinSidebarOpen` to [`Drupal.gin.toolbarExpanded`](https://git.drupalcode.org/project/gin/-/blob/8.x-3.x/js/toolbar.js#L14)

Wow thanks for flagging that @paul121!

Maybe I'll add a condition for that name as well, so both work, along with an `@todo` to remove the old one in the future.

mstenta:
@paul121 I force pushed with an additional commit for the new variable name.

paul121:
LGTM!

PR #471, opened 2022-01-05
Do not include archived assets in metrics count

mstenta:
See https://farmos.discourse.group/t/asset-metrics-on-homepage-dashboard/1029

### DIFF
--- a/modules/core/ui/metrics/src/Plugin/Block/FarmMetricsBlock.php
+++ b/modules/core/ui/metrics/src/Plugin/Block/FarmMetricsBlock.php
@@ -74,7 +74,7 @@ class FarmMetricsBlock extends BlockBase implements ContainerFactoryPluginInterf
 
     // Create a container for asset metrics.
     $output['asset'] = [
-      '#markup' => '<strong>' . Link::createFromRoute('Assets', 'view.farm_asset.page')->toString() . '</strong>',
+      '#markup' => '<strong>' . Link::createFromRoute('Assets', 'view.farm_asset.page')->toString() . ' (active)</strong>',
       'metrics' => [
         '#type' => 'container',
         '#attributes' => [
@@ -140,11 +140,16 @@ class FarmMetricsBlock extends BlockBase implements ContainerFactoryPluginInterf
 
     // Count records by type.
     foreach ($bundles as $bundle => $bundle_info) {
-      $count = $this->entityTypeManager->getStorage($entity_type)->getAggregateQuery()
+      $query = $this->entityTypeManager->getStorage($entity_type)->getAggregateQuery()
         ->accessCheck(TRUE)
-        ->condition('type', $bundle)
-        ->count()
-        ->execute();
+        ->condition('type', $bundle);
+
+      // Only include active assets.
+      if ($entity_type == 'asset') {
+        $query->condition('status', 'active');
+      }
+
+      $count = $query->count()->execute();
       $route_name = "view.farm_$entity_type.page_type";
       $metrics[] = Link::createFromRoute($bundle_info['label'] . ': ' . $count, $route_name, ['arg_0' => $bundle], ['attributes' => ['class' => ['metric', 'button']]])->toString();
     }
--- a/modules/core/ui/metrics/src/Plugin/Block/FarmMetricsBlock.php
+++ b/modules/core/ui/metrics/src/Plugin/Block/FarmMetricsBlock.php
@@ -74,7 +74,7 @@ class FarmMetricsBlock extends BlockBase implements ContainerFactoryPluginInterf
 
     // Create a container for asset metrics.
     $output['asset'] = [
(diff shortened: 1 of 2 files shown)


paul121:
I don't have a strong opinion on this. Especially considering that a user is requesting this change, I think that should be considered more important than my opinion!

...but I feel that this could create some complications if and when metrics are used more generally, especially via the API. This would no longer represent "assets" but really just "active assets" and should thus be renamed accordingly (as is proposed).

I imagine we would create a much more robust `asset` metric (perhaps showing both active & archived counts) down the road, but alas, just wanted to mention while it was on my mind.

modules/core/ui/metrics/src/Plugin/Block/FarmMetricsBlock.php
```diff
@@ -74,7 +74,7 @@ public function build() {
 
     // Create a container for asset metrics.
     $output['asset'] = [
-      '#markup' => '<strong>' . Link::createFromRoute('Assets', 'view.farm_asset.page')->toString() . '</strong>',
+      '#markup' => '<strong>' . Link::createFromRoute('Assets', 'view.farm_asset.page')->toString() . ' (active)</strong>',
```
paul121: We should translate `(active)`, and also likely `Assets` (and `Logs` below).

This might be related to this report: https://farmos.discourse.group/t/farmos-2-0-translation-localization-l10n-initiative/922/7?u=paul121

paul121 (changes requested):

Farmer-Eds-Shed:
> I don't have a strong opinion on this. Especially considering that a user is requesting this change, I think that should be considered more important than my opinion!
> 
> ...but I feel that this could create some complications if and when metrics are used more generally, especially via the API. This would no longer represent "assets" but really just "active assets" and should thus be renamed accordingly (as is proposed).
> 
> I imagine we would create a much more robust `asset` metric (perhaps showing both active & archived counts) down the road, but alas, just wanted to mention while it was on my mind.

Hi Paul, hope you don't mind me butting in here.
In farmOS 1.x the metric on the dashboard only shows active assets, it makes more sense, as a livestock farmer the vast majority of the time I'm only going to be interested in active animals. The archived animals are important maybe once or twice a year for accounts and on occasion for comparing parentage/lineage of active animals the number of them is going to be a less important as it is only going to be a metric of how many animals I've had since the database was created (might be interesting down the line but not a useful number for daily use).

mstenta:
Thanks @paul121 good feedback/considerations!

> ...but I feel that this could create some complications if and when metrics are used more generally, especially via the API. This would no longer represent "assets" but really just "active assets" and should thus be renamed accordingly (as is proposed).

I think the broader "metrics" API stuff we've been discussing will require refactoring this stuff anyway - so I'm not too worried. Right now there is no such thing as a "metric" per say. It's just a render element in the block currently. I intentionally kept this simple so that we could refactor it later: https://www.drupal.org/project/farm/issues/3217715

Here is the broader issue for refactoring: https://www.drupal.org/project/farm/issues/3176697

> I imagine we would create a much more robust `asset` metric (perhaps showing both active & archived counts) down the road, but alas, just wanted to mention while it was on my mind.

Yea glad you brought it up! That's what I'm thinking too. Maybe the `asset` metric in the future will include a *bunch* of different info in it's output (active/archived counts for each asset type, totals, etc).

Then we just need to figure out how to *display* that in the metrics block on the dashboard, which would essentially just be a simplified display of that info - to match what we have now.

So I think it will all fit together when we get there, since we'll need to refactor it all anyway when we actually create a plugin system for metrics. Definitely some steps to get there though... I'll link to this discussion from https://www.drupal.org/project/farm/issues/3176697 for posterity.

--

Updating my PR to fix the translation issue now... standby...

modules/core/ui/metrics/src/Plugin/Block/FarmMetricsBlock.php
```diff
@@ -74,7 +74,7 @@ public function build() {
 
     // Create a container for asset metrics.
     $output['asset'] = [
-      '#markup' => '<strong>' . Link::createFromRoute('Assets', 'view.farm_asset.page')->toString() . '</strong>',
+      '#markup' => '<strong>' . Link::createFromRoute($this->t('Assets'), 'view.farm_asset.page')->toString() . ' (' . $this->t('active') . ')</strong>',
```
paul121: Oh sorry... didn't remember this earlier. Should we use the entity "collection" label?

https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Entity%21EntityType.php/function/EntityType%3A%3AgetCollectionLabel/9.2.x

Was thinking back to this: https://github.com/farmOS/farmOS/pull/458#discussion_r747697108

paul121 (changes requested):

mstenta:
> Should we use the entity "collection" label?

Good point @paul121! I pushed a commit that makes this change as well.

I think this is ready to merge if it looks good to you!

paul121 (approved):

PR #475, opened 2022-01-06
farmOS 2.x l10n module

mstenta:
This is a draft PR for a new `farm_l10n` module for farmOS `2.x`, which would serve the same purpose as the module of the same name from the `7.x-1.x` branch: essentially enabling the modules we know we need for localization, and provide some default configuration for them.

### DIFF
--- a/modules/core/l10n/src/Form/L10nSettingsForm.php
+++ b/modules/core/l10n/src/Form/L10nSettingsForm.php
@@ -75,10 +75,12 @@ class L10nSettingsForm extends NegotiationSelectedForm {
   public function submitForm(array &$form, FormStateInterface $form_state) {
     parent::submitForm($form, $form_state);
 
-    // Initiate a batch operation to update the default language of users.
+    // Initiate a batch operation to update the default language of all users
+    // (except user 1).
     if ($form_state->getValue('update_existing_users')) {
       $operations = [];
-      $uids = $this->entityTypeManager->getStorage('user')->getQuery()->execute();
+      $query = $this->entityTypeManager->getStorage('user')->getQuery();
+      $uids = $query->condition('uid', '1', '!=')->execute();
       foreach ($uids as $uid) {
         $operations[] = [
           [__CLASS__, 'updateUserLanguage'],
--- a/modules/core/l10n/src/Form/L10nSettingsForm.php
+++ b/modules/core/l10n/src/Form/L10nSettingsForm.php
@@ -1,0 +1,112 @@
+<?php
+
+namespace Drupal\farm_l10n\Form;
+
+use Drupal\Core\Config\ConfigFactoryInterface;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\language\Form\NegotiationSelectedForm;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Configure the selected language negotiation method for this site.
+ *
+ * @internal
+ */
+class L10nSettingsForm extends NegotiationSelectedForm {
+
+  /**
+   * The entity type manager service.
+   *
+   * @var \Drupal\Core\Entity\EntityTypeManagerInterface
+   */
+  protected $entityTypeManager;
+
+  /**
+   * Constructor for L10nSettingsForm.
+   *
+   * @param \Drupal\Core\Config\ConfigFactoryInterface $config_factory
+   *   The factory for configuration objects.
+   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_type_manager
+   *   The entity type manager service.
+   */
+  public function __construct(ConfigFactoryInterface $config_factory, EntityTypeManagerInterface $entity_type_manager) {
+    parent::__construct($config_factory);
+    $this->entityTypeManager = $entity_type_manager;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function create(ContainerInterface $container) {
+    return new static(
+      $container->get('config.factory'),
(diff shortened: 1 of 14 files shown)


mstenta:
Ah this is actually broken right now... because it's trying to install config that already exists...

```
In PreExistingConfigException.php line 65:
                                                                                                      
  Configuration objects (language.types) provided by farm_l10n already exist in active configuration
```

So we'll need to make our config adjustments in `hook_install()` instead...

mstenta:
Huh strange... I'm also finding that I'm unable to install the core `config_translation` module just by itself due to a PHP error:

```
Error: Call to a member function getPath() on null in Drupal\config_translation\ConfigNamesMapper->getOverviewRoute() (line 247 of /opt/drupal/web/core/modules/config_translation/src/ConfigNamesMapper.php).
```

mstenta:
Force pushed to remove the `language.types` config (with a `wip` commit for modifying it via `hook_install()` which still needs to be completed). I also added `farm_l10n` to the list of optional modules so that it will be installed during automated testing. Curious if the `config_translation` install issue is unique to my local environment for some reason... let's see if the same thing happens in GitHub Actions.

mstenta:
Updates:

- Sets the default language detection rules via `hook_install()`: disables URL, and enables Session, User, and Selected language methods.
- Adds `farm_l10n` to optional modules list so it appears during install and in `/farm/settings/modules`.
- Provide a `/farm/settings/language` form for setting the "Selected language" config.
- Prevents changing the default site language (see issue I opened in farmOS drupal.org queue for future reference: https://www.drupal.org/project/farm/issues/3257430)

mstenta:
We should think through whether or not we want to provide the ability to "Add language" via the `/farm/settings/language` form too perhaps.

mstenta:
Another thing to consider: right now the `locale.settings` configuration that is installed by the `locale` module sets the `path` for downloaded translations to `sites/default/files/translations`. This works fine as a default, but in some environments it might not make sense. Perhaps this is more advanced, though, and is outside the scope of the `farm_l10n` module?

mstenta:
Rebased and added a CHANGELOG.md entry. This PR is ready for review. I will take it out of draft status...

modules/core/l10n/src/Form/L10nSettingsForm.php
```diff
@@ -0,0 +1,112 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function submitForm(array &$form, FormStateInterface $form_state) {
+    parent::submitForm($form, $form_state);
+
+    // Initiate a batch operation to update the default language of users.
+    if ($form_state->getValue('update_existing_users')) {
+      $operations = [];
+      $uids = $this->entityTypeManager->getStorage('user')->getQuery()->execute();
```
paul121: Should we exclude user id = 1 from this functionality? If I was hosting a farmOS instance for someone I wouldn't want them to change *my* language to something I don't know :-)

paul121 (changes requested):
This all looks great! I tested the following:
- Enforce `en` as the default language, verify that is skipped in the installation UI.
- With additional languages installed, test installing additional core log/asset types as a "manager" user.  I wanted to see that the additional "updating config translations" batch operation works OK - the progress bar is a little funky, goes back and forth... but not a blocker for this.
- Language settings form and the update existing users functionality.

One issue I found is that `config_translation` breaks things when `field_ui` is installed. Lots of errors, and many of the admin pages end up broken:

```shell
$ docker compose exec -u www-data www drush si --account-pass=admin farm farm.modules='all' --yes
$ docker compose exec -u www-data www drush en field_ui
Error: Call to a member function getPath() on null in /opt/drupal/web/core/modules/config_translation/src/ConfigNamesMapper.php on line 251 #0 /opt/drupal/web/core/modules/config_translation/src/Routing/RouteSubscriber.php(39): Drupal\config_translation\ConfigNamesMapper->getOverviewRoute()
```

This issue & patch helped me get an error message that was useful: https://www.drupal.org/project/drupal/issues/3016131#comment-13645477

> Base route not available for route entity.field_config.quantity_field_edit_form 

The problem is that we specify that the `quantity` entity has a `field_ui` base route, but on the `quantity_type` config entity we never define the CRUD form or routes necessary for `field_ui`. Not an issue with this PR per say, but we do need to address it.

It is fixed by removing this line: https://github.com/farmOS/farmOS/blob/12318d642a546a368eb4f0b7697ec043c858ca6b/modules/core/quantity/src/Entity/Quantity.php#L65

mstenta:
> One issue I found is that `config_translation` breaks things when `field_ui` is installed.

Ah ha!!! I was experiencing this issue locally as well and couldn't figure out what it was! I tried replicating on a clean codebase install and couldn't, so I thought it was something wrong with my slightly hacked Drupal core. Well done tracing this @paul121!

> The problem is that we specify that the `quantity` entity has a `field_ui` base route, but on the `quantity_type` config entity we never define the CRUD form or routes necessary for `field_ui`. Not an issue with this PR per say, but we do need to address it.

Indeed! I'll open an issue...

modules/core/l10n/src/Form/L10nSettingsForm.php
mstenta: Good idea!

mstenta:
- #480

modules/core/l10n/src/Form/L10nSettingsForm.php
mstenta: Resolved via https://github.com/farmOS/farmOS/pull/475/commits/a52eb225062f24976bd55bd609c1f4a85b9fa2f6

paul121 (approved):

mstenta:
Merged!

Issue #472, opened 2022-01-06
Translation in farmOS 2.x: Wierd behavior


SirSundays:
Some of the farmOS stuff is not translated, even when there is a translation in the po-file.
After a while, some of the translations seem to appear in the UI, but there are also some appearing in the list of missing translations after visiting their corresponding site.

I have some more details here on the forum: https://farmos.discourse.group/t/farmos-2-0-translation-localization-l10n-initiative/922/7?u=sirsundays

The only two important things that seem to be missing in the translation are on the dashboard the button in the top right corner "Add asset" and "Add log". When I click on them the right translations are displayed in the top row.
I would expect that these are the same as the ones displayed on the button. But still one is working while the other is not.

**Screenshots**
![grafik](https://user-images.githubusercontent.com/49527200/148398691-06822ed4-ab0e-4342-a9a2-6987240ea499.png)
![grafik](https://user-images.githubusercontent.com/49527200/148399201-1028bd99-8489-49ec-8040-649b496bd2c3.png)

Here is what I mean. Everything is **now** translated (maybe it was because of a new browser startup (only tried to reload with clear cache)), except the button in the top right. Drupal shows 100% translation. Can't be trusted in my opinion.

**Desktop (please complete the following information):**
 - Browser Firefox
 - Version 2.x-beta1

But this doesn't seem to be a major bug so don't give it too much of a priority. 

-- Jan

mstenta:
Thanks @SirSundays!

I'll admit I'm very inexperienced with Drupal 8/9 localization, so we'll be learning together. :-)

I *may* have a hunch what's causing this problem, though... if you would be willing to test it?

Here is the line of code that is defining the "Add asset" and "Add log" (also "Add plan") link titles:

https://github.com/farmOS/farmOS/blob/8530496e7effb0383f4cda43bc8c7239384843e9/modules/core/ui/action/src/Plugin/Derivative/FarmActions.php#L39

We may just need to change that line to:

```
$this->derivatives[$name]['title'] = $this->t('Add :entity_type', [':entity_type' => Unicode::ucfirst($type)]);
```

Note that this also requires adding `use StringTranslationTrait;` to the top of the class definition so that it looks like this:

```
class FarmActions extends DeriverBase {

  use StringTranslationTrait;

  /**
   * {@inheritdoc}
   */
```

Would you like to give that a try and see if it works?

mstenta:
Oh and be sure to clear the cache after making that change so that it takes effect!

[...]

PR #477, opened 2022-01-07
Fix: Buttons not translateable

SirSundays:
Pull request mentioned here:
https://github.com/farmOS/farmOS/issues/472#issuecomment-1006725357

I´m still not 100% happy with this. The Button on the Dashboard still uses ucfirst, but I thought it's fine there for now because there are just two words.
The other one should maybe be changed more. Right now I removed the lowercase because I think this is language-dependent and should be done in translation. I know while I was translating I came across the phrase "Animal assets" and all the other assets as well. Maybe something like this should be used.


mstenta:
Thanks @SirSundays ! I went ahead and merged this, along with one additional change:

> The Button on the Dashboard still uses ucfirst, but I thought it's fine there for now because there are just two words.

I changed this to use `getLabel()` instead, so no more need for `ucfirst`.

PR #481, opened 2022-01-12
Log categories are not migrated to v2

mstenta:
I'm not sure how this went unnoticed, but it looks like the v2 migration of Log categories data is broken. The Log category taxonomy terms themselves are migrated properly, but none of them get assigned to logs created during the log migrations.

The bug is on this line of code:

https://github.com/farmOS/farmOS/blob/143967356080ff2259e53ece5a827be5aa1412d7/modules/core/migrate/config/install/migrate_plus.migration_group.farm_migrate_log.yml#L107

It is trying to assign categories to a field named `log_category` on logs, but that field does not exist. The field name should be simply `category`.

https://github.com/farmOS/farmOS/blob/143967356080ff2259e53ece5a827be5aa1412d7/modules/core/entity/modules/fields/farm_entity_fields.base_fields.inc#L77

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- [Log categories are not migrated to v2 #481](https://github.com/farmOS/farmOS/pull/481)
 - Make local action buttons translatable.
 - Fix permission for map settings form (/farm/settings/map).
 - Patch `jsonapi_schema` module to fix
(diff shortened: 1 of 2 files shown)


mstenta:
We may need to create a follow-up PR that provides an additional migration to fix this, if anyone has already gone through the migration process and lost their log categories due to this bug. It probably wouldn't be too hard to create a migration like that. But since it is still so early in the beta cycle (`2.0.0-beta1` was released less than 2 weeks ago), I say we wait and see if anyone comes forward that was affected. We may have caught it early enough that we don't need to do anything more. (With this in mind I would like to get `2.0.0-beta2` released ASAP.)

mstenta:
> We may need to create a follow-up PR that provides an additional migration to fix this, if anyone has already gone through the migration process and lost their log categories due to this bug.

I created a module: https://github.com/mstenta/farm_migrate_category_fix

To apply to the fix, install the module and run this Drush command (similar to the way migrations are run originally):

`drush migrate:import farm_migrate_log_field_category`

PR #482, opened 2022-01-14
Grant data stream and notification permissions.

### LINKED ISSUE #479: Data stream and notification permissions are not granted to managed roles
The `farm_role` module is responsible for granting a standard set of entity permissions to managed roles. It does this already for `asset`, `log`, `plan`, `taxonomy_term`, and `quantity` entity types. It does not cover `data_stream` entities. Nor does it cover data stream notifications. So neither of those are accessible to non-admin roles.

paul121:
Closes #479 

- Adds `data_stream` as a managed entity type.
- Adds all `{operation} data_stream_notification` permissions to roles with config access.

In the future it would be nice if the "managed entities" could support config entities and bundle-less entities as well. Right now our `role.*.third_party.farm_role` schema expects an array of bundles when declaring more granular permission sets.

Noting this here because I investigated a little bit. The logic for this could be fairly simple, based off `$entity_type->getBundleEntityType()`:
```php
          // ManagedRolePermissionsManager::getManagedPermissionsForRole() 
          switch ($entity_type) {

            case 'asset':
            case 'log':
            case 'plan':
            case 'quantity':
            case 'data_stream':
            case 'data_stream_notification':

              // Only include the bundle if necessary.
              if ($this->entityTypeManager->getDefinition($entity_type)->getBundleEntityType()) {
                if (array_intersect(['all', $bundle], $allowed_bundles)) {
                  $perms[] = $operation . ' ' . $bundle . ' ' . $entity_type;
                }
              } else {
                $perms[] = $operation . ' ' . $entity_type;
              }
              break;
 ```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- [Data stream and notification permissions are not granted to managed roles. #479](https://github.com/farmOS/farmOS/issues/479)
 - [Log categories are not migrated to v2 #481](https://github.com/farmOS/farmOS/pull/481)
 - Make local action buttons translatable.
 - Fix permission for map settings form (/farm/settings/map).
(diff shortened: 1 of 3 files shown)


paul121:
Force pushed to fix a typo and expand the switch statement explanations!

mstenta:
Merged - thanks @paul121 !

PR #487, opened 2022-01-16
Remove data_stream collection link to non-existent route. Fixes #486

### LINKED ISSUE #486: Route "entity.data_stream.collection" does not exist.
Encountered this error when trying to delete a data stream:

```
Symfony\Component\Routing\Exception\RouteNotFoundException: Route "entity.data_stream.collection" does not exist.
```

It appears that we declare a `collection` link path in the `data_stream` entity definition, but do not actually implement that route. We don't really need that route right now, so the easy fix is to just remove this line:

https://github.com/farmOS/farmOS/blob/a5c393b13b53b47fb5948b1d531a2722da442695/modules/core/data_stream/src/Entity/DataStream.php#L63

mstenta:
See https://github.com/farmOS/farmOS/issues/486

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- [Route "entity.data_stream.collection" does not exist. #486](https://github.com/farmOS/farmOS/issues/486)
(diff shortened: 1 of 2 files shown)


mstenta:
Force-pushed `CHANGELOG.md` entry.

paul121 (approved):

mstenta:
Merged: https://github.com/farmOS/farmOS/commit/ebde335b00618ec6aca816256f9cfebb3fb71a41

Issue #485, opened 2022-01-16
Data streams cannot be deleted or removed from sensor assets


mstenta:
It appears that the Entity Reference Integrity module is making it impossible to delete `data_stream` entities when they are referenced by Sensor assets. And when you try to "Remove" the data stream from the sensor asset, it tries to delete the data stream, which then runs into the same issue. So it's a catch-22, and it is not possible to delete data streams without first disabling the Entity Reference Integrity constraint for `data_stream` entities via `/admin/config/content/entity-reference-integrity`.

mstenta:
FWIW I only found this issue after fixing https://github.com/farmOS/farmOS/issues/486

mstenta:
Ah we have a relevant @todo:

https://github.com/farmOS/farmOS/blob/a5c393b13b53b47fb5948b1d531a2722da442695/modules/core/data_stream/data_stream.module#L91

It seems that this `hook_ENTITY_TYPE_delete()` intends to clean up references on sensor assets, but this code can't actually work because of Entity Reference Integrity currently.

mstenta:
I suppose we *could* consider removing the Entity Reference Integrity constraint for `data_stream` entities, knowing that that `hook_ENTITY_TYPE_delete()` logic is in place.

Although... that only covers `data_stream` entities referenced by `asset` entities in a field named `data_stream`. So it would miss any additional data stream reference fields added downstream.

Another idea: the "Remove" option on the Inline Entity Form is currently *deleting* the referenced entity - not just *removing* it. I imagine it's possible to configure it such that it only *removes* and does not *delete*. This would at least allow you to first remove it, and then delete it. The disadvantage of this, of course, is it assumes a user knows they need to do both. Simply *removing* the data stream from an asset without deleting the data stream means there would be orphaned data streams in the system. Not the end of the world... and maybe it's time we consider adding a `collection` route to show all data streams (along with assets that reference them)? Related: #486

paul121:
> Another idea: the "Remove" option on the Inline Entity Form is currently deleting the referenced entity - not just removing it.

Hmmm interesting. It seems important to alert the user that the data stream data will be deleted as well. Another option would be to exclude the `Remove` option from the inline entity form (removing and deleting from the sensor asset). We could add a note to the effect of "You can delete data stream entities from their individual pages." *Although this might conflict with reference integrity. Hmmmm

> and maybe it's time we consider adding a collection route to show all data streams (along with assets that reference them)? Related

[...]

PR #489, opened 2022-01-16
Delete data when basic data streams are deleted. Fixes #488

### LINKED ISSUE #488: Data is not deleted when a data stream entity is deleted
Steps to reproduce:

1. Create a sensor asset.
2. Copy `curl` command from "Developer information" and run it to post a data point.
3. Delete the `data_stream` entity that was created (need to disable entity reference integrity constraint first, see #485)
4. Look in the database, the data is still there.

mstenta:
See #488

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- [Data is not deleted when a data stream entity is deleted #488](https://github.com/farmOS/farmOS/issues/488)
 - [Data does not immediately appear when posting to data streams #484](https://github.com/farmOS/farmOS/issues/484)
 - [Route "entity.data_stream.collection" does not exist. #486](https://github.com/farmOS/farmOS/issues/486)
 - Fix Quick Form help text so that it works with new multi-route approach.
--- a/modules/core/data_stream/data_stream.module
+++ b/modules/core/data_stream/data_stream.module
@@ -88,12 +88,6 @@ function data_stream_data_stream_view_alter(array &$build, DataStreamInterface $
  */
 function data_stream_data_stream_delete(DataStreamInterface $data_stream) {
 
-  // @todo Considerations for improved entity reference integrity?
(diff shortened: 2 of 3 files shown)


paul121 (changes requested):
It seems that we had this when we had a separate `legacy` data stream bundle, but the `hook_data_stream_delete` implementation got dropped with this commit: https://github.com/farmOS/farmOS/commit/e528813330b58faa09282e8b8e5b71385f19c902

We should add a test for this so we don't forget it in the future.

This PR is nearly identical to the original code though, so LGTM otherwise!

```php
function farm_sensor_listener_data_stream_delete(DataStreamInterface $data_stream) {

  // Remove all legacy listener data provided by the data stream.
  if ($data_stream->bundle() == 'legacy_listener') {
    \Drupal::database()->delete('data_stream_legacy')
      ->condition('id', $data_stream->id())
      ->execute();
  }
}
```

paul121:
Added a test for this!

In the process, noticed that we have the logic for removing references to data streams in the wrong place, it should be in farm_sensor. Moved the logic and added another test for this, but running into the entity reference integrity issue... so I marked the test to be skipped for now, but this still causes our test runner to fail. Gotta run but wanted to push here for now...

mstenta:
Thanks @paul121! I separated this into two branches...

1. `2.x-delete-data-stream-data` with my original commit + your test for deleting data
2. `2.x-data-stream-references` with your two commits re: asset reference fields

I'll point to (2) from https://github.com/farmOS/farmOS/issues/485...

Running tests on this PR (`2.x-delete-data-stream-data`) now... I will merge if it passes.

mstenta:
Ah tests failed because `SQLSTATE[42P01]: Undefined table: 7 ERROR:  relation "test36388637asset__data_stream" does not exist` - this is probably what you ran into, huh @paul121 :-)

Whelp... I suppose we can just delete that original code from `hook_ENTITY_TYPE_delete()` (which is causing the test to fail), since we discovered in #485 that it isn't working as intended anyway (and maybe we'll end up taking a different approach to solve that issue, removing the need for that code).

paul121:
Precisely - that's perfect!

PR #490, opened 2022-01-17
Invalidate basic data stream cache tags when data is received. Fixes #484

### LINKED ISSUE #484: Data does not immediately appear when posting to data streams
I need to test this again, but earlier today I was trying to debug a data stream issue and was getting confused because the `POST` was succeeding, but data was not appearing in the UI. However, I came back later and the data *was* there. So either I did something wrong (very possible, I was distracted while testing :sweat_smile:) or there's some caching bug at play. I'm *pretty sure* I cleared the cache to see if that was the cause, but I need to test again to confirm. Just documenting this here so we can come back to it...

mstenta:
See #484

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- [Data does not immediately appear when posting to data streams #484](https://github.com/farmOS/farmOS/issues/484)
 - [Route "entity.data_stream.collection" does not exist. #486](https://github.com/farmOS/farmOS/issues/486)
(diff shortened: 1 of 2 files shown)


paul121:
Alternatively, we could disable caching on the `data_stream_basic_data` view altogether. That should solve this issue.

Listener sensor data wasn't cached previously, and caching this block view could be quite large when there is lots of data. I think it would be more correct to provide an additional cache tag just for data stream data eg: `data_stream:100:data` as a separate feature request

mstenta:
> Alternatively, we could disable caching on the `data_stream_basic_data` view altogether. That should solve this issue.

That makes sense to me... no need to cache this. I've updated my branch to take this approach. I'll merge that - it's very simple. Closing this.

PR #491, opened 2022-01-17
Improvements to sensor and data stream developer info

mstenta:


paul121 (approved):
Good improvements!

I have slight hesitation to *always* show developer information on sensor assets. Two ideas:
- When a sensor asset has different data stream types associated with it and `basic` is not being used. Perhaps these other data stream types should be able to provide their own "developer information".
- The use case when a farmOS site has been "configured" behind the scenes by an admin user, and the farm user does not need to see developer information.

But unless you have quick solutions to either of these, I don't think these ideas should be a blocker right now.

mstenta:
Hmm yea that's a good point. The disadvantage right now is that there is no indication that data can be pushed directly to sensors, and data streams can be automatically generated. This will still be possible even if non-basic data streams are used.

Perhaps we could add a special permission for the developer information visibility? And assign it automatically to "config" users, but not others?

Either that or have a site-wide configuration for turning them off...

It seems to me that they should be visible by default, because that will represent the majority of use-cases. But allowing them to be hidden or turned off does make sense for other cases.

paul121:
A permission could work.. something like `configure data stream`.. although just thinking.. maybe we only display the developer information if the user has `edit` access on the data stream?

> The use case when a farmOS site has been "configured" behind the scenes by an admin user, and the farm user does not need to see developer information.

In this case the user would not need `edit` permission on the data stream, they only need permission to view it. And really, they only need permission to view the data - not the public/private keys (or additional configuration about the data stream).

Oohh, maybe **more importantly** if you were to configure a `farm_viewer` user, they shouldn't be posting data... and not have access to the developer info..

So a dedicated permission could work, but until we have additional reasons for such a permission, should we just use the `edit` permission for now?

mstenta:
> maybe we only display the developer information if the user has `edit` access on the data stream?

This makes more sense and solves the other issues you mentioned! :-)

mstenta:
Merged: https://github.com/farmOS/farmOS/commit/274f4d40473be83bdc214f9fdf9220f91067e3be

PR #492, opened 2022-01-19
Port Planting Quick Form to v2

mstenta:


modules/core/quick/tests/modules/farm_quick_test/src/Plugin/QuickForm/Test.php
```diff
@@ -65,7 +65,7 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
         [
           'measure' => 'count',
           'value' => $value,
-          'units' => $this->t('tests'),
+          'units' => 'tests',
```
paul121, line 68: Interesting.. I was curious about sending the result of `t('string')` (translatable markup) to `QuickStringTrait::trimString()` or `QuickStringTrait::prioritizedString()` and if that would break things, but it seems like it doesn't.

But now seeing this, I'm curious why did *this* change, did it break something? Is this something that needs to be supported? Maybe if you have a quick form that has a `unit` string hard-coded, but you also want that unit to be translated? (but also... translating here may not be a good idea. in this case the `unit` term content should be translated I think..)

> \Drupal\Core\StringTranslation\TranslatableMarkup An object that, when cast to a string, returns the translated string.
https://api.drupal.org/api/drupal/core%21includes%21bootstrap.inc/function/t/10.0.x

Looking at the docs on this maybe it is because the `QuickStringTrait` functions are casting the translatable markup to a string? My hunch is that this would work: `'units' => (string) $this->t('tests')`. That seems like a fair workaround. Mostly thinking out loud here, I don't think there is any problem!

paul121 (approved):
:+1: 

The only thing we chatted about was maybe changing the `date` form element to `datetime` with `'#date_time_element' => 'none'` to remove the time widget. In this case we would need to make sure the time defaults to midnight. But this shouldn't be a blocker.

https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Datetime%21Element%21Datetime.php/class/Datetime/9.4.x

modules/core/quick/tests/modules/farm_quick_test/src/Plugin/QuickForm/Test.php
mstenta: Yes it did break the test because the `\Drupal\Core\StringTranslation\TranslatableMarkup` is not a string.

> My hunch is that this would work: 'units' => (string) $this->t('tests').

Yes that would work, as well as `$this->t('tests')->toString()` I believe.

PR #495, opened 2022-01-26
Provide a helper function for loading flag options and allowed values

paul121:
See https://www.drupal.org/project/farm/issues/3253433

As mentioned in the issue this is kind of the "reverse approach" that the existing `farm_flag_applies` takes. This change makes the existing implementation of `farm_flag_applies` much simpler, although I'm not sure we can remove the function altogether (breaking change?)

The terminology of this is a little confusing, so definitely open to ideas here, but this ability to load the "intersection" of flags that "apply" to *any* or *every* bundle is the key difference here. I think the tests should demonstrate this pretty well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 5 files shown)


mstenta:
Merged. See https://www.drupal.org/project/farm/issues/3253433#comment-14428604

PR #499, opened 2022-02-04
Correct list formatting in module development documentation.

paul121:
Some of the long text in markdown lists is not fully indented, leading to some wonky formatting. Curious if there is a linter that could check for this type of thing? It's not easy to catch this in code, especially when it displays fine in the IDE preview (at least in PHPStorm for me - maybe I can change the markdown/formatter I'm using?)

consumer_doc_bug


paul121:
I did a quick run through the rest of the docs pages and did not see this anywhere else.

Only a couple issues within the farmOS.org community/trademark page itself: https://github.com/farmOS/farmOS.org/pull/37

mstenta:
Merged - thanks @paul121!

(PS: I left out the `CHANGELOG.md` update because this is pretty minor.)

PR #500, opened 2022-02-06
Update documentation for a list of community modules.

paul121:
Point to the Drupal ecosystem list since we no longer include a list of community modules on farmOS.org.


mstenta:
Good idea - thanks @paul121 !

PR #505, opened 2022-02-19
Improve fields documentation

jgOhYeah:
Hello,
As a new user to farmOS / drupal, I have added a few extra hints to the fields page for creating custom fields.
I have also added documentation for options available when specifying fields. I'm not sure if this is the best way of representing it, but hopefully it is readable (feel free to improve or suggest improvements). Forum thread discussing this: https://farmos.discourse.group/t/is-it-possible-to-use-custom-fields-and-filtering-in-asset-views-and-filter-by-a-range-of-birth-dates/1139/5

After reading [this](https://farmos.discourse.group/t/custom-module-help/1149) thread, I also corrected a typo / copy paste error in the organic flag example.

### DIFF
--- a/docs/development/module/fields.md
+++ b/docs/development/module/fields.md
@@ -102,46 +102,103 @@ function mymodule_farm_entity_bundle_field_info(EntityTypeInterface $entity_type
 
 ## Field options
 The following keys are available to use when specifying fields:
+### `'label'`
+What the field should be called when displayed to the user. 
 
-| Option | Description |
-|:-:|:--|
-| `'label'` | What the field should be called when displayed to the user. |
-| `'description'` | A description of the field and how to use it. |
-| `'computed'` | Whether the field is computed. |
-| `'required'` | Specify this to force the user to enter a value into this field. |
-| `'revisionable'` | Make the field revisionable, unless told otherwise. |
-| `'cardinality'` | Set cardinality, if specified. If neither `'cardinality'` or `'multiple'` is specified, this defaults to 1. |
-| `'multiple'` | If `'cardinality'` is not specified and `'multiple'` is, set the cardinality to unlimited. |
-| `'translatable'` | Only makes the field translatable if it is specified |
-| `'default_value_callback'` | Sets the default value callback, if specified |
-| `'type'` | The datatype of the field, represented as one of the following strings: <ul><li>[`'boolean'`](#boolean)</li><li>[`'entity_reference'`](#entity-reference)</li><li>[`'entity_reference_revisions'`](#entity-reference-revisions)</li><li>[`'file'`](#file-and-image)</li><li>[`'image'`](#file-and-image)</li><li>[`'fraction'`](#fraction)</li><li>[`'geofield'`](#geofield)</li><li>[`'id_tag'`](#id-tag)</li><li>[`'inventory'`](#inventory)</li><li>[`'list_string'`](#list-string)</li><li>[`'string'`](#string)</li><li>[`'string_long'`](#long-string)</li><li>[`'timestamp'`](#timestamp)</li></ul> Click one of the above or scroll down for options relating to each specific type.  |
-| `'hidden'` | Hide the field in form and view displays, if specified. The hidden option can either be set to `true`, which will hide it in both form and view displays, or it can be set to `'form'` or `'view'`, which will only hide it in the form or view display. |
-| `'form_display_options'` & `'view_display_options'` | Override form and view display options, if specified. |
-| `'weight'` | Contains an associative array with the keys `'form'` and `'view'`, giving hints on whereabouts the field should appear. |
+### `'description'`
+A description of the field and how to use it.
 
-### Datatype specific options
+### `'computed'`
+Whether the field is computed.
+
+### `'required'`
+Specify this to force the user to enter a value into this field.
+
+### `'revisionable'`
+Make the field revisionable, unless told otherwise.
+
+### `'cardinality'`
+Set cardinality, if specified. If neither `'cardinality'` or `'multiple'` is
+specified, this defaults to 1.
+
+### `'multiple'`
+If `'cardinality'` is not specified and `'multiple'` is, set the cardinality to
+unlimited.
+
+### `'translatable'`
+Only makes the field translatable if it is specified.
+
+### `'default_value_callback'`
+Sets the default value callback, if specified.
+
+### `'type'`
+The datatype of the field, represented as one of the following strings:
+- [`'boolean'`](#boolean)
+- [`'entity_reference'`](#entity-reference)
+- [`'entity_reference_revisions'`](#entity-reference-revisions)
+- [`'file'`](#file-and-image)
+- [`'image'`](#file-and-image)
+- [`'fraction'`](#fraction)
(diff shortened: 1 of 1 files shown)


docs/development/module/fields.md
```diff
@@ -140,8 +212,8 @@ dependencies:
   enforced:
     module:
       - my_module
-id: monitor
-label: Monitor
+id: organic
```
Farmer-Eds-Shed, line 215: Great Job,
Perhaps an additional comment on id in file name should match the `id` in the file `farm_flag.flag.[id].yml` ?

docs/development/module/fields.md
```diff
@@ -140,8 +212,8 @@ dependencies:
...
+label: Organic
 entity_types: null
 ```
```
Farmer-Eds-Shed, line 277: Great Job,
Perhaps an additional comment on id in file name should match the `id `in the file `farm_flag.flag.[id].yml` ?

jgOhYeah: I've added an additional comment. Took an educated guess that the file name is generally in the form `[select_module_name].[select_field].[id].yml`, so hopefully that is close enough :).

docs/development/module/fields.md
```diff
@@ -15,8 +15,9 @@ If the field should be added to all bundles of a given entity type (eg: all log
 types), then they should be added as "base fields" via
 `hook_entity_base_field_info()`.
 
-A `farm_field.factory` helper service is provided to make this easier:
+A `farm_field.factory` helper service is provided to make this easier.
 
+To get started, place the following in the `[modulename].module` file:
```
mstenta, line 20: newline

docs/development/module/fields.md
```diff
@@ -60,6 +61,7 @@ The format for bundle field definitions is identical to base field definitions
 (above), but the `bundleFieldDefinition()` method must be used instead of
 `baseFieldDefinition()`.
 
+To get started, place the following in the `[modulename].module` file:
```
mstenta, line 68: newline

docs/development/module/fields.md
```diff
@@ -120,12 +192,16 @@ These include:
     - Tattoo (`tattoo`, on Animal assets)
 
 These options are provided as configuration entities by farmOS modules in the
-form of YAML files.
+form of YAML files.
```
mstenta: undo this

docs/development/module/fields.md
```diff
@@ -252,6 +334,7 @@ label: Ear tag
 bundles:
   - animal
 ```
+Note that the file name is in the form `farm_flag.flag.ear_tag.[id].yml`
```
mstenta, line 391: This should be `farm_id_tag.id_tag.[id].yml`

And I see that there is a typo in the existing docs (which you copied from, so my fault!) - can you fix that too while you're at it?

`animal/config/install/farm_flag.flag.ear_tag.yml` should be `animal/config/install/farm_id_tag.id_tag.ear_tag.yml`

mstenta:
This is great @jgOhYeah - thank you so much!!

I like the additional comments included in the code samples! I wonder if we should move them to their own lines, above the code lines they are commenting on? This would be consistent with farmOS coding standards - which isn't necessarily important here in the docs, but would be one less cleanup step required after a copy+paste. What do you think? I haven't seen what this looks like rendered in farmOS.org HTML, I'm only looking at the Git changes, so I'd also be curious how very-wide lines of code render.

The field options are a good addition too! I haven't read through them in detail yet - can you describe how you compiled it? I worry a little bit that this will be difficult to keep in sync if/when new field types or options are added in the code. I wish we could automatically generate it somehow. :-)

In either case, I also want to see what this will look like in the farmOS.org template rendering - I'm not sure if we've used Markdown tables yet, so I want to be sure it renders properly. I may need @jgaehring's advice/assistance with this (we recently moved farmOS.org to a new system, still learning the new workflows)... Perhaps it would make sense to split this addition out to its own PR so it doesn't block all the other small improvements you've made?

The only other feedback I have is minor - whitespace fixes, mainly. I'll do a quick GitHub review to be more specific... :-)

mstenta (changes requested):

jgaehring:
> In either case, I also want to see what this will look like in the farmOS.org template rendering - I'm not sure if we've used Markdown tables yet, so I want to be sure it renders properly. I may need @jgaehring's advice/assistance with this (we recently moved farmOS.org to a new system, still learning the new workflows)... 

Yea, it would be nice to create the preview branch deployments on farmOS.org like we talked about. For the time being, here are some screenshots from the local build I ran by setting the source repo to @jgOhYeah's fork:

image

We don't have any CSS for tables yet, and the tables have a lot of text in their cells that don't really display very well with default styles, so I would probably recommend a different layout for now. I'll spin up a branch deployment momentarily so others can look at the preview for themselves.

jgaehring:
Here's the branch deployment:

https://source-farmos--gracious-brattain-bdd606.netlify.app/development/module/fields/

FYI, this still isn't automated in anyway, so not a permanent solution, but we can leave this branch set up for such deployments in the future and just change git url in `source-repos.js`.

mstenta:
Wow thanks @jgaehring !!

jgOhYeah:
Thanks everyone for looking at it and getting a preview running. I didn't think that the website would be that different from Github's style. I have converted the table layout to a list and headings - hopefully the inline code doesn't mess with the table of contents too much. If it does, I can work something else out.

Edit: Also hopefully addressed @mstenta 's feedback.

jgaehring:
Cool, I just triggered a new branch deploy, so that preview should be updated in a few minutes.

mstenta:
Thanks @jgOhYeah and @jgaehring!! I'll review this ASAP - just need some time to go through and double-check all the field options...

mstenta:
Sorry it's taken me so long to get back to this @jgOhYeah! I've merged your commits, and went one step farther...

https://github.com/farmOS/farmOS/commit/271caf0aeacc0efd09faa523a2bb305d05ee7e63

I decided to move the "field options" information to the [Services](https://farmos.org/development/module/services/) documentation page, in a new "Field factory service" section, along with an explanation of what this service is, why we provide it, etc etc. I felt it was necessary to give this some context, and emphasize that it's not the only way to produce field definitions.

Ultimately, the field factory service is a farmOS-specific layer on top of Drupal core's `BaseFieldDefinition` (and Entity API module's `BundleFieldDefinition`). So it is actually possible to completely bypass our service and use those classes directly - which provides a lot more flexibility. And because of that, I felt a bit torn about whether or not we should actually try to document ALL the options... because in order to do so we would essentially need to replicate Drupal core's documentation: https://www.drupal.org/docs/drupal-apis/entity-api/fieldtypes-fieldwidgets-and-fieldformatters

So, I decided to instead document *some* of the options that the field factory provides (omitting some of the "advanced" types and options), and refer the reader to the official upstream Drupal docs for more information.

I feel like this will achieve a good balance... it adds some more detail to our docs, but lessens the burden on us to maintain every detail (which I fear would be easy to forget in the future, leading to outdated docs).

Thanks again for your contributions! And sorry for the long delay in merging them in. :-)

jgOhYeah:
Thanks for sorting this out and organising it.

PR #507, opened 2022-02-26
Added weight quick form

Skipper-is:
Added the quick form module for weighing animals.

Need to still sort the setting of default units - Currently it includes the id of the unit taxonomy.
Also - uninstall hook - need to clean up the state after through _uninstall hook.


mstenta:
Nice thanks for starting this @Skipper-is !

> Need to still sort the setting of default units - Currently it includes the id of the unit taxonomy.

One thought I have: maybe we should present the user with some hard-coded options, based on their system of measurement. For example, if they use metric, show a dropdown with "kg" and "g", and if they use "us/imperial" then show "lbs" and "oz".

We should coordinate this decision with @paul121 and how the Animal Weight Report module is going to work: https://drupal.org/project/farm_animal_weight - to make sure that logs created with this quick form work with that.

Other things we'll need to do before we can merge this:

- Coding standards cleanup (you can see the output of that here: https://github.com/farmOS/farmOS/runs/5345101220?check_suite_focus=true - or run it locally: https://farmos.org/development/environment/code/)
- Automated tests (we should include tests for all quick forms that farmOS core provides)

I'll try to look at this closer in the near future to see if there's any other feedback... happy to help push this forward - when I have time. :-)

mstenta:
> Automated tests (we should include tests for all quick forms that farmOS core provides)

To whomever it may concern: in my `2.x-quick-birth` branch I just created a base class that can be used for quick form kernel tests, and refactored the Planting quick form tests to extend from it. So we can use that here too.

Skipper-is:
Have attempted to mirror the testing of the quick-birth form, with additional `assertEquals` functions.

mstenta:
That's awesome @Skipper-is - thanks! I'll add this to my list to review after the birth form.

mstenta:
Sorry for the long silence on this! I developed a plan in my head but never commented here to propose it. :-)

I propose we implement this in the farmOS Animal Weight module here: https://www.drupal.org/project/farm_animal_weight

During farmOS v2 development, we split all of the animal weight stuff out to this new contrib module, so it can serve as the shepherd of that convention. Makes perfect sense for it to provide a quick form too IMO!

My plan was to transfer @Skipper-is's work over to that repo, but I never got around to it. Once we move it over, we can close this core PR. I didn't want to close it until then so I don't forget.

How does that sound?

Issue #510, opened 2022-03-15
Error importing KML when placemarks have empty geometries


paul121:
**Describe the bug**
> TypeError: Argument 1 passed to Drupal\farm_geo\GeometryWrapper::__construct() must be an instance of Geometry, bool given, called in /opt/drupal/web/profiles/farm/modules/core/kml/src/Normalizer/KmlNormalizer.php on line 138 in Drupal\farm_geo\GeometryWrapper->__construct() (line 39 of /opt/drupal/web/profiles/farm/modules/core/geo/src/GeometryWrapper.php)

**To Reproduce**
1. Import a KML file that has placemarks with "empty" geometries. I discovered this with a placemark that looks like the following:
```
  <Placemark>
	<name>Placemark name</name>
      <MultiGeometry></MultiGeometry>
  </Placemark>
 ```
2. Click on "Parse"; nothing is displayed to the user but this log message is generated.

**Expected behavior**
If a KML file has empty geometries we should notify the user of these placemarks rather than just skip them. In the KML importer I think this would be accomplished by rendering a fieldset for the given placemark, just with an empty `geometry` field. The placemarks attributes (id, name, description, etc) can still be populated.

A later step could be to add a checkbox in the import form to "Skip placemarks that have an empty or invalid geometry.".

mstenta:
Fixed by #511

PR #511, opened 2022-03-15
Allow importing KML with empty geometries. #510

paul121:
A simple fix for the bug identified in #510 

Attaching a screenshot that shows the result of parsing the following placemark:

```kml
<Placemark>
	<name>Placemark name</name>
      <MultiGeometry></MultiGeometry>
  </Placemark>
```
Screenshot from 2022-03-15 16-50-57

### DIFF
--- a/modules/core/kml/src/Normalizer/KmlNormalizer.php
+++ b/modules/core/kml/src/Normalizer/KmlNormalizer.php
@@ -4,6 +4,7 @@ namespace Drupal\farm_kml\Normalizer;
 
 use Drupal\farm_geo\GeometryWrapper;
 use Drupal\geofield\GeoPHP\GeoPHPInterface;
+use GeometryCollection;
 use Symfony\Component\Serializer\Normalizer\DenormalizerInterface;
 use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
 
@@ -118,6 +119,11 @@ class KmlNormalizer implements NormalizerInterface, DenormalizerInterface {
 
       // Load KML into a Geometry object.
       $geometry = $this->geoPHP->load($placemark['xml'], 'kml');
+
+      // Create an empty collection if no geometry was loaded.
+      if (empty($geometry)) {
+        $geometry = new GeometryCollection();
+      }
 
       // Build properties.
       $properties = [];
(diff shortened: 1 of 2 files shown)


paul121:
Force push to add fix to README.md

modules/core/kml/src/Normalizer/KmlNormalizer.php
```diff
@@ -119,6 +119,11 @@ public function denormalize($data, $type, $format = NULL, array $context = []) {
       // Load KML into a Geometry object.
       $geometry = $this->geoPHP->load($placemark['xml'], 'kml');
 
+      // Create an empty collection if no geometry was loaded.
+      if (empty($geometry)) {
+        $geometry = new \GeometryCollection();
```
mstenta: I wonder if it makes sense to add `use GeometryCollection;` to the top of the file and change this to:

`$geometry = new GeometryCollection();`

Same effect, but feels like it would be more consistent to add a `use` statement in this case, even though `phayes/GeometryCollection` doesn't use namespaces.

I wonder if there's any benefit to consistency in this regard, like IDE integration, code sniffing, etc...

Curious your thoughts on this (very minor) thought @paul121 - doesn't need to be a blocker for merging this, I don't think, but worth considering?

paul121: Oh interesting, yeah I don't see any problem with that, as long as it works! Normally PHPStorm would add that automatically for me but it's possible I typed it in differently without using any type helpers.

mstenta: Cool! I can make that change and amend your commit.

mstenta:
Pushed that change - do you want to give it one more test @paul121? I didn't test it, but PHPStorm recognizes `GeometryCollection`, so I assume it works.

paul121:
Great, tested & works great!

paul121:
> I wonder if it makes sense to add use GeometryCollection; to the top of the file and change this to:

Hey @mstenta this change actually created a code sniffer violation:

[see this failing action](https://github.com/farmOS/farmOS/runs/5636463942?check_suite_focus=true)

```
FILE: .../profiles/farm/modules/core/kml/src/Normalizer/KmlNormalizer.php
----------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
----------------------------------------------------------------------
 7 | ERROR | [x] Non-namespaced classes/interfaces/traits should not
   |       |     be referenced with use statements
----------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
----------------------------------------------------------------------

Time: 6.38 secs; Memory: 26MB
```

mstenta:
Of course it did. :roll_eyes: 

Should have left well enough alone...

mstenta:
Pushed a commit that changes this back to the way you had it originally @paul121 :+1:

PR #513, opened 2022-03-18
Use farm_land_type_options() in KmlImporter

paul121:
`farm_land_type_options` should be used here instead of `farm_land_type_field_allowed_values`. We recently added this function with https://github.com/farmOS/farmOS/commit/1a35a65c66744e9cfcfeb9d1c80eab555f68cf87


paul121:
I think these phpunit tests are failing for another reason.

mstenta:
Thanks @paul121! I rebased this and the `CHANGELOG.md` conflicted - but I decided to just remove that bit since this is an internal change.

PR #515, opened 2022-03-21
Only require a name to build map popups

paul121:
Greg from OurSci and I ran into an issue where location assets with a `MULTIPOLYGON (((...` or `GEOMETRYCOLLECTION ( POLYGON((...` geometry were not "clickable" on the dashboard map, meaning they did not create a popup.

Chatting with @mstenta we identified the issue in the `farmOS.map.behaviors.popup.js` logic where a `name` *and* a `measurement` or `description` property is required. Openlayers does not compute the measurement for multipolygons, so this popup was not being created. I think this would also be the case for location assets with a single `point` geometry.

This change allows popups to be rendered for multipolygon geometries that will not have a measurement calculated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


paul121:
>  I think this would also be the case for location assets with a single point geometry.

Confirmed this as well

paul121:
Rebase and force pushed for changelog

PR #516, opened 2022-03-23
Include list_string, timestamp and boolean bundle fields in views.

paul121:
Follow up to #512 

The `timestamp` and `boolean` had cases to add only add `field` or `filter` handlers (respectively), but not both.

Additionally, `list_string` fields were not included at all. This was particularly bad because the dashboard map did not filter the different map layers by the proper `land_type`; land assets were being included for every layer.

Would be good to get a test in that captures this for all of the "supported" field types that we add to views... but just a simple fix for now..


paul121:
I went through all our asset and log types and believe we have all the core bundle field types supported now :-)

Issue #514, opened 2022-03-21
selenium/chrome-standalone:latest causes automated tests to fail


paul121:
**Describe the bug**
Automated tests have been failing the past couple days. It seems like it is just two test cases:

```
There were 2 errors:

1) Drupal\Tests\farm_location\Functional\LocationTest::testLocationFieldVisibility
Behat\Mink\Exception\ResponseTextException: The text "Current geometry" was not found anywhere in the text of the current page.

/opt/drupal/vendor/behat/mink/src/WebAssert.php:785
/opt/drupal/vendor/behat/mink/src/WebAssert.php:262
/opt/drupal/web/core/tests/Drupal/Tests/WebAssert.php:898
/opt/drupal/web/profiles/farm/modules/core/location/tests/src/Functional/LocationTest.php:87
/opt/drupal/vendor/phpunit/phpunit/src/Framework/TestResult.php:726

2) Drupal\Tests\farm_settings\Functional\ModulesFormTest::testInstallFunctionality
Drupal\Core\Extension\Exception\UnknownExtensionException: The module farm_plant does not exist or is not installed.

/opt/drupal/web/core/lib/Drupal/Core/Extension/ExtensionList.php:346
/opt/drupal/web/profiles/farm/modules/core/settings/tests/src/Functional/ModulesFormTest.php:162
/opt/drupal/web/profiles/farm/modules/core/settings/tests/src/Functional/ModulesFormTest.php:73
/opt/drupal/vendor/phpunit/phpunit/src/Framework/TestResult.php:726
There was 1 failure:

1) Drupal\Tests\farm_group\Functional\GroupTest::testGroupFieldVisibility
Failed asserting that a NULL is not empty.

/opt/drupal/vendor/phpunit/phpunit/src/Framework/Constraint/Constraint.php:[121](https://github.com/farmOS/farmOS/runs/5623699703?check_suite_focus=true#step:12:121)
/opt/drupal/vendor/phpunit/phpunit/src/Framework/Constraint/Constraint.php:55
/opt/drupal/web/profiles/farm/modules/asset/group/tests/src/Functional/GroupTest.php:76
/opt/drupal/vendor/phpunit/phpunit/src/Framework/TestResult.php:726

FAILURES!
Tests: 88, Assertions: 1723, Failures: 1, Errors: 2.
Error: Process completed with exit code 2.
```

**Additional context**
[...]

PR #517, opened 2022-03-31
Pin selenium/standalone-chrome to 4.1.2-20220217. #514

paul121:


mstenta:
Thanks @paul121! There is one other place where `selenium/chrome-standalone:latest` is referenced (in the documentation): https://github.com/farmOS/farmOS/blob/11cc36accf9590426d846fa9cc766de26892e84a/docs/development/environment/tests.md?plain=1#L29

If you don't mind, I'm going to amend your commit to change that as well. My thinking is: we don't know how long this pin will be in place, so might as well point devs in the right direction in our development docs too. And then, when #514 gets truly resolved, we can simply revert the commit from this PR to go back to `latest` in all three places. (Unless of course we decide against using `latest`, as you mentioned in https://github.com/farmOS/farmOS/issues/514#issuecomment-1084081496.

PR #518, opened 2022-04-05
Increase map size so delete button is shown.

paul121:
Right now when editing geometries in the map the delete button is covered by the snapping grid toggle. This is a quick fix to increase the map height. I think we should consider this because it's pretty poor UX right now

I know there is additional discussion about this here https://github.com/farmOS/farmOS-map/issues/146 and here https://github.com/farmOS/farmOS-map/pull/152

### DIFF
--- a/modules/core/map/css/farm_map.css
+++ b/modules/core/map/css/farm_map.css
@@ -1,5 +1,5 @@
 .farm-map {
-  height: 400px;
+  height: 450px;
 }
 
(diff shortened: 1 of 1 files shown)


mstenta:
We discussed this on the weekly dev call today and decided to make a small tweak to the farmOS-map CSS to fix this issue. So I don't think we need to mess with the height anymore. There are a few possible next-steps to further improve the edit buttons in the farmOS-map queue.

PR #521, opened 2022-04-07
Encourage GEOS PHP extension use

mstenta:
Add `hook_requirements()` to check if the GEOS PHP extension is installed. This is copied/modified from https://git.drupalcode.org/project/geophp/-/blob/8.x-1.x/geophp.module#L79

Here is the message displayed in `/admin/reports/status` when GEOS is present:

Screenshot from 2022-04-07 14-11-50

And when it is not present:

Screenshot from 2022-04-07 14-09-02

Also changed the wording of the server requirements to strongly encourage GEOS.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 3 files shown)


symbioquine (approved):

paul121 (approved):

mstenta:
Oops forgot to run code sniffer before pushing. :-)

Let's see how this one goes...

PR #523, opened 2022-04-08
Merge test and release workflows into a unified delivery workflow

mstenta:
Following on the changes made to our GitHub Actions `run-tests.yml` workflow in #522, I would like to take a next step by merging the `run-tests.yml` and `package-release.yml` workflows into a single unified `deliver.yml` workflow.

The `package-release.yml` workflow currently only runs when tags are pushed (as opposed to `run-tests.yml` which runs when branches are pushed, when pull requests are opened, and on a weekly schedule). It builds a Docker image, runs it, and then copies the `/opt/drupal` directory (the farmOS codebase) out of it, creates a tarball of that directory, creates a GitHub release, and uploads the tarball to it. This serves to automate the creation of our "packaged releases", which are offered as an alternative to Docker image deployments (see https://farmos.org/hosting/install/#packaged-releases).

Here is the current `package-release.yml`: https://github.com/farmOS/farmOS/blob/430a810da23c598ddf7432b5ad0539df554ab811/.github/workflows/package-release.yml

This works fine, but there are some downsides to the current implementation. For one, we're maintaining two separate workflows that build Docker images. But perhaps more critically: the `package-release.yml` workflow does NOT currently run any tests before packaging a release. So the implied assumption is that maintainers will always push `2.x` first, wait for those tests to pass, and THEN push a tag to trigger the release build process.

Now that #522 separated our `run-tests.yml` workflow into multiple jobs, it opens up the possibility for the `package-release.yml` workflow to simply become another conditional job within that, which is only triggered if a tag is pushed.

Of course, if we merge them, then the filename `run-tests.yml` is no longer an accurate representation of the workflow. It will be doing more than just running tests. It essentially becomes a [continuous delivery](https://en.wikipedia.org/wiki/Continuous_delivery) pipeline. So I decided to rename `run-tests.yml` to `deliver.yml` at the same time.

Notably, the next step will be to use this same workflow to publish Docker images to Docker Hub. See https://www.drupal.org/project/farm/issues/3203129

To illustrate what this looks like, we can see how GitHub Actions visualizes it. This is what it looks like now (after #522 but before this PR's changes):

Screenshot from 2022-04-07 15-05-41

With this PR, when a branch is pushed, it will run the same jobs as before, but you can also see the `release` job that does NOT run.

Screenshot from 2022-04-08 12-15-04

And when a tag is pushed, it triggers the `release` job to run:

Screenshot from 2022-04-08 12-15-19

Here is what the release looks like that is created:

Screenshot from 2022-04-08 12-41-49


paul121 (approved):
Looks good!

IMO combining the test and release here is little weird since release will usually be greyed out. It seems you can have a workflow be dependent on another workflow: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run

But maybe we can't cache between workflows? It seems we can, just need to make sure the cache key is the same for both workflows, but different from other runs?

mstenta:
That's an interesting idea @paul121 - and a good point to highlight that "release will usually be greyed out". That isn't a deal-breaker IMO, but I'm open to exploring the `workflow_run` trigger as well - to separate into two workflows again.

My plan is to tackle the "publish images to Docker Hub" next, as part of this same `deliver.yml` - so I might proceed with that, just to keep things in one workflow for simplicity. But maybe we can consider a follow-up that splits them after that?

Thinking through the `workflow_run` idea...

- We would need to split `deliver.yml` into `test.yml` and `deliver.yml`.
- `test.yml` would be the initial workflow triggered by branch/tag push, pull requests, and weekly schedule
- `deliver.yml` would be triggered by a successfully completed run of `test.yml`
- `deliver.yml` could handle both packaging the GitHub release AND publishing images to Docker Hub, but will need to have conditional logic to only perform those actions under certain conditions (NOT pull requests, only `2.x` branch pushes or tag pushes, etc)

I suppose we wouldn't entirely escape the fact that this "extra" thing is going to be visible either way. As this PR currently stands, we'll have an extra greyed out "Create GitHub release" in the visualization (shown above). If we split into two workflows, then we'll see TWO actions running for every push, even if the second one bails early most of the time. Which is better/worse?

mstenta:
PS: Not sure why the PHPUnit tests failed on this PR - no PHP was changed and these same tests have been passing for me all morning. For example: https://github.com/farmOS/farmOS/actions/runs/2115758265

I'll try re-running them - maybe it was a fluke. :crossed_fingers:

.github/workflows/deliver.yml
```diff
@@ -17,18 +19,28 @@ jobs:
...
+      - name: Set default FARMOS_REPO.
+        run: echo "FARMOS_REPO=${GITHUB_REPOSITORY}" >> $GITHUB_ENV
+      - name: Set FARMOS_VERSION for branch push event.
+        if: github.event_name == 'push' && github.ref_type == 'branch'
+        run: echo "FARMOS_VERSION=${GITHUB_REF:11}" >> $GITHUB_ENV
+      - name: Set FARMOS_VERSION for tag push event.
+        if: github.event_name == 'push' && github.ref_type == 'tag'
+        run: echo "FARMOS_VERSION=${GITHUB_REF:10}" >> $GITHUB_ENV
+      - name: Set FARMOS_VERSION and FARMOS_REPO for pull request event.
+        if: github.event_name == 'pull_request'
```
symbioquine, line 31: I'm curious if there's an advantage/reason to/for switching to this string comparison vs the `github.event.pull_request` boolean check it's replacing on old line #&#x2060;26?

symbioquine (approved):
Looks good to me + minor question :)

symbioquine:
> I'll try re-running them - maybe it was a fluke.

image

paul121:
> As this PR currently stands, we'll have an extra greyed out "Create GitHub release" in the visualization (shown above). If we split into two workflows, then we'll see TWO actions running for every push, even if the second one bails early most of the time. 

Actually I think the second workflow would only be run (or at least only shown in the UI) when the conditions match. Basically the same as the workflow to build docs: https://github.com/farmOS/farmOS/actions/workflows/netlify-build-docs.yml

+1 for pushing to docker hub tho... don't let this block that. I agree it could be a follow up.

.github/workflows/deliver.yml
mstenta: I just did this for consistency. Originally we only had two cases: "is not a pull request" and "is a pull request", so the boolean worked for both.

Now, we've got three: "is a push event of a branch", "is a push event of a tag", and "is a pull request event". I could have left the last one alone, but it felt better to use consistent logic in the conditions.

mstenta:
Ah good point @paul121 - in that case splitting into two workflows seems like a good next step.

> +1 for pushing to docker hub tho... don't let this block that. I agree it could be a follow up.

Sounds good! We can cross that bridge next. :-)

PR #527, opened 2022-04-21
Update toolbar logo spacing for gin beta

paul121:
Two changes:
- Decrease the top/bottom margin to 20px. This centers the logo vertically.
- Add padding to the wrapper around the logo. Without this padding the logo touches the edges of the toolbar. Not sure if we can fix this without adding the `!important` here.


paul121:
Before:

open_before
collapsed_before

After:
open_after
collapsed_after

PR #529, opened 2022-04-24
Only show active plans by default

paul121:
I noticed that right now the Plan views display both active and archived plans by default. I think we should make this only show active plans by default for consistency with the other views. Once a plan is archived you likely don't want to see it all the time.

After I first made this change and exported config there were lots of changes. So I went back and re-exported the `farm_plan` view to capture only these config schema/ordering changes (I believe the same as
c4d439c204895577af6a39a12408e89e2373f435)

Finally I exported the simple change for the `status` filter.


mstenta:
Thanks @paul121 !

PR #530, opened 2022-04-24
Remove create() implementations from classes that do not implement ContainerInjectionInterface

paul121:
The `create()` function is not needed for services that define their dependencies as in the `arguments` of `module.services.yml`. All of these services should be covered by our tests, removing this should be OK!

I noticed this because my IDE highlighted that the `create()` function's `{@inheritdoc}` was not inheriting from anywhere. The `create()` function is defined on the [`ContainerInjectionInterface`](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21DependencyInjection%21ContainerInjectionInterface.php/function/ContainerInjectionInterface%3A%3Acreate/9.3.x) which is only implemented by classes that need dependency injection but are not services themselves (things like form & plugin classes in Drupal).


mstenta:
Thanks @paul121! TIL!

PR #531, opened 2022-05-05
Remove GinSidebarOpen check.

paul121:
As we had noted in a comment this is no longer needed with Gin ^beta.

### DIFF
--- a/modules/core/ui/theme/js/toolbar.js
+++ b/modules/core/ui/theme/js/toolbar.js
@@ -5,10 +5,4 @@
   if (localStorage.getItem(itemName) === null) {
     localStorage.setItem(itemName, 'true');
   }
-  // @todo Remove this when new Gin version is released.
-  // Gin changed the name after 8.x-3.0-alpha37.
-  itemName = 'GinSidebarOpen';
-  if (localStorage.getItem(itemName) === null) {
-    localStorage.setItem(itemName, 'true');
-  }
 }());


paul121:
I'm experiencing a bug where my toolbar "open state" is not being remembered: after I have collapsed the toolbar, refresh the page and the toolbar opens. I've just noticed this for the first time which seems weird to me but its possible I've just gotten used to the toolbar/menu being open and haven't noticed.

It seems to be some kind of a race condition that makes reproducing this hard. I think removing this "legacy" code we have will fix the issue.

I'm just seeing this gin code that actually checks for the legacy `GinSidebarOpen` in local storage and changes it to the new `Drupal.gin.toolbarExpanded`: https://git.drupalcode.org/project/gin/-/blob/8.x-3.x/js/init.js#L12

```js
// Legacy Check: Transform old localStorage items to newer ones.
function checkLegacy() {
  if (localStorage.getItem('GinDarkMode')) {
    localStorage.setItem('Drupal.gin.darkmode', localStorage.getItem('GinDarkMode'));
    localStorage.removeItem('GinDarkMode');
  }

  if (localStorage.getItem('GinSidebarOpen')) {
    localStorage.setItem('Drupal.gin.toolbarExpanded', localStorage.getItem('GinSidebarOpen'));
    localStorage.removeItem('GinSidebarOpen');
  }
}
```

mstenta:
Thanks @paul121! Hopefully this solves the issue you described. We need to remove it regardless. :-)

PR #534, opened 2022-05-13
Update composer.json

Farmer-Eds-Shed:
Infer target_revision_id to be Latest Revision when Only a target_id is Provided": "https://www.drupal.org/files/issues/2022-05-13/3267304-9.patch"

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
         "drupal/entity": "1.3",
         "drupal/entity_browser": "^2.6",
         "drupal/entity_reference_integrity": "^1.1",
-        "drupal/entity_reference_revisions": "^1.8",
+        "drupal/entity_reference_revisions": "1.9",
         "drupal/entity_reference_validators": "^1.0@alpha",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -60,6 +60,9 @@
             "drupal/entity": {
                 "Issue #3206703: Provide reverse relationships for bundle plugin entity_reference fields.": "https://www.drupal.org/files/issues/2022-05-11/3206703-10.patch"
             },
+            "drupal/entity_reference_revisions": {
```
paul121, line 63: We should pin this dependency in composer.josn to its latest version. Otherwise our builds will start failing if/when they include this patch themselves.

paul121 (changes requested):
Looks good, but we should prob pin this dependency though?

mstenta:
Oh yes, thanks for the reminder @paul121. I will push a commit to this...

mstenta:
Merged! Thanks again @Farmer-Eds-Shed!

PR #536, opened 2022-05-20
Add images to asset-related taxonomies

c4ndel4:
With the addition of entity views on taxonomy terms (https://github.com/farmOS/farmOS/pull/458) this pages are more used and having a image could be helpful.

### DIFF
--- a/modules/taxonomy/animal_type/config/install/core.entity_form_display.taxonomy_term.animal_type.default.yml
+++ b/modules/taxonomy/animal_type/config/install/core.entity_form_display.taxonomy_term.animal_type.default.yml
@@ -1,0 +1,50 @@
+langcode: en
+status: true
+dependencies:
+  config:
+    - field.field.taxonomy_term.animal_type.image
+    - image.style.thumbnail
+    - taxonomy.vocabulary.animal_type
+  enforced:
+    module:
+      - farm_animal_type
+  module:
+    - image
+    - text
+id: taxonomy_term.animal_type.default
+targetEntityType: taxonomy_term
+bundle: animal_type
+mode: default
+content:
+  description:
+    type: text_textarea
+    weight: 1
+    region: content
+    settings:
+      placeholder: ''
+      rows: 5
+    third_party_settings: {  }
+  image:
+    type: image_image
+    weight: 6
+    region: content
+    settings:
+      progress_indicator: throbber
+      preview_image_style: thumbnail
+    third_party_settings: {  }
+  name:
+    type: string_textfield
+    weight: 0
+    region: content
+    settings:
+      size: 60
+      placeholder: ''
+    third_party_settings: {  }
+  status:
+    type: boolean_checkbox
+    settings:
+      display_label: true
+    weight: 7
+    region: content
+    third_party_settings: {  }
+hidden: {  }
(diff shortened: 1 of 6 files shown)


paul121 (commented):
This is a great start @c4ndel4 ! Some thoughts:

The image field definitions `field.field.taxonomy_term.*.image` look good to me. As you mentioned on the forum I think you are correct that this won't require any update hooks for the field storage since we're reusing the existing `field.storage.taxonomy_term.image` config and thus the `taxonomy_term__image` table. Although I do have a few questions about this:
- Right now `field.storage.taxonomy_term.image` is provided by the `farm_plant_type` module. So this means that `farm_animal_type` and `farm_material_type` would need to depend on `farm_plant_type`. Should we (can we?) move this field storage to a more common core module?
- I think we would still need an update hook to create this `field.field.taxonomy_term.*.image` field config on existing installs.

The entity view and form displays are another beast.. I think we might need an update hook to create them as well. Ideally we could use bundle fields on taxonomy terms like we do for our assets/logs/plans. That would effectively solve all the above problems. But iirc it isn't this easy to turn on bundle fields for taxonomy terms... hmm.

paul121:
Linking to a related issue: https://www.drupal.org/project/farm/issues/3244880

Perhaps we could add a file field to these taxonomies at the same time.

mstenta:
Thanks @c4ndel4!

I agree w/ @paul121 - we should add `file` field at the same time.

And we need update hook(s) to update config on existing sites.

c4ndel4:
Thanks for the replies, but you have named the beast that terrifies me the most... The update_hook

> * Right now `field.storage.taxonomy_term.image` is provided by the `farm_plant_type` module. So this means that `farm_animal_type` and `farm_material_type` would need to depend on `farm_plant_type`. Should we (can we?) move this field storage to a more common core module?

Yes, probably move the crop_family to its own module and define all common fields in there. I did something similar with a new taxonomy "species" that would replace "crop_family". But maybe is better to keep the existing "crop_family" and if a nerd wants to call it species he (I) can change the label of that field.

> The entity view and form displays are another beast.. I think we might need an update hook to create them as well.

I already hate them. I made a module to add the species taxonomy without deleting the crop_family (https://github.com/c4ndel4/farm_species) and this is where all problems come from. The configuration items already exists and I must delete them before installing. And when I uninstall the module, the core views are lost. I thought I could read the yml files from the filesystem but I don't really know where the module it's going to be installed...

> Ideally we could use bundle fields on taxonomy terms like we do for our assets/logs/plans. That would effectively solve all the above problems. But iirc it isn't this easy to turn on bundle fields for taxonomy terms... hmm.

Could you point me to an example? Adding fields to an asset or a log was really easy and would be great to provide the same simplicity to other entities.

mstenta:
> Yes, probably move the crop_family to its own module and define all common fields in there.

No, `crop_family` should stay where it is. And it wouldn't make sense to package these common fields in the same module as it anyway. I'm not sure where the best place would be for them, honestly. The basic need would be a module for term field storage config, since that's the only place we need it (all other entity types are covered by bundle plugins provided by the `farm_entity` module).

> Ideally we could use bundle fields on taxonomy terms like we do for our assets/logs/plans. That would effectively solve all the above problems. But iirc it isn't this easy to turn on bundle fields for taxonomy terms... hmm.

I agree this would be ideal but I am very skeptical that it is worth the effort and potential maintenance cost moving forward.

mstenta:
I've opened a new PR that adds `file` and `image` fields to all taxonomy terms via `hook_entity_base_field_info()` (so we don't need config entities for every bundle):

- #807 

I'm going to close this one. @c4ndel4 if you are using these changes in your own instance right now then my PR may cause problems for you. Happy to help you transition to the new PR's approach if you need!

PR #532, opened 2022-05-07
Add 'owner' field to assets

jgOhYeah:
Hello,
This pull request adds an owner field to assets, based off the assign to functionality of logs.
For discussion and further details, please see here: https://farmos.discourse.group/t/is-it-possible-to-use-custom-fields-and-filtering-in-asset-views-and-filter-by-a-range-of-birth-dates/1139

# Note on installing and updating / why this shouldn't be merged just yet
Currently, attempting to install the updates to the `farm_owner` module using `update.php` causes `update.php` to crash (see [here](https://farmos.discourse.group/t/is-it-possible-to-use-custom-fields-and-filtering-in-asset-views-and-filter-by-a-range-of-birth-dates/1139/23) for @mstenta's investigations).

This does work when updated using drush or installing a fresh instance.

PR #537, opened 2022-05-30
Add owner field to assets

jgOhYeah:
Hello,
This is a cleaned up branch. The previous pull request with old branch is https://github.com/farmOS/farmOS/pull/532.

This pull request adds an owner field to assets, based off the assign to functionality of logs.
For discussion and further details, please see here: https://farmos.discourse.group/t/is-it-possible-to-use-custom-fields-and-filtering-in-asset-views-and-filter-by-a-range-of-birth-dates/1139

### DIFF
--- a/modules/core/owner/config/optional/system.action.asset_assign_action.yml
+++ b/modules/core/owner/config/optional/system.action.asset_assign_action.yml
@@ -1,0 +1,11 @@
+langcode: en
+status: true
+dependencies:
+  module:
+    - farm_owner
+    - asset
+id: asset_assign_action
+label: 'Assign asset'
+type: asset
+plugin: 'asset_assign_action'
+configuration: {  }
--- a/modules/core/owner/config/optional/system.action.asset_assign_action.yml
+++ b/modules/core/owner/config/optional/system.action.asset_assign_action.yml
@@ -5,7 +5,7 @@ dependencies:
     - farm_owner
     - asset
 id: asset_assign_action
-label: 'Assign asset'
+label: 'Set owner'
 type: asset
 plugin: 'asset_assign_action'
 configuration: {  }
--- a/modules/core/owner/config/optional/system.action.asset_assign_action.yml
+++ b/modules/core/owner/config/optional/system.action.asset_assign_action.yml
@@ -5,7 +5,7 @@ dependencies:
     - farm_owner
     - asset
 id: asset_assign_action
-label: 'Set owner'
+label: 'Assign owners'
 type: asset
 plugin: 'asset_assign_action'
 configuration: {  }
--- a/modules/core/owner/config/schema/farm_owner.schema.yml
+++ b/modules/core/owner/config/schema/farm_owner.schema.yml
@@ -1,4 +1,4 @@
 # Schema for actions.
-action.configuration.log_assign_action:
+action.configuration.assign_action:
   type: action_configuration_default
-  label: 'Configuration for the log assign action'
+  label: 'Configuration for the assign action'
--- a/modules/core/owner/config/schema/farm_owner.schema.yml
+++ b/modules/core/owner/config/schema/farm_owner.schema.yml
@@ -1,4 +1,7 @@
 # Schema for actions.
-action.configuration.assign_action:
+action.configuration.asset_assign_action:
   type: action_configuration_default
-  label: 'Configuration for the assign action'
+  label: 'Configuration for the asset assign action'
+action.configuration.log_assign_action:
+  type: action_configuration_default
+  label: 'Configuration for the log assign action'
--- a/modules/core/owner/farm_owner.module
+++ b/modules/core/owner/farm_owner.module
@@ -26,6 +26,13 @@ function farm_owner_entity_base_field_info(EntityTypeInterface $entity_type) {
         'view' => -70,
       ],
     ];
+
+    // Different description and label for assets.
+    if ($entity_type->id() == 'asset') {
+      $field_info['label'] = t('Owner');
+      $field_info['description'] = t('Optionally specify an owner for this asset.');
+    }
+
     $fields['owner'] = \Drupal::service('farm_field.factory')->baseFieldDefinition($field_info);
   }
 
--- a/modules/core/owner/farm_owner.module
+++ b/modules/core/owner/farm_owner.module
@@ -17,8 +17,8 @@ function farm_owner_entity_base_field_info(EntityTypeInterface $entity_type) {
   if (in_array($entity_type->id(), ['asset', 'log'])) {
     $field_info = [
       'type' => 'entity_reference',
-      'label' => t('Assigned to'),
-      'description' => t('Optionally assign this task to one or more people.'),
+      'label' => t('Owner'),
+      'description' => t('Optionally specify who is in charge of this record.'),
       'target_type' => 'user',
       'multiple' => TRUE,
       'weight' => [
@@ -26,12 +26,6 @@ function farm_owner_entity_base_field_info(EntityTypeInterface $entity_type) {
         'view' => -70,
       ],
     ];
-
-    // Different description and label for assets.
-    if ($entity_type->id() == 'asset') {
-      $field_info['label'] = t('Owner');
-      $field_info['description'] = t('Optionally specify an owner for this asset.');
-    }
 
     $fields['owner'] = \Drupal::service('farm_field.factory')->baseFieldDefinition($field_info);
   }
--- a/modules/core/owner/farm_owner.module
+++ b/modules/core/owner/farm_owner.module
@@ -13,8 +13,8 @@ use Drupal\Core\Entity\EntityTypeInterface;
 function farm_owner_entity_base_field_info(EntityTypeInterface $entity_type) {
   $fields = [];
 
-  // Add owner field to logs.
-  if ($entity_type->id() == 'log') {
+  // Add owner field to logs and assets.
+  if (in_array($entity_type->id(), ['asset', 'log'])) {
     $field_info = [
       'type' => 'entity_reference',
       'label' => t('Assigned to'),
--- a/modules/core/owner/farm_owner.post_update.php
+++ b/modules/core/owner/farm_owner.post_update.php
@@ -1,0 +1,42 @@
+<?php
+
+/**
+ * @file
+ * Updates farm_owner module.
+ */
+
+use Drupal\system\Entity\Action;
+
+/**
+ * Add 'owner' field to assets.
+ */
+function farm_owner_post_update_add_asset_owner(&$sandbox = NULL) {
+  $entity_type = 'asset';
+  $module_name = 'farm_owner';
+  $field_name = 'owner';
+
+  $field_info = [
+    'type' => 'entity_reference',
+    'label' => t('Assigned to'),
+    'description' => t('Optionally assign this task to one or more people.'),
+    'target_type' => 'user',
+    'multiple' => TRUE,
+    'weight' => [
+      'form' => -70,
+      'view' => -70,
+    ],
+  ];
+  $field_definition = \Drupal::service('farm_field.factory')->baseFieldDefinition($field_info);
+  \Drupal::entityDefinitionUpdateManager()
+    ->installFieldStorageDefinition($field_name, $entity_type, $module_name, $field_definition);
+
+  // Create action for assigning assets to users.
+  $action = Action::create([
+    'id' => 'asset_assign_action',
+    'label' => t('Assign asset'),
+    'type' => 'asset',
+    'plugin' => 'asset_assign_action',
+    'configuration' => [],
+  ]);
+  $action->save();
+}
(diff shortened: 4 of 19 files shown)


modules/core/owner/farm_owner.module
```diff
@@ -13,8 +13,8 @@ use Drupal\Core\Entity\EntityTypeInterface;
 function farm_owner_entity_base_field_info(EntityTypeInterface $entity_type) {
   $fields = [];
 
-  // Add owner field to logs.
-  if ($entity_type->id() == 'log') {
+  // Add owner field to logs and assets.
+  if (in_array($entity_type->id(), ['asset', 'log'])) {
     $field_info = [
       'type' => 'entity_reference',
       'label' => t('Assigned to'),
```
paul121: Same, maybe update this field description as well.

modules/core/owner/farm_owner.post_update.php
```diff
@@ -0,0 +1,42 @@
...
+ */
+function farm_owner_post_update_add_asset_owner(&$sandbox = NULL) {
+  $entity_type = 'asset';
+  $module_name = 'farm_owner';
+  $field_name = 'owner';
+
+  $field_info = [
+    'type' => 'entity_reference',
+    'label' => t('Assigned to'),
+    'description' => t('Optionally assign this task to one or more people.'),
```
paul121: Should we change this field description since assets are not "tasks" ? Is there more general language we could use here?

modules/core/owner/config/schema/farm_owner.schema.yml
```diff
@@ -1,4 +1,4 @@
 # Schema for actions.
-action.configuration.log_assign_action:
+action.configuration.assign_action:
```
paul121: I think we need to implement a hook to update existing schema on sites if we change this here. I haven't done this before but think I have seen it in other projects, maybe the `log` module most recently?

modules/core/owner/config/optional/system.action.asset_assign_action.yml
```diff
@@ -0,0 +1,11 @@
+langcode: en
+status: true
+dependencies:
+  module:
+    - farm_owner
+    - asset
+id: asset_assign_action
+label: 'Assign asset'
```
paul121: While "Assign log" makes sense, I'm not sure that "Assign asset" makes as much sense. Maybe "Assign owner"?

modules/core/owner/src/Form/AssignActionForm.php
```diff
@@ -136,8 +137,29 @@ public function getConfirmText() {
...
+        $users_description = $this->t("Select people to assign these logs to.");
+        $operation_description = $this->t('Select "Append" if you want to add users to the logs, but keep the existing assignments. Select "Replace" if you want to replace existing assignments with the ones specified above.');
+        break;
+
+      case 'asset':
+        $this->entityType = $this->entityTypeManager->getDefinition('asset');
+        $title = $this->t("Assign asset(s) to");
+        $users_description = $this->t("Select people to assign these assets to.");
+        $operation_description = $this->t('Select "Append" if you want to add users to the assets, but keep the existing assignments. Select "Replace" if you want to replace existing assignments with the ones specified above.');
+        break;
```
paul121: Instead of the switch statement I think we could use single `t()` function but insert `asset` or `log` in as a parameter. This should work for both the title, users description, and operation description because all of the strings here are nearly identical! The benefit is that it is less strings to translate.

Although if we did want to make major changes to the strings, this switch statement looks great.

modules/core/ui/views/config/install/views.view.farm_asset.yml
```diff
@@ -479,6 +479,71 @@ display:
...
+          id: owner_target_id
+          table: asset__owner
+          field: owner_target_id
+          relationship: none
+          group_type: group
+          admin_label: ''
+          entity_type: asset
+          entity_field: owner
+          plugin_id: field
+          label: 'Assigned to'
```
paul121: "Assigned to" or "owner" ?

modules/core/owner/src/Plugin/Action/AssetAssign.php
```diff
@@ -0,0 +1,17 @@
...
+ * Action that assigns users to assets.
+ *
+ * @Action(
+ *   id = "asset_assign_action",
+ *   label = @Translation("Assign users to assets."),
+ *   type = "asset",
+ *   confirm_form_route_name = "farm_owner.asset_assign_action_form"
+ * )
+ */
+class AssetAssign extends AssignBase {
```
paul121, line 15: This is great, love the idea to use the `AssignBase` class.

paul121 (dismissed):
This looks pretty good @jgOhYeah !

Generally I'm curious about the language of this field, "Assigned assets" or "Owned assets", etc. It comes up a few places here. Apologies if that was already "decided" elsewhere, just something that came to mind.

jgOhYeah:
Thanks for looking at it. The 'assign asset' language was mainly because I was trying to keep the wording fairly similar between logs and assets - it can be changed. Would 'Owner' be the more correct label in this case? As a side note, [this Drupal issue](https://www.drupal.org/project/farm/issues/2400065) discusses some of the various interpretations of 'ownership' of assets, which may affect what the field gets called.

modules/core/owner/src/Plugin/Action/AssetAssign.php
jgOhYeah: Thanks to @mstenta for the suggestion.

modules/core/owner/src/Form/AssignActionForm.php
jgOhYeah: I did have the parameter substitution set up like you suggest at one point, however the php codesniffer tests didn't really like the idea of putting variable substitutions in `t()` functions. I'm not a language expert, but I recall that some languages have words in different orders to others, which I'm not sure how well the variable substitution would handle. I can change it back if this isn't an issue.

modules/core/owner/config/schema/farm_owner.schema.yml
jgOhYeah: Would it be possible to link to a previous example if you know roughly where one is please? I've had a look for update hooks and post update hooks in the `/modules/log/` and `/modules/core/log` folders to base this off, but haven't come across any. I could be missing something subtle (or pretty darn obvious :)), but the assign log action still seems to behave alright on an existing site, even without this update hook.

mstenta:
Just thought about this again while adding some records to my farmOS recently... sorry for the long delay - been very busy with things.

I wonder: should we change the label of the field on Logs to "Owner" as well? Perhaps that's what you were implying too @jgOhYeah:

> Would 'Owner' be the more correct label in this case?

It is more general than "Assigned to", which means it works in more scenarios. "Assigned to" is nice because it's easy to understand *in the context of delegated work assignments*, so "Owner" is a bit of a step down in that context, but maybe the tradeoff is acceptable? It generalizes the meaning a bit, which allows it to be more flexible depending on context and convention.

There are contexts where "Assigned to" feels weird - for example if you are a one-person operation. I suppose "Owner" is odd in that context as well, though. The idea is "who performed this task?" What is the best label for that? "Owner" is the right *idea* I still believe (and therefore machine name). So perhaps it's the best *label* as well.

jgOhYeah:
> Perhaps that's what you were implying too @jgOhYeah:
I was originally confirming that 'Owner' was the correct replacement for 'Assigned to' for assets, but it wasn't particularly clearly worded and looking back again see that or something similar is what @paul121 was suggesting in the first place.

I don't really have a preference as to the naming. The only risk with changing the log field label is consistency for those who are used to the old label, but this isn't a major change.

I agree that depending on the situation, 'Assigned to' vs 'Owner' have their own advantages and disadvantages. For our installation it can sometimes be a presumed who did it as well :).

mstenta:
Finally got a chance to dig in and review/test this! Thanks again @jgOhYeah and I appreciate your patience. :-)

> I wonder: should we change the label of the field on Logs to "Owner" as well?

I think I am leaning more towards this. It would allow us to keep a lot more of these labels and strings consistent, which has some benefits (eg: less code, easier to maintain translations).

I rebased the working branch, squashed the codesniffer commits into the first commit, and then added a new commit that updates the language to this effect.

I used "owner" in noun contexts, and "assign" in verb contexts, roughly speaking. It still uses "Assigned assets" and "Assigned logs" in the Views of records owned by users, but I think that's OK.

In this sense, we can think of it as "_record_ ownership" - as in "who is in charge of this record" (whatever that is taken to mean). Some may use this to mean "task assignment" (in the case of logs), or "who is in charge of this" (in the case of logs/assets), others may use it to mean "the legal owner" (of assets).

Lastly, I want to raise one last idea/question: should we auto-assign asset owners in the same way we auto-assign log owners (when a log is saved, if no owner is assigned, it assigns it to the current user)? I actually went ahead and implemented this (which required also opening this MR: https://www.drupal.org/project/farm/issues/3306227) - but then in testing it I realized this means it's going to add "Owner" to all existing assets - and maybe we don't really want that. Maybe we should leave it up to the user to decide how ownership is assigned for assets. Although it feels weird for logs and assets to behave differently in this aspect, doesn't it? I kept this change out of the branch for now, but I'd like to hear what others think. We could also punt on this decision and open a followup issue for consideration (I can add my commit there).

I feel like this is ready to merge, pending thoughts/review of all of the above.

mstenta:
Meant to respond to this @paul121:

> I think we need to implement a hook to update existing schema on sites if we change this here.

I'm not sure anything is needed for schema changes, is it? Maybe I'm not understanding.

symbioquine:
> it's going to add "Owner" to all existing assets - and maybe we don't really want that. Maybe we should leave it up to the user to decide how ownership is assigned for assets. Although it feels weird for logs and assets to behave differently in this aspect, doesn't it? I kept this change out of the branch for now, but I'd like to hear what others think.

My gut feeling is that it's a little odd to make the logic assume the asset creator is the asset owner.

The best rational for that stance I can come up with is that the lifecycle of assets and logs is often quite different - many assets are perennial or are long-lived equipment/land-features/etc. Contrasted with logs where I believe the default of self-assigning logs is pretty sane since the person who created the log presumably has logical ownership of the task - at least insofar as capturing it as a log goes - and may already have completed it themselves.

Maybe it would be better to provide a "default interpretation" of the no-owner state for assets (and make that the default) which somehow describes them as "belonging to the farm as a whole"?

mstenta:
I like that reasoning @symbioquine - and leaving this as-is means that "Owner" will not appear on assets unless it is filled in. (There is still an "Owner" column" in the Views, which will be empty, but I think that's OK for now.)

jgOhYeah:
Thanks for fixing this up. I would be inclined to agree with @symbioquine 's reasoning at this stage. In our case we have certain assets like implements that require more research on who exactly owns them or they don't really have an owner, so being able to leave than field blank by default would be useful.

mstenta:
Ah hmm tests are failing:

```
Drupal\Core\Config\Schema\SchemaIncompleteException: Schema errors for system.action.asset_assign_action with the following errors: system.action.asset_assign_action:configuration missing schema
```

mstenta:
> Ah hmm tests are failing:

I pushed a commit that attempts to fix this... waiting for results from GitHub Actions...

https://github.com/mstenta/farmOS/actions/runs/2934888499

mstenta:
Getting a different schema error now, related to the changes to Views config:

`Drupal\Core\Config\Schema\SchemaIncompleteException: Schema errors for views.view.farm_people with the following errors: views.view.farm_people:display.default.display_options.nothing missing schema`

Will need to debug that another time...

mstenta:
I figured out the other schema error. @jgOhYeah it appears you may have copied and pasted a chunk of code into the `views.view.farm_people.yml` file, does that sound correct? There were two chunks added to that file, but the second one was invalid. Deleting it allowed the tests to pass. It does not affect the intention of the change, so the View still has the new "Assigned assets" column. Let me know if you have any concerns, but I think that explains what happened.

I pushed a fixup commit for this, but I will squash it into the first commit of the branch before merging. I just kept it separate for review.

jgOhYeah:
> I figured out the other schema error. @jgOhYeah it appears you may have copied and pasted a chunk of code into the `views.view.farm_people.yml` file, does that sound correct? There were two chunks added to that file, but the second one was invalid. Deleting it allowed the tests to pass. It does not affect the intention of the change, so the View still has the new "Assigned assets" column. Let me know if you have any concerns, but I think that explains what happened.
> 
> I pushed a fixup commit for this, but I will squash it into the first commit of the branch before merging. I just kept it separate for review.

I think that must have happened during my attempt at unscrambling some of the merges I'd done and making a fresh branch. I recall that involved a couple of manual merges and something obviously went wrong. Thanks for fixing it, sorry it caused so much bother.

mstenta:
No worries! Easy fix. :-)

I think this is ready for final review and merge now. All tests are passing.

mstenta:
(Just merged my fixup into the initial commit and rebased onto latest `2.x` to resolve `CHANGELOG.md` conflict.)

mstenta:
Rebased and added one more commit to document the new owner field in the user guide and data model docs.

If tests pass I'll merge this.

mstenta:
Merged! At long last!!!

Thanks again @jgOhYeah! :-D :tada:

jgOhYeah:
Thanks for all the effort you have put into this @mstenta and others. This has / will make farmOS much more useful for us and hopefully will for others.

PR #544, opened 2022-06-19
Issue #3290929: Provide a farmOS map form element

mstenta:
See https://www.drupal.org/project/farm/issues/3290929


docs/development/module/maps.md
```diff
@@ -0,0 +1,358 @@
...
+   *
+   * @param \Drupal\farm_map\Event\MapRenderEvent $event
+   *   The MapRenderEvent.
+   */
+  public function onMapRender(MapRenderEvent $event) {
+    $event->addBehavior('mybehavior');
+  }
+
+}
+```
```
symbioquine, line 358: Since the `farm_map` render element has its own documentation page now, I wonder whether it would be good to have a section showing how to use the element as a "prototype" (i.e. with a custom instantiator) also. It would be pretty easy to copy my example from [this commit](https://github.com/farmOS/farmOS/commit/c92fe90985b65e09ec92fc802273c0b959974e0c), but I can also add it as a follow-up if you prefer...

modules/core/map/farm_map.post_update.php
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * @file
+ * Post update hooks for the farm_map module.
+ */
+
+use Drupal\farm_map\Entity\MapBehavior;
+use Drupal\farm_map\Entity\MapType;
+
+/**
+ * Generalize geofield map types and behavior.
```
symbioquine, line 12: Is there a pattern / best practice of linking back to the relevant issue here or do we just rely on the git history to provide the justification for these?

modules/core/map/farm_map.post_update.php
```diff
@@ -0,0 +1,40 @@
...
+  $input_behavior->save();
+
+  // Delete the geofield behavior.
+  $geofield_behavior = MapBehavior::load('geofield');
+  $geofield_behavior->delete();
+
+  // Delete the geofield_widget map type.
+  $geofield_widget_map_type = MapType::load('geofield_widget');
+  $geofield_widget_map_type->delete();
+}
```
symbioquine, line 40: Is this idempotent or does Drupal take care of only running this once?

docs/development/module/maps.md
mstenta: Ah good point! I actually forgot about the custom instantiator functionality. :-D

If you wouldn't mind following up in a separate PR that would be great to add.

As I was writing this page I was also curious about including some information about how to build a behavior with Webpack, so that it could use more OpenLayers features if it wanted to. I've never done that myself, though, so figured it could be a potential follow-up as well.

modules/core/map/farm_map.post_update.php
mstenta: Drupal takes care of only running it once. And it only runs on *existing* sites - not on new installs.

modules/core/map/farm_map.post_update.php
mstenta: I've just relied on Git to provide the context. These `[module].post_update.php` files will be relatively linear in their histories (only additions, no refactors, no moving code between files), so `git blame` can get you to the history you need fairly easily. But it's a good point/question to raise, and maybe we should start including the relevant issue in comments for reference.

symbioquine (approved):
LGTM!

modules/core/map/src/Element/FarmMapInput.php
```diff
@@ -0,0 +1,149 @@
...
+   * @return array
+   *   The processed element.
+   */
+  public static function processElement(array $element, FormStateInterface $form_state, array &$complete_form) {
+    $element['#tree'] = TRUE;
+
+    // Merge provided map behaviors into defaults. Enable wkt and input
+    // behaviors if #disabled is not TRUE.
+    $default_behaviors = !$element['#disabled'] ? ['wkt', 'input'] : [];
+    $behaviors = array_merge($default_behaviors, $element['#behaviors']);
```
paul121, line 65: This excludes the `wkt` behavior when the element is `#disabled`, but I think we might want to include it? Idea being you want to show the geometry in the map, but not have edit controls. I see we set `behaviors.wkt.edit` when the element is disabled, but that won't do anything if the `wkt` behavior isn't added.

I seem to remember that we add the `wkt` in an event subscriber as well, maybe this is covered there.

paul121 (approved):

modules/core/map/src/Element/FarmMapInput.php
mstenta: Yes, the `MapRenderEventSubscriber` still adds the `wkt` behavior whenever WKT is passed in:

https://github.com/mstenta/farmOS/blob/c60b7d159757e8659965fd865b0853ef9e6ec606/modules/core/map/src/EventSubscriber/MapRenderEventSubscriber.php#L61

It is a bit odd to have that logic in two different places now, perhaps. So might be worth bringing that over too in a follow-up commit?

PR #547, opened 2022-06-27
Generate unique form IDs for quick forms

paul121:
Currently all quick forms get the same form ID `quick_form`. We can implement the core `BaseFormIdInterface` to make the base form ID be `quick_form` and build a unique form ID for each quick form from the `id` route parameter.


mstenta:
Good idea thanks @paul121!

Issue #539, opened 2022-06-02
Add button/menu item for data stream notifications collection page


paul121:
This page is not easy to find right now.

mstenta:
@Farmer-Eds-Shed opens a pull request for this: #555

mstenta:
Closing this now that the PR was merged. Thanks again @Farmer-Eds-Shed!

PR #555, opened 2022-08-06
Add button/menu item for data stream notifications collection page #539

Farmer-Eds-Shed:
Add button/menu item for data stream notifications collection page #539

Adds to settings menu
image

### DIFF
--- a/modules/core/data_stream/modules/notification/data_stream_notification.info.yml
+++ b/modules/core/data_stream/modules/notification/data_stream_notification.info.yml
@@ -5,3 +5,4 @@ package: farmOS
 core_version_requirement: ^9
 dependencies:
   - farm:data_stream
+  - farm:farm_notification
--- a/modules/core/notification/farm_notification.routing.yml
+++ b/modules/core/notification/farm_notification.routing.yml
@@ -1,0 +1,7 @@
+farm_notification.settings:
+  path: '/farm/settings/notification'
+  defaults:
+    _form: '\Drupal\farm_notification\Form\NotificationSettingsForm'
+    _title: 'Notification settings'
+  requirements:
+    _permission: 'administer farm notification'
--- a/modules/core/notification/src/Form/NotificationSettingsForm.php
+++ b/modules/core/notification/src/Form/NotificationSettingsForm.php
@@ -1,0 +1,41 @@
+<?php
+
+namespace Drupal\farm_notification\Form;
+
+use Drupal\Core\Form\FormBase;
+use Drupal\Core\Form\FormStateInterface;
+
+/**
+ * Provides a farm_notification settings form.
+ */
+class NotificationSettingsForm extends Formbase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getFormId() {
+    return 'farm_noication_settings';
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildForm(array $form, FormStateinterface $form_state) {
+
+    // Add a placeholder message.
+    $form['placeholder'] = [
+      '#type' => 'markup',
+      '#markup' => $this->t('Select the tab for the type of notification to configure. Notification modules can add additional tabs to this page.'),
+    ];
+
+    return $form;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function submitForm(array &$form, FormStateInterface $form_state) {
+
+  }
+
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 - [Fix for Autocomplete dropdown not showing in Chrome on Android #552](https://github.com/farmOS/farmOS/issues/552)
+- [Add button/menu item for data stream notifications collection page #555](https://github.com/farmOS/farmOS/pull/555)
 
 ## [2.0.0-beta6] 2022-07-30
 
--- a/modules/core/data_stream/modules/notification/data_stream_notification.links.task.yml
+++ b/modules/core/data_stream/modules/notification/data_stream_notification.links.task.yml
@@ -1,0 +1,5 @@
+data_stream_notification.settings:
+  base_route: farm_settings.settings_page
+  route_name: entity.data_stream_notification.collection
+  title: 'Data Stream Notifications'
+  weight: 10
--- a/modules/core/data_stream/modules/notification/data_stream_notification.links.task.yml
+++ b/modules/core/data_stream/modules/notification/data_stream_notification.links.task.yml
@@ -1,5 +1,5 @@
 data_stream_notification.settings:
-  base_route: farm_settings.settings_page
   route_name: entity.data_stream_notification.collection
-  title: 'Data Stream Notifications'
+  parent_id: farm_notification.settings
+  title: 'Data Streams'
   weight: 10
--- a/modules/core/data_stream/modules/notification/data_stream_notification.post_update.php
+++ b/modules/core/data_stream/modules/notification/data_stream_notification.post_update.php
@@ -1,0 +1,15 @@
+<?php
+
+/**
+ * @file
+ * Post update functions for Data Stream Notifications module.
+ */
+
+/**
+ * Install farmOS Notifications module.
+ */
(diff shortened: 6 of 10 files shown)


mstenta:
Looks OK to me! Requesting @paul121's review since he is more familiar with data streams.

One small thought (not a blocker): I could see this being a sub-tab of a new general "Notifications" tab in the future, which may also include other notification type settings (eg: worker task notifications).

FYI I'll plan to rebase and squash these commits before merging.

Farmer-Eds-Shed:
I'll not be offended if it's better suited elsewhere, was bugging me with having nowhere though.
Funny enough I had considered a new notifications tab as I am also working on some notifications modules that could occupy a sub tab.

mstenta:
> Funny enough I had considered a new notifications tab as I am also working on some notifications modules that could occupy a sub tab.

The main question is: what module should be responsible for the "Notifications" tab, if multiple modules need to rely on it. Doesn't make sense to have it in the Data Stream module. And we don't really want it to show up all the time, otherwise it would be empty without any child tabs.

Farmer-Eds-Shed:
Could it not just be it's own module in this case? Not installed by default but required by any new notifications module.

paul121:
Interesting thanks for working on this @Farmer-Eds-Shed !! My first thought is that this seems a little weird in the Settings pages, but it's not awful. Agree that it would look better under a "Notifications" tab...

> The main question is: what module should be responsible for the "Notifications" tab, if multiple modules need to rely on it. Doesn't make sense to have it in the Data Stream module. And we don't really want it to show up all the time, otherwise it would be empty without any child tabs.

Agreed. But perhaps this could just be provided by `farm_settings` ? I do believe that the parent tab won't be displayed if there are no children tabs available or accessible (due to permissions). If it's possible for a "local task" to be provided that doesn't actually link to anything (there wouldn't be a canonical "notifications" page), then `farm_settings` could just provide the notification "Namespace" and let other modules populate it.

Docs for reference: https://www.drupal.org/docs/drupal-apis/menu-api/providing-module-defined-local-tasks#s-define-static-local-tasks

> route_name (required): The machine name of the local task route - this also determines where it's displayed.

Hmm but looks like a `route_name` is required. Not sure if there is a clever way to get around that... maybe we could make a dummy route of some sort? Even if it was simply "Select the tab for the type of notification you would like to configure." I think we would either need a canonical notification page of some sort or dynamic logic to make the Notification tab go to the lowest-weight notification tab that is provided. Hmmm. Any :bulb: ideas @Farmer-Eds-Shed @mstenta ?

_Alternatively...._ at one point I think we had discussed adding it as an Action Link on the data stream view tab. So when you're viewing a data stream, there could be a button "Configure notifications" in the top right. This would bring you to this page to configure *all* notifications, not only notifications for the single data stream, so perhaps that is confusing anyways.

Farmer-Eds-Shed:
> Interesting thanks for working on this

No Problem, It's a feature I want to use myself, so I felt it needs to go somewhere. 
> My first thought is that this seems a little weird in the Settings pages, but it's not awful

Yea, probably can't disagree, it's the least worst of the menu places I experimented with though 😆.

> _Alternatively...._ at one point I think we had discussed adding it as an Action Link on the data stream view tab. So when you're viewing a data stream, there could be a button "Configure notifications" in the top right. This would bring you to this page to configure _all_ notifications, not only notifications for the single data stream, so perhaps that is confusing anyways.

Not sure I like this one either as I agree it could be confusing, it would be better to link to notifications for the single data stream this way, but then it would still make sense for a notifications menu to manage all notifications.

> Hmm but looks like a `route_name` is required. Not sure if there is a clever way to get around that... maybe we could make a dummy route of some sort? Even if it was simply "Select the tab for the type of notification you would like to configure." I think we would either need a canonical notification page of some sort or dynamic logic to make the Notification tab go to the lowest-weight notification tab that is provided. Hmmm. Any 💡 idea

If creating an empty menu is an issue, then may be a basic notifications module is an option, similar to the data streams one but a little less obscure for the average user. I'd imagine the option of notifications for log creation/update could be popular and not weird to be activated alongside any other notification module that could be wanted in the future?  

I still propose using the settings menu for now if implementing the notifications menu may be a bit off,  as somewhere is better than nowhere, I opened the PR as I want to keep in line with farmOS without maintaining minor customizations on my server, but I can just as easily create a module to achieve the same. There are others interested in data streams but are probably unable to find this feature.

paul121:
> I still propose using the settings menu for now if implementing the notifications menu may be a bit off, as somewhere is better than nowhere, I opened the PR as I want to keep in line with farmOS without maintaining minor customizations on my server, but I can just as easily create a module to achieve the same. There are others interested in data streams but are probably unable to find this feature.

Totally agree, thanks for expressing this. Given that this will only appear if that module is enabled, I agree that this is better than nowhere. In the future we could move this without it being a "breaking" change as well.

paul121 (approved):
I just wanted to review which permission is required to access this `entity.data_stream_notification.collection` route..

I believe it is this `access data_stream_notification overview` which we only grant to `config` roles, so this seems safe to me :+1: 

https://github.com/farmOS/farmOS/blob/fc77c86abe6b142702e7b103cd85e0c207333b79/modules/core/data_stream/modules/notification/data_stream_notification.managed_role_permissions.yml#L3

mstenta:
Thanks for reviewing @paul121! And all the good thoughts @Farmer-Eds-Shed!

>  may be a basic notifications module is an option

I started an initial branch for this on yesterday's dev call: https://github.com/mstenta/farmOS/tree/2.x-notification

We discussed that this core `farm_notification` module could eventually become the home for general notification logic (perhaps extracted/generalized from the `data_stream_notification` module). For now it just provides a "Notifications" tab under settings, which the data stream notification module can depend on.

I think I'll merge this, then rebase this PR on top and put the new menu link under that. Then we can look at generalizing the data stream notification code as a future step.

mstenta:
Updated screenshot:

Screenshot from 2022-08-12 09-21-51

modules/core/notification/farm_notification.routing.yml
```diff
@@ -0,0 +1,7 @@
+farm_notification.settings:
+  path: '/farm/settings/notification'
+  defaults:
+    _form: '\Drupal\farm_notification\Form\NotificationSettingsForm'
+    _title: 'Notification settings'
+  requirements:
+    _permission: 'administer farm notification'
```
paul121, line 7: I believe we need to add this to `managed_role_permissions` like here: https://github.com/farmOS/farmOS/blob/fc77c86abe6b142702e7b103cd85e0c207333b79/modules/core/data_stream/modules/notification/data_stream_notification.managed_role_permissions.yml#L3

modules/core/data_stream/modules/notification/data_stream_notification.info.yml
```diff
@@ -5,3 +5,4 @@ package: farmOS
 core_version_requirement: ^9
 dependencies:
   - farm:data_stream
+  - farm:farm_notification
```
paul121, line 8: I think we need an update hook to enable `farm_notification` on sites that already have `data_stream_notification` ?

modules/core/notification/src/Form/NotificationSettingsForm.php
```diff
@@ -0,0 +1,41 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildForm(array $form, FormStateinterface $form_state) {
+
+    // Add a placeholder message.
+    $form['placeholder'] = [
+      '#type' => 'markup',
+      '#markup' => $this->t('Notification modules can add additional tabs to this page.'),
```
paul121: Could we add a prompt for the user? On mobile or small screens the tab options won't be visible. eg: "Select the tab for the type of notification to configure. Notification modules can add additional tabs to this page."

paul121 (changes requested):

mstenta:
Thanks @paul121! Good eyes...

I made all the changes you requested in two "fixup" commits (to be squashed into the appropriate original commits before merging).

I also made one other small change: I made "Notifications" plural in the tab titles.

Requesting re-review before merge.

mstenta:
PS: I squashed the commits as described and pushed to a new `2.x-notification` branch if you want to see the final result: https://github.com/farmOS/farmOS/compare/2.x...mstenta:farmOS:2.x-notification

paul121 (approved):
LGTM!

mstenta:
Merged - thanks again @Farmer-Eds-Shed and @paul121!

PR #556, opened 2022-08-19
Correct and standardize OAuth curl examples

paul121:
These are still using 1.x conventions for the `client_id` and `scope`. A few other minor changes to standardize these docs.


mstenta:
Thanks @paul121!

PR #557, opened 2022-08-19
Improve API docs

paul121:
The root API index.md page doesn't include very much useful information about the API. Most of the JSON:API features are documented in `changes.md`, but a new user would have no reason to go to this page. I think we should pull some of the high-level JSON:API concepts into `index.md` for simplicity.

I also added explicit links to various Drupal core JSON:API documentation pages to *emphasize* that all of the specific features are already documented there. In the future it would be nice to have our own examples of common API CRUD requests.

Lastly, included a simple link to the client libraries & documentation at the top. Because that is easy and short.

Feedback welcome just wanted to take a quick stab at improving this page :-)


mstenta:
Much better! Thanks @paul121!!

PR #564, opened 2022-09-09
Simplify location query processing of asset ids.

paul121:
Just a little improvement here to remove the iteration over each result record.

I also removed the `array_unique` by adding a `SELECT DISTINCT a.id`. I don't think passing unique ids is a requirement for the `entityStorage->loadMultiple($asset_ids)` but it doesn't hurt.


paul121:
The first test runner attempt failed due to mariadb connection. Re-ran the jobs and they are passing.

paul121:
Changelog!

mstenta:
Thanks @paul121 - LGTM! Generally I try to avoid `DISTINCT` but in this case I think it makes good sense. :+1:

PR #494, opened 2022-01-25
Change getGroupMembers to return an array of assets keyed by their ID

paul121:
See https://www.drupal.org/project/farm/issues/3259245

> Add/update tests to ensure asset IDs are returned as the keys, with both $recurse options.

It didn't seem like we had any explicit tests for the `recursive` parameter so I added a new dedicated test method for this.

PR #565, opened 2022-09-09
Update AssetLocationInterface::getAssetsByLocation to return asset objects keyed by ID

paul121:
I noticed this would be a super simple change to implement. Similar to #494 

The `farm_location` module already returns assets keyed by ID, but `farm_group` was using loosing this structure with its use of `array_merge`. Because `GroupMembershipInterface::getGroupMembers` already returns asset objected keyed by ID, we can simply change `array_merge` to use the `+=` assignment. I've done this in the last commit of this PR.

The bulk of this PR is actually updating the tests to accommodate this change. When revisiting the location and group membership tests I found it hard to follow the logic. In many places we're simply asserting that the `count()` of the asset array is correct, but not necessarily the actual contents of the array. I think this is OK once you understand the complete logic of the tests, but it makes understanding the logic of the tests a bit harder.

To (hopefully) make things easier I've added a new `FarmAssetTestTrait` with a helper function `assertCorrectAssets()` for asserting the correct contents of "asset arrays". The function allows you to *optionall*y check that the asset ids are included in the array keys, too. The large benefit of this is that it makes the tests more readable:

```diff
-    $this->assertEquals(1, count($this->groupMembership->getGroupMembers([$second_group])), 'Future group assignment logs do not affect members.');
+    $this->assertCorrectAssets([$animal], $this->groupMembership->getGroupMembers([$second_group]), TRUE, 'Future group assignment logs do not affect members.');
```

and also reduces the duplication of asserting multiple characteristics of the "asset array" while making our tests more consistent:
```diff
    $group_members = $this->groupMembership->getGroupMembers([$first_group, $second_group]);
-    $this->assertEquals(2, count($group_members), 'Group members from multiple groups can be queried together.');

    // Test that asset IDs are preserved as the array keys.
-    $this->assertTrue(in_array($animal->id(), array_keys($group_members)), 'The ID of the first animal is preserved.');
-    $this->assertTrue(in_array($second_animal->id(), array_keys($group_members)), 'The ID of the second animal is preserved.');
+    $this->assertCorrectAssets([$animal, $second_animal], $group_members, TRUE, 'Group members from multiple groups can be queried together.');
```

I added and implemented this trait in the first two commits of this PR (tests passing) to demonstrate how this can be used even for arrays that do not return asset IDs as their keys. I imagine there are various places in our tests that could be refactored to use this trait, however I only considered the location and group membership tests for now. I don't think there should too many other requirements for such a helper function. The message passing is a little odd, but effective enough for reporting test failures.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [Improve API docs #557](https://github.com/farmOS/farmOS/pull/557)
 - [Issue #3305724 by pcambra, m.stenta: Allow map type and behaviors to be configurable in map blocks](https://www.drupal.org/project/farm/issues/3305724)
 - [Update Drupal core to 9.4.x](https://github.com/farmOS/farmOS/pull/566)
+- [Update AssetLocationInterface::getAssetsByLocation to return asset objects keyed by ID #565](https://github.com/farmOS/farmOS/pull/565)
 
 ### Fixed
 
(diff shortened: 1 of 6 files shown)


mstenta:
Thanks @paul121! Good improvements!

PR #567, opened 2022-09-14
Update excluded configuration example to not use module settings

paul121:
Our automated update docs describe how farm_update "does not touch simple" configuration (eg: module settings) - only configuration entities (eg: Views)" *but* the example hook implementations use the `mymodule.settings` simple config. 
https://farmos.org/development/module/updates/#exclude-specific-configuration

Lets update this to be a custom view provided by mymodule, `views.view.mymodule_custom`.


mstenta:
Good catch! Thanks!

Issue #541, opened 2022-06-07
adding circular membership in groups


botlfarm:
**Describe the bug**
Adding a group to itself breaks group member views

**To Reproduce**
1. create a new group with no assets
2. create activity log selecting the new group as an asset
3. make the log a group assignment and select the group
4. mark as done and save the log
5. try to view the group members (hangs or gets fatal error)

**Expected behavior**
Expect the group membership page to display. 

To fix I deleted the log and it acts as normal. 

Tested on https://farmos-demo.rootedsolutions.io/demo/2.x

This was initially found in my working environment but when I tried deleting the log to fix the problem it appeared to not delete the log completely. /logs and /logs/activity showed error `The website encountered an unexpected error. Please try again later.` Troubleshooting with @mstenta showed parts of the deleted log still in the DB. But I can't reproduce this in the test environment. Some of this initial troubleshooting in [chat](https://matrix.to/#/!rxZezKXfxOKJxTgNGs:matrix.org/$16546229706531lgEbv:matrix.org?via=matrix.org&via=libera.chat&via=kernelpanic.cafe)

mstenta:
Thanks for opening this @botlfarm! I thought we had some protection against circular group membership - but I guess I was remembering this: https://www.drupal.org/project/farm/issues/3179016 (and maybe the old 1.x logic we had for preventing group circular references: https://www.drupal.org/project/farm/issues/2915217).

We can fix this in the `farm_group` module's event subscriber, I think: https://github.com/farmOS/farmOS/blob/2.x/modules/asset/group/src/EventSubscriber/LogEventSubscriber.php

Along with an automated test...

mstenta:
> We can fix this in the farm_group module's event subscriber, I think

Actually a constraint validator on the `asset` field is the better approach.

Also thinking - we probably need this for location as well - to prevent a location from being located in itself. :-P

mstenta:
And they need to be recursive - to prevent Group A from being a member of Group B which is a member of Group A (and for locations).

mstenta:
> Also thinking - we probably need this for location as well - to prevent a location from being located in itself. :-P

Spun off a separate issue for this: https://github.com/farmOS/farmOS/issues/561

mstenta:
I have most of the logic for this working, as well as automated tests for it. I'm just working through some minor issues with the dependencies for automated tests... hope to have those resolved soon so we can get this fix in.

mstenta:
[...]

PR #568, opened 2022-09-17
Prevent circular asset location

### LINKED ISSUE #561: Prevent circular asset location
The (potential) need for this came up in the related issue https://github.com/farmOS/farmOS/issues/541.

Needs testing and steps to reproduce...

mstenta:
Fixes #561

### DIFF
--- a/modules/core/location/farm_location.module
+++ b/modules/core/location/farm_location.module
@@ -45,6 +45,18 @@ function farm_location_entity_base_field_info(EntityTypeInterface $entity_type) 
 }
 
 /**
+ * Implements hook_entity_base_field_info_alter().
+ */
+function farm_location_entity_base_field_info_alter(&$fields, EntityTypeInterface $entity_type) {
+  /** @var \Drupal\field\Entity\FieldConfig[] $fields */
+
+  // Prevent creating circular asset location.
+  if ($entity_type->id() == 'log' && !empty($fields['asset'])) {
+    $fields['asset']->addConstraint('CircularAssetLocation');
+  }
+}
+
+/**
  * Sets the default value for asset is_location boolean field.
  *
  * @param \Drupal\Core\Entity\ContentEntityInterface $entity
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - [Update Drupal core to 9.5.x](https://github.com/farmOS/farmOS/pull/621)
 
+### Fixed
+
+- [Prevent circular asset location #568](https://github.com/farmOS/farmOS/pull/568)
+
 ## [2.0.0-rc1] 2022-12-15
 
 PHP 8+ is now the recommended minimum version requirement for farmOS. The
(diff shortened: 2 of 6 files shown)


mstenta:
I think this needs to be postponed until https://www.drupal.org/project/farm/issues/3310286 is done.

See this comment for reasoning: https://www.drupal.org/project/farm/issues/3310286#comment-14697132

modules/core/location/farm_location.module
```diff
@@ -44,6 +44,18 @@ function farm_location_entity_base_field_info(EntityTypeInterface $entity_type)
...
 }
 
+/**
+ * Implements hook_entity_base_field_info_alter().
+ */
+function farm_location_entity_base_field_info_alter(&$fields, EntityTypeInterface $entity_type) {
+  /** @var \Drupal\field\Entity\FieldConfig[] $fields */
+
+  // Prevent creating circular asset location.
+  if ($entity_type->id() == 'log' && !empty($fields['asset'])) {
```
symbioquine, line 54: Should this also do the `is_movement` check?

symbioquine (commented):
I think I brought up before, the question of whether this kind of constraint checking is vulnerable to race conditions.

In other words, does the constraint check happen within a database transaction with locking such that two write calls can't both pass the constraint check, then write changes which together create a circular relationship, but where neither individually would have?

Were you able to check that? I think before we talked about setting a breakpoint at the end of the constraint check, then creating two such requests letting them get past the constraint validation, then allowing them to both proceed...

modules/core/location/farm_location.module
mstenta: I suppose we could include `&& !empty($fields['is_movement'])` here too, but all log types have that field, so it wouldn't add much (unless that changes in the future I suppose). Although to be fair, all log types also have an `asset` field, so that check is equally unnecessary here I suppose.

We do check the SPECIFIC value of `is_movement` in `CircularAssetLocationConstraintValidator` (below): https://github.com/farmOS/farmOS/pull/568/files#diff-a423c1299c8e264f45b3abda4e862ea21b300de6946241c5618168fdb587d0b9R54

mstenta:
> In other words, does the constraint check happen within a database transaction with locking such that two write calls can't both pass the constraint check, then write changes which together create a circular relationship, but where neither individually would have?

This is a great thing to bring up @symbioquine, and I believe the answer is NO: this does NOT protect against the kind of race condition you describe.

There are potentially other situations that could result in circulate asset location too, I imagine. These are some notes I left to myself to follow-up on:

```
    // @todo
    // Does reverting to a previous entity revision perform validation?
    // If not, it would be possible to get into a circular location situation.
    // I could see this not being covered by core because most validation does
    // not involve checking other entities like this does.
    // Come to think of it... is it possible to create the same condition by
    // deleting a log? Imagine creating logs for two assets that move them
    // through being located within each other, and the logs all validate
    // initially, but then you delete one that causes a later one to become
    // invalid...
    // what if you shuffle around the timestamps of a bunch of movement logs?
```

So this PR is not solving ALL potential situations, I don't think. But in theory it should prevent the most likely one (and the one that has happened in the wild already: #541).

I propose we merge this as a first step and create a follow-up issue to document, test, and think through other hypothetical situations. How does that sound?

paul121 (approved):
> I propose we merge this as a first step and create a follow-up issue to document, test, and think through other hypothetical situations. How does that sound?

:+1:

symbioquine:
> I propose we merge this as a first step and create a follow-up issue to document, test, and think through other hypothetical situations. How does that sound?

Sounds good to me.

symbioquine (approved):

mstenta:
Thanks @symbioquine and @paul121! I rebased and merged.

PR #569, opened 2022-09-20
Add log asset filter to all displays of farm_quantity view

paul121:
This has been requested by Rothamsted: https://github.com/Rothamsted-Ecoinformatics/farm_rothamsted/issues/281 I just dropped in a quick log.asset entity reference filter and it seems to be working well! I tested all displays, including CSV export.

One thing I tested was if this created duplicate quantity results in the view: indeed if a log references multiple assets, and multiple assets are included in the views filter, then the same quantity will be seen twice in the views output (easily identified by the first `id` column). **We should decide if this is a bug or a feature??** Generally, I think we should not show duplicates. This is the current behavior for the "Log assets" column, multiple log assets are included in the same row, and the row is not duplicated.

I then turned on the "Reduce duplicates" option for the log.asset entity reference filter. This worked as advertised and removed the duplicates! But this option does have a warning:

> This filter can cause items that have more than one of the selected options to appear as duplicate results. If this filter causes duplicate results to occur, this checkbox can reduce those duplicates; however, the more terms it has to search for, the less performant the query will be, so use this with caution. Shouldn't be set on single-value fields, as it may cause values to disappear from display, if used on an incompatible field.

"The more terms it has to search for" - in our case I'm assuming this means the "assets" but I'm not certain it doesn't mean the logs or quantities themselves. I wouldn't expect a user to manually filter by so many assets that performance becomes a major issue. But it is possible that there are many logs & quantities on an instance, which might make the query less performant anyways.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [Issue #3309234: Add PHPStan to test and delivery workflow](https://www.drupal.org/project/farm/issues/3309234)
 - [Issue #3309198: Allow users to override Gin theme settings](https://www.drupal.org/project/farm/issues/3309198)
 - [Add owner field to assets #537](https://github.com/farmOS/farmOS/pull/537)
+- [Add log asset filter to all displays of farm_quantity view #569](https://github.com/farmOS/farmOS/pull/569)
 
 ### Changed
 
--- a/modules/core/ui/views/config/install/views.view.farm_quantity.yml
+++ b/modules/core/ui/views/config/install/views.view.farm_quantity.yml
@@ -1334,6 +1334,56 @@ display:
             default_group: All
             default_group_multiple: {  }
             group_items: {  }
+        asset_target_id:
+          id: asset_target_id
+          table: log__asset
+          field: asset_target_id
+          relationship: reverse__log__quantity
+          group_type: group
+          admin_label: ''
+          entity_type: log
+          entity_field: asset
+          plugin_id: entity_reference
+          operator: or
+          value: {  }
+          group: 1
+          exposed: true
+          expose:
+            operator_id: asset_target_id_op
+            label: 'Log assets'
+            description: ''
+            use_operator: false
+            operator: asset_target_id_op
+            operator_limit_selection: false
+            operator_list: {  }
+            identifier: asset_target_id
+            required: false
+            remember: false
+            multiple: true
+            remember_roles:
+              authenticated: authenticated
+              anonymous: '0'
+              farm_manager: '0'
+              farm_operator: '0'
+              farm_viewer: '0'
+              farm_worker: '0'
+            reduce: false
+          is_grouped: false
+          group_info:
+            label: ''
+            description: ''
+            identifier: ''
+            optional: true
+            widget: select
+            multiple: false
+            remember: false
+            default_group: All
+            default_group_multiple: {  }
+            group_items: {  }
+          reduce_duplicates: true
+          handler: 'default:asset'
+          widget: autocomplete
+          handler_settings: {  }
       filter_groups:
         operator: AND
         groups:


mstenta:
Thanks @paul121! Unfortunately if this causes duplicate rows then I don't think we can merge it as-is.

I spent more time than I'd like to admit "fixing" duplicate rows in farmOS v1 because of the confusion they caused to users. And in the case of the quantity View it's even more critical because this is the primary way to export quantity data to CSV via the UI. If someone exports quantities and isn't aware that there might be duplicates (why should they be?) it can cause miscalculations if they use that data for analysis.

IMO if we are going to add a filter like this we'll have to figure out how to do it in a way that does not cause duplicates in the first place. I don't feel comfortable leaning on the "reduce duplicates" option.

Duplicates generally imply a problem with the query, and usually can be solved with a better designed query. In Drupal Views this often means writing a custom handler, though, but I think that's where we would need to go with this. I'd probably start by getting the query that Views is generating and playing with that in raw SQL until we have something that works. Then build a Views handler to match.

paul121:
> I don't feel comfortable leaning on the "reduce duplicates" option.

Any particular reason for this?

I had thought that "reduce duplicates" meant some post-processing after the query is run but it looks like selecting this option actually changes the original query. Here are the queries I'm seeing for both `reduce_duplicates` on and off, each filtered to a single asset and multiple assets (duplicates only happen when filtering to multiple assets but interesting to see both)

The reason this seems to work is because `reduce_duplicates` removes the single `INNER JOIN log__asset` with a separate `LEFT JOIN log__asset` for each asset that is being filtered. There is then a `WHERE` statement that limits the results from the relevant join to only some asset IDs. This makes me feel better about using it. I also tested in a sql console and confirmed the query does not cause any duplicate rows. In a custom views handler we might end up doing the same thing?

Now I also understand the warning message "the more terms it has to search for..." means each additional asset is another left join, but I think it is uncommon someone would be filtering by *many* assets that this would make much of a difference?

```
# Reduce duplicates OFF:
## Filter to single asset:

SELECT "quantity"."value__numerator" AS "quantity_value__numerator", "quantity"."value__denominator" AS "quantity_value__denominator", "quantity_quantity"."id" AS "quantity_quantity_id", "quantity"."id" AS "id"
FROM
{quantity} "quantity"
LEFT JOIN {log__quantity} "log__quantity" ON quantity.id = log__quantity.quantity_target_id
LEFT JOIN {log_field_data} "quantity_quantity" ON log__quantity.entity_id = quantity_quantity.id
INNER JOIN {log__asset} "quantity_quantity__log__asset" ON quantity_quantity.id = quantity_quantity__log__asset.entity_id AND quantity_quantity__log__asset.deleted = '0'
WHERE (quantity_quantity__log__asset.asset_target_id = '28')
ORDER BY "quantity_quantity"."timestamp" DESC NULLS LAST, "quantity_quantity_id" DESC NULLS LAST
LIMIT 50 OFFSET 0

## Filter to multiple assets:

SELECT "quantity"."value__numerator" AS "quantity_value__numerator", "quantity"."value__denominator" AS "quantity_value__denominator", "quantity"."id" AS "id", "quantity_quantity"."id" AS "quantity_quantity_id"
FROM
{quantity} "quantity"
LEFT JOIN {log__quantity} "log__quantity" ON quantity.id = log__quantity.quantity_target_id
LEFT JOIN {log_field_data} "quantity_quantity" ON log__quantity.entity_id = quantity_quantity.id
INNER JOIN {log__asset} "quantity_quantity__log__asset" ON quantity_quantity.id = quantity_quantity__log__asset.entity_id AND quantity_quantity__log__asset.deleted = '0'
WHERE (quantity_quantity__log__asset.asset_target_id IN('28', '29'))
ORDER BY "quantity"."id" DESC NULLS LAST
LIMIT 50 OFFSET 0

# Reduce duplicates ON:

## Filter to single asset:
SELECT "quantity"."value__numerator" AS "quantity_value__numerator", "quantity"."value__denominator" AS "quantity_value__denominator", "quantity_quantity"."id" AS "quantity_quantity_id", "quantity"."id" AS "id"

FROM
{quantity} "quantity"
LEFT JOIN {log__quantity} "log__quantity" ON quantity.id = log__quantity.quantity_target_id
LEFT JOIN {log_field_data} "quantity_quantity" ON log__quantity.entity_id = quantity_quantity.id
LEFT JOIN {log__asset} "log__asset_value_0" ON quantity_quantity.id = log__asset_value_0.entity_id AND log__asset_value_0.asset_target_id = '28'
WHERE "log__asset_value_0"."asset_target_id" = '28'
ORDER BY "quantity_quantity"."timestamp" DESC NULLS LAST, "quantity_quantity_id" DESC NULLS LAST
LIMIT 50 OFFSET 0

## Filter to multiple assets:
SELECT "quantity"."value__numerator" AS "quantity_value__numerator", "quantity"."value__denominator" AS "quantity_value__denominator", "quantity_quantity"."id" AS "quantity_quantity_id", "quantity"."id" AS "id"
FROM
{quantity} "quantity"
LEFT JOIN {log__quantity} "log__quantity" ON quantity.id = log__quantity.quantity_target_id
LEFT JOIN {log_field_data} "quantity_quantity" ON log__quantity.entity_id = quantity_quantity.id
LEFT JOIN {log__asset} "log__asset_value_0" ON quantity_quantity.id = log__asset_value_0.entity_id AND log__asset_value_0.asset_target_id = '28'
LEFT JOIN {log__asset} "log__asset_value_1" ON quantity_quantity.id = log__asset_value_1.entity_id AND log__asset_value_1.asset_target_id = '29'
WHERE ("log__asset_value_0"."asset_target_id" = '28') OR ("log__asset_value_1"."asset_target_id" = '29')
ORDER BY "quantity_quantity"."timestamp" DESC NULLS LAST, "quantity_quantity_id" DESC NULLS LAST
LIMIT 50 OFFSET 0
```

mstenta:
> I had thought that "reduce duplicates" meant some post-processing after the query is run but it looks like selecting this option actually changes the original query.

Oh well that's good to know. I had also assumed that it meant post-processing in PHP or use of `DISTINCT`. Thanks for posting the actual queries.

> `reduce_duplicates` removes the single `INNER JOIN log__asset` with a separate `LEFT JOIN log__asset` for each asset that is being filtered. There is then a `WHERE` statement that limits the results from the relevant join to only some asset IDs.

Ah interesting. This approach seems like a good one. I understand the performance concerns and agree with you that in practice that shouldn't be an issue.

Also, seeing that "reduce duplicates" is a Drupal core feature (of the `ManyToOneHelper` class) gives me some more confidence that it has been battle tested.

How about we add this to our agenda for this week's dev call? Might be good to get some other eyes and thoughts on this.

Just for reference, and to relive past trauma (j/k :sweat_smile:), here are some of the bug reports for duplicate rows that have come up in the past (notice the first one is THIS one):

- https://www.drupal.org/project/farm/issues/3104106
- https://www.drupal.org/project/farm/issues/3049153
- https://www.drupal.org/project/farm/issues/2958576
- https://www.drupal.org/project/farm/issues/2954290
- https://www.drupal.org/project/farm/issues/2955393
- https://www.drupal.org/project/farm/issues/3240953

So you can see why I might be a little over-cautious here. :sweat_smile:

mstenta:
Whelp. Just tested our existing Views and we already have potential duplicate issues with some of the filters. Basically any filter that is a reference field on the entities which allow multiple values, including: flags, log category, assigned to, assets, locations, etc...

I suppose this is another reason not to block this particular addition, AND a reason to open a dedicated issue to consider applying "reduce duplicates" to the other filters as well. I'm testing them locally, and that does fix the duplicates. Makes me wonder what the edge cases of this might be... and where the limits are, if any, in practice...

mstenta:
> open a dedicated issue to consider applying "reduce duplicates" to the other filters as well

PR: https://github.com/farmOS/farmOS/pull/571

paul121:
> Makes me wonder what the edge cases of this might be... and where the limits are, if any, in practice...

Yes I'm curious about this as well!

PR #573, opened 2022-09-23
Minor improvements to views task links

paul121:
This PR is mostly to consolidate and simplify how our views task links are created. There is only one functional change (the last commit) which fixes a minor bug where the primary tab is not marked `active` when viewing secondary tabs other than "All". For example, on an Asset page, looking at `activity` logs referencing the asset, the "Activity" secondary tab is active, but the "Logs" primary tab is not.

While looking at the `drupal/entity` task links that are created for our entities, I realized that task links created through a derivative class end up with a prefix in their ID. This is a gotcha when trying to reference another task link as a `parent_id`. To further confuse things, our views tasks links are split between the static definition and derivative class. Consolidating all of these tasks links to the derivative class makes it easier to understand the relationship between them, as well as providing a consistent prefix to the task link IDs.

As an example, the views of logs that are included on asset pages. The taxonomy term asset/log views follow a similar pattern.

Before:
- "Logs" primary tab: `farm.asset.logs`
- "All" default secondary tab: `farm.asset.logs.all`
- "Activity" secondary tab: `farm.asset.logs.type:farm.asset.logs.activity`

After:
- "Logs" primary tab: `farm.asset.logs.type:logs`
- "All" default secondary tab: `farm.asset.logs.type:all`
- "Activity" secondary tab: `farm.asset.logs.type:activity`

### DIFF
--- a/modules/core/ui/views/farm_ui_views.links.task.yml
+++ b/modules/core/ui/views/farm_ui_views.links.task.yml
@@ -15,20 +15,6 @@ farm.asset.logs.all:
 farm.asset.logs.type:
   deriver: Drupal\farm_ui_views\Plugin\Derivative\FarmLogViewsTaskLink
 
-# Add asset/log view tabs to taxonomy terms.
-farm.taxonomy_term.assets:
-  title: 'Assets'
-  route_name: view.farm_asset.page_term
-  route_parameters:
-    entity_bundle: 'all'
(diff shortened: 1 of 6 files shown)


mstenta:
Merged thanks @paul121! https://github.com/farmOS/farmOS/commit/9a207bdc541458073b0006a4c0e32f9563528df6

PR #575, opened 2022-09-26
Fix [warning] Invalid argument supplied for foreach() EntityViewsDataTaxonomyFilterTrait.php:26

### LINKED ISSUE #574: [warning] Invalid argument supplied for foreach() EntityViewsDataTaxonomyFilterTrait.php:26
We are experiencing this warning multiple times whenever we try to rebuild caches via Drush in a custom project based on farmOS:

```
$ drush cr
 [warning] Invalid argument supplied for foreach() EntityViewsDataTaxonomyFilterTrait.php:26
 [warning] Invalid argument supplied for foreach() EntityViewsDataTaxonomyFilterTrait.php:26
 [warning] Invalid argument supplied for foreach() EntityViewsDataTaxonomyFilterTrait.php:26
 [warning] Invalid argument supplied for foreach() EntityViewsDataTaxonomyFilterTrait.php:26
 [warning] Invalid argument supplied for foreach() EntityViewsDataTaxonomyFilterTrait.php:26
 [warning] Invalid argument supplied for foreach() EntityViewsDataTaxonomyFilterTrait.php:26
 [warning] Invalid argument supplied for foreach() EntityViewsDataTaxonomyFilterTrait.php:26
 [warning] Invalid argument supplied for foreach() EntityViewsDataTaxonomyFilterTrait.php:26
 [warning] Invalid argument supplied for foreach() EntityViewsDataTaxonomyFilterTrait.php:26
 [warning] Invalid argument supplied for foreach() EntityViewsDataTaxonomyFilterTrait.php:26
 [warning] Invalid argument supplied for foreach() EntityViewsDataTaxonomyFilterTrait.php:26
 [warning] Invalid argument supplied for foreach() EntityViewsDataTaxonomyFilterTrait.php:26
 [warning] Invalid argument supplied for foreach() EntityViewsDataTaxonomyFilterTrait.php:26
 [warning] Invalid argument supplied for foreach() EntityViewsDataTaxonomyFilterTrait.php:26
 [warning] Invalid argument supplied for foreach() EntityViewsDataTaxonomyFilterTrait.php:26
 [warning] Invalid argument supplied for foreach() EntityViewsDataTaxonomyFilterTrait.php:26
 [warning] Invalid argument supplied for foreach() EntityViewsDataTaxonomyFilterTrait.php:26
 [warning] Invalid argument supplied for foreach() EntityViewsDataTaxonomyFilterTrait.php:26
 [success] Cache rebuild complete.
```

I haven't had a chance to dig in and figure out what's happening. It doesn't seem to be causing any issues.

Here is the relevant line of code producing the warning: https://github.com/farmOS/farmOS/blob/f60478f6df89b966fc89637400b8e3d92ca4d4c2/modules/core/entity/modules/views/src/EntityViewsDataTaxonomyFilterTrait.php#L26

mstenta:
Fixes https://github.com/farmOS/farmOS/issues/574

### DIFF
--- a/modules/core/entity/modules/views/src/EntityViewsDataTaxonomyFilterTrait.php
+++ b/modules/core/entity/modules/views/src/EntityViewsDataTaxonomyFilterTrait.php
@@ -19,14 +19,37 @@ trait EntityViewsDataTaxonomyFilterTrait {
     // Configure the taxonomy_term reference field filter.
     // Logic derived form taxonomy_field_views_data_alter().
     foreach ($fields as $field) {
+
+      // If this is not a taxonomy term reference field, skip it.
       if ($field->getSettings()['target_type'] !== 'taxonomy_term') {
         continue;
       }
-      $table_name = $this->tableMapping->getFieldTableName($field->getName());
-      foreach ($data[$table_name] as $table_field_name => $field_data) {
-        $field_name = $field->getName();
-        if (isset($field_data['filter']) && strpos($table_field_name, $field_name) === 0) {
-          $data[$table_name][$table_field_name]['filter']['id'] = 'taxonomy_index_tid';
+
+      // Get the field name.
+      $field_name = $field->getName();
+
+      // Iterate through the Views data tables and columns.
+      foreach ($data as $table_name => $table_data) {
+        foreach ($table_data as $table_field_name => $field_data) {
+
+          // If this field doesn't have a filter handler, skip it.
+          if (!isset($field_data['filter'])) {
+            continue;
+          }
+
+          // Ensure that we are only altering the Views field we want.
+          // This will either be the field name itself, or the field name plus
+          // a `_target_id` suffix (depending on whether the field is a base or
+          // bundle field, single or multiple values, etc).
+          $table_field_names = [
+            $field_name,
+            $field_name . '_target_id',
+          ];
+          if (in_array($table_field_name, $table_field_names)) {
+
+            // Set the filter handler ID.
+            $data[$table_name][$table_field_name]['filter']['id'] = 'taxonomy_index_tid';
+          }
         }
       }
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [Incorrect translation placeholders for asset names #540](https://github.com/farmOS/farmOS/issues/540)
 - Update farmOS-map to [v2.0.5](https://github.com/farmOS/farmOS-map/releases/tag/v2.0.5) to fix [Uncaught (in promise) TypeError: o.getChangeEventType is not a function #551](https://github.com/farmOS/farmOS/issues/551)
 - [Prevent circular group membership #562](https://github.com/farmOS/farmOS/pull/562)
+- [Fix [warning] Invalid argument supplied for foreach() EntityViewsDataTaxonomyFilterTrait.php:26 #575](https://github.com/farmOS/farmOS/pull/575)
 
 ## [2.0.0-beta6] 2022-07-30
 


modules/core/entity/modules/views/src/EntityViewsDataTaxonomyFilterTrait.php
```diff
@@ -22,11 +22,12 @@ protected function addReverseRelationships(array &$data, array $fields) {
...
       }
-      $table_name = $this->tableMapping->getFieldTableName($field->getName());
-      foreach ($data[$table_name] as $table_field_name => $field_data) {
-        $field_name = $field->getName();
-        if (isset($field_data['filter']) && strpos($table_field_name, $field_name) === 0) {
-          $data[$table_name][$table_field_name]['filter']['id'] = 'taxonomy_index_tid';
+      $field_name = $field->getName();
+      foreach ($data as $table_name => $table_data) {
+        foreach ($table_data as $table_field_name => $field_data) {
+          if (isset($field_data['filter']) && strpos($table_field_name, $field_name) === 0) {
```
paul121: > `strpos($table_field_name, $field_name) === 0`

I'm curious why we can't check that `$table_field_name == $field_name` ? I think it could be possible that two different fields have the same prefix and this logic could inadvertently change the filter ID for the wrong field.

Consider two field names: Taxonomy reference `custom_field` and String `custom_field_raw`. I think this logic would change `custom_field_raw` to use the `taxonomy_index_tid` filter.

paul121: but I think this approach should work! :+1:

mstenta: > Consider two field names: Taxonomy reference `custom_field` and String `custom_field_raw`. I think this logic would change `custom_field_raw` to use the `taxonomy_index_tid` filter.

That's a good point.

I'm not sure why we used `strpos($table_field_name, $field_name) === 0` originally. Here is the commit that added it, if it rings any bells: https://github.com/farmOS/farmOS/commit/7358e4d64c97f156994ddc1f821109d78d740eaf

This PR doesn't change that bit, just restructures the `foreach` so that we don't need to figure out `$table_name`.

I wonder if there was a good reason for using `strpos`... :thinking:

mstenta: > This PR doesn't change that bit, just restructures the `foreach` so that we don't need to figure out `$table_name`.

Ah but I see how that would introduce this issue, as we are no longer limiting which tables this logic can potentially run on...

mstenta: Ah ok... just ran this through the debugger and quickly realized why we used `strpos()`...

Example:

- `$table_field_name` is `animal_type_target_id`
- `$field_name` is `animal_type`

mstenta: But not for all of them...

Example:

- `$table_field_name` is `units`
- `$field_name` is `units`

mstenta: Perhaps we could do:

`if (isset($field_data['filter']) && in_array($table_field_name, [$field_name, $field_name . '_target_id'])) {`

... basically checking if it's either `$field_name` or `$field_name . '_target_id'`.

mstenta: >  I think it could be possible that two different fields have the same prefix and this logic could inadvertently change the filter ID for the wrong field.

Ah ha! Sure enough I found a case where this happens in the custom codebase I'm working in, which has both `field_input` and `field_input_date` fields.

mstenta:
OK. Force-pushed the commit to implement this change. I also restructured the code to make the logic more readable with extra comments etc.

paul121 (approved):
Thanks for adding the comments :-)

PR #576, opened 2022-09-27
Attempt to update the map size until it is rendered

paul121:
This fixes a bug where the map is not rendered in modal and off-canvas dialogs until the map size is updated. This is a challenging edge case because it is hard to determine when the dialog is done loading and the map size can be updated. Here is what I ran into:
- There are jQuery events fired for when the dialog is created/closed ([change record](https://www.drupal.org/node/1852224)), but unfortunately we cannot subscribe to these with simple native JS event subscribers, we have to use jQuery. I don't think we should add a dependency on jQuery for this and even when I tried this in my console it didn't seem sufficient to update the map:
```js
jQuery(window).on('dialog:aftercreate', () => window.farmOS.map.instances[0].map.updateSize());
```
- When the map is initially created it is not a child of the `<div role="dialog">` parent. This makes it hard to identify which maps may be included in a dialog, unlike our existing code that can check for maps included in a form or details element quite easily. I figured this out by logging the following a few times after the map is created:

```js
window.farmOS.map.instances[0].map.getTargetElement().closest('div[role="dialog"]')
```

-  It seems that very soon after the map becomes a child of the dialog element it can finally have its size updated to render in the dialog. I'm not sure that there is an easy way to determine *when* this parent change happens. As a result, I've implemented a simple for loop that iterates every 0.5 seconds for 3 seconds attempting to update the map size. Once rendered, the loop is exited. I feel that this is a pretty safe addition and shouldn't have any noticeable impact on client side performance, even when a map is intentionally rendered in a hidden state.
- I also tested this with the internet connection throttled to slow 3G and saw that this still worked even though it took much  longer than 3 seconds after `farm_map.js` and `farmOS-map.js` were loaded before the map was rendered. My thought was the map might be created way before being added to the dialog but I don't think this is the case. Soon after the dialog was displayed the map appeared & rendered as expected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [Fix map ui being unreadable on dark mode #642](https://github.com/farmOS/farmOS/pull/642)
 - Patch Drupal core to fix [Issue #3266341: Views pagers do math on disparate data types, resulting in type errors in PHP 8](https://www.drupal.org/project/drupal/issues/3266341)
 - [Implement FarmBreadcrumbBuilder::applies() to only affect desired routes](https://github.com/farmOS/farmOS/pull/644)
+- [Attempt to update the map size until it is rendered #576](https://github.com/farmOS/farmOS/issues/576)
 
 ## [2.0.1] 2023-02-08
 
--- a/modules/core/map/js/farm_map.js
+++ b/modules/core/map/js/farm_map.js
@@ -75,6 +75,24 @@
         });
       }
 
+      // Handle other cases where the map may not have a correct size when
+      // it is initially rendered. This happens for modal and off-canvas
+      // dialogs where there is no easy way to detect when the dialog is
+      // finished loading and the map can be updated.
+      // Iterate every 0.5 seconds for 3 seconds until the map
+      // can be rendered and gets a size. Once rendered end the loop.
+      const delay = ms => new Promise(res => setTimeout(res, ms));
+      async function updateSize(instance) {
+        for (let i = 0; i < 6; i++) {
+          if (!instance.map.getSize().includes(0)) {
+            return;
+          }
+          instance.map.updateSize();
+          await delay(500);
+        }
+      }
+      updateSize(instance);
+
       return instance;
     }
 


mstenta:
I need to look closer at this, but in the meantime it reminds me of some issues we had in the past (in farmOS 1.x). Here's one:

- https://www.drupal.org/project/openlayers/issues/2644580

And this one from earlier in 2.x development:

- https://www.drupal.org/project/farm/issues/3189747

Generally I prefer to avoid using JS loops+timeouts if we can avoid it (can't always). I'd be curious if there's an event we can trigger on instead...

mstenta:
> I'd be curious if there's an event we can trigger on instead...

Of course you considered this already @paul121 😄 ...

I'll review in more detail soon...

modules/core/map/js/farm_map.js
```diff
@@ -75,6 +75,24 @@
...
+      // it is initially rendered. This happens for modal and off-canvas
+      // dialogs where there is no easy way to detect when the dialog is
+      // finished loading and the map can be updated.
+      // Iterate every 0.5 seconds for 3 seconds until the map
+      // can be rendered and gets a size. Once rendered end the loop.
+      const delay = ms => new Promise(res => setTimeout(res, ms));
+      async function updateSize(instance) {
+        for (let i = 0; i < 6; i++) {
+          instance.map.updateSize();
+          if (!instance.map.getSize().includes(0)) {
```
paul121, line 87: I should clarify how this works. `map.getSize()` returns an array with the width and height of of the map viewport ([docs](https://openlayers.org/en/latest/apidoc/module-ol_Map-Map.html#updateSize), [source](https://github.com/openlayers/openlayers/blob/v7.1.0/src/ol/Map.js#L1668))

by using `map.getSize().includes(0)` this checks to see if either the width or height of the viewport is 0. Most of the time this will return `[0,0]` (when not visible) but I saw one instance where it only had a width, something like `[1225,0]`

mstenta:
Some relevant discussion in chat about this today: https://irc.farmos.org/bot/log/farmOS/2022-09-27#T81745

mstenta:
Dusting off old open PRs...

I still don't think we should use `setTimeout()`. It is a workaround, which will probably work most of the time, but there is a chance that it won't (it depends on network timing that we can't control), which means it is an inherently imperfect solution. Maybe I'm wrong about that, based on your investigations @paul121, and it will always work - but I can't shake the "code smell" of it.

> There are jQuery events fired for when the dialog is created/closed ([change record](https://www.drupal.org/node/1852224)), but unfortunately we cannot subscribe to these with simple native JS event subscribers, we have to use jQuery. I don't think we should add a dependency on jQuery for this...

jQuery is already being included on the page when off canvas dialogs are used, so it seems reasonable that downstream code that wants to fix this issue *in those circumstances* can also depend on jQuery.

That said, I agree we shouldn't add jQuery as a dependency *broadly*. Perhaps this shouldn't be solved in `farm_map.js` at all, since off-canvas dialogs are not a common use-case (and aren't used at all in farmOS core currently).

Maybe we can figure out a more targeted fix for them? I still see value in providing a solution to this in farmOS core, and not forcing module developers to add extra JS to fix it in each downstream implementation.

Is it possible to add extra JS to the page ONLY WHEN modals or off-canvas dialogs are used? Can we add a simpler JS library that gets added to the page automatically and targets this problem specifically, rather than adding it to `farm_map.js` broadly?

> even when I tried this in my console it didn't seem sufficient to update the map
> ...
> It seems that very soon after the map becomes a child of the dialog element it can finally have its size updated to render in the dialog. I'm not sure that there is an easy way to determine when this parent change happens.

Sounds like you've looked pretty closely at the event-driven approach, and it's unfortunate that it's not working easily. I'd still be curious to see if we can make it work, but if not... and if we have a way to only add JS in the contexts where this fix is necessary... then I might be willing to hold my nose and merge a `setTimeout()`... :sweat_smile:

paul121:
> It is a workaround, which will probably work most of the time, but there is a chance that it won't (it depends on network timing that we can't control), which means it is an inherently imperfect solution.

Actually I don't think this depends on network timing. It's been awhile, but I did mention above that I tested on a slow connection where it took longer than 3 seconds for the page to load and this still worked just fine. I believe this is because this code is not executed until the (likely AJAX?) request returns and content is added to the page, at which point this `farm_map.js` is run & attached to a new context in the document. SO the 3 second delay would only fail if a device/browser does not render the map within the modal/off canvas dialog within 3 seconds. We could also expand this time, I do not think it would have a negative impact.

> Is it possible to add extra JS to the page ONLY WHEN modals or off-canvas dialogs are used? Can we add a simpler JS library that gets added to the page automatically and targets this problem specifically, rather than adding it to farm_map.js broadly?

This might be challenging because you can load any page in modal/dialog. We can't assume a page will always be intended for that purpose and include the JS there. However, we could likely attach some JS to every modal/dialog... but then this JS would come with every modal/dialog, not only ones with a map.

FWIW I think this would have little to no impact including it in `farm_map.js` on existing maps. My current PR would call `instance.map.updateSize();` one additional time which shouldn't have any effect if the map is already rendered. A simple change could be to move `instance.map.updateSize()` down below the if statement that checks for the current map size: `if (!instance.map.getSize().includes(0)) {` in which case the dreaded timeout would ONLY run if the map doesn't have a size. I believe if we ever reach this point in the code and the map still doesn't have a size then the map will not be rendered, so at the end of the day it's a safe (albeit moderately ugly) check to include.

mstenta:
Alright I concede. :smile:

It's really not a big deal, to all of your points above @paul121. Thanks for indulging my critiques anyway. It is a good issue to fix, and if this is the all-around best (including easiest) way to do it, so be it!

I'll approve this PR, unless anyone else feels strongly and wants to object more to the timeout approach.

> A simple change could be to move `instance.map.updateSize()` down below the if statement that checks for the current map size

Want to make this change? Sounds like a simple improvement to me.

mstenta:
Out of curiosity, I seem to recall seeing a demo recently by @pcambra that included a map in an off-canvas dialogue. @pcambra did you have to work around this issue also?

pcambra:
Oh it's been a while since I implemented that, what we've done is a bunch of tweaks on how the CSS for off canvas renders, see https://github.com/National-Forestry-Authority/forests/blob/1139f05be3e68eb9c575f4598d6f1a6f18618d23/web/modules/custom/farm_nfa/farm_nfa.module and we have a custom farm_gin theme https://github.com/National-Forestry-Authority/forests/blob/1139f05be3e68eb9c575f4598d6f1a6f18618d23/web/themes/custom/farm_gin/farm_gin.theme

CHANGELOG.md
```diff
@@ -175,6 +176,7 @@ If you would like to skip the automatic fix, add the following line to your
 - [Set reduce_duplicates: true in Views exposed filters for multivalue fields #571](https://github.com/farmOS/farmOS/pull/571)
 - [Update core map behaviors to properly depend on core/drupalSettings #578](https://github.com/farmOS/farmOS/pull/578)
 - [Issue #3243922 by paul121, Symbioquine, m.stenta: farmOS-map chunks fail to load when map is rendered via AJAX](https://www.drupal.org/project/farm/issues/3243922)
+- [Prevent circular group membership #562](https://github.com/farmOS/farmOS/pull/562)
```
mstenta: Just noticed this snuck in... :-)

pcambra:
We also have https://www.drupal.org/project/gin_lb that fixes a lot of what was going on at the offcanvas level

mstenta (approved):

paul121:
@pcambra anything that changes modals? This same issue was happening for modals. Really curious why this may not be happening for you and what is different... but since you're removing that much CSS for the off-canvas it might be hard to chase down the cause in the case specifically... :-)

paul121:
@pcambra or maybe it has to do with how you initiate the modal/off-canvas being opened? I've been adding the `data-dialog-*` attributes like I see you do here: https://github.com/National-Forestry-Authority/forests/blob/1139f05be3e68eb9c575f4598d6f1a6f18618d23/web/modules/custom/farm_nfa/farm_nfa.module#L211-L217

But wonder if perhaps there are other ways to "open" a modal/off-canvas and insert the map in such a way that it does get rendered?

symbioquine (approved):
It's a bit of a hack, but, being so simple and self-contained, it's probably the most robust path forward.

PR #562, opened 2022-08-29
Prevent circular group membership

mstenta:
Fixes https://github.com/farmOS/farmOS/issues/541

PR #577, opened 2022-09-27
Prevent circular group membership

mstenta:
We decided to revert the PR that originally added this (#562), which addressed the bug reported in #541.

We discovered a critical flaw in the logic. The circular group membership check was not taking the timestamp of the log into account. This means that it might prevent saving an old log if it didn't pass validation against the CURRENT state of data.

Dependent on https://www.drupal.org/project/farm/issues/3310286

Remaining todo:

- Use $timestamp from https://www.drupal.org/project/farm/issues/3310286
- Changelog

### DIFF
--- a/modules/asset/group/src/Plugin/Validation/Constraint/CircularGroupMembershipConstraintValidator.php
+++ b/modules/asset/group/src/Plugin/Validation/Constraint/CircularGroupMembershipConstraintValidator.php
@@ -1,0 +1,107 @@
+<?php
+
+namespace Drupal\farm_group\Plugin\Validation\Constraint;
+
+use Drupal\Core\DependencyInjection\ContainerInjectionInterface;
+use Drupal\farm_group\GroupMembershipInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\Validator\Constraint;
+use Symfony\Component\Validator\ConstraintValidator;
+
+/**
+ * Validates the CircularGroupMembership constraint.
+ */
+class CircularGroupMembershipConstraintValidator extends ConstraintValidator implements ContainerInjectionInterface {
+
+  /**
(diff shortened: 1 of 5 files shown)


symbioquine (approved):

modules/asset/group/src/Plugin/Validation/Constraint/CircularGroupMembershipConstraintValidator.php
```diff
@@ -0,0 +1,93 @@
...
+        continue;
+      }
+
+      // Load members of this group (recursively).
+      $members = $this->groupMembership->getGroupMembers([$asset], TRUE, $timestamp);
+
+      // Make sure that none of the group(s) are members of this asset.
+      foreach ($groups as $group) {
+        foreach ($members as $member) {
+          if ($group->id() == $member->id()) {
```
paul121: I think we need to also check that a group is not referenced twice on the same log. This is the original simple case (not recursive) reported in #541

eg:

```php
      foreach ($groups as $group) {
      
        // Check for group double reference.
        if ($group->id == $asset->id()) {
          // violation
        }
        
        // Check the group's members.
        foreach ($members as $member) {
          if ($group->id() == $member->id()) {

mstenta: Good point! I will add a check for this to the constraint logic, and a test for it in the automated tests.

mstenta: Added a `fixup` commit that I'll merge into the original commit after review: https://github.com/farmOS/farmOS/pull/577/commits/a336b9d9e424d8eb99d2be76627dd67a1a0c8d00

paul121 (approved):

PR #580, opened 2022-09-29
Use GitHub Actions to build Docker Hub images

mstenta:
Draft commit to begin publishing our images to Docker Hub via GitHub Actions, per https://www.drupal.org/project/farm/issues/3203129

This is currently untested. Opened as a draft PR for initial sanity check and review of the logic with @symbioquine and @paul121.

### DIFF
--- a/.github/workflows/deliver.yml
+++ b/.github/workflows/deliver.yml
@@ -159,3 +159,54 @@ jobs:
           files: /tmp/farmOS-${{ env.FARMOS_VERSION }}.tar.gz
           draft: false
           prerelease: false
+  publish:
+    name: Publish to Docker Hub
+    # We only publish to Docker Hub if this is a tag or 2.x branch push event
+    # to the official repository.
+    if: github.repository == 'farmOS/farmOS' && github.event_name == 'push' && (github.ref_type == 'tag' || (github.ref_type == 'branch' && needs.build.outputs.farmos_version == '2.x'))
+    runs-on: ubuntu-latest
+    needs:
+      - build
+      - sniff
+      - test
+    steps:
+      - name: Set FARMOS_VERSION from previous output.
+        run: echo "FARMOS_VERSION=${{ needs.build.outputs.farmos_version }}" >> $GITHUB_ENV
+      - name: Restore farmOS 2.x Docker image from cache
+        uses: actions/cache@v3
+        with:
+          path: /tmp/farmos-2x.tar
+          key: farmos-2x-${{ github.run_id }}
+      - name: Load farmos/farmos:2.x image
+        run: docker load < /tmp/farmos-2x.tar
+      - name: Restore farmOS 2.x-dev Docker image from cache
+        uses: actions/cache@v3
+        with:
+          path: /tmp/farmos-2x-dev.tar
+          key: farmos-2x-dev-${{ github.run_id }}
+      - name: Load farmos/farmos:2.x-dev image
+        run: docker load < /tmp/farmos-2x-dev.tar
+      - name: Login to Docker Hub.
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+      # If the 2.x branch was pushed...
+      - name: Publish farmos/farmos:2.x image to Docker Hub.
+        if: github.ref_type == 'branch' && env.FARMOS_VERSION == '2.x'
+        run: docker push farmos/farmos:2.x
+      - name: Publish farmos/farmos:2.x-dev image to Docker Hub.
+        if: github.ref_type == 'branch' && env.FARMOS_VERSION == '2.x'
+        run: docker push farmos/farmos:2.x-dev
+      # If a tag was pushed, tag the Docker image and push to Docker Hub.
+      # If the tag is a valid semantic versioning string, also tag "latest".
+      # Semver regex from https://github.com/semver/semver/issues/199#issuecomment-43640395
+      - name: Tag and publish farmos/farmos:{tag} image to Docker Hub.
+        if: github.ref_type == 'tag'
+        run: |
+          docker tag farmos/farmos:2.x farmos/farmos:${{ env.FARMOS_VERSION }}
+          docker push farmos/farmos:${{ env.FARMOS_VERSION }}
+          if echo ${{ env.FARMOS_VERSION }} | grep -Pq '^(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$'; then
+            docker tag farmos/farmos:2.x farmos/farmos:latest
+            docker push farmos/farmos:latest
+          fi
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [Make translatable strings consistent between fields and actions #594](https://github.com/farmOS/farmOS/pull/594)
 - [Issue #3316925: Mark certain classes as @internal to indicate non-public APIs](https://www.drupal.org/project/farm/issues/3316925)
 - [Render link to taxonomy terms in farm entity views #595](https://github.com/farmOS/farmOS/pull/595)
+- [Issue #3203129: Use GitHub Actions to build Docker Hub images](https://www.drupal.org/project/farm/issues/3203129)
 
 ### Fixed
 


.github/workflows/deliver.yml
```diff
@@ -155,3 +155,58 @@ jobs:
           files: /tmp/farmOS-${{ env.FARMOS_VERSION }}.tar.gz
           draft: false
           prerelease: false
+  publish:
+    name: Publish to Docker Hub
+    # We only publish to Docker Hub if this is a tag or 2.x branch push event
+    # to the official repository.
+    if: |
+      github.repository == 'farmOS/farmOS' &&
```
paul121: Could also add a check to make sure the docker username/password are available before continuing

paul121 (commented):
Looks good, seems like this would work! Is there anything in particular to look at?

I'm curious how long the `docker push` takes, would it make sense to do these concurrently? Not a big deal anyways.

mstenta:
Thanks @paul121! Was just hoping for a quick sanity check. I think the next step is to spin up an `mstenta/farmos` Docker Hub repo and test it on that.

> Could also add a check to make sure the docker username/password are available before continuing

Not a bad idea.

Strange... it seems like it tried to run this when I pushed the `2.x-publish` branch to my fork, even though it's not the `farmOS/farmOS` repo... that check should have prevented it. :thinking: 

 https://github.com/farmOS/farmOS/actions/runs/3155701515/jobs/5134778747

mstenta:
> Strange... it seems like it tried to run this when I pushed the `2.x-publish` branch to my fork, even though it's not the `farmOS/farmOS` repo... that check should have prevented it. :thinking: 

Oh that was for the pull request... not the push to my fork.

Still... points to the fact that we should prevent all PR branches from running this!

mstenta:
> Could also add a check to make sure the docker username/password are available before continuing

Force pushed with a check for docker credentials (let's see if this works)...

mstenta:
Noting a few related things we'll need to do outside of code for this:

- Create a Docker Hub user with access to https://hub.docker.com/r/farmos/farmos/ that will be used for pushes.

After switching over to GitHub Actions we will want to disable/clean up the existing build process:

- Remove Docker Hub webhooks from https://github.com/farmOS/farmOS/settings/hooks (this should prevent new builds from being initiated in Docker Hub automatically)
- Delete "Build rules" and "Build triggers" from https://hub.docker.com/repository/docker/farmos/farmos/builds/edit
- Remove `farmOS-docker` GitHub user from "Linked Accounts" in farmOS organization settings.
  - Deactivate `farmOS-docker` GitHub user? I don't think this is used for anything else.

mstenta:
> Force pushed with a check for docker credentials (let's see if this works)...

Hmm can't seem to get this to work. First I tried:

`([[ ! -z ${{ secrets.DOCKER_USERNAME }} ]] && [[ ! -z ${{ secrets.DOCKER_PASSWORD }} ]])`

Which gave me: `The workflow is not valid. .github/workflows/deliver.yml (Line: 163, Col: 9): Unrecognized named-value: 'secrets'. Located at position 1 within expression: secrets.DOCKER_USERNAME`

https://github.com/farmOS/farmOS/actions/runs/3158714074

Then I tried:

`(secrets.DOCKER_USERNAME && secrets.DOCKER_PASSWORD)`

And that didn't prevent the action from running, resulting in the same error as before: `Username and password required`

https://github.com/farmOS/farmOS/actions/runs/3158757041/jobs/5141241052

I'm going to abandon this for now. I'm not sure it makes sense for this to be a primary means for preventing this from running, and having the error message is useful when the credentials aren't present IMO.

If anyone feels otherwise, and wants to suggest other logic to try I'm happy to give it another go!

I do want to prevent this from running on PRs, though... so I'll try to figure that out...

mstenta:
Well good news! I managed to successfully build and push a `2.x` image to my `mstenta/farmos` repo!

https://hub.docker.com/layers/mstenta/farmos/2.x/images/sha256-bf9bd6856780c3e3814aa241384ff7f471e6b876f4fecbc372979e4a87df10e9?context=repo

Still some outstanding issues:

```
    # We only publish to Docker Hub if this is a tag or 2.x branch push event
    # to the official repository.
    if: |
      github.repository == 'farmOS/farmOS' &&
      github.event_name == 'push' &&
      (github.ref_type == 'tag' || (github.ref_type == 'branch' && ${{ needs.build.outputs.farmos_version }} == '2.x'))
```

This doesn't seem to be working at all. I would not expect this job to run in my fork, but it is. I would also not expect it to run on a non-`2.x` branch (it worked on `2.x-publish` on my fork). I would ALSO not expect it to run on pull requests, which it did in `farmos/farmos`: https://github.com/farmOS/farmOS/actions/runs/3159008118/jobs/5141741979

```
      - name: Tag and publish mstenta/farmos:latest image to Docker Hub.
        if: github.ref_type == 'tag' && [[ ${{ env.FARMOS_VERSION }} =~ ^(?!0\d)\d+(?:\.(?!0\d)\d+){2}(?:-(?:(?!0\d)(?:(?!_)[\w-])+|\.(?!\.|$))+)?(?:\+(?:(?:(?!_)[\w-])+|\.(?!\.|$))+)?$ ]]
```

This condition ALSO doesn't seem to work. I would not expect this to run, but it did (and failed): https://github.com/mstenta/farmOS/actions/runs/3159007903/jobs/5141746415

There may be others... I have a lot of the logic commented out in my branch right now just to isolate testing.

mstenta:
> This doesn't seem to be working at all. I would not expect this job to run in my fork, but it is.

Reformatting the `if` so that it doesn't use a pipe w/ multiple lines fixed this.

> This condition ALSO doesn't seem to work. I would not expect this to run, but it did (and failed)

I worked around this by moving it out of the `step` condition and into a condition within the `run` command themselves.

mstenta:
Alright! This is working pretty well so far! I have it pushing `2.x` and `2.x-dev` images when the `2.x` branch is pushed, and tagged images when tags are pushed.

The only thing that isn't working yet is the logic for pushing the `latest` tag when a semver tag is pushed... still experimenting with that.

I'll force-push the current state of things shortly (I'm testing in another branch to avoid too much noise here).

Here are the images that have been successfully pushed to my `mstenta/farmos` Docker Hub repo: https://hub.docker.com/r/mstenta/farmos/tags

mstenta:
Waiting to confirm that this does NOT run on forks (only on the `farmOS/farmOS` repo)... https://github.com/mstenta/farmOS/actions/runs/3160119083

Also waiting to confirm that this NOT run on pull request branches... https://github.com/farmOS/farmOS/actions/runs/3160234329

mstenta:
> The only thing that isn't working yet is the logic for pushing the latest tag when a semver tag is pushed... still experimenting with that.

Got this working thanks to @symbioquine's ninja regex + grep skills!

> Waiting to confirm that this does NOT run on forks (only on the farmOS/farmOS repo)... https://github.com/mstenta/farmOS/actions/runs/3160119083
> 
> Also waiting to confirm that this NOT run on pull request branches... https://github.com/farmOS/farmOS/actions/runs/3160234329

Unfortunately something changed upstream causing our tests to stop passing generally... so I can't confirm these last two bits until that's resolved.

But as soon as we confirm those then I think this is ready to move out of "draft" status.

mstenta:
> Unfortunately something changed upstream causing our tests to stop passing generally... so I can't confirm these last two bits until that's resolved.

Fixed: https://github.com/farmOS/farmOS/commit/7694b2cbd3ee02884ddceba3f8e0bae5e695dfe2

Confirmed that this does NOT run on forks (only on the farmOS/farmOS repo): https://github.com/mstenta/farmOS/actions/runs/3161524273

Also confirmed that this NOT run on pull request branches: https://github.com/farmOS/farmOS/actions/runs/3161524455

mstenta:
Last steps to deploy this:

- Remove Docker Hub webhooks from https://github.com/farmOS/farmOS/settings/hooks (this should prevent new builds from being initiated in Docker Hub automatically)
- Merge the PR.
- Confirm that `2.x` and `2.x-dev` images are build and pushed to Docker Hub.
- Delete "Build rules" and "Build triggers" from https://hub.docker.com/repository/docker/farmos/farmos/builds/edit
- Remove farmOS-docker GitHub user from "Linked Accounts" in farmOS organization settings.
  - Deactivate farmOS-docker GitHub user? I don't think this is used for anything else.

.github/workflows/deliver.yml
```diff
@@ -155,3 +155,54 @@ jobs:
...
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+      # If the 2.x branch was pushed...
+      - name: Publish farmos/farmos:2.x image to Docker Hub.
+        if: github.ref_type == 'branch' && env.FARMOS_VERSION == '2.x'
+        run: docker push farmos/farmos:2.x
+      - name: Publish farmos/farmos:2.x-dev image to Docker Hub.
+        if: github.ref_type == 'branch' && env.FARMOS_VERSION == '2.x'
+        run: docker push farmos/farmos:2.x-dev
```
symbioquine, line 200: It looks like the `2.x` and `2.x-dev` docker tags will always have the same image - as opposed to the `2.x` tag only being updated for tagged releases. Am I reading that right? If so, does that match user expectations?

mstenta: `docker push farmos/farmos:2.x` will push the `2.x` image and `docker push farmos/farmos:2.x-dev` will push the `2.x-dev` image. Both get pushed whenever the `2.x` branch gets new commits. So I think this is correct, unless I'm overlooking something?

symbioquine: Isn't the `2.x` docker tag a little redundant - either with the `2.x-dev` or `latest` docker tags? It seems like it could be a source of issues for folks who unintentionally end up running pre-release versions when they mean to be using `latest` (and actually probably should be using tagged releases)?

I don't think it needs to block this PR or anything since I suspect this is in line with the existing behavior from Docker Hub builds, I'm just asking the question whether it actually makes sense to keep the `2.x` tag when there's also the `2.x-dev` and `latest` tags that are more explicit.

mstenta: Oh gotcha - yea this just maintains the existing behavior, which is also a continuation of the old 1.x behavior, which was defined in the build rules on our Docker Hub.

The `2.x` image is different from the `2.x-dev` image, which includes things like XDebug. It is the difference between these two `Dockerfile`s:

- https://github.com/farmOS/farmOS/blob/2.x/docker/Dockerfile
- https://github.com/farmOS/farmOS/blob/2.x/docker/dev/Dockerfile

The `latest` image will always be the same as the most recent semantic version tag.

So the `2.x` image is distinct from both `2.x-dev` and `latest`, in that it is the _non-dev_ image with the latest commit of the `2.x` branch. Someone _may_ want to use this if they need a commit that hasn't been included in a tagged release yet, but don't want dev dependencies included. Although hopefully in practice this won't be common.

symbioquine: Oh right! Forgot about the dev vs non-dev distinction. That makes sense!

symbioquine (approved):

paul121 (approved):

mstenta:
Alright! Looks like we're good to go with this.

I think I'll wait until we are ready for a new release before merging, so I remember to monitor/test everything at the same time.

PR #583, opened 2022-10-11
Issue #3314741: EntityFieldManager::getFieldMap() doesn't show farmOS bundle fields

mstenta:
See https://www.drupal.org/project/farm/issues/3314741

### DIFF
--- a/modules/core/entity/farm_entity.module
+++ b/modules/core/entity/farm_entity.module
@@ -15,6 +15,121 @@ use Drupal\farm_entity\BundlePlugin\FarmEntityBundlePluginHandler;
 use Drupal\farm_entity\Routing\DefaultHtmlRouteProvider;
 
 /**
+ * Implements hook_modules_installed().
+ */
+function farm_entity_modules_installed($modules, $is_syncing) {
+
+  // Add bundle fields to the bundle field map.
+  _farm_entity_rebuild_bundle_field_map('install', $modules);
+}
+
+/**
+ * Implements hook_modules_uninstalled().
+ */
+function farm_entity_modules_uninstalled($modules, $is_syncing) {
+
+  // Remove bundle fields from the bundle field map.
+  _farm_entity_rebuild_bundle_field_map('uninstall', $modules);
+}
+
+/**
+ * Helper function for rebuilding bundle field maps.
+ *
+ * This runs when modules are installed/uninstalled, loads bundle fields that
+ * are defined by the module via hook_farm_entity_bundle_field_info(), and
+ * updates the entity field map accordingly.
+ *
+ * @param string $op
+ *   Operation being performed. Must be either 'install' or 'uninstall'.
+ * @param array $modules
+ *   The list of modules that are being installed/uninstalled.
+ */
+function _farm_entity_rebuild_bundle_field_map(string $op, array $modules) {
+
+  // If the operation is not "install" or "uninstall", bail.
+  if (!in_array($op, ['install', 'uninstall'])) {
+    return;
+  }
+
+  // If none of the modules implement hook_farm_entity_bundle_field_info(),
+  // bail.
+  if (!\Drupal::moduleHandler()->hasImplementations('farm_entity_bundle_field_info', $modules)) {
+    return;
+  }
+
+  // Iterate through entity types.
+  $entity_type_definitions = \Drupal::service('entity_type.manager')->getDefinitions();
+  foreach ($entity_type_definitions as $entity_type => $entity_type_definition) {
+
+    // Only proceed for entity types that use bundle plugins.
+    if (!in_array($entity_type, ['asset', 'log', 'plan', 'quantity'])) {
+      continue;
+    }
+
+    // Get the bundle field map key value collection.
+    $bundle_field_map = \Drupal::service('keyvalue')->get('entity.definitions.bundle_field_map')->get($entity_type) ?? [];
+
+    // Get a list of installed bundles for this entity type.
+    $bundles = \Drupal::service('entity_type.bundle.info')->getBundleInfo($entity_type);
+
+    // Iterate through bundles.
+    foreach ($bundles as $bundle => $bundle_info) {
+
+      // Iterate through the modules.
+      foreach ($modules as $module) {
+
+        // Invoke hook_farm_entity_bundle_field_info() to get bundle field
+        // definitions added by the module.
+        /** @var \Drupal\entity\BundleFieldDefinition[] $bundle_fields */
+        $bundle_fields = \Drupal::service('module_handler')->invoke($module, 'farm_entity_bundle_field_info', [$entity_type_definition, $bundle]) ?? [];
+
+        // If bundle fields are empty, skip.
+        if (empty($bundle_fields)) {
+          continue;
+        }
+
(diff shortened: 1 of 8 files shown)


mstenta:
Note that this PR currently includes two commits that change the `equipment` and `quick` fields from bundle fields to base fields.

If we want to split those out to review separately that's fine with me.

modules/core/entity/farm_entity.module
```diff
@@ -14,6 +14,121 @@ use Drupal\entity\EntityPermissionProvider;
...
+  if (!\Drupal::moduleHandler()->hasImplementations('farm_entity_bundle_field_info', $modules)) {
+    return;
+  }
+
+  // Iterate through entity types.
+  $entity_type_definitions = \Drupal::service('entity_type.manager')->getDefinitions();
+  foreach ($entity_type_definitions as $entity_type => $entity_type_definition) {
+
+    // Only proceed for entity types that use bundle plugins.
+    if (!in_array($entity_type, ['asset', 'log', 'plan', 'quantity'])) {
```
symbioquine, line 65: I'm seeing [this array](https://github.com/farmOS/farmOS/blob/c92d181c45b2158b360318edb277ff175fb78cd7/modules/core/entity/farm_entity.module#L73) in a lot of places - sometimes in different orders or with/without the 'data_stream' entry.

Would it make sense to refactor it into a constant - or maybe just additional attributes on the entity definitions themselves?

e.g. here;

```php
if (!$entity_type_definition->supportsBundlePlugins()) {
    continue;
}
```

modules/core/entity/farm_entity.module
```diff
@@ -14,6 +14,121 @@ use Drupal\entity\EntityPermissionProvider;
...
+            if (empty($bundle_field_map[$field_name]['bundles'])) {
+              unset($bundle_field_map[$field_name]);
+            }
+          }
+        }
+      }
+    }
+
+    // Set the bundle field map key value collection.
+    if (!empty($bundle_field_map)) {
```
symbioquine, line 123: What if it became empty as part of the changes above? Shouldn't it still get set in that case?

modules/core/entity/farm_entity.post_update.php
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    // Set the bundle field map key value collection.
+    if (!empty($bundle_field_map)) {
+      \Drupal::service('keyvalue')->get('entity.definitions.bundle_field_map')->set($entity_type, $bundle_field_map);
+    }
+  }
+
+  // Delete the entity field map cache entry.
+  \Drupal::service('cache.discovery')->delete('entity_field_map');
```
symbioquine, line 71: Why implement this all again? Couldn't it just call `_farm_entity_rebuild_bundle_field_map('install', $modules)` with a list of all the modules that implement `hook_farm_entity_bundle_field_info`?

modules/core/entity/farm_entity.post_update.php
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * @file
+ * Post update functions for farmOS entity module.
+ */
+
+/**
+ * Add hook_farm_entity_bundle_field_info() fields to bundle field maps.
+ */
+function farm_entity_post_update_add_farm_entity_bundle_field_maps(&$sandbox = NULL) {
```
symbioquine, line 11: What's this `$sandbox` parameter? Do we need to use/honor it?

symbioquine, line 11: Probably just my Drupal naivete showing here, but does this implement a hook or is it called some other way?

modules/core/entity/tests/src/Functional/FarmEntityBundleFieldTest.php
```diff
@@ -59,6 +59,49 @@ protected function setUp():void {
...
+
+    // Reload the entity field map.
+    $this->entityFieldManager->setFieldMap([]);
+    \Drupal::service('cache.discovery')->delete('entity_field_map');
+    $field_map = $this->entityFieldManager->getFieldMap();
+
+    // Confirm that the 'test_contrib_hook_bundle_field' no longer exists in the
+    // log field map.
+    $this->assertArrayNotHasKey('test_contrib_hook_bundle_field', $field_map['log']);
+  }
```
symbioquine, line 103: Should this also test the case where more than one module provides a given bundle field? i.e. that the bundle gets added, the field mapping doesn't get overwritten or prematurely deleted?

modules/core/quick/tests/src/Kernel/QuickFormTest.php
```diff
@@ -94,13 +95,13 @@ public function testQuickFormSubmission() {
     $this->assertNotEmpty($storage['assets'][0]->id());
 
     // Confirm that the asset is linked to the quick form.
-    $this->assertEquals('test', $storage['assets'][0]->quick[0]);
+    $this->assertEquals('test', $storage['assets'][0]->get('quick')[0]->value);
```
symbioquine, line 98: Why does this need to change?

modules/core/entity/farm_entity.module
mstenta: Oh good idea! Didn't know about the `supportsBundlePlugins()` method!

paul121 (commented):
Just to capture what we chatted about on the monthly call..

This reminded me of this issue which provides a solution for a problem I have run into in update hooks adding/removing a bundle field in a custom module: https://www.drupal.org/project/drupal/issues/3129179

So considering that, it would be nice if there was a helper function we could call at any time to rebuild this field map during update hooks. To simplify things this helper function could rebuild the entire mapping, regardless of modules installed/uninstalled.

This issue I referenced may provide a solution for Drupal core, but we may need to determine if it works for the farmOS bundle field hook. It sounds like providing our own helper method for farmOS core would be the easiest near term solution (and call this helper function to do this logic as is being done by this PR)

modules/core/entity/farm_entity.module
symbioquine: image

paul121: I think we could use `EntityTypeInterface::getBundleEntityType()`

```

  /**
   * Gets the name of the entity type which provides bundles.
   *
   * @return string|null
   *   The name of the entity type which provides bundles, or NULL if the entity
   *   type does not have a bundle entity type.
   */
  public function getBundleEntityType();
 ```

mstenta:
@paul121 raised an important question on the dev call today: if we convert the `equipment` field to a base field, will that stop it from showing up in Log Views?

mstenta:
Ran into this again with the new Convention taxonomy + reference field provided by this module: https://github.com/mstenta/farm_convention/

Worked around it by converting it from a bundle field to a base field: https://github.com/mstenta/farm_convention/commit/a7fa1bcf8902a909e9f15276eaefea7a5cbe20de

mstenta:
I'm going to close this PR and start a new one that is based off of farmOS 3.x. I will explore whether we can patch Drupal core based on the work being done in https://www.drupal.org/project/drupal/issues/3129179.

Thanks for the reviews and feedback @symbioquine and @paul121! Hopefully the next PR will be simpler. :-)

mstenta:
Here is the new PR: https://github.com/farmOS/farmOS/pull/854

PR #584, opened 2022-10-13
Adapt csv export logic to support a type filter allowing single or multiple values

paul121:
The current logic is expecting the exposed type filter to always be an array. This causes csv exports to break when exporting from the all quantities page because the quantity type filter does not allow multiple values. When no quantity type is provided the type filter is set to 'All'.

Another solutiong would be to change the quantity type filter to allow multiple values... but more generally we should allow a single type to be provided as well. It's possible someone could manually build the CSV export link w/ query parameters and provide the type as a single value eg: `?type=animal` vs `?type[]=animal`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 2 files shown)


mstenta:
Thanks @paul121!

PR #588, opened 2022-10-21
Add action for bulk assigning asset parent

paul121:
This was originally requested here: https://github.com/Rothamsted-Ecoinformatics/farm_rothamsted/issues/283

I've implemented the asset parent action in the Rothamsted module and opening this PR to add to farmOS core. They will depend on the feature being in farmOS core if/when this PR is merged.

At first it wasn't clear which core module should host this asset parent action. Chatting with @mstenta we decided we could create a new `farm_parent` module similar to what we did for `farm_owner`. This should work well, it's extra nice because the `parent` field depended on `entity_reference_validators` which means we can move this dependency off of `farm_entity_fields` to `farm_parent` in the process as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - [Add action to bulk categorize logs #590](https://github.com/farmOS/farmOS/pull/590)
+- [Add action for bulk assigning asset parent #588](https://github.com/farmOS/farmOS/pull/588)
 
 ### Changed
 
--- a/modules/core/entity/farm_entity.info.yml
+++ b/modules/core/entity/farm_entity.info.yml
@@ -6,7 +6,6 @@ core_version_requirement: ^9
 dependencies:
   - entity:entity
   - entity_reference_integrity:entity_reference_integrity_enforce
-  - entity_reference_validators:entity_reference_validators
   - exif_orientation:exif_orientation
   - farm:farm_entity_fields
   - farm:farm_entity_views
--- a/modules/core/entity/modules/fields/farm_entity_fields.base_fields.inc
+++ b/modules/core/entity/modules/fields/farm_entity_fields.base_fields.inc
@@ -43,28 +43,12 @@ function farm_entity_fields_asset_base_fields() {
         'view' => 10,
       ],
     ],
-    'parent' => [
-      'type' => 'entity_reference',
-      'label' => t('Parents'),
-      'description' => t('If this asset descends from other assets, they can referenced here to create a lineage relationship.'),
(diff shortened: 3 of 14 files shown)


paul121:
Tests are passing! There were a few tests that needed the `parent` field and I also was missing schema for the action.

The only other consideration I had with this feature is that we're starting to introduce a lot bulk actions. It would be nice if we could allow users to toggle which are available. But for now I think it is better to have this feature included

mstenta:
This LGTM - thanks @paul121!

I rebased onto `2.x`, made the `CHANGELOG.md` line into a link to this PR, and added one commit:

https://github.com/farmOS/farmOS/commit/23ef72c3955578f6f73254d7c22c72f84a0e0a23

Does that seem sensible @paul121? Basically just copied the title and description from the existing "Parents" field, so that it can use the same translations. Maybe the `#description` isn't perfect (in *either* context), but perhaps that can be changed as a next step. For now (just adding an action), seems like it's best to just reuse the strings so we get existing translations.

If you approve, we're good to merge!

paul121 (commented):
> If this asset descends from other assets, they can referenced here to create a lineage relationship.

That sounds good, although just realizing we have a typo in that existing #description.. "they can _**be**_ referenced here"

mstenta:
Ah jeez. Good eye @paul121 - maybe we should just change both then since fixing it would invalidate existing translations anyway.

paul121:
There is a similar consideration for the "Assign owner" action - right now it has a different `#description` between the field and action form:

https://github.com/farmOS/farmOS/blob/cbef96909123f87e95b2681f5709205103fa2c2e/modules/core/owner/farm_owner.module#L16-L28

https://github.com/farmOS/farmOS/blob/cbef96909123f87e95b2681f5709205103fa2c2e/modules/core/owner/src/Form/AssignActionForm.php#L171-L177

I don't have too strong of an opinion here.. using the same text is convenient, but it seems that the form could have text that is a bit more useful. The "Append/Replace" functionality of these action forms is particularly interesting here... I think it influences if the action field should include language like "optional" ?

mstenta:
Oh man good point. The scope is creeping a bit on this haha. I think we probably *should* make the owner assignment field and action consistent too... and fix the grammar issue in the `#description`. But maybe we should just revert my commit and follow up. I want to get this action merged. :-)

mstenta:
All of this has me thinking... maybe we should have a "gate" for PRs that asks: "does this add or change any translatable strings?" so we can be more mindful of how that affects translations moving forward. Since we aren't "stable" yet I think we can be a little more loose and fix these bits.

mstenta:
Alright I reverted my change to the `#description` field, but kept the other two changes (and just amended the commit that adds the action, since they were very minor). We can follow-up on the `#description` here and the other action strings... let's get this one merged first!

PR #591, opened 2022-11-03
Fix click sort columns expand filters fieldset

### LINKED ISSUE #586: Click sort columns expand filters fieldset
Reported in this forum comment: https://farmos.discourse.group/t/farmos-2-0-0-beta7-has-been-released/1366/4?u=mstenta

Summary: if you click a column name in farmOS Views to sort that column, the (normally collapsed) "Filters" box opens up.

We started collapsing "Filters" by default in https://www.drupal.org/project/farm/issues/3306344, but included logic to expand it if any URL query parameters are set.

paul121:
Closes #586 

I think this is the easiest approach here. When a click sort is applied all of the default filter values are passed as query params. We can check the provided exposed input against the default filter values to see if there are any non-default filter values. If so, we can open filters.


mstenta:
Thanks @paul121!

PR #593, opened 2022-11-12
Check PHP compatibility with codesniffer

mstenta:
Recommended by @pcambra. This will give us the ability to quickly check that all our code is compatible with the version(s) of PHP we claim to support.

### DIFF
--- a/.github/workflows/deliver.yml
+++ b/.github/workflows/deliver.yml
@@ -75,6 +75,10 @@ jobs:
         run: docker run farmos/farmos:2.x-dev phpcs /opt/drupal/web/profiles/farm
       - name: Run PHPStan
         run: docker run farmos/farmos:2.x-dev phpstan analyze /opt/drupal/web/profiles/farm
+      - name: Check PHP compatibility of contrib modules and themes (ignore warnings).
+        run: |
+          docker run farmos/farmos:2.x-dev phpcs --standard=PHPCompatibility --runtime-set testVersion 7.4- --warning-severity=0 /opt/drupal/web/modules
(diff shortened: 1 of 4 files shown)


mstenta:
I also added a commit that runs checks against our contrib modules and themes, in hopes that we can catch potential issues in the future.

symbioquine (approved):

mstenta:
Oops accidentally left a bit of test code in there (return type on a method I was using to test if PHP 5.6 check failed). Removed and force pushed.

mstenta:
FYI @paul121 I had to rename the `Numeric` class in the data stream notification module to fix the only warning that came up... see this commit for the details: https://github.com/farmOS/farmOS/pull/593/commits/4b149a6e143a7511ed904d5ed3542592c43785ce

paul121 (commented):
Nice! We had talked about doing this check without failing the workflow. But I assume it could be quite concerning if this found anything other than warning errors?

paul121:
Curious what types of things the warnings were finding too. I can look later myself, just curious tho

mstenta:
Yea I figured for these it would be good to know if they fail, and I put it after the sniff/stan of our own code so it won't block those. I believe we also talked about running PHPStan on contrib to look for deprecations. I think the main thing is we wouldn't want ALL phpcs and phpcs output to fail the overall jobs, so we'd need to tailor them to target specific things like this is doing (only running the PHPCompatibility sniffs).

All of the warnings were similar to the one I fixed in our code, and IIRC all were in the migrate module's use of the "resource" reserved keyword.

PR #594, opened 2022-11-16
Make translatable strings consistent between fields and actions

mstenta:
Follow-up PR based on conversation in #588

### DIFF
--- a/modules/core/owner/farm_owner.module
+++ b/modules/core/owner/farm_owner.module
@@ -17,8 +17,8 @@ function farm_owner_entity_base_field_info(EntityTypeInterface $entity_type) {
   if (in_array($entity_type->id(), ['asset', 'log'])) {
     $field_info = [
       'type' => 'entity_reference',
-      'label' => t('Owner'),
-      'description' => t('Optionally specify who is in charge of this record.'),
+      'label' => t('Owners'),
+      'description' => t('Assign ownership to one or more users.'),
       'target_type' => 'user',
       'multiple' => TRUE,
       'weight' => [
--- a/modules/taxonomy/log_category/src/Form/LogCategorizeActionForm.php
+++ b/modules/taxonomy/log_category/src/Form/LogCategorizeActionForm.php
@@ -155,6 +155,7 @@ class LogCategorizeActionForm extends ConfirmFormBase {
     $form['category'] = [
       '#type' => 'select',
       '#title' => $this->t('Log category'),
+      '#description' => $this->t('Use this to organize your logs into categories for easier searching and filtering later.'),
       '#options' => $options,
       '#multiple' => TRUE,
     ];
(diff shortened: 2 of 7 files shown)


modules/core/owner/farm_owner.module
```diff
@@ -17,8 +17,8 @@ function farm_owner_entity_base_field_info(EntityTypeInterface $entity_type) {
   if (in_array($entity_type->id(), ['asset', 'log'])) {
     $field_info = [
       'type' => 'entity_reference',
-      'label' => t('Owner'),
-      'description' => t('Optionally specify who is in charge of this record.'),
+      'label' => t('Owners'),
+      'description' => t('Assign ownership to the selected user(s).'),
```
paul121: The "(s)" just bugs me, what about "Assign ownership to one or more users"? That's more explicit that one or more can be selected?

modules/taxonomy/log_category/src/Form/LogCategorizeActionForm.php
```diff
@@ -155,6 +155,7 @@ public function buildForm(array $form, FormStateInterface $form_state) {
     $form['category'] = [
       '#type' => 'select',
       '#title' => $this->t('Log category'),
+      '#description' => $this->t('Use this to organize your logs into categories for easier searching and filtering later.'),
```
paul121, line 158: Could we condense this? It seems a bit verbose and inconsistent from other descriptions, maybe don't include "your"?

"Organize logs into categories for easier searching and filtering."

This is also nearly identical to the flags description. Can we emphasize that you can build your own log category hierarchy? In Rothamsted modules we have included a link to the hierarchy, but not sure how viable that may be here.

paul121 (commented):
Overall I approve, this does make the strings consistent, but I wonder about revisiting these strings more generally...

modules/core/owner/farm_owner.module
mstenta: Good suggestion @paul121 - it does mean a change to BOTH strings, but I guess we're already deciding to do that with parents. Better now than post-stable I guess. I'll change this and re-push shortly...

mstenta: (We also JUST recently changed these ownership strings, so we broke translations anyway. I don't feel as bad doing it again immediately.)

modules/taxonomy/log_category/src/Form/LogCategorizeActionForm.php
mstenta: These feel like bigger considerations than just consistency. Not that I disagree - but maybe we can tackle them in a different PR?

> In Rothamsted modules we have included a link to the hierarchy, but not sure how viable that may be here.

This would probably need an access check too - not everyone who can edit fields can access the hierarchy, I think.

PR #595, opened 2022-11-17
Render link to taxonomy terms in farm entity views

paul121:
I think this would be a nice addition since: https://github.com/farmOS/farmOS/pull/458


mstenta:
LGTM! I actually thought this was already done. :-)

PR #597, opened 2022-11-18
Change default client secret to be NULL to avoid issue #3322325

paul121:
It seems like a good idea to use `NULL` for the consumer/client `secret` going forward.. here's a PR for this approach if we'd like to do so

https://www.drupal.org/project/simple_oauth/issues/3322325


mstenta:
Thanks @paul121!

What do you think about also applying the `simple_oauth` patch from [#3322325](https://www.drupal.org/project/simple_oauth/issues/3322325) in this same PR?

I added a commit before yours to that effect, rebased and force-pushed... thoughts?

paul121:
We need to see tests pass on that path/MR before we should merge it. Unfortunately I'm not sure how to trigger the simple_oauth Gitlab tests pipeline...

Keeping them separate would be nice because right now the `farm` client in the 2.x-dev image is broken. With this PR we could fix the issue for new installs on 2.x-dev which is used in automated tests for other modules and farmOS.py/farmOS.js

mstenta:
Roger that - I'll rebase+merge this without my patch commit, and create a follow-up PR for that. Thanks for the explanation!

mstenta:
Follow-up: https://github.com/farmOS/farmOS/pull/600

PR #598, opened 2022-11-19
Fix input log quantity material migration

### LINKED ISSUE #579: Input material migration issue
We identified a bug with the v1->v2 input log migration logic and how it migrates material taxonomy terms into material quantities.

Originally discovered here: https://github.com/Rothamsted-Ecoinformatics/farm_rothamsted/issues/300#issuecomment-1252833290

This issue affects input logs that referenced multiple material terms in farmOS v1. During the migration to v2, these material terms are migrated into new "material quantity" entities that are referenced by the log. The original issue that added this logic is here (with lots of background): https://www.drupal.org/project/farm/issues/3196834

The result is that multiple material quantities are created in the v2 database (one per material term referenced in the v1 log), but only the FIRST material term is referenced in all of them.

We need to fix this in the migration logic.

We will also need to create a followup migration/script to fix this on existing sites, if possible.

mstenta:
Fixes #579

### DIFF
--- a/modules/core/migrate/farm_migrate.post_update.php
+++ b/modules/core/migrate/farm_migrate.post_update.php
@@ -1,0 +1,264 @@
+<?php
+
+/**
+ * @file
+ * Post update hooks for the farm_migrate module.
+ */
+
+use Drupal\Core\Database\Database;
+use Drupal\Core\Site\Settings;
+use Drupal\Core\Utility\UpdateException;
+
+/**
+ * Fix migrated input log quantity materials.
+ */
+function farm_migrate_post_update_fix_input_log_quantity_materials(&$sandbox) {
+
+  // For more information see https://github.com/farmOS/farmOS/issues/579.
+
+  // This update can be skipped by adding:
+  // $settings['farm_migrate_skip_input_log_migration_fix'] = TRUE;
+  // to settings.php.
+  if (Settings::get('farm_migrate_skip_input_log_migration_fix', FALSE) === TRUE) {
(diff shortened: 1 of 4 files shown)


modules/core/migrate/farm_migrate.post_update.php
```diff
@@ -0,0 +1,264 @@
...
+function farm_migrate_post_update_fix_input_log_quantity_materials(&$sandbox) {
+
+  // For more information see https://github.com/farmOS/farmOS/issues/579.
+
+  // This update can be skipped by adding:
+  // $settings['skip_input_log_migration_fix'] = TRUE;
+  // to settings.php.
+  if (Settings::get('skip_input_log_migration_fix', FALSE) === TRUE) {
+    return t("Skipping automatic fix of input log material quantities because of this line in settings.php: \$settings['skip_input_log_migration_fix'] = TRUE;");
+  }
```
paul121: Not a biggie but maybe we could prefix this settings variable with `farm_update_*`. Just helps to make it extra clear that this update hook & settings variable is only needed if the `farm_update` module is enabled. But that could be explained in release notes too

paul121 (approved):
LGTM

modules/core/migrate/farm_migrate.post_update.php
mstenta: Yea good call. I'll make that change (add `farm_migrate_` prefix) and then merge this. Thanks for the review @paul121!

PR #602, opened 2022-11-25
Add "Date received" and "Date processed" fields to lab test logs

mstenta:
This adds two new optional fields to lab test logs, to represent when the lab received the sample, and when they processed it.

This comes out of the work we are doing with OpenTEAM to support importing soil test data in the Modus forum. Modus includes these fields and farmOS does not.


mstenta:
Note that this PR is built on top of #601, so it currently includes all the commits in that PR as well. I will rebase this after that is merged.

mstenta:
Merged. https://github.com/mstenta/farmOS/commit/fbd754e867bc261c5cf0e0086efd0133a2a480c2

PR #603, opened 2022-11-25
Convert lab field to term reference on lab test logs

mstenta:
Currently "Lab test" logs have a "Laboratory" field on them that takes a simple string. This PR adds a new "Lab" taxonomy, changes the "Laboratory" field to a term reference, and migrates all existing values to terms.


mstenta:
Note that this PR builds on top of both #601 and #602, but only adds a single commit. All of those PR commits are shown with this one, but they will be merged first ideally.

PR #604, opened 2022-11-25
Add a test quantity type with test method taxonomy

mstenta:
This PR adds a new "Test" quantity type and makes it the default for "Lab test" log types. This new quantity type has a "Test method" field that references terms in a new "Test methods" taxonomy.

The intended use-case for this is to record what laboratory testing method was used for each lab test quantity measurement. This will allow better use of the Modus nomenclatures (https://bitbucket.org/modus/analysis-nomenclature/src/master/), but is also flexible so it can be used in other ways as well.


mstenta:
This PR builds on top of these other PRs, so it includes their commits as well, but only adds 5 commits of its own: #601, #602, #603

mstenta:
Merged: https://github.com/mstenta/farmOS/commit/90e9e2aad7ada53eabacd61cc22a8bb4b725ca33

PR #605, opened 2022-11-25
Add tissue lab test type

mstenta:
This simply adds a new test type for "Tissue" samples (eg: plant tissue analysis).


mstenta:
Note that this PR is just a single commit, but is built off of a few other open PRs, so those commits are also shown here: #601, #602, #603, #604

PR #608, opened 2022-11-28
Correct hook_farm_update_exclude_config API docs

paul121:
The `.yml` is not needed here. This is already correct in the farmos.org development docs: https://farmos.org/development/module/updates/#hook_farm_update_exclude_config


mstenta:
Good catch - thanks @paul121!

PR #609, opened 2022-11-29
Correct CSS classname for priority flag

paul121:
The `.flag--monitor` class was duplicated in our flag.css file! I believe one of these flags was supposed to be `.flag--priority`

Also included a commit to chain the flag specific class `.flag--priority` after the generic `.flag` class. This is a better pattern to follow and ensure that flags will receive their intended styles. This chaining was necessary for the farm_organic module to provide styles for its flags: https://git.drupalcode.org/project/farm_organic/-/merge_requests/1


mstenta:
Rebased and merged - thanks @paul121!

PR #610, opened 2022-11-29
Fix user admin permissions form alter for managed roles

paul121:
We have a form alter to check and disable checkboxes for managed roles on the role permissions page. This works for the form with all role permissions `admin/people/permissions` but doesn't work correctly on the individual role permissions page eg: `admin/people/permissions/anonymous`. Interestingly both of these forms use the same form ID `user_admin_permissions` which is why both forms are being altered.

On the individual role permissions page it states that *all* roles are "managed" and does not check or disable the permission checkboxes as it should. This happens because the `array_search($role->id(), $role_names)` returns `FALSE` when a role ID is not found and the 0 index is used in the `$managed_role_indices` array.

Fixing the index calculation solves the issue and all this logic continues to work on both permission forms :rocket: 

Screenshot from 2022-11-29 11-30-31

Screenshot from 2022-11-29 11-46-21

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ soon as possible.
 
 - [Correct hook_farm_update_exclude_config API docs #608](https://github.com/farmOS/farmOS/pull/608)
 - [Correct CSS classname for priority flag #609](https://github.com/farmOS/farmOS/pull/609)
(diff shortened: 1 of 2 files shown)


mstenta (approved):

mstenta:
Awesome thanks @paul121!

PR #614, opened 2022-12-06
Include GEOS version in status report

paul121:
I've been working on getting both my local and a few production environments up to PHP 8.0+. In the process I revisited the PHP-GEOS extension and libgeos "GEOS" requirements.

I noticed in my local environment the docker container ends up with GEOS 3.9.0. While the production environment on platformsh was failing to build PHP-GEOS with GEOS 3.11.1. The fix was to use GEOS 3.11.0 in the production environment. I'm not sure why this wasn't working.. but that is kinda beside the point.

In my debugging process it would have been convenient if the GEOS version was available right in the status report. It is currently available in the PHP info. Many other status report items include the relevant version

This `GEOSVersion()` function is included in php-geos here: https://github.com/libgeos/php-geos/blob/master/geos.c#L44

Screenshot from 2022-12-05 16-42-31
Screenshot from 2022-12-05 16-42-02

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ soon as possible.
 ### Added
 
 - [Issue #3186530: farmOS 2.x PHP 8 support](https://www.drupal.org/project/farm/issues/3186530)
+- [Include GEOS version in status report #614](https://github.com/farmOS/farmOS/pull/614)
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


mstenta (changes requested):
I like this idea! Looks like it just needs a slight tweak to appease codesniffer.

paul121:
Oops thanks, I missed that. Added a `phpcs:ignore Squiz.PHP.LowercasePHPFunctions.CallUppercase` for that line

mstenta (approved):

mstenta:
Thanks @paul121!

PR #615, opened 2022-12-06
Fix enable/disable buttons on data stream notifications page

### LINKED ISSUE #613: Enable/disable data stream notification not working from the quick-menu.
In the notification's list view, the deactivate menu item is not working.
It works if it's deactivated (or activated) while editing the notifacation itself.

Also, it's not working to re-activate a notification directly from the list of deactivated notifications.

paul121:
Closes #613 

The correct `entity_access` permission operation is `update` not `edit`


mstenta:
Thanks @paul121 !

PR #617, opened 2022-12-06
Shorten name of tests and delivery workflow

paul121:
This should make it easier to read the results of individual jobs in this workflow. See screenshot. With a shorter workflow name the individual job names should appear without being cut off :-)

Screenshot from 2022-12-06 10-06-02


paul121:
haha you can even see this change in this pr:

Screenshot from 2022-12-06 10-35-57

mstenta:
Thanks @paul121!

PR #618, opened 2022-12-08
Reorder fields in farm_log view

paul121:
This reorders the fields in the farm_log table view as described in this forum post: https://farmos.discourse.group/t/change-column-order-in-logs-view/1419?u=paul121

Before: status, ID, timestamp, log name, log type**
After: log name, log type**, timestamp, status, ID (moved to very last)
_**the log type column is only displayed on the `/logs page`, not `/logs/{log_type]`_

Unfortunately because this PR only changes the field order, this is blocked by this issue: https://www.drupal.org/project/farm/issues/3325911

I would like to consider removing the log ID from these table views, but I think we do need the log ID in the CSV export. We could override the CSV export to include the log ID but then then we lose the convenience of the "default views fields" in the CSV export display. If we did make this change for the log ID then this PR would not be blocked by the above issue.

What do we think?

### DIFF
--- a/modules/core/ui/views/config/install/views.view.farm_log.yml
+++ b/modules/core/ui/views/config/install/views.view.farm_log.yml
@@ -87,144 +87,6 @@ display:
           action_title: Action
           include_exclude: exclude
           selected_actions: {  }
-        status:
-          id: status
-          table: log_field_data
-          field: status
-          relationship: none
-          group_type: group
-          admin_label: ''
-          entity_type: log
-          entity_field: status
-          plugin_id: field
-          label: Status
-          exclude: false
-          alter:
-            alter_text: false
-            text: ''
-            make_link: false
-            path: ''
-            absolute: false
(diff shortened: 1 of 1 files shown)


mstenta:
I'll try to add thoughts to the forum topic so we don't split the conversation too much, but one question:

> If we did make this change for the log ID then this PR would not be blocked by the above issue.

Not sure I understand, aren't all of these changes related to moving the columns around, not just the log ID?

mstenta:
I'm going to close this because it is quite outdated at this point.

Ultimately, I think the best long-term solution to this would be to expose some configuration for the Views provided by `farm_ui_views` which allows the Config Admin role to reorder and show/hide columns. That would enable the kind of customization that we've heard requests for, without getting bogged down in debates over which column order and which fields to include. We can ship with some sane defaults (like we try to now) and then let individual instances configure to their taste.

Issue #497, opened 2022-01-31
Quantities include newlines in Log CSV exports


mstenta:
When exporting a set of Logs as CSVs, the "Quantity" columns currently summarizes all quantities associated with the log, but many newlines are included which break the formatting:

![Screenshot from 2022-01-31 06-08-25](https://user-images.githubusercontent.com/95381/151783707-5bee1677-7875-44b2-a024-c2283ee1ed2b.png)

mstenta:
Related: the View of logs gets pretty unwieldy when there are lots of quantities associated with the logs, because they all get included in the output.

Screenshot for example:

![Screenshot from 2022-09-08 09-16-31](https://user-images.githubusercontent.com/95381/189131925-535530ac-ac73-4263-95e1-392835f346c2.png)

The same code is used to generate both this View and the CSV exports... so it seems to me we should consider both in whatever solution we come up with.

mstenta:
Or maybe they should be two separate issues... uncertain yet. We can split this up if we need to...

mstenta:
Maybe we can brainstorm this on today's dev call.

mstenta:
> Related: the View of logs gets pretty unwieldy when there are lots of quantities associated with the logs, because they all get included in the output.

I opened a very simple PR for this one specifically, in hopes that we can take a small step before 2.0.0: https://github.com/farmOS/farmOS/pull/619

mstenta:
Opened a PR with a proposed solution for this newlines problem: https://github.com/farmOS/farmOS/pull/622

PR #619, opened 2022-12-13
Only show the first quantity in log Views

mstenta:
This is a very simple change to resolve the issue illustrated in this comment: https://github.com/farmOS/farmOS/issues/497#issuecomment-1240706646

We discussed more elaborate potential fixes for this, but in the interest of getting *something* done before 2.0.0 stable I propose we take this initial small step and open up a follow-up if there is desire.

Another benefit of this is that it will improve performance in lists of logs that have lots of quantities associated with them, by only loading one quantity for each log.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ soon as possible.
 
 - [Remove default_measure configuration from quantity type definitions #612](https://github.com/farmOS/farmOS/pull/612)
 - [Shorten name of tests and delivery workflow #617](https://github.com/farmOS/farmOS/pull/617)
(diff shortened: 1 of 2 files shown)


mstenta:
Worth noting that this is how it worked in farmOS v1. When we built these Views in v2, it was an "easy win" to show all quantities, but I think we've learned that it needs more intentional UX/UI work to really make it work well.

paul121 (commented):
This is good with me! But do want to confirm: I think this will only show a single quantity in the CSV export of logs, is that intended?

mstenta:
@paul121 That is true, and was also true in v1.

Users can get all quantities in CSV via the quantity Views, and that's generally better anyway because it separates them out to multiple rows.

PR #622, opened 2022-12-16
Remove line breaks and whitespace from log CSV quantity cells

### LINKED ISSUE #497: Quantities include newlines in Log CSV exports

mstenta:
Fixes https://github.com/farmOS/farmOS/issues/497

I took the approach described in this comment: https://www.drupal.org/project/csv_serialization/issues/2914214#comment-12575679

I decided to add this to a new `farm_csv` module, similar to our existing `farm_kml` module. This module uses a `ServiceModifierInterface` class to override the class used for the CSV encoder provided by `csv_serialization` with our own `FarmCsvEncoder`, which extends from the original and simply performs some extra formatting via the `formatValue()` method.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [Prevent circular asset location #568](https://github.com/farmOS/farmOS/pull/568)
 - [Prevent circular group membership #562](https://github.com/farmOS/farmOS/pull/562)
 - [Issue #3328419: Uninstalling farm_ui_views module breaks site](https://www.drupal.org/project/farm/issues/3328419)
+- [Remove line breaks and whitespace from log CSV quantity cells #622](https://github.com/farmOS/farmOS/pull/622)
 
 ## [2.0.0-rc1] 2022-12-15
 
--- a/modules/core/quantity/config/install/core.entity_view_mode.quantity.plain_text.yml
+++ b/modules/core/quantity/config/install/core.entity_view_mode.quantity.plain_text.yml
@@ -1,0 +1,12 @@
+langcode: en
+status: true
+dependencies:
+  enforced:
+    module:
+      - quantity
+  module:
+    - quantity
+id: quantity.plain_text
+label: Plain text
+targetEntityType: quantity
+cache: false
--- a/modules/core/quantity/quantity.post_update.php
+++ b/modules/core/quantity/quantity.post_update.php
@@ -1,0 +1,31 @@
+<?php
+
+/**
+ * @file
+ * Post update hooks for the quantity module.
+ */
+
+use Drupal\Core\Entity\Entity\EntityViewMode;
+
+/**
+ * Create plain text view mode for quantities.
+ */
+function quantity_post_update_plain_text_view_mode(&$sandbox) {
+  $view_mode = EntityViewMode::create([
+    'id' => 'quantity.plain_text',
+    'label' => 'Plain text',
+    'targetEntityType' => 'quantity',
+    'cache' => FALSE,
+    'dependencies' => [
+      'enforced' => [
+        'module' => [
+          'quantity',
+        ],
+      ],
+      'module' => [
+        'quantity',
+      ],
+    ],
+  ]);
+  $view_mode->save();
+}
--- a/modules/core/quantity/src/Entity/Quantity.php
+++ b/modules/core/quantity/src/Entity/Quantity.php
@@ -30,7 +30,7 @@ use Drupal\user\EntityOwnerTrait;
  *     "inline_form" = "\Drupal\quantity\Form\QuantityInlineForm",
  *     "list_builder" = "\Drupal\quantity\QuantityListBuilder",
  *     "permission_provider" = "\Drupal\entity\UncacheableEntityPermissionProvider",
- *     "view_builder" = "Drupal\Core\Entity\EntityViewBuilder",
+ *     "view_builder" = "Drupal\quantity\QuantityViewBuilder",
  *     "views_data" = "Drupal\quantity\QuantityViewsData",
  *     "form" = {
  *       "default" = "Drupal\Core\Entity\ContentEntityForm",
--- a/modules/core/quantity/src/QuantityViewBuilder.php
+++ b/modules/core/quantity/src/QuantityViewBuilder.php
@@ -1,0 +1,54 @@
+<?php
+
+namespace Drupal\quantity;
+
+use Drupal\Component\Utility\Html;
+use Drupal\Core\Entity\EntityInterface;
+use Drupal\Core\Entity\EntityViewBuilder;
+use Drupal\Core\Security\TrustedCallbackInterface;
+
+/**
+ * Quantity entity view builder.
+ */
+class QuantityViewBuilder extends EntityViewBuilder implements TrustedCallbackInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function view(EntityInterface $entity, $view_mode = 'full', $langcode = NULL) {
+    $build = parent::view($entity, $view_mode, $langcode);
+
+    // If the view mode is plain_text, add a post render callback to strip tags
+    // and whitespace.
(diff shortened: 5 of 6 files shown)


modules/core/csv/src/FarmCsvServiceProvider.php
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Overrides the CSV Serialization service class.
+ */
+class FarmCsvServiceProvider implements ServiceModifierInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function alter(ContainerBuilder $container) {
```
symbioquine: Would it make sense for this to decorate the existing implementation rather than be a straight replacement so other customizations could also be applied without being mutually exclusive with this one?

symbioquine (approved):
One question, otherwise looks good.

paul121 (dismissed):
Similar to @symbioquine 's comment... altering this `csv` encoder seems a bit heavy handed. If anything else uses this csv_serialization `csv` encoder, it will now have newlines and whitespace removed too. As I understand newlines are allowed in CSVs... 

I know our current views don't include longer text fields like `notes` but if/when they do, I do think that newlines should be included in the CSV output. It just wouldn't be my preferred format for exporting that type of data :-)

Instead of altering this encoder, can we just provide our own encoder service for csv? The class could still extend `CsvEncoder` and include this same logic. Then we just configure our view displays to use the new encoder we provide.

mstenta:
Thanks for taking a look @paul121 @symbioquine. I am not completely satisfied with this approach either (because it is heavy-handed, as you both pointed out), but it worked and was easy, so I figured I'd open a PR to start the conversation at least. :-)

> Instead of altering this encoder, can we just provide our own encoder service for csv? The class could still extend `CsvEncoder` and include this same logic. Then we just configure our view displays to use the new encoder we provide.

We could do this. It feels a bit weird to me, though. It would mean having options like this in the Views config:

- CSV
- CSV (remove all newlines)

Also, that wouldn't prevent this issue:

> I know our current views don't include longer text fields like `notes` but if/when they do, I do think that newlines should be included in the CSV output.

So it might just be kicking the can.

I think ideally we would be able to target the `quantity` field in our Views specifically, and fix the specific issue at hand, rather than removing all newlines everywhere. I'm not sure if there's an easy way to do this, however, without adding undo complexity.

The first thing I tried was removing newlines from our quantity entity twig template file itself, and that worked *partially*, but there were other newlines still present from outside of that template. I also feel like that isn't the "right" solution to this either, because it is not an *explicit* targeted fix, and could be accidentally undone in the future (or if someone provides their own quantity template override and isn't aware of this).

I also tried using `hook_views_pre_render()` (too early in the process to alter the output), `hook_views_post_render()` (only gives us access to the ENTIRE output, so targeting the quantity column requires post-processing the CSV output in its entirety), and `hook_views_preprocess_views_view_field()` (which seemed like it should work, but doesn't fire for the CSV export display - maybe because it is a REST display and so that's handled differently, I'm not sure).

Another option might be to provide a new view mode/display for quantities, and use that in the View. This might give us more flexibility in targeting and altering it via a hook. In order to do this, however, I think we would have to override the fields config for the CSV display of the View, which we've managed to avoid up until now.

mstenta:
I pushed some new commits that change how this is done, so that it targets the quantity cells specifically.

The new `farm_csv` module now provides a "CSV" view mode for quantity entities. It adds a post-render callback when quantities are rendered with this view mode, which strips tags, newlines, and whitespace from the output. Then, in the `farm_ui_views` module, for the `farm_log` View, the quantity view mode is programmatically overridden from `default` to `csv` when the "CSV export" display is used.

modules/core/csv/src/FarmCsvServiceProvider.php
mstenta: Resolving this conversation because a different approach is being taken, which no longer replaces this class.

mstenta:
Looking for feedback on the approach taken, as well as *where* I put the code for this. I've been too close to it, so it could use some fresh eyes. Does it make sense? Would it be better suited somewhere else? Or does the `farm_csv` module make sense for this?

mstenta:
Here is what the CSV export for a test activity log with a single quantity looks like before/after. I installed a fresh farmOS 2.x instance, created an activity log with a standard quantity, exported the activity logs CSV, then checked out the `2.x-csv` branch, ran `drush updb`, and exported the same CSV again.

Before:

```
"Bulk update",Status,id,Timestamp,"Log name","Log type",Assets,Location,Quantity,"Equipment used",Flags,"Log category",Owner
,Pending,1,2022-12-21,"Activity log 1",Activity,,,"My Label
    (


  Count
)    


  10
 


  item(s)",,,,admin
```

After:

```
"Bulk update",Status,id,Timestamp,"Log name","Log type",Assets,Location,Quantity,"Equipment used",Flags,"Log category",Owner
,Pending,1,2022-12-21,"Activity log 1",Activity,,,"My Label ( Count ) 10 item(s)",,,,admin
```

paul121 (commented):
This looks good to me. And I think it makes sense for this to be in `farm_csv` as implemented here.

However, I must admit that after seeing all of this it feels like a lot of complexity just to enable exporting a single quantity for logs... not to dig up an old question but maybe worth revisiting... is having a single quantity particularly useful? what if we just don't include quantities in this export?

mstenta:
It's a valid question... which implies two options:

1. We remove the Quantity field from `farm_log` View entirely (so that it also doesn't end up in CSV). I don't think we should do this because it would severely reduce the usefulness of log Views (think about looking at a list of harvest logs, input logs, etc and being able to see the quantity at a glance).
2. We override the "CSV Export" display and remove the quantity column from that, but keep it in the default View config. This also feels strange to me... because I would expect clicking "Export CSV" from the log View will give me a CSV that matches what I'm looking at. It feels arbitrary to just have the quantity missing. (begs the question "why?")

Neither option can really be justified, IMO.

> I must admit that after seeing all of this it feels like a lot of complexity just to enable exporting a single quantity for logs...

Yeaaa like I said above (with emphasis added):

> I think ideally we would be able to target the `quantity` field in our Views specifically, and fix the specific issue at hand, rather than removing all newlines everywhere. I'm not sure if there's an easy way to do this, however, **without adding undo complexity**.

All things considered, it actually involved less code than I anticipated - but I agree it still feels like a hack to fix this one thing. I don't love it, but I feel like I've exhausted the options available.

mstenta:
Looked at this together with @paul121 @symbioquine and @Farmer-Eds-Shed on the dev call today and we decided to go ahead and merge this, despite the bit of cruft it adds.

The justification is basically that this is the best way to "polish" the existing functionality, but that there may be a bigger question about whether or not we should simply remove the quantity from log CSV exports entirely, to encourage users to export those by other means.

This discussion evolved into this new idea, to make it clear to users that CSVs are inherently limited: https://www.drupal.org/project/farm/issues/3328886

Ultimately we decided that the cruft necessary for polishing the current functionality is isolated enough that we could potentially remove it in the future if we take a different approach. But for now it makes sense to include this ahead of the 2.0.0 release.

mstenta:
Came up with a last-minute idea in chat w/ @paul121 and @symbioquine, which cleans this up even more. Full chat log here: https://irc.farmos.org/bot/log/farmOS/2022-12-22#T86498

Tl;dr:

Rather than creating a new `farm_csv` module, with a `csv` Quantity view mode, and a hook to add a `#post_render` callback to quantities to strip tags and whitespace, now we are providing a `plain_text` Quantity view module directly in the `quantity` module, and extending the `EntityViewBuilder` class to add the `#post_render` callback in the same class, so "Plain text" it is a core feature of Quantity entities everywhere. And then we simply use that in CSV exports.

This feels much better, and resolves the "cruft" that the previous approach created. It is essentially the same logic, but simply renaming the view mode and moving it to the `quantity` module made all the difference.

Thanks again to @paul121 and @symbioquine for helping think this through. I didn't expect this one to be so complicated! :smile:

PR #623, opened 2022-12-20
Make sure log asset != log location in circular location constraint logic

mstenta:
Follow-up to #568 to check for an even simpler circular relationship scenario, where a movement log is created with the same asset in both the `asset` and `location` log fields. Also updated automated test to check for this.


mstenta:
I'm not going to add a `CHANGELOG.md` line for this, since it really should have been part of #568 in the first place, and that one is already covered in the `CHANGELOG.md`.

paul121 (approved):

PR #627, opened 2023-01-17
Replace all usages of docker-compose with native docker compose

penyaskito:
docker-compose up -d gives an error, docker compose up -d works though.

### DIFF
--- a/.github/workflows/deliver.yml
+++ b/.github/workflows/deliver.yml
@@ -118,17 +118,17 @@ jobs:
           DB_URL: ${{ matrix.DB_URL }}
         run: |
           cp docker/docker-compose.testing.* .
-          docker-compose -f docker-compose.testing.common.yml -f docker-compose.testing.${{ matrix.dbms }}.yml config > docker-compose.yml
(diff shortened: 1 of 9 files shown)


mstenta:
Oh right - I think that is the new way of doing things, eh? Has Docker officially deprecated `docker-compose` in favor of `docker compose`? I think I heard that was coming...

If so, there may be other places in our docs that need updating too. I will `grep` and maybe add some additional commits. Thanks for getting this started @penyaskito!

symbioquine:
This link should probably be in the commit message: https://www.docker.com/blog/announcing-compose-v2-general-availability/

mstenta:
I added two commits (ignore my first force-push that merged them into one commit... decided to keep them separate).

The first replaces ALL usages of `docker-compose` with `docker compose` everywhere in the codebase. Second commit just adds a CHANGELOG.md entry.

> This link should probably be in the commit message: https://www.docker.com/blog/announcing-compose-v2-general-availability/

@symbioquine Once we have all this settled I will add that to the merge commit that I create.

symbioquine (approved):

mstenta:
Merged! https://github.com/farmOS/farmOS/commit/92281a373eeebb984e102a8fe9029f0b7133767f

Thanks again @penyaskito :-)

PR #629, opened 2023-01-20
Issue #3335267: 405 Method Not Allowed when trying to patch entities as a user with non-default language

mstenta:
See https://www.drupal.org/project/farm/issues/3335267

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [Add folder creation and clear caches to Configure private filesystem in the Hosting and Environment docs #628](https://github.com/farmOS/farmOS/pull/628)
 - [Issue #3336698: Add "project: farm" to farm.info.yml to fix drupal.org usage statistics](https://www.drupal.org/project/farm/issues/3336698)
 - [Fix type error when un-assigning asset from group #631](https://github.com/farmOS/farmOS/pull/631)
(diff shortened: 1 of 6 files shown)


symbioquine (approved):
My only nits to pick here are:

* Is it worth adding a test for this?
* I'd personally put more of the "Proposed resolution" description from the Drupal issue into the commit message itself.

Otherwise, LGTM - pending manual testing by @Farmer-Eds-Shed

mstenta:
> Is it worth adding a test for this?

Yes it probably is. I'll set the status to "Needs work" on the d.o issue and mention that...

> I'd personally put more of the "Proposed resolution" description from the Drupal issue into the commit message itself.

Makes sense. I generally think that as long as the issue number is referenced in the commit message, that is the best link back to the full context and reasoning. That covers `git blame` forensics. But even `git blame` history can become lost/fragmented over time as things change or move around, so in this case I thought it worth also linking back to the original issue in a code comment directly:

https://github.com/farmOS/farmOS/pull/629/files#diff-3846dce09e5e8e499c65c7337a81d86c0edd9535592c6d023e28555e7e55e588R36

But that's just my convention. :-)

mstenta:
FWIW I probably won't have time to write tests for this right away... so if anyone wants to pick that up feel free! I'll post an update when I'm able to, if no one else has in the meantime.

Farmer-Eds-Shed:
Looks good,
I tried a few different log an asset types with no issues.

mstenta:
Force pushed this branch with one additional commit before the original commit, which adds kernel tests that demonstrate the issue. The tests fail without the fix commit.

Note that this depends on #638, so it includes all of the commits from that branch too. We should get that merged first, and then review this.

mstenta:
Reviewed with @paul121 - this is ready to merge.

PR #630, opened 2023-01-20
Change asset action date pickers to use HTML5 calendar widgets

### LINKED ISSUE #509: Bulk actions using different date format
When selecting multiple assets from a list and applying bulk actions (for example at the bottom action-> group assets) the simple form does not have a place for time. It also is using the older style date picker selecting year, month, and day separately as apposed to selecting from a calendar. For consistency it would be nice to see the date/time selection to be the same across all input locations.

erlarese:
Fixes #509 - this changes the date picker in the "move asset" and "group asset" panels to use the calendar selector widget from the browser.  This makes these panels more consistent with date selection in other areas of the UI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - [Move land and structure type fields higher up for consistency #632](https://github.com/farmOS/farmOS/pull/632)
+- [Change asset action date pickers to use HTML5 calendar widgets #630](https://github.com/farmOS/farmOS/pull/630)
 
 ### Fixed
 
--- a/modules/asset/group/src/Form/AssetGroupActionForm.php
+++ b/modules/asset/group/src/Form/AssetGroupActionForm.php
@@ -135,12 +135,11 @@ class AssetGroupActionForm extends ConfirmFormBase {
         ->toString());
     }
 
+    $date = new DrupalDateTime();
     $form['date'] = [
-      '#type' => 'datelist',
+      '#type' => 'date',
       '#title' => $this->t('Date'),
-      '#default_value' => new DrupalDateTime(),
-      '#date_part_order' => ['year', 'month', 'day'],
-      '#date_year_range' => '-15:+15',
+      '#default_value' => date('Y-m-d', $date->getTimestamp()),
       '#required' => TRUE,
     ];
 
@@ -197,8 +196,6 @@ class AssetGroupActionForm extends ConfirmFormBase {
         $groups = $this->entityTypeManager->getStorage('asset')->loadMultiple($group_ids);
       }
 
-      /** @var \Drupal\Core\Datetime\DrupalDateTime $date */
-      $date = $form_state->getValue('date');
       $done = (bool) $form_state->getValue('done', FALSE);
 
       // Generate a name for the log.
@@ -210,7 +207,7 @@ class AssetGroupActionForm extends ConfirmFormBase {
       $log = Log::create([
         'name' => $log_name,
         'type' => 'observation',
-        'timestamp' => $date->getTimestamp(),
+        'timestamp' => strtotime($form_state->getValue('date')),
         'asset' => $accessible_entities,
         'is_group_assignment' => TRUE,
         'group' => $groups,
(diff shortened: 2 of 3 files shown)


mstenta:
I merged your PHPStan fix commit into `2.x` as a first step: https://github.com/farmOS/farmOS/commit/69f12631e23e9dd6e268344ea64fbc069d65beb2

Then I rebased the rest on top of that. I merged the `CHANGELOG.md` commit into the first one and titled it to match this PR.

Automated tests passed, so I just need to manually test it and it's good to go! Will do that later this afternoon hopefully. :-)

Thanks for the work on this @erlarese and your patience in the review process!

mstenta:
Hmm this doesn't seem to be working.

I see two issues (which are related I think):

1. The `#default_value` no longer populates the field with the current date.
2. Pressing the submit button shows the following error (eg: with the "Move" action):

```
Error: Call to a member function getTimestamp() on string in Drupal\farm_location\Form\AssetMoveActionForm->submitForm() (line 241 of profiles/farm/modules/core/location/src/Form/AssetMoveActionForm.php). 
```

The same thing happened when I was testing this in the Log module (the asset actions are based off of that, so that makes sense).

@erlarese I will figure out the proper solution in Log, and then hand this back over to you to replicate the changes here if you want.

mstenta:
I reworked the `#default_value` to be a string in the format `Y-m-d`, and the submit logic to parse that back into a timestamp using `strtotime()`. Both actions work now.

paul121 (commented):
I've tested and confirmed this does work :+1: 

However, looking into this issue a bit more I have a couple thoughts.

1. I think we should be using the `datetime` type here with the `'#date_time_element' => 'none'` option. `datetime` uses `date` itself, but seems to have additional logic for specifying a correct HTML5 date format, so this might be more robust in supporting HTML5 date inputs too? Documented here: https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Datetime%21Element%21Datetime.php/function/Datetime%3A%3AprocessDatetime/8.3.x

2. Another reason for `datetime` is because it uses the `DrupalDateTime` object as both the default value and submitted value. It is also more flexible because it can allow a `time` input as well. These are important considerations if someone wanted to alter this form to include time - using `datetime` would make that possible, and allow the form submission to rely on a standardized return type of `DrupalDateTime` that covers many use-cases, not only `Y-m-d`.

3. ** I'm not suggesting people alter these particular forms... Basically, I think this is good to think about for farmOS more generally. If possible we should standardize on a single form element for date/time input across farmOS.**

4. Looking back to the original #509 I'm curious *if we can* support time in these forms too, as that was a part of the feature request. The `datelist` element made entering a time very cumbersome. I think this is much better with the HTML5 inputs. Can we just enable both?

paul121:
Perhaps we should go ahead and merge this, it is a good improvement, a good contribution from @erlarese !

We could have a follow up change to use `datetime` more thoroughly afterwards and review other usages in the process.

mstenta:
Thanks for reviewing and suggesting `datetime` + `'#date_time_element' => 'none'` @paul121! I wasn't aware of that option. It seems like the better choice - the `date('Y-m-d')` + `strtotime()` felt weird to me in these changes.

> Perhaps we should go ahead and merge this, it is a good improvement, a good contribution from @erlarese !

Generally I would agree - but since we are also making a related change in the Log module (https://www.drupal.org/project/log/issues/3336721), which requires tagging a new version, it seems like we might as well do it altogether so we don't have to handle both again in a followup.

I will give this a try in both projects soon...

> I think this is good to think about for farmOS more generally. If possible we should standardize on a single form element for date/time input across farmOS.

I think I `grep`ed the codebase when I was looking at this earlier for `'#type' => 'date'` and there was only one other place: the planting quick form. So maybe I'll make that change at the same time for consistency. Then moving forward we have a standard we can try to stick to.

mstenta:
> it seems like we might as well do it altogether so we don't have to handle both again in a followup.

OK so looking at all of this... I think what I'll do is test out the `datetime` approach you described @paul121 in the Log module, get that merged. Then merge this. Then followup with a PR to change these action forms AND the planting quick form to `datetime`.

mstenta:
Merged - thanks again @erlarese and @paul121!

PR #631, opened 2023-01-31
Fix type error when un-assigning asset from group

paul121:
Submitting the group asset action form without a group yields the following error:

> Message TypeError: array_column(): Argument 1 ($array) must be of type array, null given in array_column() (line 195 of /var/www/html/web/profiles/farm/modules/asset/group/src/Form/AssetGroupActionForm.php)

One fix is to use the null coalescing operator to return an empty array when the form value is null.

We made this change in the move action form as well: https://github.com/farmOS/farmOS/commit/6d13a732d704c0312f4106871948af724ed4931d


mstenta:
Thanks @paul121!

PR #632, opened 2023-02-01
Move land and structure type fields higher up for consistency

paul121:
Currently the land and structure type field come after status, flags and owners. This is different than animal and plant assets whose type field comes immediately after the asset name. Since the land and structure type are required fields it seems more appropriate to have them included higher in the form.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- [Move land and structure type fields higher up for consistency #632](https://github.com/farmOS/farmOS/pull/632)
+
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


paul121:
I noticed this while demoing farmOS at EcoFarm. Because the land and structure type are lower down in the form it makes demoing this a bit awkward... "So here you can create a new land asset. You can also add a status, flag and owners just like on all asset types.  **scroll down screen* Oh, yeah, this is a land asset, so it has a required land type, with these options...".

It just seems more natural to include these fields higher up. The land and structure type also are not visible on a tablet until you scroll down:

Screenshot from 2023-02-01 15-16-13



Compare the current animal and land asset forms:

Screenshot from 2023-01-31 11-01-32
Screenshot from 2023-01-31 11-00-12

mstenta (approved):

mstenta:
Makes good sense to me! Thanks @paul121!

PR #634, opened 2023-02-04
Move default entity tabs to secondary tabs

paul121:
I believe we are starting to push the limits of how many primary tabs (technically called primary "local tasks") are being displayed on some of our entity pages. As such, I propose we move the "default" tabs (View, Edit and Revisions) into secondary tabs under a common primary tab for the entity. A few reasons for this:

Note: I originally stumbled upon this idea when thinking about using contextual links for edit/revision/delete actions: https://farmos.discourse.group/t/use-contextual-links-for-entity-edit-revision-and-delete-links/1227

1. Additional tabs will continue to be out of context with the default tabs that directly relate to viewing + editing fields about the entity. Additional tabs are often used to display information related to the entity (eg: logs referencing an asset). Moving the default tabs under a common primary tab keeps this "entity context" in one place.
2. We can improve the context of a given page (eg: "What am I looking at?") by improving the name of this proposed "common primary tab" to be 1) the entity type label (eg: "Asset") or 2) the entity bundle (eg: "Animal").
3. There is a limited amount of space for primary tabs. Moving the default tabs to secondary tabs creates space for more primary tabs.

Right now we only have a few "additional tabs" that we add to entity pages but I foresee more being added overtime to fulfill more specific use-cases (especially on plans!). Here is a quick overview of current tabs I can recall:

Logs:
- View, edit, revisions (default)

Assets:
- View, edit, revisions (default)
- Logs (shows logs referencing an asset, visible on all asset types)
- Assets (shows assets located within an asset, visible on location assets)
- Locations (shows children locations, visible on location assets that have children)

Taxonomy terms:
- View, edit, delete (default)
- Assets (shows assets that have a field referencing this term, eg: plant + plant type)
- Logs (shows logs that have field a field referencing this term, eg: seeding log + plant type)

Plans:
- View, edit, revisions (default)
- farmOS core does not have any plan types. However, I believe many plan types will use additional tabs to display specialized UIs and relationships.

This means that right now location assets can have up to 3 tabs (logs, assets and locations), taxonomy terms can have 2 tabs (plant types have both asset and logs tabs).

An important aspect of this decision is that the placement on these default tabs is "universal" across all bundles of the entity. For example, we cannot only move these default tabs for land assets, they must be moved for all asset types. I'm currently reaching this conundrum in the Rothamsted experiment plan I'm working on. It currently has 3 additional tabs and I would like to add a couple more. I could move these default tabs for plans in the Rothamsted module... but as soon as they would like to enable additional plan types, they would be affected as well.

We can also decide which entity types to make this change on. At minimum, I'd like to see this implemented for plans. But I think there is value in implementing this for assets and *taxonomy terms, and even logs as well. I think it's forward thinking to "create the space" for additional primary tabs before other people/modules/use-cases need them and the UI is not well-suited to add them.


paul121:
Wanted to open a PR to show the code. I've also copied this PR description into a forum post for further discussion: https://farmos.discourse.group/t/move-default-entity-tabs-to-secondary-tabs/1508

mstenta:
Rebased onto latest 2.x ahead of v2.1.0.

mstenta:
One thing I'm noticing in testing this is that Logs *only* have secondary tabs now:

Screenshot from 2023-05-11 14-58-33

Presumably this is because they only have a single primary tab, and the primary tabs are only shown if there are more than one.

I suppose this is OK. Just noting it. If additional tabs ever get added to Logs then this will work consistently with others.

mstenta:
Ah I see you noted this already in the forum @paul121 : https://farmos.discourse.group/t/move-default-entity-tabs-to-secondary-tabs/1508/2?u=mstenta

paul121:
I've fixed the test failure after collapsing secondary tabs... two issues:

1. First, the tests are currently using the `classy` theme which is deprecated and removed from D10.

Screenshot from 2023-05-17 15-42-17

2. I believe `classy` was originally set so that we could test those class names for menu local tasks. I'm not quite sure why the default `stark` theme doesn't work... but anyways, with this PR I'm adding a template that uses the gin/claro class names, so it no longer matters which theme is set, it will use the class names from the template

mstenta:
We moved the Gin-specific stuff from `farm_ui_menu` to `farm_ui_theme` on the dev call today, which resolved the issue with tests in the `farm_ui_views` module, and @paul121 opened a follow-up PR to resolve that separately: https://github.com/farmOS/farmOS/pull/680

Thanks again for this @paul121! It it's a good UX improvement!

PR #636, opened 2023-02-05
Remove time service from planting quick form

mstenta:
#635 changed the way that `#default_value` was populated for the date fields in the Planting quick form. The `time.datetime` service is no longer used. This PR removes it from the Planting quick form class's injected dependencies.

I opened this as a follow-up because there could be an argument that changing this class's signature is a breaking change. If someone were extending this quick form to create their own, they might need to update their class signature accordingly, otherwise an error would occur.

This is a very small one - and probably a very small chance of any issue. But it brings up a broader question, perhaps, about what we consider breaking. Curious to hear thoughts on this, and whether or not we should merge this.

### DIFF
--- a/modules/quick/planting/src/Plugin/QuickForm/Planting.php
+++ b/modules/quick/planting/src/Plugin/QuickForm/Planting.php
@@ -29,6 +29,8 @@ use Psr\Container\ContainerInterface;
  *     "create plant asset",
  *   }
  * )
+ *
+ * @internal
  */
 class Planting extends QuickFormBase {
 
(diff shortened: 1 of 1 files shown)


paul121 (approved):
I agree, this one is probably OK. Maybe we should consider making it `internal` at the same time? Maybe we have an existing issue open for this?

mstenta:
> Maybe we should consider making it `internal` at the same time?

I thought about suggesting this as well.

mstenta:
Added a commit that adds `@internal` to the planting quick form: https://github.com/farmOS/farmOS/pull/636/commits/baf294d3bd091adafd790380fdcad15181016cf0

PR #638, opened 2023-02-07
Add farmOS API kernel tests

mstenta:
This adds some basic automated kernel tests to the `farm_api` module. They test the following:

- The root API path is `/api`.
- farmOS meta info is added to `/api` in the `meta.farm` section.
- Creating an asset.
- Creating a log that references the asset.
- Retrieving individual assets and logs.
- Retrieving collections of assets and logs.
- Updating assets and logs.
- Deleting assets and logs.

My motivation for adding these tests started here: https://github.com/farmOS/farmOS/pull/629. I want to be able to test the fix for that issue in a kernel test, but we didn't have any existing API kernel tests, so I figured I would add some basic ones, and then build off of them for the `l10n` tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- [Add farmOS API kernel tests #638](https://github.com/farmOS/farmOS/pull/638)
+
 ### Changed
 
 - [Move land and structure type fields higher up for consistency #632](https://github.com/farmOS/farmOS/pull/632)
--- a/modules/core/api/tests/modules/farm_api_test/config/install/asset.type.test.yml
+++ b/modules/core/api/tests/modules/farm_api_test/config/install/asset.type.test.yml
@@ -1,0 +1,7 @@
+langcode: en
+status: true
+id: test
+label: Test
+description: ''
+workflow: asset_default
+new_revision: true
--- a/modules/core/api/tests/modules/farm_api_test/config/install/log.type.test.yml
+++ b/modules/core/api/tests/modules/farm_api_test/config/install/log.type.test.yml
@@ -1,0 +1,8 @@
+langcode: en
+status: true
+id: test
+label: Test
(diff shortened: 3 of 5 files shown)


mstenta:
> farmOS meta info is added to `/api` in the `meta.farm` section.

There is some overlap here with our existing `EntryPointTest` (which is a functional test): https://github.com/farmOS/farmOS/blob/2.x/modules/core/api/tests/src/Functional/EntryPointTest.php

But I figure it doesn't hurt to have a basic check in the kernel test too.

mstenta:
(Pushed a small codesniffer fix.)

mstenta:
(...aaaaand a PHPStan fix :laughing:)

mstenta:
Ah need to make a slight adjustment to my test so it passes for branches other than `2.x`...

mstenta:
I just removed the failing test - it isn't really necessary to test every part of the `meta.farm` information because that is covered more thoroughly by the functional test anyway.

paul121 (approved):
This LGTM!

I do have the slightest worry that a kernel test might miss more obscure bugs a functional test would catch with "all" modules enabled... the only thing I can think that might be important to include that is missing here is the `entity` module. But these tests are quite simple (not using many fields on asset or logs) so its probably OK.

This is probably something that should be considered by other modules if they choose to extend this test class. It would be nice to support "all use cases" from the start but I think that might be hard with kernel tests

mstenta:
> I do have the slightest worry that a kernel test might miss more obscure bugs a functional test would catch with "all" modules enabled...

Yes I totally agree. My thought is it provides a basis to start from, and we can add to it if/when we discover bugs that these tests are missing. But at the very least it provides a safety net for the very broad strokes.

> the only thing I can think that might be important to include that is missing here is the `entity` module. But these tests are quite simple (not using many fields on asset or logs) so its probably OK.

Glad you mention this! My followup PR #629 for localization-enabled API testing actually adds `entity` and `farm_entity` to these tests, because I found they were necessary to replicate the issue that PR addresses. So that's a good example of what I just described above ("we can add to it if/when we discover bugs that these tests are missing").

PR #640, opened 2023-02-10
Convert geometry values to WKT in GeofieldWidget

### LINKED ISSUE #639: Geometry fields allow raw KML
**Describe the bug**

It is possible to paste raw KML (not WKT as expected) in the text area below a map (eg: in a land asset), and it gets accepted as valid when it shouldn't.

**To Reproduce**
1. Go to /asset/add/land
2. Paste KML into the textarea below the "intrinsic geometry" map.
3. Save the asset.
4. Observe that you are allowed to save it.

Notably, the View tab of the asset shows the correct geometry in the map (!) which is odd, but then if you click "Edit" again, it does NOT show in the map of the edit form. And if you create a log that references this land asset in the `location` field, you get an error when it tries to copy the geometry from the asset to the log.

**Expected behavior**

A validation warning should be displayed notifying the user that only WKT can be pasted into the input.

**Screenshots**

Here is a screenshot of the KML in the Edit screen - notice that it does NOT show a geometry (this was after it was already saved and I clicked "Edit" again):

Screenshot 2023-02-07 at 14-39-35 Edit Test invalid geometry Test farmOS 2 x

And here is a screenshot of the View page - notice that it shows the correct geometry:

Screenshot 2023-02-07 at 14-39-11 Test invalid geometry Test farmOS 2 x

Here is the KML I tested with:

```
<?xml version="1.0" encoding="UTF-8"?>
<kml xmlns="http://earth.google.com/kml/2.1">
  <Document>
    <Placemark id="ab7dcf9b-9617-44f3-b809-b4a6e52c59b6">
      <Polygon>
        <outerBoundaryIs>
          <LinearRing>
            <coordinates>-58.973305281048,37.35047489177 -58.852934104224,26.445056162378 -39.593545812412,25.253487002766 -39.352803458765,38.489916251467 -58.973305281048,37.35047489177</coordinates>
          </LinearRing>
        </outerBoundaryIs>
      </Polygon>
      <name>Test invalid geometry</name>
      <entity_type>asset</entity_type>
      <bundle>land</bundle>
      <internal_id>402</internal_id>
    </Placemark>
  </Document>
</kml>
```

### LINKED ISSUE #693: Filter option allow reset after apply with no input
Hi, I noticed that all filter options, when I clicked Apply without any additional filter criteria, still allow me to click Reset. I believe it should be addressed to prevent confusion when using filter options.

mstenta:
Fixes #639  - see that issue for full context and decision process.

Note that this includes two additional commits ahead of the final directly relevant commit:

> Fix config schema for field.widget.settings.farm_map_geofield.

This fixes an outdated setting name in `farm_map.schemal.yml` and adds a missing one. This was discovered when I was writing an automated test.

> Add an automated test for the farmOS Geofield widget. 

This adds an automated functional test for our custom Geofield widget. This is copied and minimally modified from the upstream Geofield module's widget tests. Note that this commit ONLY tests WKT data. The next (and final) commit adds tests for GeoJSON and WKB (also copied from upstream module's test) and fixes the bug described in #639

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- [Convert geometry values to WKT in GeofieldWidget #640](https://github.com/farmOS/farmOS/pull/640)
+
 ## [2.0.1] 2023-02-08
 
 ### Added
--- a/modules/core/map/config/schema/farm_map.schema.yml
+++ b/modules/core/map/config/schema/farm_map.schema.yml
@@ -67,7 +67,10 @@ field.widget.settings.farm_map_geofield:
   type: mapping
   label: 'Farm map geofield settings'
   mapping:
-    geocode_file_field:
+    display_raw_geometry:
+      type: boolean
+      label: 'Whether or not to display the raw geometry field underneath map.'
+    populate_file_field:
       type: string
       label: 'The name of the file field to geocode into the geofield.'
 block.settings.map_block:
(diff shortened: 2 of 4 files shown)


paul121 (approved):
This LGTM!

PR #644, opened 2023-02-25
Breadcrumb improvements

paul121:
We should be implementing `FarmBreadcrumbBuilder::applies()` - it defaults to TRUE which means we may be building breadcrumbs for all routes! Since we inherit the default `PathBasedBreadcrumbBuilder` this probably isn't causing any issues. However I think this could cause issues if another breadcrumb builder was trying to do something on another route and had a lower priority.

Also adding a simple "People" breadcrumb on user pages:

Screenshot from 2023-02-24 17-10-51


mstenta:
Thanks @paul121! Rebased and merged.

PR #645, opened 2023-02-27
Improve farmOS version guidance in install/update docs

ghost:


mstenta:
Sorry for the delay @lordeddi - I saw this and then forgot it!

I made a few small adjustments (described below) and merged it. Thanks again!!

Adjustments:

- Moved the markdown link URLs inline for consistency with the rest of our docs.
- I dropped the change to `docs/development/environment/index.md` because dev environments use `2.x`, so there's no need to set a specific version.
- I moved the change to `install.md` down to the relevant section of the guide (where `2.x.y` is mentioned) and added a bit more specific guidance, along with a note that `latest` is not recommended.

PR #650, opened 2023-03-08
Create a QuickFormTestBase class that other quick form tests can extend from

mstenta:
I made this last year when I was starting work on a v2 Birth quick form. Dusting that stuff off again now, and figured that this commit could be merged by itself first.

The idea is that it provides a base class for modules to use so they can write their own quick form kernel tests without replicating a whole bunch of boilerplate stuff. I figure we can add more to it over time too to make quick form testing more standardized.

This also refactors the Planting quick form test to use the new base class.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - [Document building farmOS with Composer #648](https://github.com/farmOS/farmOS/pull/648)
+- [Create a QuickFormTestBase class that other quick form tests can extend from #650](https://github.com/farmOS/farmOS/pull/650)
 
 ### Fixed
 
--- a/modules/core/quick/tests/src/Kernel/QuickFormTestBase.php
+++ b/modules/core/quick/tests/src/Kernel/QuickFormTestBase.php
@@ -1,0 +1,126 @@
+<?php
+
+namespace Drupal\Tests\farm_quick\Kernel;
+
+use Drupal\Core\Form\FormState;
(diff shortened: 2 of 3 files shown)


paul121 (approved):
This LGTM!

Trying to think of implications/considerations of this.. don't have much. I guess going forward we will need to consider if changes to the abstract `QuickFormTestBase` would break downstream uses. As an example, I imagine that installing an entity schema or config twice could be bad. I like that this only includes the core entity type and no bundles.

Along those lines, maybe we should include the base `plan` module + entity schema here as well (would there be anything else needed to support plans?) I don't have any quick forms that really require a plan at this moment but I imagine we may have some by the end of 2023. An implementing module could do it themselves while extending this class. But also seems like it wouldn't hurt to include that here, although it is not necessary right now.

mstenta:
> Trying to think of implications/considerations of this.. don't have much. I guess going forward we will need to consider if changes to the abstract `QuickFormTestBase` would break downstream uses.

Yup. We could mark it as `@internal` if we want to just use it for our own tests (and contrib can use it at their own risk). Maybe not a bad idea...

> Along those lines, maybe we should include the base plan module + entity schema here as well (would there be anything else needed to support plans?) I don't have any quick forms that really require a plan at this moment but I imagine we may have some by the end of 2023. An implementing module could do it themselves while extending this class. But also seems like it wouldn't hurt to include that here, although it is not necessary right now.

I'd rather keep it as-is for now. We can consider adding plan stuff later, but I have a feeling that MOST of the time it will make more sense to do that in the downstream test... most quick forms won't have plan integrations... at least not for a long time.

mstenta:
> We could mark it as `@internal`

Amended my commit to add this.

PR #654, opened 2023-03-11
Add a QuickFormElementsTrait with a buildInlineContainer() method

mstenta:
This adds a `farm_quick/quick` CSS library to quick forms. The thought is that this library can be used for common CSS styles used in quick forms over time.

As a first step, the only style this adds is an `inline` class that is helpful for rendering form elements inline.

The second commit refactors the Planting quick form to use the `inline` style for its quantity fields.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [Include Fraction bundle fields in default Views #664](https://github.com/farmOS/farmOS/pull/664)
 - [Allow map to be resized vertically #663](https://github.com/farmOS/farmOS/pull/663)
 - [Add integer, decimal, and email field support to field factory service #666](https://github.com/farmOS/farmOS/pull/666)
+- [Add a QuickFormElementsTrait with a buildInlineContainer() method #654](https://github.com/farmOS/farmOS/pull/654)
 
 ### Changed
 
--- a/modules/core/quick/css/quick.css
+++ b/modules/core/quick/css/quick.css
@@ -1,0 +1,5 @@
+.inline {
+  display: flex;
+  flex-wrap: wrap;
+  column-gap: 2em;
+}
--- a/modules/core/quick/farm_quick.libraries.yml
+++ b/modules/core/quick/farm_quick.libraries.yml
@@ -1,4 +1,0 @@
-quick:
-  css:
-    theme:
-      css/quick.css: { }
--- a/modules/core/quick/farm_quick.libraries.yml
+++ b/modules/core/quick/farm_quick.libraries.yml
@@ -1,0 +1,4 @@
+quick:
+  css:
+    theme:
+      css/quick.css: { }
--- a/modules/core/quick/src/Form/QuickForm.php
+++ b/modules/core/quick/src/Form/QuickForm.php
@@ -116,6 +116,9 @@ class QuickForm extends FormBase implements BaseFormIdInterface {
       ];
     }
 
+    // Attach quick form CSS/JS.
+    $form['#attached']['library'][] = 'farm_quick/quick';
+
     return $form;
   }
 
--- a/modules/core/quick/src/Form/QuickForm.php
+++ b/modules/core/quick/src/Form/QuickForm.php
@@ -116,9 +116,6 @@ class QuickForm extends FormBase implements BaseFormIdInterface {
       ];
     }
 
-    // Attach quick form CSS/JS.
-    $form['#attached']['library'][] = 'farm_quick/quick';
-
     return $form;
   }
 
--- a/modules/core/quick/src/Traits/QuickFormElementsTrait.php
+++ b/modules/core/quick/src/Traits/QuickFormElementsTrait.php
@@ -1,0 +1,27 @@
+<?php
+
+namespace Drupal\farm_quick\Traits;
+
+/**
+ * Provides methods for building common quick form elements.
+ */
+trait QuickFormElementsTrait {
+
+  /**
+   * Build an inline container element.
+   *
+   * @return array
+   *   Returns a render array.
+   */
+  public function buildInlineContainer() {
+    return [
+      '#type' => 'container',
+      '#attributes' => [
+        'class' => [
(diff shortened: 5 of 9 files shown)


mstenta:
Here are before and after screenshots of the planting quick form to demonstrate the changes to quantity fields:

**Before**

Screenshot 2023-03-11 at 11-40-43 Planting Drush Site-Install

**After**

Screenshot 2023-03-11 at 11-41-30 Planting Drush Site-Install

mstenta:
We discussed this on one of the last dev calls. The idea of creating a special `inline-container` element was proposed instead. I looked into that and it felt like overkill (see chat logs below for more details). In chat @paul121 pointed to a strategy he's using in the Rothamsted module, which provides a helper function that encapsulates the boilerplate styles: https://github.com/Rothamsted-Ecoinformatics/farm_rothamsted/blob/43563c906d6ccac16f3ac7b307306ec65f6c6bc2/modules/farm_rothamsted_quick/src/Plugin/QuickForm/QuickExperimentFormBase.php#L1037-L1050

Chat logs: https://irc.farmos.org/bot/log/farmOS/2023-03-16#T88931

mstenta:
I pushed a commit that moves the `quick.css` library from the `farm_quick` module to the `farm_ui_theme` module, since it is presentation-layer details. The idea being that the theme layer should be responsible for deciding what an "inline container" looks like. This may require different CSS depending on the theme, so it shouldn't be part of the quick form module.

I also renamed the CSS class from `inline` to `inline-container` to be more specific. And I added a `quick` class to the `<form>` element of quick forms, so the CSS rule can specifically target inline containers in quick forms with `form.quick .inline-container {...}`. This makes these changes very targeted to quick forms. If we want to make them available more generally, we can consider that as a next step, but this keeps the considerations simpler for now.

I started to implement a `QuickFormElementsTrait` with a `buildInlineContainer()` method, but backed out of it for two reasons:

1. With the styles moved to `farm_ui_theme`, the only thing the `buildInlineContainer()` method would be encapsulating is setting of the `inline-container` class on the element. This doesn't seem like it is worth a helper. Developers who are building quick forms can designate any `container` element as inline by adding an `inline-container` class to it. The theme layer (eg: `farm_ui_theme`) is then responsible for deciding what that means.
2. The Planting quick form code would need to be refactored to use a `buildInlineContainer()` method, because of the way it builds the log form arrays currently. See: https://github.com/farmOS/farmOS/blob/f68887cc454009ec7853535a68ce09fabc343008/modules/quick/planting/src/Plugin/QuickForm/Planting.php#L358

I figure we can consider adding a `QuickFormElementsTrait` again in the future, if there are additional helper methods needed, and we can reconsider whether or not it makes sense to `buildInlineContainer()` at that time to further encapsulate the `inline-container` class. In the meantime, adding a CSS class seems sufficient to me.

mstenta:
> And I added a `quick` class to the `<form>` element of quick forms, so the CSS rule can specifically target inline containers in quick forms with `form.quick .inline-container {...}`

In testing I realized we already have a `quick-form` class on the `<form>` element, so I removed the extra `quick` class that I had added and adapted the CSS rule to be `form.quick-form .inline-container {...}`, and then force-pushed.

symbioquine (approved):
LGTM - I like how the new inline class is scoped to only apply to quick forms!

paul121 (commented):
TLDR; This looks good and I don't want to block this PR. I'm glad we're starting small and experimenting with this in quick forms. But I do think there will be more general uses for this flavor of inline container and as such it would be smart have an abstraction to make this implementation less fragile to future changes.

> I also renamed the CSS class from inline to inline-container to be more specific. 

Maybe something to consider... Drupal core already has a `container-inline` class. So this naming could be confusing. As I described in chat this `container-inline` class won't work for our needs specifically because it makes the field labels + descriptions inline, while we want to only make the broader field "blocks" inline. I think this goes to show that a desired "inline container" does not always mean the same thing. Perhaps we can be even more specific, something like : `form-container-inline`, if not `quickform-container-inline`.

> With the styles moved to farm_ui_theme, the only thing the buildInlineContainer() method would be encapsulating is setting of the inline-container class on the element. This doesn't seem like it is worth a helper. 

I agree that this isn't a huge "helper", it actually creates a bit more work to implement. But the main thing I want to advocate with `buildInlineContainer() ` is some kind of abstraction for the inline container beyond only the `inline-container` CSS class.  An abstraction would allow us to change the inline container implementation in the future without requiring quick forms to modify their code. This may be thinking far ahead, but it does seem that this inline container implementation could be applicable more generally as well. It's hard to say how & where this could be implemented (drupal core, gin or only farmOS?) but having this abstraction would allow for an easier upgrade path.

Actually here is a Drupal core issue: Provide API for inline containers https://www.drupal.org/project/drupal/issues/3059593 which really only notes two things, but is basically where I have come to conclusion of all this:

> Currently containers are marked as inline containers by applying container-inline class. While this might work fine for some use cases, there are use cases where other changes might have to be applied to the markup based on whether it's displayed inline.

Yes - there is a need for different types of inline containers!

> Instead of using the container-inline class as an API, provide explicit API for marking containers inline.

Ultimately I think that render elements + templates are the correct way to build this abstraction/API and let themes add CSS styles and override templates as needed. A CSS class may be sufficient for our current theme, but in theory a theme could want to render inline containers with a different element or logic within a template.

Obviously all of this leaves us with a lot "unknowns" - having an abstraction method gives us flexibility to adopt these in the future

mstenta:
These are all great considerations @paul121.

I pushed two new commits:

- Reformat Planting quick form log field info arrays. No functional change. (To set the stage for the next commit.)
-  Abstract inline container structure into new QuickFormElementsTrait::buildInlineContainer() trait+method.

I renamed this PR to "Add a QuickFormElementsTrait with a buildInlineContainer() method" and updated `CHANGELOG.md` accordingly.

@paul121 what do you think?

mstenta:
(Force-pushed to fix a minor PHP Codesniffer complaint.)

paul121 (approved):
Nice LGTM! I really do hope/believe this will be useful one day in the future :smile:

symbioquine (approved):

PR #656, opened 2023-03-11
Birth quick form (ported from farmOS 1.x)

mstenta:
This ports the Birth quick form to farmOS v2. It works almost exactly the same, but it adds the ability to differentiate the "birth mother" from the "genetic mother" per @Farmer-Eds-Shed's thought in the forum:

> I can see that it’s probably possible to add a third parent but this will be too ambiguous for genetic lineage purposes. So dose the quick form you are working on allow for this scenario? I would think there should be a check box and field for surrogate mother that allows the birth log to be assigned to the animal which gave birth while adding only the 2 genetic parents to the Asset that is created.
> 
> When I register Claves using the Department of Agricultures App it has 3 fields, Dam, Genetic Dam and Sire.

And my suggestion:

> Ah great point! No the quick form did not allow for this scenario! But it could! How about if the v2 quick form have three fields: “Birth mother”, “Genetic mother”, and “Genetic father” - and when you submit it, it saves the “Birth mother” to the Birth log, but saves the “Genetic mother/father” to the child asset “Parents”? I think that would cover it nicely.

Full thread: https://farmos.discourse.group/t/birth-log-time-date-stamp/1161

This PR depends on both #654 and #655 to work as intended. I will leave it as a "draft" until those are merged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [Refresh map edit layer when WKT is pasted into data input field #670](https://github.com/farmOS/farmOS/pull/670)
 - [Add QuickStringTrait::entityLabelsSummary() method for summarizing entity labels #675](https://github.com/farmOS/farmOS/pull/675)
 - [Add asset inventory views field #679](https://github.com/farmOS/farmOS/pull/679)
+- [Birth quick form (ported from farmOS 1.x) #656](https://github.com/farmOS/farmOS/pull/656)
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


mstenta:
Here is a screenshot of the form:

Screenshot 2023-03-10 at 11-48-07 Birth Drush Site-Install

mstenta:
Pushed a fixup commit to use the current user's timestamp in the default date: https://github.com/farmOS/farmOS/pull/656/commits/69a960648fc962fa2a94446bb822860b62206136

(@paul121 reminded me of this in #677 :smile: )

paul121 (approved):

PR #662, opened 2023-03-23
Simplify all map resize logic to use ResizeObserver

paul121:
This simplifies things.... wish I would have found this sooner!

`ResizeObserver` seems to have full support in browsers since 2018-2020
https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver
https://caniuse.com/resizeobserver


mstenta:
Wow! That really simplifies things!

mstenta (approved):

symbioquine (approved):
Beautiful!

PR #663, opened 2023-03-23
Allow map to be resized vertically

paul121:
Simple change allows the map to be resized vertically! `resize: vertical` adds a little drag icon to the bottom right corner of the map the same thing that is usually in a text area. `overflow: hidden` prevents scroll bars from appearing around the map.

As is this would be a simple change that allows all maps to be resized. Some questions:
- Should this be configurable?
- Should we also allow resizing horizontally?
- Should this be implemented in a map behavior?
- ...

This is dependent on #662 for proper resizing logic

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - [Add "Speed" to the list of quantity measures #658](https://github.com/farmOS/farmOS/pull/658)
 - [Include Fraction bundle fields in default Views #664](https://github.com/farmOS/farmOS/pull/664)
+- [Allow map to be resized vertically #663](https://github.com/farmOS/farmOS/pull/663)
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


mstenta:
> - Should this be configurable?
> - Should we also allow resizing horizontally?
> - Should this be implemented in a map behavior?
> - ...

We looked at this on the dev call today, and it seems like we can merge this as a first step regardless of these questions. If anyone *doesn't* want a map to be vertically resizable, they can tweak the CSS for that instance. If we find we want to disable this in a lot of places, we can make it configurable. But I don't really foresee that being a need, and this is a very simple first step. I don't see any downside.

The other question that came up was: should we have a `min-height`. In testing it didn't seem like it was really necessary. If someone makes it too small, they lose access to buttons etc, but they can just make it bigger again to fix that. We can't easily determine what a minimum size should be based on buttons, so I don't see any reason to make this more complicated than it needs to be.

mstenta (approved):

symbioquine (approved):

PR #668, opened 2023-04-13

mstenta:
Draft PR for https://www.drupal.org/project/farm/issues/3352183 - see that issue for more details.


modules/core/ui/theme/farm_ui_theme.module
```diff
@@ -22,6 +22,13 @@ function farm_ui_theme_theme($existing, $type, $theme, $path) {
   ];
 }
 
+/**
+ * Implements hook_theme_registry_alter().
+ */
+function farm_ui_theme_theme_registry_alter(&$theme_registry) {
+  $theme_registry['comment']['path'] = \Drupal::service('extension.list.module')->getPath('farm_ui_theme') . '/templates';
+}
+
```
paul121, line 31: We should use a theme suggestion for comment templates instead of overriding `comment.html.twig` eg: `comment--asset`, `comment--log`, `comment--asset--land`. I would hope these suggestions exist but if not we could add them

modules/core/comment/farm_comment.module
```diff
@@ -0,0 +1,93 @@
...
+  // (CommentItemInterface::HIDDEN) instead of 2 (CommentItemInterface::OPEN),
+  // because the Drupal\comment\Plugin\Field\FieldType\CommentItem::schema()
+  // default is 0.
+  $field = BundleFieldDefinition::create('comment');
+
+  // Set the field label.
+  $field->setLabel(t('Comments'));
+
+  // We assume that the comment type is "[entity-type]_comment".
+  $field->setSetting('comment_type', $entity_type . '_comment');
```
paul121: I don't think the `_comment` suffix is necessary for the comment type bundle ID, it will always be in the context of the comment entity type. Could we simplify this so the comment type is only the entity type?

As an example, this is the API path right now: `/api/comment/log_comment`

modules/core/ui/theme/farm_ui_theme.module
paul121: Wow.... there's only a suggestion for the referenced entity type's bundle. Only thinking about nodes, I see :disappointed: 

Screenshot from 2023-04-14 13-41-52

mstenta:
Just realized I forgot to push the commit that splits out the `farm_comment_asset` and `farm_comment_log` modules.

Thanks for the reviews @paul121! All make sense. I'll work on those changes and push more commits soon.

modules/core/comment/farm_comment.module
mstenta: https://github.com/mstenta/farmOS/commit/0354fe2d3702d716da7a3fc5fc5a4674b96b8c23

modules/core/ui/theme/farm_ui_theme.module
mstenta: > We should use a theme suggestion for comment templates instead of overriding `comment.html.twig`

Does this mean we would need to provide a template for each comment type? And contrib modules would need to provide templates for their types as well? :thinking: 

I think the intention here (in the `farm_ui_theme` module) is to override ALL comment theming in farmOS. So it's heavy handed, but that's the goal.

paul121: If we can avoid the heavy handed approach that would be great, template suggestions are kinda designed for this. Instead of a template for each entity type I think we could provide a single generic `comment--farm` template and add this as a template suggestion for asset/log/etc comment types that would like to use the farmOS convention. Should also be possible for templates to inherit from one another, so having multiple templates may not be a big deal (no duplication).

paul121: Happy to investigate this and give it a try!

mstenta: That makes sense! Please do! Feel free to push commit(s) to this branch.

mstenta:
Closing this in favor of #798.

PR #669, opened 2023-04-13
Fix bulk move/group action log names when unsetting location/group

mstenta:
If you use the bulk actions for "Move assets" or "Group assets", but you leave the "Location" or "Group" fields blank (which means you intend to unset the location / group of the assets), the logs that are created have an auto-generated name that doesn't make sense.

For example:

"Move Tractor to "

"Group Tractor into "

This is because the log name patterns are simply:

`Move @assets to @locations`

`Group @assets into @groups`

This PR fixes this by checking if `@locations` / `@groups` is empty, and offering a different pattern if it is:

`Clear location of @assets`

`Clear group membership of @assets`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


paul121 (approved):

mstenta:
Merged.

PR #670, opened 2023-05-04
Refresh map edit layer when WKT is pasted into data input field

mstenta:
This adds a bit of logic to our farmOS-map `input` behavior, which is used in all maps that have the `edit` behavior enabled (eg: `farm_map_input` form elements, including our `geofield` widget). It allows you to paste WKT into the `textarea` below the map, replacing the geometry in the editable layer.

### DIFF
--- a/modules/core/map/js/farmOS.map.behaviors.input.js
+++ b/modules/core/map/js/farmOS.map.behaviors.input.js
@@ -1,10 +1,38 @@
 (function () {
   farmOS.map.behaviors.input = {
     attach: function (instance) {
+
+      // Get the data input form element.
+      var input = instance.map.getTargetElement().parentElement.querySelector('[data-map-geometry-field]')
+
+      // When features change in the edit layer, write WKT to the input form element.
       instance.editAttached && instance.editAttached.then(() => {
         instance.edit.wktOn('featurechange', function(wkt) {
-          instance.map.getTargetElement().parentElement.querySelector('[data-map-geometry-field]').value = wkt;
+          if (input.value !== wkt) {
+            input.value = wkt;
+          }
         });
+      });
+
+      // Add an event listener to the input element, which will attempt to
+      // import new WKT when it changes.
+      input.addEventListener('input', (e) => {
+
+        // If the value is empty, only clear features from the layer.
+        if (!input.value) {
+          instance.edit.layer.getSource().clear();
+          return;
+        }
+
+        // Clear features from the layer.
+        instance.edit.layer.getSource().clear();
+
+        // Read features from WKT and add them to the layer.
+        var features = instance.readFeatures('wkt', input.value);
+        instance.edit.layer.getSource().addFeatures(features);
+
+        // Zoom to the layer.
+        instance.zoomToLayer(instance.edit.layer);
       });
     },
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- [Refresh map edit layer when WKT is pasted into data input field #670](https://github.com/farmOS/farmOS/pull/670)
+
 ### Changed
 
 - Update farmOS-map to [v2.2.0](https://github.com/farmOS/farmOS-map/releases/tag/v2.2.0)


modules/core/map/js/farmOS.map.behaviors.input.js
```diff
@@ -1,11 +1,56 @@
...
+          input.value = wkt;
         });
       });
+
+      // Run a handleInput() callback when input changes and attempt to
+      // import the new WKT.
+      input.oninput = handleInput;
+      function handleInput(e) {
+        if (input.value) {
+
```
symbioquine: Reverse conditional and return early so reader doesn't need to track as much indentation for the rest of the function?

```suggestion
        // If the value is empty, only clear features from the edit layer.
        if (!input.value) {
          instance.edit.layer.getSource().clear();
          return;
        }
```

modules/core/map/js/farmOS.map.behaviors.input.js
```diff
@@ -1,11 +1,56 @@
...
+            title: 'WKT',
+            wkt: input.value,
+            visible: false,
+          });
+
+          // Get the new layer's features. Bail if there are none.
+          var features = layer.getSource().getFeatures()
+
+          // If there are no features, bail.
+          if (!features) {
```
symbioquine: I believe `ol/source/Vector::getFeatures()` always returns an array - even if the source is empty... https://github.com/openlayers/openlayers/blob/9613cb97617ce2e3af082c236d5f63a9dc7e93f3/src/ol/source/Vector.js#L700-L712

```suggestion
          if (!features || !features.length) {
```

modules/core/map/js/farmOS.map.behaviors.input.js
```diff
@@ -1,11 +1,56 @@
...
         instance.edit.wktOn('featurechange', function(wkt) {
-          instance.map.getTargetElement().parentElement.querySelector('[data-map-geometry-field]').value = wkt;
+          input.value = wkt;
         });
       });
+
+      // Run a handleInput() callback when input changes and attempt to
+      // import the new WKT.
+      input.oninput = handleInput;
+      function handleInput(e) {
```
symbioquine:
```suggestion
      input.addEventListener('input', (e) => {
```

See https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener

image

modules/core/map/js/farmOS.map.behaviors.input.js
```diff
@@ -1,11 +1,56 @@
...
+        if (input.value) {
+
+          // Add a new temporary invisible WKT layer.
+          var layer = instance.addLayer('wkt', {
+            title: 'WKT',
+            wkt: input.value,
+            visible: false,
+          });
```
symbioquine: What happens if the WKT is invalid? Currently this is handled server-side and an error is surfaced, but this would silently fail as-written right?

I'm guessing we need to catch the errors and surface them on/around the input text field... possibly using `input.setCustomValidity(message)`. See https://developer.mozilla.org/en-US/docs/Learn/Forms/Form_validation#validating_forms_using_javascript

modules/core/map/js/farmOS.map.behaviors.input.js
```diff
@@ -1,11 +1,56 @@
...
+
+      // Get the data input form element.
+      var input = instance.map.getTargetElement().parentElement.querySelector('[data-map-geometry-field]')
+
+      // When features change in the edit layer, write WKT to the input form element.
       instance.editAttached && instance.editAttached.then(() => {
         instance.edit.wktOn('featurechange', function(wkt) {
-          instance.map.getTargetElement().parentElement.querySelector('[data-map-geometry-field]').value = wkt;
+          input.value = wkt;
         });
```
symbioquine: Just curious how we avoid an infinite loop if the input updates the features and the features update the input?

symbioquine (commented):
I'm really happy to see this change happening!

modules/core/map/js/farmOS.map.behaviors.input.js
mstenta: I wondered this too, but apparently this does not trigger the `featurechange` event, and so it doesn't cause the input field to get updated again.

modules/core/map/js/farmOS.map.behaviors.input.js
mstenta: It silently fails currently.

Can we tackle this in a followup, and just let Drupal surface any errors on form save for now?

modules/core/map/js/farmOS.map.behaviors.input.js
mstenta: Good idea! Fixup: https://github.com/farmOS/farmOS/pull/670/commits/34a609e0547cf53e5b5fffcf6cad546238327760

modules/core/map/js/farmOS.map.behaviors.input.js
mstenta: Thanks! I had a feeling there was a more "proper" way to do this. :-)

Fixup: https://github.com/farmOS/farmOS/pull/670/commits/9bdb4c39c535e87dc25d158faf3718047ff8eb55

modules/core/map/js/farmOS.map.behaviors.input.js
mstenta: I actually removed this condition and forgot to force push my branch. It had a logical error which could have led to the code returning before cleaning up the temporary layer. This also changes slightly when we are able to use this: https://github.com/farmOS/farmOS-map/pull/194

So it's worth reviewing again with these changes. I just rebased with this change, plus some other fixup commits to address your other feedback.

modules/core/map/js/farmOS.map.behaviors.input.js
paul121: Could check to see if the input already has the same value that it would be replacing. If it is the same, don't update the value

mstenta: @paul121 That make sense. Will it be challenging to compare them, though? We have the new WKT, and we have the old map layer source features. I don't think we have a public method of converting those features to WKT, so we can't compare WKT to WKT. Maybe we can convert the new WKT to features and compare OL features to OL features? I wonder if that will get complicated if there are multiple features and they end up in different orders?

modules/core/map/js/farmOS.map.behaviors.input.js
symbioquine: Seems reasonable

symbioquine (approved):
LGTM

modules/core/map/js/farmOS.map.behaviors.input.js
paul121: Hmm I was just thinking to do this check in the wkt event listener right before replacing the input value, when we do have the WKT string?

```js
if (input.value != wkt) {
  input.value = wkt;
}
```

paul121: But maybe the browser already handled this situation and won't fire a change event if the value doesn't change?

mstenta: Oooh I see. I thought you meant in the new `input` event listener... Yea that makes perfect sense. I'll push a fixup to add that...

mstenta: https://github.com/farmOS/farmOS/pull/670/commits/16642b62cf25fb34f4c6c09becd5965cd4cf520f

paul121 (approved):

PR #671, opened 2023-05-05
Add QuickStringTrait::entityLabelsSummary() and deprecate farm_log_asset_names_summary()

mstenta:
Opening this PR to start a discussion...

Currently the `farm_log_asset_names_summary()` function is only used in the `AssetMoveActionForm` and `AssetGroupActionForm` classes.

If we were to make both of those into quick forms instead, and redirect to them with assets prepopulated from the bulk action (via https://farmos.org/development/module/quick/#quick-form-actions), we could further unify helper functions like this one into quick form traits. And those bulk actions can delegate to more generally useful quick forms.

This PR doesn't do any of that, but could be a first step in that direction.

### DIFF
--- a/modules/core/quick/tests/src/Kernel/QuickStringTest.php
+++ b/modules/core/quick/tests/src/Kernel/QuickStringTest.php
@@ -88,9 +88,9 @@ class QuickStringTest extends KernelTestBase {
   }
 
   /**
-   * Test assetNamesSummary() method.
+   * Test entityLabelsSummary() method.
    */
-  public function testAssetNamesSummary() {
+  public function testEntityLabelsSummary() {
 
     // Create a test asset type.
     $asset_type = AssetType::create([
@@ -114,12 +114,12 @@ class QuickStringTest extends KernelTestBase {
 
     // Test default with a cutoff of 3.
     $expected = $assets[0]->label() . ', ' . $assets[1]->label() . ', ' . $assets[2]->label() . ' (+ 7 more)';
-    $name_summary = $this->assetNamesSummary($assets);
+    $name_summary = $this->entityLabelsSummary($assets);
     $this->assertEquals($expected, $name_summary);
 
     // Test with a cutoff of 1.
     $expected = $assets[0]->label() . ' (+ 9 more)';
-    $name_summary = $this->assetNamesSummary($assets, 1);
+    $name_summary = $this->entityLabelsSummary($assets, 1);
     $this->assertEquals($expected, $name_summary);
 
     // Test with a cutoff of 0.
@@ -128,7 +128,7 @@ class QuickStringTest extends KernelTestBase {
       $labels[] = $asset->label();
     }
     $expected = implode(', ', $labels);
-    $name_summary = $this->assetNamesSummary($assets, 0);
+    $name_summary = $this->entityLabelsSummary($assets, 0);
     $this->assertEquals($expected, $name_summary);
   }
 
--- a/modules/core/quick/tests/src/Kernel/QuickStringTest.php
+++ b/modules/core/quick/tests/src/Kernel/QuickStringTest.php
@@ -122,7 +122,7 @@ class QuickStringTest extends KernelTestBase {
     $name_summary = $this->assetNamesSummary($assets, 1);
     $this->assertEquals($expected, $name_summary);
 
-    // Test default with a cutoff of 0.
+    // Test with a cutoff of 0.
     $labels = [];
     foreach ($assets as $asset) {
       $labels[] = $asset->label();
--- a/modules/core/quick/tests/src/Kernel/QuickStringTest.php
+++ b/modules/core/quick/tests/src/Kernel/QuickStringTest.php
@@ -2,6 +2,8 @@
 
 namespace Drupal\Tests\farm_quick\Kernel;
 
+use Drupal\asset\Entity\Asset;
+use Drupal\asset\Entity\AssetType;
 use Drupal\farm_quick\Traits\QuickStringTrait;
 use Drupal\KernelTests\KernelTestBase;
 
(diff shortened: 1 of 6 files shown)


mstenta:
There is more to do here, if we pursue this:

- Needs to be documented in https://farmos.org/development/module/quick/#traits-and-helper-methods
- Needs automated tests.

mstenta:
> - Needs to be documented in https://farmos.org/development/module/quick/#traits-and-helper-methods
> - Needs automated tests.

Force-pushed the commit with these included.

When I was writing the automated test, it uncovered a bug with the current `farm_log_asset_names_summary()` logic. So I added a commit to fix that before this commit. If we don't pursue this, we should separate that commit out and merge it separately (or just go ahead and merge that commit right now).

modules/core/quick/tests/src/Kernel/QuickStringTest.php
```diff
@@ -72,4 +87,49 @@ public function testPrioritizedString() {
...
+    $expected = $assets[0]->label() . ', ' . $assets[1]->label() . ', ' . $assets[2]->label() . ' (+ 7 more)';
+    $name_summary = $this->assetNamesSummary($assets);
+    $this->assertEquals($expected, $name_summary);
+
+    // Test with a cutoff of 1.
+    $expected = $assets[0]->label() . ' (+ 9 more)';
+    $name_summary = $this->assetNamesSummary($assets, 1);
+    $this->assertEquals($expected, $name_summary);
+
+    // Test default with a cutoff of 0.
```
symbioquine: This comment is inaccurate (in calling zero the default) right?

mstenta: Ah good catch @symbioquine - I'll push a fixup commit.

mstenta:
Pushed two fixup commits: one to fix the comment in tests that @symbioquine pointed out above, and one to add `protected` visibility to the new method to fix PHPCodesniffer complaint.

There is still another set of PHPCodesniffer errors/warning that need to be resolved:

```
 FILE: /opt/drupal/web/profiles/farm/modules/core/log/farm_log.module
----------------------------------------------------------------------
FOUND 1 ERROR AND 2 WARNINGS AFFECTING 1 LINE
----------------------------------------------------------------------
 74 | WARNING | The deprecation-version 'farmOS:2.x' does not match
    |         | the lower-case machine-name standard: drupal:n.n.n or
    |         | project:n.x-n.n or project:n.x-n.n-label[n] or
    |         | project:n.n.n or project:n.n.n-label[n]
 74 | WARNING | The removal-version 'farmOS:3.x' does not match the
    |         | lower-case machine-name standard: drupal:n.n.n or
    |         | project:n.x-n.n or project:n.x-n.n-label[n] or
    |         | project:n.n.n or project:n.n.n-label[n]
 74 | ERROR   | Each @deprecated tag must have a @see tag immediately
    |         | following it
----------------------------------------------------------------------
```

This is the first time we're using `@deprecated` in farmOS 2.x, so I think we need to establish a process for this. We probably want to create a change record (CR) node on drupal.org and point to that with the `@see` tag.

paul121 (commented):
In the spirit of this being moved to a general quick form helper trait... could this method be be generalized to just be `entityNamesSummary`?  May be useful for other entities once this is in quick form world.

mstenta:
@paul121 That makes sense to me. I pushed a commit that renames it (separate from the original commit). And I renamed this PR accordingly.

This commit also copies the code from `farm_log_asset_names_summary()` instead of just delegating to it, because the function signature of `farm_log_asset_names_summary()` explicitly requires an array of `AssetInterface` instances. I had an `@todo` to copy that code into the new helper method anyway, so we might as well just do it now. It means the code is duplicated, but it's not very much. And knowing that `farm_log_asset_names_summary()` is deprecated, we can just assume it will go away in the future.

Last thing to figure out here is still the change record stuff. But if you approve of the changes then at least we know this PR is code complete (with the exception of the Codesniffer error/warning).

mstenta:
I created a dedicated issue for deprecating `farm_log_asset_names_summary()`: https://www.drupal.org/project/farm/issues/3359452

mstenta:
... and a change record: https://www.drupal.org/node/3359456

mstenta:
I've decided that I'm going to split this PR up into multiple. So I'm closing this one and will be opening some new ones...

Reasoning: I think we should wait to officially deprecate `farm_log_asset_names_summary()` until we've actually refactored the Move and Group actions to use quick forms. There are still some considerations around that idea, which means there is uncertainty. If we deprecate the function before figuring that out it may put us in a more difficult position.

So, I think the best order of operations for this should be:

1. Create movement quick form (using `farm_log_asset_names_summary()`)
2. Create group membership quick form (also using `farm_log_asset_names_summary()`)
3. Refactor both Move and Group actions to use the quick forms
4. Add `QuickStringTrait::entityLabelsSummary()` and use it in the quick forms
5. Deprecate `farm_log_asset_names_summary()`

I am working on 1 right now and plan to open a PR soon. I was hoping to use the new `QuickStringTrait::entityLabelsSummary()` in that, but in light of these thoughts I will just use `farm_log_asset_names_summary()` (with an explicit dependency on `farm_log`) for now, and we can plan to replace that in the future.

> When I was writing the automated test, it uncovered a bug with the current `farm_log_asset_names_summary()` logic. So I added a commit to fix that before this commit. If we don't pursue this, we should separate that commit out and merge it separately (or just go ahead and merge that commit right now).

New PR for this commit: https://github.com/farmOS/farmOS/pull/674

mstenta:
> 4. Add `QuickStringTrait::entityLabelsSummary()`

Draft PR: https://github.com/farmOS/farmOS/pull/675

> 5. Deprecate `farm_log_asset_names_summary()`

Draft PR: https://github.com/farmOS/farmOS/pull/676

PR #673, opened 2023-05-09
Remove field links from asset entity browser view

paul121:
Our asset entity browser opens a table view of assets. Clicking on links within this table opens the link within the same entity browser modal and leaves the user without any clear way to "go back" to the entity browser and select assets.

Removing links from the asset name and current location columns is a simple fix that prevents this from happening.

Screenshot from 2023-05-09 13-56-52

Screenshot from 2023-05-09 13-57-04

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Update farmOS-map to [v2.2.1](https://github.com/farmOS/farmOS-map/releases/tag/v2.2.1)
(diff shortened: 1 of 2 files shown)


mstenta:
I'm OK with this change. If we hear back afterwards that users want the ability to open assets from this View we could explore other solutions. Maybe just opening them in a new tab by default, or having a little "open in new tab" icon in a column on the right that does that.

paul121:
> Maybe just opening them in a new tab by default, or having a little "open in new tab" icon in a column on the right that does that.

Agreed! I was happy to see that fields in this view display were already overridden... so adding `target = _blank` might be fairly easy on only this page?

mstenta (approved):

PR #674, opened 2023-05-10
Fix farm_log_asset_names_summary() $cutoff parameter

mstenta:
There is a bug in the way the `$cutoff` parameter of the `farm_log_asset_names_summary()` function works.

https://github.com/farmOS/farmOS/blob/2b3bb21683396fc0b5ec0cd2f5107dc4e4e4d16a/modules/core/log/farm_log.module#L74

It works as expected if the `$cutoff` is `0` (all names are included), or `1` (only the first name is included). Everything else is treated as if `$cutoff` is `1` (only the first name is included).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - [Fix bulk move/group action log names when unsetting location/group #669](https://github.com/farmOS/farmOS/pull/669)
+- [Fix farm_log_asset_names_summary() $cutoff parameter #674](https://github.com/farmOS/farmOS/pull/674)
 
 ### Security
 
--- a/modules/core/log/farm_log.module
+++ b/modules/core/log/farm_log.module
@@ -63,10 +63,9 @@ function farm_log_entity_prepare_form(EntityInterface $entity, $operation, FormS
  *   An array of assets.
  * @param int $cutoff
  *   The number of asset names to include before summarizing the rest.
- *   If the number of assets exceeds the cutoff, only the first asset's
- *   name will be included, and the rest will be summarized as "(+ X more)".
- *   If the number of assets is less than or equal to the cutoff, or if the
- *   cutoff is 0, all asset names will be included.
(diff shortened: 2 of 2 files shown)


paul121 (changes requested):
I think this is working as described by the cutoff parameter. Specifically that one asset name will be included when the cutoff is exceeded:

https://github.com/farmOS/farmOS/blob/0093d48b84a88aa5402404331051807ed12e3cc2/modules/core/log/farm_log.module#L64-L69

However I agree in that this is maybe not how I expected this function to behave, cutoff of 1 includes 1 asset name, cutoff of 2 includes two asset names. I'm OK with changing the documented behavior but we should update the docs.

Thought: maybe we only change this in the new `entityLabelsSummary` method? Prevent a "breaking change"?

mstenta:
> I think this is working as described by the cutoff parameter.

Wow. Good reading @paul121 lol - I didn't even notice that the docblock is describing the current (wrong) behavior (and therefore the behavior is right??) :face_with_spiral_eyes: 

Although actually, the docblock says it both ways:

> The number of asset names to include before summarizing the rest.

This describes the intended behavior IMO. And if you only read that without reading the rest, the expectation is that it would include as many asset names as the `$cutoff` defines.

But then if you keep reading...

> If the number of assets exceeds the cutoff, only the first asset's 
> name will be included, and the rest will be summarized as "(+ X more)". 
> If the number of assets is less than or equal to the cutoff, or if the 
> cutoff is 0, all asset names will be included. 

Super strange. I will take the blame for it! We were deep in the 2.x upgrade when it was written. :-)

https://github.com/farmOS/farmOS/commit/835209e5a2e705cbd4bcd19c482606350db701aa

> However I agree in that this is maybe not how I expected this function to behave
> I'm OK with changing the documented behavior but we should update the docs.

Agreed. I'll push a fixup commit.

> Thought: maybe we only change this in the new entityLabelsSummary method? Prevent a "breaking change"?

This one's tricky because the "intended behavior" got so muddled by the docblock comments. But I think we should still consider this a "bug". And if we're going to have both `entityLabelsSummary()` and `farm_log_asset_names_summary()` in parallel for a time, I think we should at least keep them in sync.

paul121 (approved):

PR #675, opened 2023-05-10
Add QuickStringTrait::entityLabelsSummary() method for summarizing entity labels

mstenta:
Originally discussed here, and ultimately decided to split it out to a separate PR: https://github.com/farmOS/farmOS/pull/671


mstenta:
Note to self: after the Movement quick form is merged (opening PR soon), switch to using this new method in it and remove the explicit dependency on `farm_log`.

paul121 (approved):
Just remove the `farm_log` dependency in tests

PR #677, opened 2023-05-11
Movement quick form (ported from farmOS 1.x)

mstenta:
Screenshot:

Screenshot 2023-05-11 at 15-11-48 Movement Drush Site-Install

### DIFF
--- a/modules/quick/movement/src/Plugin/QuickForm/Movement.php
+++ b/modules/quick/movement/src/Plugin/QuickForm/Movement.php
@@ -1,0 +1,326 @@
+<?php
+
+namespace Drupal\farm_quick_movement\Plugin\QuickForm;
+
+use Drupal\Core\Datetime\DrupalDateTime;
+use Drupal\Core\Entity\EntityInterface;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Messenger\MessengerInterface;
+use Drupal\Core\Session\AccountInterface;
+use Drupal\farm_geo\Traits\WktTrait;
+use Drupal\farm_location\AssetLocationInterface;
+use Drupal\farm_quick\Plugin\QuickForm\QuickFormBase;
+use Drupal\farm_quick\Plugin\QuickForm\QuickFormInterface;
+use Drupal\farm_quick\Traits\QuickFormElementsTrait;
+use Drupal\farm_quick\Traits\QuickLogTrait;
+use Drupal\farm_quick\Traits\QuickStringTrait;
+use Psr\Container\ContainerInterface;
+
+/**
+ * Movement quick form.
+ *
+ * @QuickForm(
+ *   id = "movement",
+ *   label = @Translation("Movement"),
+ *   description = @Translation("Record the movement of assets."),
+ *   helpText = @Translation("Use this form to record the movement of assets to a new location."),
+ *   permissions = {
+ *     "create activity log",
+ *   }
+ * )
+ */
+class Movement extends QuickFormBase implements QuickFormInterface {
+
+  use QuickLogTrait;
+  use QuickFormElementsTrait;
+  use QuickStringTrait;
+  use WktTrait;
+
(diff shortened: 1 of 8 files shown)


modules/quick/movement/src/Plugin/QuickForm/Movement.php
```diff
@@ -0,0 +1,304 @@
...
+      ],
+      '#display_raw_geometry' => TRUE,
+    ];
+
+    // Hidden fields to store asset and location geometry.
+    $form['asset_geometry'] = [
+      '#type' => 'hidden',
+      '#value' => $this->combinedAssetGeometries($this->loadEntityAutocompleteAssets($form_state->getValue('asset'))),
+      '#prefix' => '<div id="asset-geometry" data-drupal-selector="asset-geometry">',
+      '#suffix' => '</div>',
```
paul121: Minor thing, but I've started using a `container` element instead of prefix/suffix for wrapper IDs. I prefer how this code looks/works using form API but curious what you think:

```php
    $form['asset_geometry_wrapper'] = [
      '#type' => 'container',
      '#attributes' => [
        'id' => 'asset-geometry',
        'data-drupal-selector' => 'asset-geometry',
      ],
      'asset_geometry' => [
        '#type' => 'hidden',
        '#value' => $this->combinedAssetGeometries($this->loadEntityAutocompleteAssets($form_state->getValue('asset'))),
      ],
    ];
```

modules/quick/movement/src/Plugin/QuickForm/Movement.php
```diff
@@ -0,0 +1,304 @@
...
+    ];
+    $form['location_geometry'] = [
+      '#type' => 'hidden',
+      '#value' => $this->combinedAssetGeometries($this->loadEntityAutocompleteAssets($form_state->getValue('location'))),
+      '#prefix' => '<div id="location-geometry" data-drupal-selector="location-geometry">',
```
paul121: Is `data-drupal-selector` important here? This just feels like a special data ID that drupal may actually use for other things. I doubt there is any conflict here, but maybe we could rename to something like `data-movement-geometry` to make it clear we aren't using a special drupal form API keyword?

modules/quick/movement/src/Plugin/QuickForm/Movement.php
```diff
@@ -0,0 +1,304 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function buildForm(array $form, FormStateInterface $form_state, string $id = NULL) {
+
+    // Date.
+    $form['date'] = [
+      '#type' => 'datetime',
+      '#title' => $this->t('Date'),
+      '#default_value' => new DrupalDateTime('midnight'),
```
paul121: This should use the user's timezone for reasons described here: https://github.com/farmOS/farmOS/pull/665

paul121 (changes requested):

modules/quick/movement/src/Plugin/QuickForm/Movement.php
mstenta: https://github.com/farmOS/farmOS/pull/677/commits/cc7e2638a5d6c6e8c8fbc4becec211a6007cc350

modules/quick/movement/src/Plugin/QuickForm/Movement.php
mstenta: https://github.com/farmOS/farmOS/pull/677/commits/842d0031fe98aaf8820f2deaa015b967abed8f61

modules/quick/movement/src/Plugin/QuickForm/Movement.php
mstenta: https://github.com/farmOS/farmOS/pull/677/commits/f1d5a18c64833eaf17c95242c8946ccddc54ca17

mstenta:
Thanks for the review+feedback @paul121! I've addressed all your comments above with fixup commits. Please take another look when you have a chance. :-)

paul121 (approved):

PR #678, opened 2023-05-12
Changes modules directory so mounted modules are recognized

braughtg:
Documentation indicated that modules should be added to `/var/opt/drupal/web/sites/all/modules` within the -dev container. However, modules mounted into that location were not recognized by drupal. This PR changes the directory for modules to be `/opt/drupal/web/sites/all/modules`. When modules are mounted into the container in this directory they are recognized by drupal.


mstenta:
Good find - thanks @braughtg! Merged!

PR #679, opened 2023-05-15
Add asset inventory views field

wotnak:
These changes make the inventory based asset quantity computed field available in views (in the same way as the asset location field is handled).


mstenta (changes requested):
Thanks @wotnak! This will be a nice addition.

One request: can you replace all occurrences of the phrase "quantity" with "inventory"? I think that will be more explicit, and avoid confusion with the more general use of "quantities" in farmOS. This field is specifically about inventory.

Also, I think we can leave the `views` dependency as optional. This should work if the Views module is enabled, and any modules that need to use it can depend on `views`. But we can leave this module with minimal dependencies to make automated testing easier etc.

paul121 (changes requested):
LGTM, we just need a schema definition for the views field: https://github.com/farmOS/farmOS/blob/2.0.4/modules/core/location/config/schema/farm_location.views.schema.yml

mstenta (approved):

paul121 (approved):

mstenta:
Merged - thanks @wotnak!

PR #681, opened 2023-05-19
Use a single process for MariaDB tests

mstenta:
MariaDB tests fail a large percentage of the time in GitHub Actions with the following error:

```
Failed to run installer database tasks: Failed to connect to your database server. The server reports the following message: SQLSTATE[HY000] [2002] Connection refused.
- Is the database server running?
- Does the database exist or does the database user have sufficient privileges to create the database?
- Have you entered the correct database name?
- Have you entered the correct username and password?
- Have you entered the correct database hostname and port number?
```


mstenta:
Tests are still failing - this doesn't fix it. :-(

PR #682, opened 2023-05-23
Fix undeclared dependency on farm_group in birth quick form

mstenta:
Discovered a bug with the new birth quick form today. It is supposed to optionally add support for Group Membership, but if the `farm_group` module is not enabled the following error occurs and it is impossible to open the quick form (or even the `/quick` index page):

```
Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException: You have requested a non-existent service "group.membership". in Drupal\Component\DependencyInjection\Container->get() (line 157 of /opt/drupal/web/core/lib/Drupal/Component/DependencyInjection/Container.php).
```

This is because the birth quick form tries to inject the `group.membership` service as a dependency.

This PR fixes the issue by removing the injected dependency and using `\Drupal::service('group.membership')` instead.

A `// @phpstan-ignore-next-line` is added above this line to allow PHPStan tests to pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- [Fix undeclared dependency on farm_group in birth quick form #682](https://github.com/farmOS/farmOS/pull/682)
+
 ## [2.1.0] 2023-05-19
 
 This is the first minor release of the farmOS 2.x branch, following
--- a/modules/quick/birth/src/Plugin/QuickForm/Birth.php
+++ b/modules/quick/birth/src/Plugin/QuickForm/Birth.php
@@ -9,7 +9,6 @@ use Drupal\Core\Extension\ModuleHandlerInterface;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Messenger\MessengerInterface;
 use Drupal\Core\Session\AccountInterface;
-use Drupal\farm_group\GroupMembershipInterface;
 use Drupal\farm_location\AssetLocationInterface;
 use Drupal\farm_quick\Plugin\QuickForm\QuickFormBase;
 use Drupal\farm_quick\Traits\QuickAssetTrait;
@@ -67,13 +66,6 @@ class Birth extends QuickFormBase {
   protected $assetLocation;
 
   /**
-   * Group membership service.
-   *
-   * @var \Drupal\farm_group\GroupMembershipInterface
-   */
-  protected $groupMembership;
-
-  /**
    * Current user object.
    *
    * @var \Drupal\Core\Session\AccountInterface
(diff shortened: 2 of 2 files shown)


symbioquine (approved):
image

```txt
symbioquine

    [mstenta@mstenta:matrix.org](https://matrix.to/#/!rxZezKXfxOKJxTgNGs:matrix.org/$168485543897468knRuX:matrix.org?via=matrix.org&via=libera.chat&via=chamba.cloud)
    [Discovered a bug with the new birth quick form: ](https://matrix.to/#/!rxZezKXfxOKJxTgNGs:matrix.org/$168485543897468knRuX:matrix.org?via=matrix.org&via=libera.chat&via=chamba.cloud)https://github.com/farmOS/farmOS/pull/682

The bug fix seems fine, but I wonder if the behavior of defaulting to the mother's group should happen when building the form...
The way it's written, it looks like there's no way to not have the child in a group if the mother is in one.
Whereas if the form were just pre-populated with the mother's group, then the value could just be cleared to achieve that case.

mstenta@mstenta:matrix.org
That is true - and was an intentional decision (albeit a shortcut - because autopopulating that field would take more work) - that should be a followup feature request IMO
```

paul121 (approved):

wotnak:
Instead of using static \Drupal::service and ignoring phpstan maybe a better approach would be to:
- declare $groupMembership field as nullable
- in constructor move $group_membership parameter to the end and set its default value to NULL
- in create() pass group_membership as group.membership service only if it exists othervide NULL
	this can be done in two ways:
	- `$contianer->has('group.membership') ? $container->get('group.membership') : NULL`
	- or by changing $container typehint to Symfony\Component\DependencyInjection\ContainerInterface and using `$container->get('group.membership', ContainerInterface::NULL_ON_INVALID_REFERENCE)`
- then before $groupMembership is used check if it is not null
	existing check for moduleExists should be enough but to be more explicit additional `$this->groupMembership !== NULL` condition in the if statement in which groupMembership is used could be added

The described approach would still fix the issue, but it would allow continuing using dependency injection and to not ignore phpstan rules.

```patch
diff --git a/modules/quick/birth/src/Plugin/QuickForm/Birth.php b/modules/quick/birth/src/Plugin/QuickForm/Birth.php
index 3381bc80..b0578ca3 100644
--- a/modules/quick/birth/src/Plugin/QuickForm/Birth.php
+++ b/modules/quick/birth/src/Plugin/QuickForm/Birth.php
@@ -9,6 +9,7 @@ use Drupal\Core\Extension\ModuleHandlerInterface;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Messenger\MessengerInterface;
 use Drupal\Core\Session\AccountInterface;
+use Drupal\farm_group\GroupMembershipInterface;
 use Drupal\farm_location\AssetLocationInterface;
 use Drupal\farm_quick\Plugin\QuickForm\QuickFormBase;
 use Drupal\farm_quick\Traits\QuickAssetTrait;
@@ -72,6 +73,13 @@ class Birth extends QuickFormBase {
    */
   protected $currentUser;
 
+  /**
+   * Group membership service.
+   *
+   * @var \Drupal\farm_group\GroupMembershipInterface|null
+   */
+  protected $groupMembership = NULL;
+
   /**
    * Constructs a QuickFormBase object.
    *
@@ -93,14 +101,17 @@ class Birth extends QuickFormBase {
    *   Asset location service.
    * @param \Drupal\Core\Session\AccountInterface $current_user
    *   Current user object.
+   * @param \Drupal\farm_group\GroupMembershipInterface|null $group_membership
+   *   Group membership service.
    */
-  public function __construct(array $configuration, $plugin_id, $plugin_definition, MessengerInterface $messenger, EntityTypeManagerInterface $entity_type_manager, ModuleHandlerInterface $module_handler, ConfigFactoryInterface $config_factory, AssetLocationInterface $asset_location, AccountInterface $current_user) {
+  public function __construct(array $configuration, $plugin_id, $plugin_definition, MessengerInterface $messenger, EntityTypeManagerInterface $entity_type_manager, ModuleHandlerInterface $module_handler, ConfigFactoryInterface $config_factory, AssetLocationInterface $asset_location, AccountInterface $current_user, ?GroupMembershipInterface $group_membership = NULL) {
     parent::__construct($configuration, $plugin_id, $plugin_definition, $messenger);
     $this->entityTypeManager = $entity_type_manager;
     $this->moduleHandler = $module_handler;
     $this->configFactory = $config_factory;
     $this->assetLocation = $asset_location;
     $this->currentUser = $current_user;
+    $this->groupMembership = $group_membership;
   }
 
   /**
@@ -117,6 +128,7 @@ class Birth extends QuickFormBase {
       $container->get('config.factory'),
       $container->get('asset.location'),
       $container->get('current_user'),
+      $container->has('group.membership') ? $container->get('group.membership') : NULL,
     );
   }
 
@@ -480,8 +492,7 @@ class Birth extends QuickFormBase {
         $group = [$this->entityTypeManager->getStorage('asset')->load($group)];
       }
       if (empty($group)) {
-        // @phpstan-ignore-next-line
-        $group = \Drupal::service('group.membership')->getGroup($birth_mother, $birthdate->getTimestamp());
+        $group = $this->groupMembership->getGroup($birth_mother, $birthdate->getTimestamp());
       }
       if (!empty($group)) {
         $birth_log_values['group'] = $group;
```

mstenta:
Thanks for the tip @wotnak! I like that!

Would you like to open a PR from your diff above? It LGMT! (The only other suggestion you had was to add a check for `$this->groupMembership !== NULL` which I agree would be good)

PR #699, opened 2023-07-07
Improve asset bundle logic in ID tag widget

paul121:
This fixes a bug where assets cannot be used in an inline entity form because the ID Tag widget assumes it is being used in the context of an entity form_display:

> Warning: Undefined array key "form_display" in Drupal\farm_id_tag\Plugin\Field\FieldWidget\IdTagWidget->formElement() (line 39 of /var/www/html/web/profiles/farm/modules/core/id_tag/src/Plugin/Field/FieldWidget/IdTagWidget.php)

Inline entity forms and likely other entity forms may not use the form_display key in form storage so using the entity from the FieldItemList is more robust.


paul121:
The above warning leads to the following error which crashes the request:

> Error: Call to a member function get() on null in Drupal\farm_id_tag\Plugin\Field\FieldWidget\IdTagWidget->formElement() (line 42 of /var/www/html/web/profiles/farm/modules/core/id_tag/src/Plugin/Field/FieldWidget/IdTagWidget.php)

mstenta:
Thanks @paul121!

PR #701, opened 2023-07-10
Use access content permission in farm_ui_views test

paul121:
The administer taxonomy permission should not be used in tests because we do not grant the permission to standard farmOS roles. Taxonomy terms can be viewed with the access content permission.

https://github.com/farmOS/farmOS/blob/a1b315c9026e1dca54d31fb1a20e69c4e3eb59e8/modules/core/role/src/ManagedRolePermissionsManager.php#L290-L293


mstenta:
Thanks @paul121!

PR #702, opened 2023-07-18
Fix composer.json version constraints for migrate_plus and migrate_tools

mstenta:
This fixes a small issue introduced by PR #700.

See this comment: https://github.com/farmOS/farmOS/pull/700#issuecomment-1640986423


paul121 (approved):
Good catch!

PR #703, opened 2023-08-02
Misc quick form code and documentation improvements

mstenta:
This is a collection of small improvements to the core farmOS quick form code and documentation, which came out of earlier work I was starting on "[Configurable quick forms](https://www.drupal.org/project/farm/issues/3354935)". None of these changes are specific to that work, and are general improvements that I figured we could merge independently. I'm submitting them all in one PR because most of them are small, but happy to split into multiple PRs if anyone prefers.

The commit messages for each describe the changes in more detail.


modules/core/quick/src/Form/QuickForm.php
```diff
@@ -61,7 +61,7 @@ public function getBaseFormId() {
    */
   public function getFormId() {
     $id = $this->getRouteMatch()->getParameter('id');
-    return $this->getBaseFormId() . "_$id";
+    return $this->quickFormManager->createInstance($id)->getFormId();
```
paul121: This should use the base form ID

```suggestion
    return $this->getBaseFormId() . '_' . $this->quickFormManager->createInstance($id)->getFormId();
```

paul121:
A general thought re: [Extend QuickFormInterface from FormInterface](https://github.com/farmOS/farmOS/pull/703/commits/7b3b845ceab38b9e3dafb4238e8ec49e1f874b1a)... if forms are all going to implement this interface I think we should remove `Drupal\farm_quick\Form\QuickForm` entirely because it doesn't really do anything. Instead, we can dynamically build routes for all of the individual quick forms since they are now implementing the proper `FormInterface`.

Probably need some additional thought here re: configurable quick forms... but still I think we should remove the `QuickForm` class if possible

modules/core/quick/src/Form/QuickForm.php
mstenta: The `QuickFormBase` class actually already prepends the form ID w/ `quick_form_` in `getFormId()`, so it already ends up as `quick_form_[id]`. The change you suggested would result in a form ID of `quick_form_quick_form_[id]`.

https://github.com/mstenta/farmOS/blob/0c8864f265581faeb98c1ee3fc83100d1284e19b/modules/core/quick/src/Plugin/QuickForm/QuickFormBase.php#L63

But maybe it would be better for `QuickFormBase` to just return the quick form ID, and not prepend `quick_form_`. Combining that with your suggested change would have the same end result (form id of `quick_form_[id]`).

I will prepare a fixup commit for consideration...

mstenta: https://github.com/mstenta/farmOS/commit/d6e23246eaf9c217d921e46e38e2eb4c179f9b90

mstenta:
> Probably need some additional thought here re: configurable quick forms... but still I think we should remove the `QuickForm` class if possible

I'll follow up with you in chat about this @paul121 - we may need a wrapper form class for the configurable quick forms anyway. But maybe we can figure out a clever way to remove both. Let's tackle that in the configurable quick form branch after this.

mstenta:
Pushed one more fixup commit to fix a failing test: https://github.com/farmOS/farmOS/pull/703/commits/a90cf00ee965b3ab0bef38417f6d713b114fa28c

mstenta:
I converted this back to draft because there may be some more commits added as we're working on the configurable quick forms branch.

docs/development/module/quick.md
```diff
@@ -43,12 +72,26 @@ class Egg extends QuickFormBase {
    */
   public function buildForm(array $form, FormStateInterface $form_state, string $id = NULL) {
 
-    // Egg quantity.
+    // Date+time selection field (defaults to now).
+    $form['timestamp'] = [
+      '#type' => 'datetime',
+      '#title' => $this->t('Date'),
+      '#default_value' => new DrupalDateTime(),
```
wotnak: It might be a good idea to use the current user timezone when setting the default datetime, since by default system timezone or UTC will be used, which might be different from the timezone the user has configured in their account. The same [was done](https://github.com/farmOS/farmOS/pull/665) in quick forms included in farmOS core, so it would be nice to follow these best practices also in this example.

docs/development/module/quick.md
```diff
@@ -160,69 +222,186 @@ Available traits and the methods that they provide include:
...
+ *   confirm_form_route_name = "farm.quick.harvest"
  * )
  */
-class EggHarvest extends QuickFormActionBase {
+class Harvest extends QuickFormActionBase {
 
   /**
    * {@inheritdoc}
    */
   public function getQuckFormId(): string {
```
wotnak: There's a typo in the method name, the `i` in `Quck` is missing. It is however introduced in the [QuickFromActionBase](https://github.com/farmOS/farmOS/blob/2.x/modules/core/quick/src/Plugin/Action/QuickFormActionBase.php#L73) definition directly so would require fixing the code first, but it might be a nice opportunity to this now while already making changes in quick forms.

docs/development/module/quick.md
mstenta: Good call @wotnak - I'll add that and force push the branch shortly.

docs/development/module/quick.md
mstenta: Oops! Good catch @wotnak! That went unnoticed.

Unfortunately it's a bit tricky to fix because there may already be implementations of this base class in the wild. One that I'm aware of is this: https://github.com/Rothamsted-Ecoinformatics/farm_rothamsted/blob/2d8e76f28df8da6dd5a9e108e5bc833e89a5ff2b/modules/farm_rothamsted_quick/src/Plugin/Action/QuickExperimentAction.php#L23

That *may* be the only one (since it was the reason this base class was added originally, so chances are few people know about it), but best practice would be for us to not change it in a way that potentially breaks downstream. I'll chat with @paul121 about it and figure out what we can do.

Thanks for pointing it out!

mstenta: I added a commit to fix this. This IS technically a breaking change, because the method is declared as `abstract` - and there isn't a very easy way to deprecate that and add another, because that would also need to be declared as `abstract`, which would also serve to break existing implementations.

I'm inclined to just fix it, under the assumption that this isn't being used widely downstream yet. I searched ALL code on GitHub (pretty neat that you can do that!), and the only references to it were farmOS core (and forks of it), and the Rothamsted module, which we control: https://github.com/search?q=getQuckFormId&type=code

We'll need to fix it in the Rothamsted module before this gets merged so that doesn't break.

mstenta: Fixed in Rothamsted: https://github.com/Rothamsted-Ecoinformatics/farm_rothamsted/commit/e1e8861776db73fac2b99edac01eb52d698e979c

docs/development/module/quick.md
mstenta: I added timezone in this commit: https://github.com/farmOS/farmOS/pull/703/commits/744fae9bf2e331e2604d2ef095d242c642dc6849

It's a bit of a shortcut because I didn't use dependency injection. I'm a bit torn on that because on the one hand it might be good to show how to do that properly, but on the other hand I really want these examples to be as simple as possible. I even considered removing the `timestamp` field from the examples entirely, but I think it's good to include.

Open to pushback on this, but that's my reasoning. :-)

paul121 (approved):

mstenta:
Force pushed to fix a minor PHPStan issue. This is ready to merge, but I may hold off for v2.2.x.

PR #705, opened 2023-08-09
Wrap all entity labels with Markup::create() in translated log names

### LINKED ISSUE #698: Birth log quick form apostrophe becomes &#039;
**Describe the bug**
When creating a birth log using the quick form if the childs name has an `'` in it the log title replicases the `'` with `&#039;`.

**To Reproduce**
1. go to birth quick form
2. enter childs name with an apostrophe like `suzies's1-1`
3. select any mother asset
4. save the log

**Expected behavior**
log title: `Birth: suzie's1-1`

**Actual Results**
log title: `Birth: suzie&#039;s1-1`

image

farmOS version 2.1.0
tested on windows chrome and android chrome

mstenta:
Fixes https://github.com/farmOS/farmOS/issues/698 (and similar issues in other quick forms and bulk actions).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - [Fix composer.json version constraints for migrate_plus and migrate_tools #702](https://github.com/farmOS/farmOS/pull/702)
+- [Fix birth log quick form apostrophe becomes &#039; #698](https://github.com/farmOS/farmOS/issues/698)
 
 ## [2.1.2] 2023-07-18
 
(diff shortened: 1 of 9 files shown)


paul121 (approved):
TIL! Glad this was an easy fix, will try and remember to use this going forward.

Makes we wonder about changing/adding a `label` helper method on our entities that returns a `MarkupInterface` object instead of string. That wouldn't fix this all of the time (particularly strings with multiple asset names) but might help. Definitely worth considering something like this for the `entityLabelsSummary` function

mstenta:
> Makes we wonder about changing/adding a `label` helper method on our entities that returns a `MarkupInterface` object instead of string. That wouldn't fix this all of the time (particularly strings with multiple asset names) but might help. Definitely worth considering something like this for the `entityLabelsSummary` function

I think what's tricky about doing this (and why core hasn't already) is because we don't know what the context will be where the label is used. The reason we want a `MarkupInterface` object in these cases is specifically because we want to use them inside an auto-generated log name, which we are also passing through the `t()` function. So basically we want to use one entity's label inside another entity's label. No where in this process is the text being printed to the page, which is where escaping of the label MUST happen to avoid XSS. So I'm not sure if there's any way we can make this easier, unfortunately. :-/

PR #706, opened 2023-08-11
farmOS Setup Menu

paul121:
As I proposed on the forum: https://farmos.discourse.group/t/add-setup-page-to-farmos-core/1712


paul121:
@mstenta one thing I'm pondering here is dependencies on `farm_setup`. Right now `farm_settings` does not have any dependents and it is an optional base module. Generally I think `farm_setup` should behave the same, and `farm_settings` should not need to depend on it. But the thoughts I'm having:

- How do we want to enable this module on existing installs? If it is an "optional base module" similar to settings, what is the logic to determine if it is installed? For now I have added an update hook to `farm_settings` that enables `farm_setup`.... eg: "If you previously had settings installed, the setup module will be installed. If you do not want the setup module, it can be uninstalled. "

- With this change both `farm_import` and `farm_settings` do not have a menu or toolbar item available without `farm_setup`. Generally I think this is fine as I expect `farm_setup` to be enabled most all the time... and I like the idea that you could have settings without setup, but maybe that is just a weird use-case that doesn't really matter.
  - FWIW I added a commit that adds this dependency for `farm_import`
  
- _.... Ultimately if we make `farm_settings` depend on `farm_setup` that would simplify a lot of this. But lets not do that unless we really need to?_

modules/core/ui/menu/farm_ui_menu.module
```diff
@@ -29,6 +29,12 @@ function farm_ui_menu_menu_links_discovered_alter(&$links) {
   if (!empty($links['farm.quick:farm.quick'])) {
     $links['farm.quick:farm.quick']['parent'] = 'farm.base';
   }
+
+  // Move the farm.report menu link to the farm.base parent.
```
mstenta: Comment should say "farm.setup" not "farm.report".

modules/core/settings/farm_settings.links.menu.yml
```diff
@@ -0,0 +1,13 @@
+farm.settings:
+  title: Settings
+  description: Configure settings for farmOS and installed modules.
+  parent: farm.setup
```
mstenta, line 11: I think this creates an implicit dependency on `farm_setup`?

modules/core/role/farm_role.managed_role_permissions.yml
```diff
@@ -4,5 +4,4 @@ farm_role:
     - access user profiles
     - change own username
   config_permissions:
-    - access administration pages
```
mstenta, line 7: Are we sure the only reason this permission was granted was for getting to the taxonomy lists?

mstenta (changes requested):
Awesome work @paul121!!

I haven't tested this yet but started to review the code changes...

One thing I'm wondering: do we need a "Setup" toolbar item? Maybe we should just add it to the bottom of the main menu, move "Settings" into there, and remove the separate toolbar item completely. It is a bit of an odd-ball up there as it is - and it made sense when we just had "Settings" - but makes less sense to have a like to "Setup" in two places IMO.

> @mstenta one thing I'm pondering here is dependencies on `farm_setup`. Right now `farm_settings` does not have any dependents and it is an optional base module.

As it is now, I think any module that has a menu item with a parent of `farm.setup` should technically declare it as a dependency. I added an inline comment to that effect in my code review.

I wonder if `farm_setup` should just absorb `farm_settings` into it. "Settings" are a subset of the "Setup" process. And right now the Settings module is for both settings AND enabling farmOS modules. It might be cleaner to just make all of that part of the new, broader, "Setup" module - which can take the place of "Settings" in the list of base modules. We could move all the code from `farm_settings` to `farm_setup` and just leave an empty and deprecated `farm_settings` module for existing sites, along with an update hook to uninstall it - and plan to remove it with farmOS 3.x.

modules/core/settings/farm_settings.links.menu.yml
paul121: Somewhat a dependency, yes, but does not cause any issues. If `farm_setup` is not enabled it puts `farm.settings` at the root of the main menu. But I'm coming around to just adding `farm_setup` as a dependency

paul121:
> One thing I'm wondering: do we need a "Setup" toolbar item?

It's a fair question. I like having it there separate from the menu as "quick access" like in other platforms but agree it isn't necessary. I'll ask on the forum

> I wonder if farm_setup should just absorb farm_settings into it. "Settings" are a subset of the "Setup" process.

Yes this is something we should consider as well. Overall I don't mind two separate modules even if settings is quite simple, basically just provide some infrastructure for the settings menu and tabs on settings pages, kinda like the report or import modules right now. We briefly mentioned (cc @symbioquine) on the dev call possible future ideas for setup pages like having a mechanism to track progress or notify when something is not done and that `farm_setup` could provide helpers to do that. With that in mind I do like keeping `farm_setup` separate from `farm_settings`.

> Ultimately if we make farm_settings depend on farm_setup that would simplify a lot of this. But lets not do that unless we really need to?

I'm coming around to this idea more... it's really the simplest solution as far as update goes, just add a new `farm_setup` module and install it if `farm_settings` was previously installed. Keeps a little bit of code isolation going forward between setup/settings and isn't too complex anyways?

modules/core/role/farm_role.managed_role_permissions.yml
paul121: Yes I do think so! It really only enabled the Administration menu. I've done some testing and all the taxonomy term pages still work without the permission

paul121:
Rebased and added `farm_setup` dependency for `farm_settings` since it was only two more lines

mstenta:
> Rebased and added `farm_setup` dependency for `farm_settings` since it was only two more lines

Cool! I think that makes sense. And I agree keeping `farm_setup` and `farm_settings` as separate modules makes sense.

> > One thing I'm wondering: do we need a "Setup" toolbar item?
> 
> It's a fair question. I like having it there separate from the menu as "quick access" like in other platforms but agree it isn't necessary. I'll ask on the forum

So I think this is the only remaining thing to figure out. Not much feedback on the forum (aside from @gbathree's comment here: https://farmos.discourse.group/t/add-setup-page-to-farmos-core/1712/6) but maybe we can ask folks more directly on the next dev call.

The idea of having "Setup" in both places (on the left in the main menu and in the upper right) feels superfluous. Just having it in the main menu alone will make it available on every page so "quick access" is achieved.

I could maybe see an argument for having "Setup" in the main menu and "Settings" in the toolbar, for quick access to the settings page specifically (as a subset of Setup). But I think I like the idea of simplifying the toolbar even more. :-)

paul121:
> I could maybe see an argument for having "Setup" in the main menu and "Settings" in the toolbar, for quick access to the settings page specifically (as a subset of Setup).

Ha... I would probably suggest the opposite. I think the idea is that Setup pages will be visited more frequently than Settings pages (create content vs configure one-time settings).

But easy to drop the toolbar altogether and see how it goes, it seems like that is consensus on the forum. I like Greg's point that the standard for our UI is the left menu

paul121:
Force-pushed dropping the setup toolbar

mstenta (approved):

mstenta:
Merged.

PR #709, opened 2023-08-23
Add action links to add location assets on locations page

paul121:
Simple improvement that makes it easier to add location assets from the location hierarchy page. I often find myself here wanting to create an asset and need to go back to relevant pages to do so.

location-action-link

### DIFF
--- a/modules/core/ui/location/src/Plugin/Derivative/AddLocationAssetAction.php
+++ b/modules/core/ui/location/src/Plugin/Derivative/AddLocationAssetAction.php
@@ -1,0 +1,76 @@
+<?php
+
+namespace Drupal\farm_ui_location\Plugin\Derivative;
+
+use Drupal\Component\Plugin\Derivative\DeriverBase;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\Plugin\Discovery\ContainerDeriverInterface;
+use Drupal\Core\StringTranslation\StringTranslationTrait;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Defines actions to add location assets on locations page.
+ */
+class AddLocationAssetAction extends DeriverBase implements ContainerDeriverInterface {
+
+  use StringTranslationTrait;
+
+  /**
+   * The entity type manager service.
+   *
+   * @var \Drupal\Core\Entity\EntityTypeManagerInterface
+   */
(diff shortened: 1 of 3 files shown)


paul121:
Force push for a phpstan error

modules/core/ui/location/src/Plugin/Derivative/AddLocationAssetAction.php
```diff
@@ -0,0 +1,77 @@
...
+
+    // Add links to asset/add/type for each asset type.
+    $asset_label = $this->entityTypeManager->getDefinition('asset')->getLabel();
+    foreach ($asset_storage->loadMultiple($asset_type_ids) as $asset_type_id => $asset_type) {
+      $this->derivatives[$asset_type_id] = $base_plugin_definition;
+      $this->derivatives[$asset_type_id]['title'] = $this->t('Add @bundle @entity_type', ['@bundle' => $asset_type->label(), '@entity_type' => $asset_label]);
+      $this->derivatives[$asset_type_id]['route_name'] = 'entity.asset.add_form';
+      $this->derivatives[$asset_type_id]['route_parameters'] = ['asset_type' => $asset_type_id];
+      $this->derivatives[$asset_type_id]['appears_on'][] = 'farm.locations';
+      $this->derivatives[$asset_type_id]['appears_on'][] = 'farm.asset.locations';
```
mstenta: Right now this will show the action link on `/asset/%/locations`, which is where we display the sub-hierarchy of locations under the current asset (if that asset is a location itself). But the action button will just take you to a blank asset creation form without pre-populating anything.

I think as a user, if I saw that button in that context, my expectation would be that it would be for adding more children locations, and so it would pre-populate the `parent` field in the asset creation form.

This would be similar to the way in which clicking "Add Activity log" on an asset automatically references that asset in the log edit form.

This would be nice, but I don't think we need to worry about it as part of this first-step PR. I think we should just omit this line for now and NOT show the button on `farm.asset.locations`. We can consider adding logic to pre-populate the `parent` field in the future.

mstenta (changes requested):

modules/core/ui/location/src/Plugin/Derivative/AddLocationAssetAction.php
paul121: Agreed :+1:

mstenta (approved):

PR #711, opened 2023-08-25
Condense metrics UI

paul121:
Condenses the metrics UI elements

### DIFF
--- a/modules/core/ui/metrics/src/Plugin/Block/FarmMetricsBlock.php
+++ b/modules/core/ui/metrics/src/Plugin/Block/FarmMetricsBlock.php
@@ -82,7 +82,7 @@ class FarmMetricsBlock extends BlockBase implements ContainerFactoryPluginInterf
     $output['asset']['metrics'] = [
       '#type' => 'container',
       '#attributes' => [
-        'class' => 'assets metrics-container',
+        'class' => ['assets', 'metrics-container'],
       ],
     ];
     $metrics = $this->getEntityMetrics('asset');
@@ -107,7 +107,7 @@ class FarmMetricsBlock extends BlockBase implements ContainerFactoryPluginInterf
     $output['log']['metrics'] = [
       '#type' => 'container',
       '#attributes' => [
-        'class' => 'logs metrics-container',
+        'class' => ['logs', 'metrics-container'],
       ],
     ];
     $metrics = $this->getEntityMetrics('log');
--- a/modules/core/ui/metrics/src/Plugin/Block/FarmMetricsBlock.php
+++ b/modules/core/ui/metrics/src/Plugin/Block/FarmMetricsBlock.php
@@ -72,55 +72,41 @@ class FarmMetricsBlock extends BlockBase implements ContainerFactoryPluginInterf
   public function build() {
     $output = [];
 
-    // Create a container for asset metrics.
-    $assets_label = $this->entityTypeManager->getStorage('asset')->getEntityType()->getCollectionLabel();
-    $output['asset']['label'] = [
-      '#type' => 'html_tag',
-      '#tag' => 'h6',
-      '#value' => Link::createFromRoute($assets_label, 'view.farm_asset.page')->toString() . ' (' . $this->t('active') . ')',
-    ];
-    $output['asset']['metrics'] = [
-      '#type' => 'container',
-      '#attributes' => [
+    // Create a list of asset metrics.
+    $assets_label = $this->entityTypeManager->getStorage('asset')->getEntityType()->getCollectionLabel() . ' (' . $this->t('active') . ')';
+    $output['asset'] = [
+      '#theme' => 'item_list',
+      '#title' => Link::createFromRoute($assets_label, 'view.farm_asset.page')->toRenderable(),
+      '#items' => $this->getEntityMetrics('asset'),
+      '#empty' => $this->t('No assets found.'),
+      '#wrapper_attributes' => [
         'class' => ['assets', 'metrics-container'],
       ],
-    ];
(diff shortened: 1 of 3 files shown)


mstenta:

modules/core/ui/metrics/src/Plugin/Block/FarmMetricsBlock.php
```diff
@@ -153,7 +157,7 @@ protected function getEntityMetrics($entity_type) {
 
       $count = $query->count()->execute();
       $route_name = "view.farm_$entity_type.page_type";
-      $metrics[] = Link::createFromRoute($bundle_info['label'] . ': ' . $count, $route_name, ['arg_0' => $bundle], ['attributes' => ['class' => ['metric', 'button']]])->toString();
+      $metrics[] = Link::createFromRoute($bundle_info['label'] . ': ' . $count, $route_name, ['arg_0' => $bundle], ['attributes' => ['class' => ['metric', 'button', 'button--small']]])->toString();
```
mstenta: Cross-posting @wotnak's observation related to accessibility:

> The main thing I would check for in these changes would be [target size](https://wcag.com/developers/2-5-8-target-size-minimum-level-aa/). Done some quick testing, and it looks like this requirement is met. Although for buttons in ‘Metrics’ block on the dashboard currently target size is passing AAA level (at least 44x44px) and after changes will be passing only AA level (at least 24x24px).

https://farmos.discourse.group/t/ui-improvements-condensing-space/1732/5

mstenta (approved):
This looks good to me, in general!

I flagged one issue with accessibility that @wotnak mentioned in the forum, but I don't know if that is a blocker. There are other buttons of similar size on the dashboard already, so perhaps we should consider this a wider issue and discuss broader changes to reach WCAG conformance level AAA for all our buttons.

@wotnak do you agree? Should we merge this and then consider broader changes in a follow-up? Or would that essentially undo these changes, thus making this PR pointless?

modules/core/ui/metrics/src/Plugin/Block/FarmMetricsBlock.php
```diff
@@ -74,13 +74,15 @@ public function build() {
...
-    $output['asset'] = [
-      '#markup' => '<strong>' . Link::createFromRoute($assets_label, 'view.farm_asset.page')->toString() . ' (' . $this->t('active') . ')</strong>',
-      'metrics' => [
-        '#type' => 'container',
-        '#attributes' => [
-          'class' => 'assets metrics-container',
-        ],
+    $output['asset']['label'] = [
+      '#type' => 'html_tag',
+      '#tag' => 'h6',
```
mstenta: Oh this was the other minor issue that @wotnak mentioned in the forum:

> Only other issue I found is heading tags order, on the dashboard after changes there is h1 ‘Dashboard’ and then directly after h6 for entity types in ‘Metrics’ block without other heading levels in between. While this is not strictly required to pass wcag requirements, it is a [good practice](https://www.w3.org/WAI/tutorials/page-structure/headings/#heading-ranks) to use heading levels in sequential order.

mstenta:
Sorry I'm reading through the thread on https://farmos.discourse.group/t/ui-improvements-condensing-space/1732 at the same time as I'm adding comments over there. :-)

It sounds @paul121 and @wotnak are in agreement that the button sizes should not be changed in the "default" style, but instead be made dependent on Gin's density settings.

https://farmos.discourse.group/t/ui-improvements-condensing-space/1732/7

Changing my review to "changes requested" so I remember that. :-)

mstenta (changes requested):
(see above comments)

mstenta:
I will throw this idea out there, while I'm thinking about it: maybe we don't need to use buttons at all for these metrics. What if they were just a simple unordered list of items?

Or maybe we should rethink the usefulness of the metrics dashboard pane in general. It's nice to have numbers at a glance for assets and logs, but it's a step down from what we had in v1 which also had "total acreage". I know we've talked about broader ideas of "metrics" like that, but it's never been implemented, so the dashboard block feels a bit extraneous right now.

Just thinking out loud. :-)

mstenta:
> Technically it’s actually a little challenging because we are trying to keep the gin-specific styling isolated to `farm_ui_theme` and this is provided by `farm_ui_metrics`. It might make sense to create some more dashboard CSS variables like `--dashboard-content-spacing` or ?? which we could override more easily in `farm_ui_theme` without specifically targeting the Metrics. I’m curious your thoughts on this? https://www.drupal.org/project/farm/issues/3291697

(from https://farmos.discourse.group/t/ui-improvements-condensing-space/1732/7)

Just wanted to say: I think it's OK to have some metric-specific stuff in `farm_ui_theme`. We already encapsulate a lot of other styles for specific elements in there. Although I agree it would be better to have more generic styles. Just to say that I'm not opposed to a short-term solution of adding metrics styles to `farm_ui_theme` if it makes this easier...

paul121:
> What if they were just a simple unordered list of items?
> Or maybe we should rethink the usefulness of the metrics dashboard pane in general.

Yes and yes... Didn't want to get too crazy with this PR, just want to condense things as a first step. In the long run if we do add more types of metrics I think this will need more thought. Maybe vertical tabs or details elements so not all of the metrics are visible all the time. Ideally someone could configure it down to only what they want to see.

Let's try a list though... using theme `item_list` actually simplifies the code quite a bit. Neat how you push the title & empty logic out into the component itself. Also I don't think this will require any density settings.

This screenshot doesn't have the other dashboard changes but would greatly benefit from the white background for increased contrast like @wotnak pointed out.
Screenshot from 2023-09-06 17-29-43

mstenta (approved):

PR #712, opened 2023-08-25
Dashboard improvements

paul121:
Condenses the dashboard panes and renders using the standard Gin background + drop shadow

### DIFF
--- a/modules/core/ui/dashboard/css/dashboard_layout.css
+++ b/modules/core/ui/dashboard/css/dashboard_layout.css
@@ -1,8 +1,0 @@
-.layout--twocol .layout__region {
-  box-sizing: border-box;
-  overflow-x: scroll;
-}
-
-.layout--twocol .layout__region--first {
-  padding-right: 1.5em;
-}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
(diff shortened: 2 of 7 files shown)


mstenta:

modules/core/ui/dashboard/css/dashboard_layout.css
```diff
@@ -1,8 +1,40 @@
+/* Add a gap between layout regions. */
+.layout--twocol {
+  --dashboard-spacing: 1rem;
+  gap: var(--dashboard-spacing);
```
mstenta: This creates an implicit dependency on `farm_ui_theme` because that's the only place `--dashboard-spacing` is defined.

Can we move all of these dashboard style tweaks (including the ones that were here originally) to `modules/core/ui/theme/css/dashboard.css` and not worry about styling at all in the `farm_ui_dashboard` module?

mstenta (changes requested):

modules/core/ui/dashboard/css/dashboard_layout.css
paul121: Actually `--dashboard-spacing` is defined here on line 3 as well. We're just overriding the variable in `farm_ui_theme`.

> Can we move all of these dashboard style tweaks (including the ones that were here originally) to modules/core/ui/theme/css/dashboard.css and not worry about styling at all in the farm_ui_dashboard module?

I suppose we could, it would be simpler

paul121: Pushed new commits that first move the styles to `farm_ui_theme` and then add on gin related spacing/styling

mstenta (approved):

PR #713, opened 2023-08-25
Condense views table UI

paul121:
General improvements to the views table UI, exposed filters, and condensing space on pages

### DIFF
--- a/modules/core/ui/views/config/install/views.view.farm_asset.yml
+++ b/modules/core/ui/views/config/install/views.view.farm_asset.yml
@@ -1061,7 +1061,7 @@ display:
             image_target_id:
               align: ''
               separator: ''
-              empty_column: false
+              empty_column: true
               responsive: ''
             id:
               sortable: true
@@ -1087,22 +1087,22 @@ display:
             flag_value:
               align: ''
               separator: ''
-              empty_column: false
+              empty_column: true
               responsive: ''
             parent_target_id:
               align: ''
               separator: ''
-              empty_column: false
+              empty_column: true
               responsive: ''
             owner_target_id:
               align: ''
               separator: ''
-              empty_column: false
+              empty_column: true
               responsive: ''
             location:
               align: ''
               separator: ''
-              empty_column: false
+              empty_column: true
               responsive: ''
             status:
               sortable: true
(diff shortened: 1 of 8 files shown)


mstenta:

modules/core/ui/views/config/install/views.view.farm_asset.yml
```diff
@@ -1061,7 +1061,7 @@ display:
             image_target_id:
               align: ''
               separator: ''
-              empty_column: false
+              empty_column: true
```
mstenta, line 1064: Do we want to set this on bundle-specific columns too? Currently this is only going to hide columns that are in the base View configuration, but we're also adding columns for bundle fields at runtime:

https://github.com/farmOS/farmOS/blob/399c01a177df14d4d2be547e28c6a98dee72fd63/modules/core/ui/views/farm_ui_views.views_execution.inc#L39

We aren't currently touching the `style.options.columns` configuration when we add columns in that hook, but maybe we should be - and without doing that it feels like this improvement is only half implemented. It wouldn't be obvious why some empty columns are hidden and others aren't.

modules/core/ui/theme/css/regions.css
```diff
@@ -0,0 +1,23 @@
+/* Decrease the horizontal margin on medium to large screens. Gin uses xxl. */
```
mstenta: I wonder if we should split this out to another PR so they can be considered separately? As is they would be hidden under the "Views"-specific `CHANGELOG.md` line.

mstenta (changes requested):

modules/core/ui/views/config/install/views.view.farm_asset.yml
paul121: Agreed

modules/core/ui/theme/css/regions.css
paul121: Yes - split to here https://github.com/farmOS/farmOS/pull/719

modules/core/ui/views/config/install/views.view.farm_asset.yml
paul121: This was pretty easy! Also added a commit to let bundle fields be sortable.

mstenta (approved):

Issue #1, opened 2014-07-16
Options for standardizing on vocabularies for interoperability when publishing a Web API


rjcorwin:
It looks like the UN has done some work on creating an ontology for Agriculture, it's an ontology called [Agrovoc (read their about page)](http://aims.fao.org/standards/agrovoc/about). The UN has also sponsored some work in Drupal to get Agrovoc into Drupal.   It's not entirely clear if that vocab is a clear fit for Drupal Farm, it will take some more digging. Here's what I've found so far.
- AgriDrupal installation profile 
  - doesn't look like this was ever finished, [not in the Installation Profiles list on Drupal.org](https://www.drupal.org/project/project_distribution?f[0]=&f[1]=&f[2]=&f[3]=sm_field_project_type%3Afull&text=AgriDupal&solrsort=iss_project_release_usage+desc&op=Search)
  - [about page on the UN ag site](http://aims.fao.org/tools/agridrupal)
  - [AgriDrupal demo site](http://www.agridrupal.org/Agridrupal7.x.7.9/agridrupaltest/agridrupal-current-readme-file)
  - [An announcement from 2013 that AgriDrupal is released but the download link is broken](http://aims.fao.org/community/agridrupal/blogs/agridrupal-based-drupal-7-has-been-released)
  - [An announcement that a new website is using AgriDrupal from 2014](http://aims.fao.org/community/blogs/rural-horizons-digital-library-uses-agridrupal)
- [UN sponsored Agrovoc module which has some commits from June 2014](https://www.drupal.org/project/agrovoc)
- [UN sponsored Agrovocfield module](https://www.drupal.org/project/agrovocfield)

mstenta:
This is great! Thanks RJ.

I hadn't seen AgriDrupal. I took a quick look, but I'm still not quite sure what it's purpose is. It doesn't seem to deal with record keeping. Log into the demo site and take a look at the "node/add" page (http://www.agridrupal.org/Agridrupal7.x.7.9/node/add), and it has a big list of nodetypes, but most seem to deal with management of contacts, conferences, publications, etc. Pretty cool stuff, but not much overlap with the Farm distro, it seems. I need to look closer.

[...]

PR #714, opened 2023-08-26
Add an Account Admin role with permission to administer users and assign managed roles

mstenta:
For consideration: this PR adds a new "Account Admin" role (`farm_account_admin`) to the "farmOS Default Roles" module (`farm_role_roles`) which grants the following permissions:

- `administer farm settings`
- `administer users`
- `assign [role_id] role` (for each of the "managed" farmOS roles)

This is a role that I currently provide on Farmier-hosted farmOS instances, so I thought it might be worth contributing back to core if it is generally useful.

This is helpful in cases where you want to designate certain users as "Account Admins" without giving them full Drupal admin access. Currently, the role with the highest level of permissions in farmOS is "Manager", but they do not have the ability to create new users, edit existing users, or assign roles. On a default farmOS installation on the Drupal admin (uid 1) can do that. This provides a role specifically for that purpose.

### DIFF
--- a/modules/core/role/modules/account_admin/config/install/user.role.farm_account_admin.yml
+++ b/modules/core/role/modules/account_admin/config/install/user.role.farm_account_admin.yml
@@ -1,0 +1,24 @@
+langcode: en
+status: true
+dependencies:
+  enforced:
+    module:
+      - farm_role_account_admin
+  module:
+    - farm_role
+id: farm_account_admin
+label: 'Account Admin'
+weight: 1
+is_admin: false
+permissions:
+  - 'administer farm settings'
+  - 'administer users'
+third_party_settings:
+  farm_role:
+    access:
+      config: true
+      entity:
+        view all: true
+        create all: true
+        update all: true
+        delete all: true
--- a/modules/core/role/modules/account_admin/config/install/user.role.farm_account_admin.yml
+++ b/modules/core/role/modules/account_admin/config/install/user.role.farm_account_admin.yml
@@ -18,7 +18,7 @@ third_party_settings:
     access:
       config: true
       entity:
-        view all: true
-        create all: true
-        update all: true
-        delete all: true
+        view all: false
+        create all: false
+        update all: false
+        delete all: false
--- a/modules/core/role/modules/account_admin/farm_role_account_admin.routing.yml
+++ b/modules/core/role/modules/account_admin/farm_role_account_admin.routing.yml
@@ -4,4 +4,4 @@ farm_role_account_admin.settings:
     _form: '\Drupal\farm_role_account_admin\Form\AccountAdminSettingsForm'
     _title: 'Account Admin Role settings'
   requirements:
-    _permission: 'administer farm settings'
+    _permission: 'configure account admin role'
--- a/modules/core/role/modules/account_admin/farm_role_account_admin.routing.yml
+++ b/modules/core/role/modules/account_admin/farm_role_account_admin.routing.yml
@@ -1,0 +1,7 @@
+farm_role_account_admin.settings:
+  path: 'farm/settings/account-admin'
+  defaults:
+    _form: '\Drupal\farm_role_account_admin\Form\AccountAdminSettingsForm'
+    _title: 'Account Admin Role settings'
+  requirements:
+    _permission: 'administer farm settings'
--- a/modules/core/role/modules/account_admin/tests/src/Functional/UserAccessTest.php
+++ b/modules/core/role/modules/account_admin/tests/src/Functional/UserAccessTest.php
@@ -1,0 +1,39 @@
+<?php
+
+namespace Drupal\Tests\farm_role_account_admin\Functional;
+
+use Drupal\Tests\farm_test\Functional\FarmBrowserTestBase;
+
+/**
+ * Tests access to user 1.
+ *
+ * @group farm
+ */
+class UserAccessTest extends FarmBrowserTestBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  protected static $modules = [
+    'farm_role_account_admin',
+  ];
+
+  /**
+   * Test user 1 access.
+   */
+  public function testUser1Access() {
+
+    // Create and login a user with farm_account_admin role.
+    $user = $this->createUser();
+    $user->addRole('farm_account_admin');
+    $user->save();
+    $this->drupalLogin($user);
+
+    // Confirm that the user cannot access user 1.
+    $this->drupalGet('user/1');
+    $this->assertSession()->statusCodeEquals(403);
+    $this->drupalGet('user/1/edit');
+    $this->assertSession()->statusCodeEquals(403);
+  }
+
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- [Add an Account Admin role with permission to administer users and assign managed roles #714](https://github.com/farmOS/farmOS/pull/714)
+
 ### Changed
 
 - [Use Gin border radius for the map #710](https://github.com/farmOS/farmOS/pull/710)
--- a/composer.json
+++ b/composer.json
@@ -39,6 +39,7 @@
         "drupal/log": "^2.0.2",
         "drupal/migrate_plus": "^6.0",
         "drupal/migrate_tools": "^6.0.2",
+        "drupal/role_delegation": "^1.2",
         "drupal/simple_oauth": "5.2.3",
         "drupal/state_machine": "^1.0",
         "drupal/subrequests": "^3.0.3",
--- a/farm.profile
+++ b/farm.profile
@@ -58,6 +58,7 @@ function farm_modules() {
       'farm_import_kml' => t('KML asset importer'),
       'farm_fieldkit' => t('Field Kit integration'),
       'farm_l10n' => t('Translation/localization features'),
+      'farm_role_account_admin' => t('Account Admin role'),
     ],
   ];
 }
--- a/modules/core/role/modules/account_admin/config/schema/farm_role_account_admin.schema.yml
+++ b/modules/core/role/modules/account_admin/config/schema/farm_role_account_admin.schema.yml
@@ -1,0 +1,7 @@
+farm_role_account_admin.settings:
+  type: config_object
+  label: 'farmOS Account Admin Role settings'
+  mapping:
+    allow_peer_role_assignment:
+      type: boolean
+      label: 'Allow users with the Account Admin role to assign/revoke the Account Admin role.'
--- a/modules/core/role/modules/account_admin/farm_role_account_admin.info.yml
+++ b/modules/core/role/modules/account_admin/farm_role_account_admin.info.yml
@@ -1,0 +1,9 @@
+name: farmOS Account Admin Role
+description: Provides an Account Admin role for managing users.
+type: module
+package: farmOS
+core_version_requirement: ^9
+dependencies:
+  - farm:farm_role
+  - farm:farm_settings
+  - role_delegation:role_delegation
--- a/modules/core/role/modules/account_admin/farm_role_account_admin.links.task.yml
+++ b/modules/core/role/modules/account_admin/farm_role_account_admin.links.task.yml
@@ -1,0 +1,5 @@
+farm_role_account_admin.settings:
+  base_route: farm_settings.settings_page
+  route_name: farm_role_account_admin.settings
+  title: 'Account Admin'
+  weight: 5
--- a/modules/core/role/modules/account_admin/farm_role_account_admin.managed_role_permissions.yml
+++ b/modules/core/role/modules/account_admin/farm_role_account_admin.managed_role_permissions.yml
@@ -1,0 +1,3 @@
+farm_role_account_admin:
+  permission_callbacks:
+    - Drupal\farm_role_account_admin\AccountAdminPermissions::permissions
--- a/modules/core/role/modules/account_admin/farm_role_account_admin.module
+++ b/modules/core/role/modules/account_admin/farm_role_account_admin.module
@@ -1,0 +1,22 @@
+<?php
+
+/**
+ * @file
+ * Hooks implemented by the farmOS Account Admin Role module.
+ */
+
+use Drupal\Core\Access\AccessResult;
+use Drupal\Core\Entity\EntityInterface;
+use Drupal\Core\Session\AccountInterface;
+
+/**
+ * Implements hook_ENTITY_TYPE_access().
+ */
+function farm_role_account_admin_user_access(EntityInterface $entity, $operation, AccountInterface $account) {
+
+  // Only user 1 can access user 1.
+  if ($entity->id() == 1 && $account->id() != 1) {
+    return AccessResult::forbidden();
+  }
+  return AccessResult::neutral();
+}
--- a/modules/core/role/modules/account_admin/farm_role_account_admin.permissions.yml
+++ b/modules/core/role/modules/account_admin/farm_role_account_admin.permissions.yml
@@ -1,0 +1,2 @@
+configure account admin role:
+  title: 'Configure Account Admin role'
--- a/modules/core/role/modules/account_admin/src/AccountAdminPermissions.php
+++ b/modules/core/role/modules/account_admin/src/AccountAdminPermissions.php
@@ -1,0 +1,66 @@
+<?php
+
+namespace Drupal\farm_role_account_admin;
+
+use Drupal\Core\DependencyInjection\ContainerInjectionInterface;
+use Drupal\farm_role\ManagedRolePermissionsManagerInterface;
+use Drupal\user\RoleInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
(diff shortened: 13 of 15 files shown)


mstenta:
The only other thought with this is maybe we want it to be in a separate module from `farm_role_roles`, so that it can still be used in cases where the other default roles are not (Manager, Worker, Viewer). Curious to hear what others think - happy to split it out if that would be valuable.

mstenta:
@paul121 This would also simplify your module: https://github.com/paul121/farm_site_config

Notably, I decided to name this role "Account Admin" instead of just "Admin" to avoid any confusion with Drupal core's concept of "Admin" roles, which imply much higher level of access.

mstenta:
Ah just realized this will need an update hook to create the new role on existing installs. Unless we decide to move it to a separate module...

mstenta:
I split it out to a separate optional `farm_role_account_admin` module. Ready for review!

paul121 (commented):
Cool, I agree this would be useful in core. It's really an essential part to hosting farmOS IMO. And yes to keeping it as a separate module.

What do you think about including [userprotect](https://www.drupal.org/project/userprotect) as well? With the rule to prevent users from modifying account #1. As I recall this is a fairly important thing to have in place if you're hosting farmOS. I know we don't want to make too many decisions re: hosting, but are there really any situations when you wouldn't want this?

mstenta:
> What do you think about including [userprotect](https://www.drupal.org/project/userprotect) as well? With the rule to prevent users from modifying account https://github.com/farmOS/farmOS/issues/1. As I recall this is a fairly important thing to have in place if you're hosting farmOS.

This is a really good point @paul121 - I forgot that adding `administer users` permission means that anyone with this new role can edit user 1 - VERY BAD!

In Farmier I prevent this with `hook_ENTITY_TYPE_access()`:

```
/**
 * Implements hook_ENTITY_TYPE_access().
 */
function farmier_user_access(EntityInterface $entity, $operation, AccountInterface $account) {

  // Only user 1 can access user 1.
  if ($entity->id() == 1 && $account->id() != 1) {
    return AccessResult::forbidden();
  }
  return AccessResult::neutral();
}
```

It looks like that's roughly how `userprotect` works, except it also adds a lot of other features for fine-grained user protection rules. What do you think about just adding the hook above to this new module, rather than adding the whole `userprotect` module as a dependency?

Side note: it looks like there might be a conflict between `userprotect` and `role_delegation` (found on the `userprotect` project page): https://www.drupal.org/project/userprotect/issues/1984520

mstenta:
Sorry for the noise - after I split it out to a separate module I broke the tests - I should have set it back to draft until I got those worked out. Should be green now.

I added a fixup commit that adds a dependency on `farm:farm_settings`, because this role ALSO adds the `administer farm settings` permission. This is a carry over from Farmier, and it sort of stretches the purpose of this module, so curious if we should rethink that or not. As it stands now, the `administer farm settings` permission is not added to any managed roles, so it's up to the host to decide who gets it. Feels like it makes sense to be in this "Account Admin" role to me, but open to feedback on that.

I also added a commit that implements the hook I described in my previous comment to restrict access to user 1, along with a functional test that confirms access is denied at `user/1` and `user/1/edit` (if there are other things worth checking let me know and I'll add it).

Adding this restriction on user 1 also expands the scope/purpose of this module a bit IMO. I wonder if `farm_role_account_admin` is the right name for it with these added bits? Or if we should describe it more generally. I'm also happy to commit this as is for now and cross that bridge if we come to it.

symbioquine (commented):
Does Drupal already enforce that users cannot create/modify users with permissions they don't have themselves? We may also not want this account admin permission to be "recursive"?

I'm imagining that the change as-is could let users create new users who could create new users, etc...

Even worse would be a privilege escalation where a user with account admin could grant themselves roles they aren't supposed to have - or create new users for themselves with those roles...

mstenta:
The Account Admin role would grant a user the ability to: a) create new users, b) edit existing users (including changing their email address, password, etc., and c) grant any "managed farmOS role" to any user (including themselves).

A "managed farmOS role" is one that is "managed" by the `farm_role` module, and has it's permissions automatically assigned and maintained - as opposed to a normal ("unmanaged") Drupal role, which has permissions assigned by clicking checkboxes in the admin UI.

The three "managed" roles that farmOS ships with are: Manager, Worker, Viewer. None of these have admin-level Drupal permissions.

So yea, this "Account Admin" role is VERY privileged in the sense that they have a lot of control over other users. But, critically, they cannot assign "unmanaged" roles to users, and they cannot create/edit roles themselves. They can only assign "managed" roles, which are restricted in their permissions.

Drupal has a dedicated permission called `administer permissions` that grants the user access to the form for creating new roles and assigning permissions to them. That's the really dangerous one from a privilege escalation standpoint.

Now... if someone who is hosting a farmOS site added their own "managed" role via a custom module, which granted `administer permissions`, then they would certainly be opening up a privilege escalation vulnerability. But that would also be true if they created an "unmanaged" role with that permission. In both cases that's the responsibility of the host/developer of said module.

mstenta:
@symbioquine hopefully that explains a bit more how this works... but I want to answer your questions more specifically too:

> Does Drupal already enforce that users cannot create/modify users with permissions they don't have themselves?

There are two relevant Drupal permissions: `administer users` and `administer permissions`

The `administer users` permission grants broad control over users generally. This includes the ability to edit any user. But it does not grant the ability to assign roles.

The `administer permissions` permission grants the ability to build roles, with sets of permissions assigned to them, AND to assign those roles to users. This is the really dangerous one, and it should only ever be granted to someone trusted because it easily allows for privilege escalation. Generally speaking you never assign this to anyone, unless you need to have additional "admin" users beyond just user 1.

In the context of farmOS, without the `administer users` permission, only user 1 can add new users. Assigning that permission allows them to create users, but it still doesn't allow them to grant roles. And because everything in farmOS requires a role, that is essentially useless. The new user would be able to log in but would get "Access Denied" everywhere. But... we still don't want to grant the `administer permissions` permission to anyone, because that essentially means the ability to escalate to full admin status. That is where the [Role Delegation](https://www.drupal.org/project/role_delegation) module comes in. It adds new permissions for EACH role: `assign [role] role`

So with this new Account Admin module, we provide a new "Account Admin" role, and automatically assign the `assign manager role`, `assign worker role`, and `assign viewer role` permissions to it (along with any other "managed" roles that might be in the system). But we do NOT add permission to add "unmanaged" roles.

> We may also not want this account admin permission to be "recursive"?
> I'm imagining that the change as-is could let users create new users who could create new users, etc...

This is true. Someone with the "Account Admin" role CAN create a new user AND assign them the "Account Admin" role as well, which would then allow that person to create more. But that would be their choice. They don't need to assign the "Account Admin" role, and can instead just assign a lower role like "Manager", "Worker", or "Viewer".

> Even worse would be a privilege escalation where a user with account admin could grant themselves roles they aren't supposed to have - or create new users for themselves with those roles...

Yea I think we're safe from that by virtue of the fact that we never assign the `administer permissions` permission.

symbioquine:
> hopefully that explains a bit more how this works... but I want to answer your questions more specifically too

Yeah, that's helpful and makes sense!

I'd argue that the "Account admin" role shouldn't "be recursive" by default. My gut feeling is that the average site admin granting that role wouldn't intend for it to allow an unbounded set of additional account admins without some additional input/action from the site admin.

Another similar scenario is whether the "Account admin" role should be allowed to remove that role from other account admins. If so, it creates additional risk in granting that role - i.e. not just that they'd make bad account administration decisions for new/existing regular users, but that they could modify or lock out the existing account admins.

Maybe it would be wise to add a check such that the account admin role cannot create or modify accounts that have the account admin role or grant that role to other accounts?

paul121:
> Side note: it looks like there might be a conflict between userprotect and role_delegation (found on the userprotect project page): https://www.drupal.org/project/userprotect/issues/1984520

Thanks for sharing. It is a D7 issue but could totally still apply to D8, I only sorta follow it.

> I added a fixup commit that adds a dependency on farm:farm_settings, because this role ALSO adds the administer farm settings permission... Feels like it makes sense to be in this "Account Admin" role to me, but open to feedback on that.

For simplicity and the majority of use-cases I do agree. But perhaps it makes sense to create a separate role that is specifically for `administer farm settings`, just to support an instance that wants to separate who is responsible for account admin and farm settings? I wonder what Rothamsted might think of this.

> Adding this restriction on user 1 also expands the scope/purpose of this module a bit IMO. I wonder if farm_role_account_admin is the right name for it with these added bits?

The bummer with this is that it is hard-coded behavior. User protect is nice because it is configuration that could be disabled. But I agree it is overly complex for what we need. I would be OK committing this now as well.

paul121:
> I'd argue that the "Account admin" role shouldn't "be recursive" by default.
> Maybe it would be wise to add a check such that the account admin role cannot create or modify accounts that have the account admin role or grant that role to other accounts?

Glad you brought this up @symbioquine. I agree, specifically that the account admins the "farm owner" (not user 1) promotes should not be recursive. But there is a challenge here for the "hosting provider" where you DO want the "farm owner" to be able to create more account admins. So we're back to having a distinction of the hosting specific "farm/instance admin" vs more general "account admin" I think? :thinking:

mstenta:
> The bummer with this is that it is hard-coded behavior. User protect is nice because it is configuration that could be disabled. But I agree it is overly complex for what we need.

Yea that is the drawback that was on my mind too. Then again - I think it's probably OK since it only targets user 1. So `userprotect` could also be used for other users/purposes. Or worst case you just don't use this Account Admin module and roll your own.

> But there is a challenge here for the "hosting provider" where you DO want the "farm owner" to be able to create more account admins. So we're back to having a distinction of the hosting specific "farm/instance admin" vs more general "account admin" I think? 🤔

That was my first thought too. The purpose of this module is essentially to delegate account management from user 1 to other lower-level users.

Worth mentioning: this module will NOT be installed by default in farmOS. It is listed under the "Optional" modules, so it would need to be installed manually by a site admin to use it. Not sure if that was clear or not, or if it changes any of these considerations.

mstenta:
We talked about this on the dev call today and @paul121 had a good idea...

What if we added a setting to the module: "Allow Account Admins to assign Account Admin role"

This could be disabled by default, to address the "recursive" concern that @symbioquine brought up. And then hosts could decide to allow it if they want.

symbioquine:
> What if we added a setting to the module: "Allow Account Admins to assign Account Admin role"
> 
> This could be disabled by default, to address the "recursive" concern that @symbioquine brought up. And then hosts could decide to allow it if they want.

A setting seems like the best of both worlds!

Maybe it should also protect account admins from modifying the accounts of other account admins though too? If so, that should be made clear by the text for the setting...

Maybe one of:

* "[ ] Allow Account Admins to designate or modify other Account Admins"
* "[ ] Allow peer-management for Account Admins"

The description could be something like:

"*Allows users with the Account Admin role to add that role to other users or otherwise modify existing users that have the Account Admin role*"

mstenta:
I pushed a commit that I started working on yesterday which adds a simple setting:

Screenshot from 2023-09-01 12-51-01

I chatted with @symbioquine a bunch about the idea to "restrict Account Admins from editing other Account Admins", but ultimately decided that it might be hard to achieve that without adding another heavy-handed `AccessResult::forbidden()` in `hook_user_access()`, which might produce unexpected results (eg: if `administer users` permission were assigned to a different role, access would still be denied).

We can dig deeper into that idea in a follow-up, if it comes up as a request, but this seems like a good first step.

What do you think @symbioquine @paul121?

modules/core/role/modules/account_admin/farm_role_account_admin.routing.yml
```diff
@@ -0,0 +1,7 @@
+farm_role_account_admin.settings:
+  path: 'farm/settings/account-admin'
+  defaults:
+    _form: '\Drupal\farm_role_account_admin\Form\AccountAdminSettingsForm'
+    _title: 'Account Admin Role settings'
+  requirements:
+    _permission: 'administer farm settings'
```
paul121: Hmm I wonder if this should be a different permission... as is, any account admin could come in and change this setting. Basically it doesn't prevent the recursive issue.

If we introduce a new permission that no standard/managed farmOS role provides, then this setting would be entirely under the control of the host/user 1. They could modify the setting OR provide a new role w/ this new permission that allows trusted users to modify this setting.

paul121 (changes requested):

modules/core/role/modules/account_admin/farm_role_account_admin.routing.yml
mstenta: Yes I was just commenting on that. I think that a new permission is the right idea. I'll add a commit, standby...

mstenta: https://github.com/farmOS/farmOS/pull/714/commits/e8f69d33b5a6034dad89f555e7a3f77579123c7d

paul121 (commented):
The last concern I have is that I think we need to invalidate some kind of cache when this `allow_peer_role_assignment` setting changes. This is because the setting adds/removes a permission to the role and it is possible to cache things based on the role and/or user permissions.

I'm not quite sure the best way to solve this but I do think we need to do something. This CR might be relevant and help explain: https://www.drupal.org/node/2459373

mstenta:
I actually found that same CR when I was testing this @paul121! :smile: 

I tested that changing the setting does automatically add/remove the permission when viewing the `/admin/people/permissions` page. In our kernel test we needed to rebuild the kernel simply because we are caching lists of role permissions in our `ManagedRolePermissionsManager` class: https://github.com/farmOS/farmOS/blob/81d6c61b4949916d2cf649c363d27a687c90f166/modules/core/role/src/ManagedRolePermissionsManager.php#L130

But maybe I'm not imagining other scenarios. How can we test this?

mstenta:
I think our case might be a little different than others, because we are overriding the `RoleStorage::hasPermission()` method with our own, which happens after normal role permissions (within the `user_role` config entity) get cached... I think? There are a lot of layers here so I definitely may be overlooking something.

mstenta:
@paul121 https://github.com/farmOS/farmOS/pull/714/commits/0caaf57f719a75628ebd819b4ba252784e18c471

modules/core/role/modules/account_admin/config/install/user.role.farm_account_admin.yml
```diff
@@ -0,0 +1,24 @@
...
+  - 'administer users'
+third_party_settings:
+  farm_role:
+    access:
+      config: true
+      entity:
+        view all: true
+        create all: true
+        update all: true
+        delete all: true
```
symbioquine: I wonder if it makes sense to grant these write permissions?

Without them, it opens up the possibility of having a user Alice who operates in an administrative capacity to add/modify other users, but doesn't interact with the farm data itself. Obviously Alice could modify her own account to also add the `farm_manager` role, but not having these permissions by default could act like a safety net where she would have to go out of her way to act outside that administrative capacity.

On the other hand, as-is this maintains the superset relationship between each successively more permissive role. i.e. `account_admin > manager > worker > viewer`. That's probably a bit easier to document/explain, but I personally see it as a bit unintuitive that account admin would implicitly give create/edit permissions on assets.

modules/core/role/modules/account_admin/tests/src/Functional/UserAccessTest.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+namespace Drupal\Tests\farm_role_account_admin\Functional;
+
+use Drupal\Tests\farm_test\Functional\FarmBrowserTestBase;
+
+/**
+ * Tests access to user 1.
+ *
+ * @group farmier
```
symbioquine: Wrong group?

mstenta: Good catch! I force pushed to fix that.

modules/core/role/modules/account_admin/config/install/user.role.farm_account_admin.yml
mstenta: Mm yea these are good points. I don't feel strongly either way. @paul121 do you have any preference?

It just means that for new Farmier instances I would assign both the Account Admin and Manager roles to the first user that gets created, rather than just Account Admin. Although.. I guess it would mean I would also need to bulk update all existing instances to add the Manager role to Account Admin users, which is a bit of a pain, but that's my problem not farmOS core's... just thinking out loud. :-)

mstenta: Giving this some more thought: I suppose I'm actually a bit in favor of limiting the permissions - it's nice to keep the roles focused. and it helps to demonstrate that roles "layer" on one another. I'll push a quick commit that makes this change and we can roll it back if @paul121 or other disagree. :-)

mstenta: https://github.com/farmOS/farmOS/pull/714/commits/e3228f08ac7b82015adb1634dac17fe154bab211

Note: this is leaving `config: true`, but setting all entity permissions to `false`. This is because we're also using this role for the `administer farm settings` permission, which sort shines new light on a consideration I mentioned earlier:

> this role ALSO adds the `administer farm settings` permission. This is a carry over from Farmier, and it sort of stretches the purpose of this module

symbioquine (approved):
LGTM

paul121 (approved):

mstenta:
Merged!

PR #715, opened 2023-08-31
Exclude block.block.gin_content from farm_update

paul121:
This fixes an issue I've encountered where the block layout can be configured (specifically the "Content" region) and be left in a "broken state" where the "Powered by farmOS" block is positioned *above* the main content. Also according to the comment in `farm_ui_theme_farm_update_exclude_config` I think this should have already been excluded :-)

This isn't a normal thing users will experience but is a concern for UI development because it leaves things in a "broken" state... For me, I was playing with adding a new block to the top of "Content" region. After I removed the block and tried to "reset" to normal things worked fine. BUT *after clearing the cache* my development instance was left with "Powered by farmOS" at the top each page (above "Main page content". There does not appear to be any way to fix this via the UI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - [Fix composer.json version constraints for migrate_plus and migrate_tools #702](https://github.com/farmOS/farmOS/pull/702)
 - [Fix birth log quick form apostrophe becomes &#039; #698](https://github.com/farmOS/farmOS/issues/698)
+- [Exclude block.block.gin_content from farm_update #715](https://github.com/farmOS/farmOS/pull/715)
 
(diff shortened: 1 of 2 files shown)


paul121:
I think this happens because saving the block layout form assigns new weights to the block config entities based on the user's configuration. For me, it seems that the weights all got set to negative values, not sure how, but this is just using the UI. So this makes sense that after rebuilding the cache, the `block.block.gin_content` weight is reset to `0`, which is "lower" than the other negative values. Even though we install "Powered by farmOS" with weight 1000 it seems that this gets reset/condensed later when making changes via the UI. 

Now I see that I could "fix" this by toggling the row weights in the block form and assigning proper values but that's not really a proper work around.
Screenshot from 2023-08-31 13-37-42

mstenta:
Makes sense to me - thanks @paul121!

I added a `CHANGELOG.md` and amended the commit message to include the PR number.

PR #720, opened 2023-09-07
Content form and vertical tabs

paul121:
Implements the gin content form and moves fields to vertical tabs within the form and sidebar.

More discussion here: https://farmos.discourse.group/t/ui-improvements-editing-forms/1743?u=paul121

There is one existing issue in Gin that may block this PR: https://www.drupal.org/project/gin/issues/3342164
- Without the node module I found that the Gin content form doesn't work, simply because the `node_edit_form` template isn't registered, even though Gin overrides the template. There is a simple fix that Gin should implement.
- In the meantime @wotnak showed me an even simpler way we can register the `node_edit_form` (see `farm_ui_theme_theme` in this PR)
- However... in the end a warning message is still logged because Gin depends on a Node css/js library that doesn't exist. Despite the error everything still works though, and removing this one line in Gin solves the issue: https://git.drupalcode.org/project/gin/-/blame/8.x-3.x/src/GinContentFormHelper.php#L209

### DIFF
--- a/modules/core/field/src/FarmFieldFactory.php
+++ b/modules/core/field/src/FarmFieldFactory.php
@@ -833,7 +833,7 @@ class FarmFieldFactory implements FarmFieldFactoryInterface {
 
     // Build form and view display settings.
     $field->setDisplayOptions('form', [
-      'type' => 'options_buttons',
+      'type' => 'options_select',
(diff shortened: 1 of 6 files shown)


paul121:
Closing in favor of #770 so tests can run

PR #722, opened 2023-09-30
farmOS v2 CSV import module

mstenta:
See https://www.drupal.org/project/farm/issues/3382539


modules/core/import/modules/csv/farm_import_csv.info.yml
```diff
@@ -0,0 +1,21 @@
...
+  - farm:farm_location
+  - farm:farm_log_asset
+  - farm:farm_log_quantity
+  - farm:farm_parent
+  - farm:farm_quantity_standard
+  - log:log
+  - migrate_plus:migrate_plus
+  - migrate_source_csv:migrate_source_csv
+  - migrate_source_ui:migrate_source_ui
+  - farm:quantity
```
paul121: nit, sort farm:quantity alphabetically

modules/core/import/modules/csv/tests/src/Functional/CsvImportTest.php
```diff
@@ -0,0 +1,85 @@
...
+   * {@inheritdoc}
+   */
+  protected static $modules = [
+    'farm_import_csv_test',
+  ];
+
+  /**
+   * Test CSV importers.
+   */
+  public function testCsvImporters() {
```
paul121: Rename to clarify this is only testing access logic?

modules/core/import/modules/csv/src/Access/CsvImportMigrationAccess.php
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function create(ContainerInterface $container) {
+    return new static(
+      $container->get('plugin.manager.migration'),
+    );
+  }
```
paul121: I do not believe this `create()` method is necessary since we are not implementing the ContainerFactory interface, instead the dependencies are wired in `services.yml`

paul121 (commented):
Just a few initial things, still reviewing the later UI related commits

modules/core/import/modules/csv/farm_import_csv.info.yml
mstenta: https://github.com/farmOS/farmOS/pull/722/commits/d84413ad123e8c55affe413f03e6d241560e2b26

modules/core/import/modules/csv/tests/src/Functional/CsvImportTest.php
mstenta: https://github.com/farmOS/farmOS/pull/722/commits/e35eeea3bf4c799bedfcf0f76ce267d8c495af95

modules/core/import/modules/csv/src/Access/CsvImportMigrationAccess.php
mstenta: https://github.com/farmOS/farmOS/pull/722/commits/139707b72f78a3f6af6903321b5488ca04f6d027

modules/core/import/modules/csv/src/Controller/CsvImportController.php
```diff
@@ -94,4 +136,126 @@ public function index(): array {
...
+      foreach ($migration['third_party_settings']['farm_import_csv']['columns'] as $info) {
+        if (!empty($info['name'])) {
+          $item = '<strong>' . $info['name'] . '</strong>';
+          if (!empty($info['description'])) {
+            $item .= ': ' . $info['description'];
+          }
+          $items[] = Markup::create($item);
+        }
+      }
+      $template_link = Link::createFromRoute($this->t('Download template'), 'farm.import.csv.template', ['migration_id' => $migration_id]);
```
paul121: It seems that a link to the template is only rendered if there are column names/descriptions in third party settings. Probably a safe assumption but perhaps something that could be teased out later.

modules/core/import/modules/csv/src/Controller/CsvImportController.php
```diff
@@ -94,4 +136,126 @@ public function index(): array {
...
+      $build['columns']['descriptions'] = [
+        '#theme' => 'item_list',
+        '#items' => $items,
+        '#suffix' => '<p>' . $template_link->toString() . '</p>',
```
paul121: I recently learned there is an `html_tag` render element, could we try and use this for rendering simple html elements? https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Render%21Element%21HtmlTag.php/class/HtmlTag

```php
$build['template'] = [
  '#type' => 'html_tag',
  '#tag' => 'p',
  '#value' => $template_link->toString(),
];
```

modules/core/import/modules/csv/src/Plugin/Derivative/CsvImportMigrationBase.php
```diff
@@ -119,4 +130,93 @@ public function getDerivativeDefinitions($base_plugin_definition): array {
...
+        }
+
+        break;
+    }
+
+    // If the field supports multiple values, explode on comma delimiter
+    // as a first step and describe how to format values.
+    if ($field_definition->getCardinality() === -1 || $field_definition->getCardinality() > 1) {
+      array_unshift($process, ['plugin' => 'explode', 'delimiter' => ',']);
+      $description[] = 'Multiple values can be separated by commas with the whole cell wrapped in quotes.';
```
paul121: Translate this string?

modules/core/import/modules/csv/src/Plugin/Derivative/CsvImportMigrationBase.php
```diff
@@ -119,4 +130,93 @@ public function getDerivativeDefinitions($base_plugin_definition): array {
...
+    }
+
+    // If the field is required, make note of that in the column description.
+    if ($field_definition->isRequired()) {
+      $description[] = 'Required.';
```
paul121: Translate this string?

modules/core/import/modules/csv/tests/src/Kernel/AssetCsvImportTest.php
```diff
@@ -47,18 +49,27 @@ public function testAssetCsvImport() {
     $expected_values = [
       2 => [
         'name' => 'Old tractor',
+        'manufacturer' => '',
+        'model' => '',
+        'serial_number' => '',
```
paul121, line 54: Should we provide a value for one of these bundle fields? Otherwise I don't think this test is demonstrating that bundle fields are correctly populated.

modules/core/import/modules/csv/src/Plugin/Derivative/CsvImportMigrationBase.php
```diff
@@ -171,6 +172,31 @@ protected function addBundleField($field_definition, &$mapping, &$columns): void
...
+        // Asset reference.
+        if ($field_definition->getSetting('target_type') == 'asset') {
+          $plugin = [
+            'plugin' => 'asset_lookup',
+          ];
+          if (!empty($field_definition->getSetting('handler_settings')['target_bundles'])) {
+            $plugin['bundle'] = array_keys($field_definition->getSetting('handler_settings')['target_bundles']);
+          }
+          $process[] = $plugin;
+          $description[] = 'Accepts asset names, ID tags, UUIDs, and IDs.';
```
paul121: Translate this string?

modules/core/import/modules/csv/src/Plugin/Derivative/CsvImportMigrationLog.php
```diff
@@ -32,4 +32,21 @@ protected function alterProcessMapping(array &$mapping, string $bundle): void {
...
+   * {@inheritdoc}
+   */
+  protected function alterColumnDescriptions(array &$columns, string $bundle): void {
+    parent::alterColumnDescriptions($columns, $bundle);
+
+    // Add allowed quantity measure values.
+    foreach ($columns as &$column) {
+      if ($column['name'] == 'quantity measure') {
+        $allowed_measures = array_keys(quantity_measures());
+        $column['description'] .= ' Allowed values: ' . implode(', ', $allowed_measures);
```
paul121: Translate?

modules/core/import/modules/csv/src/Plugin/Derivative/CsvImportMigrationAsset.php
```diff
@@ -30,6 +30,41 @@ protected function alterProcessMapping(array &$mapping, string $bundle): void {
...
+  /**
+   * {@inheritdoc}
+   */
+  protected function alterColumnDescriptions(array &$columns, string $bundle): void {
+    parent::alterColumnDescriptions($columns, $bundle);
+
+    // Describe the ID tag columns.
+    $columns[] = [
+      'name' => 'id tag',
+      'description' => 'ID tag.',
```
paul121: Translate these descriptions?

paul121 (commented):
This looks pretty good to me! Just a few small things

modules/core/import/modules/csv/src/Controller/CsvImportController.php
mstenta: That's true. We can't generate a template without knowing the column names.

modules/core/import/modules/csv/src/Controller/CsvImportController.php
mstenta: Awesome! Wasn't familiar with `html_tag` - that's really helpful! I will push a fixup commit.

modules/core/import/modules/csv/tests/src/Kernel/AssetCsvImportTest.php
mstenta: Oops! Forgot to finish that. Good catch! I will push a fixup.

mstenta:
Thanks for the review @paul121! I pushed fixup commits to address your feedback.

modules/core/import/modules/csv/src/Controller/CsvImportController.php
```diff
@@ -94,4 +136,126 @@ public function index(): array {
...
+      $response->headers->set('Content-Type', 'application/csv');
+      $response->headers->set('Content-Disposition', 'attachment; filename="' . $filename . '"');
+      $column_names = [];
+      if (!empty($migration['third_party_settings']['farm_import_csv']['columns'])) {
+        foreach ($migration['third_party_settings']['farm_import_csv']['columns'] as $column) {
+          if (!empty($column['name'])) {
+            $column_names[] = $column['name'];
+          }
+        }
+      }
```
paul121: Could be simplified:
```suggestion
      $column_names = array_filter(array_column($migration['third_party_settings']['farm_import_csv']['columns'] ?? [], 'name'));
```

modules/core/import/modules/csv/src/Controller/CsvImportController.php
paul121: Ah yes I see. We are not expecting all of the keys in the process pipeline to be included in the CSV. :+1:

modules/core/import/modules/csv/src/Controller/CsvImportController.php
mstenta: Ah great! Fixup on the way...

modules/core/import/modules/csv/config/install/views.view.farm_import_csv_asset.yml
paul121, line 1: The views of imported CSVs include the `private://` path in the rendered link. I don't think this provides much value to users so ideally we wouldn't need to show that. I tried using the views `File: Filename` field formatted as a link to the file. This works, but doesn't capture very well how the same filename may have been uploaded multiple times and renamed for distinctness.

filename-differences
views-filename


This could be solved by looking at the Imported timestamp. I also think we should render this timestamp in a more user-friendly format. Shown is the "Default short date format":
imported-long

imported-format

paul121: TLDR; the `Filename` should be as simple as possible without adding confusion for the user (`private://csv/...`).... but also we should be able to demonstrate that the same file uploaded multiple times results in distinct uploads. I think a "clean" `Filename` combined with row number + imported timestamp is sufficient, but don't have a strong opinion.

I can just envision someone asking "what is private://...?" ... "what does this timestamp mean? what is T? what is +0000" - it should be all be simple as possible

mstenta: Ah yea I actually started with `File: Filename` and decided to switch to URI for the reason you described (@symbioquine pointed out that it would better to show the actual filename that was created on the server on one of our dev calls). That was an improvement in that regard IMO, but it did have the tradeoff of adding the full `private://csv/..` prefix. I was able to justify that to myself by saying "well it emphasizes that the files are private! ha!" :-P

What do you think about leaving this as-is for now and tackling this as a UI/UX followup improvement?

mstenta: > Shown is the "Default short date format"

I've avoided this in v2 because of complaints we've had in the past about the default format and the placement of month vs date (in EU the standard is 06/10/2023 not 10/06/2023), so we standardized on HTML5 format for everything in v2. For this, the only extra thing we needed was time so the thought process was: use the HTML5 format that includes time.

I'd love to talk more broadly about date/time formatting, but I think we should be consistent and tackle that everywhere.

paul121 (commented):
Minor issue with the ID Tag type. This might be a separate from CSV import. I can import assets with `other`, an empty ID Tag type, and an invalid `wrong` ID tag type all without failure. However, via the UI I can only select valid ID tag type values, or leave it blank. I think the CSV import should fail with an invalid ID tag type.

After importing a tag type of `wrong` it seems to be saved and is visible in the asset form, but not visible in the asset display. When saving the asset form there is no validation error which makes me think this is a broader issue.

wrong-id-tag-type
wrong-id-tag-form

paul121 (approved):
Overall this is GREAT. Things I want to note for future followup:

- Improve imported views display (above comments)
- Improve menu link hierarchy, so not one long list
  - Where do "custom" CSV importers go?
- Since this is so easy to use, kinda want ability to import geometries via CSV. But maybe that is not what we should be encouraging

mstenta:
> Minor issue with the ID Tag type. This might be a separate from CSV import.

Oh good catch! I will open up a separate issue for that and try to take a look at it ASAP.

> Overall this is GREAT. Things I want to note for future followup:

Thanks for the review! And the list of followups! I have a list myself, which I didn't include here... I'll comment on the forum topic to update the rest of the community on this development and start brainstorming next steps together!

mstenta:
https://farmos.discourse.group/t/csv-importers-in-v2-x/942/45

Here's the raw list of follow-ups I had... I'll start a new forum topic after 2.2.0 is released so we can continue the conversation and get more feedback from folks.

- Test https://www.drupal.org/project/migrate_tools/issues/3257230 + https://www.drupal.org/project/migrate_source_ui/issues/3257936 (sounds like this could affect large imports that have errors/skipped rows)
- Preventing duplicates (came out of chat discussions)
  - Prevent creating terms that already exist (by name)
  - Prevent creating assets that already exist (by name)
- Support importing `owner` field on logs and assets
- Support importing `geometry` field on logs and `intrinsic_geometry` on assets
- Support importing boolean bundle fields.
- Support boolean-ish values for asset/log `status` fields.
- In `asset_lookup`, if there are multiple matches, take one with highest asset ID (this will require changes in `entity_lookup`)
- Move `asset_lookup` and `term_lookup` plugins to `farm_migrate` so they can be used outside of `farm_import_csv` by other migrations.
- Refresh CSV menus/routers when `migration` config entities are created/deleted/updated

mstenta:
> Minor issue with the ID Tag type. This might be a separate from CSV import.

- https://github.com/farmOS/farmOS/issues/725

PR #723, opened 2023-10-05
Add a Group membership assignment quick form

mstenta:
This adds a simple Group Membership quick form. It is basically a copy+paste of the existing Movement quick form, with the geometry/map stuff removed. Other than that all the code and tests are essentially the same.

The motivation for this is twofold:

1) Make it easier to assign group membership (and easier to find than the current bulk action).
2) Prepare to replace the existing Group and Move bulk actions with ones that redirect to the Move and Group quick forms.

I've already started on the refactoring necessary for (2), so we should be able to merge #676 ahead of v2.2.0 and then finish that up before 3.0.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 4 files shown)


paul121 (approved):

mstenta:
Merged.

PR #728, opened 2023-10-09
Fix Quick form actions cause RouteNotFoundException: Route farm.quick.[id] does not exist.

### LINKED ISSUE #727: Quick form actions cause RouteNotFoundException: Route farm.quick.[id] does not exist.
- farmOS v2.2.0 included this PR to enable "Configurable Quick Forms": https://github.com/farmOS/farmOS/pull/707
- One of the commits in that PR merged all quick form routes together into a single `farm_quick.quick_form` route: https://github.com/farmOS/farmOS/pull/707/commits/3505e80124a2fe365eab5510fd2585f9ead9e69d
- This change broke quick form actions, each of which point to a specific route via their `confirm_form_route_name` annotation.
- There isn't a simple fix for that because Drupal core doesn't seem to support passing additional parameters to `confirm_form_route_name` routes, so there is no way to refactor quick form actions so that they pass the `quick_form` route parameter (afaict): https://git.drupalcode.org/project/drupal/-/blob/df5948229a8dc10bef675acb31ee939037b863ca/core/modules/views/src/Plugin/views/field/BulkForm.php#L431

mstenta:
Fixes #727

This reverts https://github.com/farmOS/farmOS/commit/3505e80124a2fe365eab5510fd2585f9ead9e69d and adds an event subscriber to rebuild quick form routes when quick form config entities are saved.

### DIFF
--- a/modules/core/quick/tests/src/Functional/QuickFormTest.php
+++ b/modules/core/quick/tests/src/Functional/QuickFormTest.php
@@ -146,6 +146,9 @@ class QuickFormTest extends FarmBrowserTestBase {
     ]);
     $config_entity->save();
 
+    // Rebuild routes.
+    \Drupal::service('router.builder')->rebuildIfNeeded();
+
     // Go to the quick form index and confirm that the requires_entity_test
     // quick form item is visible.
     $this->drupalGet('quick');
@@ -157,6 +160,15 @@ class QuickFormTest extends FarmBrowserTestBase {
     $this->drupalGet('quick/requires_entity_test');
     $this->assertSession()->statusCodeEquals(200);
     $this->assertSession()->pageTextContains($this->t('Test field'));
+
+    // Delete the config entity and confirm that it is removed.
+    $config_entity->delete();
+    \Drupal::service('router.builder')->rebuildIfNeeded();
+    $this->drupalGet('quick');
+    $this->assertSession()->statusCodeEquals(200);
+    $this->assertSession()->pageTextNotContains($this->t('Test requiresEntity quick form'));
+    $this->drupalGet('quick/requires_entity_test');
+    $this->assertSession()->statusCodeEquals(404);
   }
 
 }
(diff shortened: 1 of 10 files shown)


modules/core/quick/src/EventSubscriber/QuickFormConfigSubscriber.php
paul121, line 1: We should also update the routes when config is deleted (I think that is separate from Save events?)

Did some searching and found an interesting way `jsonapi_extras` rebuilds JSON:API routes when their config entity changes, using `JsonapiResourceConfig::postSave`/`JsonapiResourceConfige::postDelete` functions:

https://git.drupalcode.org/project/jsonapi_extras/-/blob/8.x-3.x/src/Entity/JsonapiResourceConfig.php?ref_type=heads#L72

I think we could add `QuickFormInstance::postSave` and `QuickFormInstance::postDelete` like:

```php
  /**
   * {@inheritdoc}
   */
  public static function postDelete(EntityStorageInterface $storage, array $entities) {
    parent::postDelete($storage, $entities);
    \Drupal::service('router.builder')->setRebuildNeeded();
  }
```

modules/core/quick/tests/src/Functional/QuickFormTest.php
```diff
@@ -146,6 +146,9 @@ public function testQuickForms() {
     ]);
     $config_entity->save();
 
+    // Rebuild routes.
+    \Drupal::service('router.builder')->rebuildIfNeeded();
+
```
paul121, line 151: Would be nice if this line wasn't required in tests, but makes sense it may be. Seems like the next request in this test should trigger a rebuild of the routes but maybe that isn't the case? Just want to make sure we tried

paul121 (changes requested):

modules/core/quick/src/EventSubscriber/QuickFormConfigSubscriber.php
mstenta: Ah cool! I'll give that a try instead of the event subscriber.

modules/core/quick/tests/src/Functional/QuickFormTest.php
mstenta: Yea that line is required for the tests to pass. I just retested locally with and without to confirm.

paul121 (approved):

paul121:
Oh, codesniffer issues

PR #734, opened 2023-10-11
Create multi-stage dockerfile

AlMaVizca:
- Creating files for PHP configurations to improve legibility

I would also like to refactor the `build-farmOS.sh` script to:

- Install composer packages in another stage to only copy the vendor folder
- Remove git and zip packages from the image, using something similar to PHP_GEOS

Please let me know if it would be ok to add all that on this PR, or you prefer to create another issue for it.

Edited according to this [comment](https://github.com/farmOS/farmOS/pull/734#issuecomment-1775908092)


docker/Dockerfile
```diff
@@ -1,6 +1,32 @@
 # Inherit from the Drupal 9 image on Docker Hub.
+FROM drupal:9.5 as geos
+# Install git and unzip (needed to build PHP GEOS and needed by Composer).
+# RUN apt-get update \
+#   && apt-get install -y git
+
+ARG PHP_GEOS_VERSION=e77d5a16abbf89a59d947d1fe49381a944762c9d
```
AlMaVizca: Would it be ok to use master branch?
I've checked and that commit has been merged

mstenta: I think my preference would be to keep this as-is for now. As far as I can tell, the `master` branch does not add any new features or fix any bugs, it looks to be just development related changes, so unpinning may introduce more risk than benefit at this time. Ideally we could use a tagged release, but since they have not tagged anything since 1.0.0 in 2017 I'm not sure we can expect new releases. This bears further discussion, though, so maybe it should be a separate PR. That would also make it easier for us to link to independently, if we want to make an official request for a new release from the `php-geos` maintainers.

mstenta (changes requested):
Thanks for tackling this @AlMaVizca! It's funny because just the other day I was thinking about refactoring this to use a multi-stage build.

Can we split this up into multiple atomic commits to make it easier to review and to keep each "intention" separate? I like to try to keep the changes in each commit very focused (even to the point of separating "whitespace" changes out to their own commits).

It seems like the primary changes are:

- Moving PHP configurations to separate files to improve legibility
- Reduce image size with multi-stage build process
- Run container as www-data user

> I would also like to refactor the `build-farmOS.sh` script to:
> - Install composer packages in another stage to only copy the vendor folder
> - Remove git and zip packages from the image, using something similar to PHP_GEOS
> 
> Please let me know if it would be ok to add all that on this PR, or you prefer to create another issue for it.

I'm curious what this would look like, but maybe we should keep it separate for now.

I also need to think a little deeper about how all of these changes may affect images that extend this base image.

And lastly, right now the focus is on `docker/Dockerfile`, but we also have `docker/dev/Dockerfile`, which may also need to be considered to make sure nothing breaks.

Thanks again for diving in and proposing next steps!

docker/Dockerfile
```diff
@@ -10,38 +29,20 @@ ARG PROJECT_VERSION=2.x
...
 #Add custom configurations
 COPY conf.d/ /usr/local/etc/php/conf.d
 
-# Install postgresql-client so Drush can connect to the database.
 RUN apt-get update \
-  # See https://stackoverflow.com/questions/51033689/how-to-fix-error-on-postgres-install-ubuntu
-  && mkdir -p /usr/share/man/man1 \
-  && mkdir -p /usr/share/man/man7 \
-  && apt-get install -y postgresql-client
+    && apt-get install -y git unzip postgresql-client\
```
AlMaVizca: Git and unzip are required until composer is taken to another building stage and farmOS is fetched from a released package

docker/Dockerfile
```diff
@@ -10,38 +29,20 @@ ARG PROJECT_VERSION=2.x
...
-  && docker-php-ext-enable geos
-
 #Add custom configurations
 COPY conf.d/ /usr/local/etc/php/conf.d
 
-# Install postgresql-client so Drush can connect to the database.
 RUN apt-get update \
-  # See https://stackoverflow.com/questions/51033689/how-to-fix-error-on-postgres-install-ubuntu
-  && mkdir -p /usr/share/man/man1 \
-  && mkdir -p /usr/share/man/man7 \
```
AlMaVizca, line 66: I've checked and the error is not happening

AlMaVizca:
It's my pleasure to cooperate, also it's my way to understand the system and see if I can use it :)

> I'm curious what this would look like, but maybe we should keep it separate for now.
> 
> I also need to think a little deeper about how all of these changes may affect images that extend this base image.
> 
> And lastly, right now the focus is on `docker/Dockerfile`, but we also have `docker/dev/Dockerfile`, which may also need to be considered to make sure nothing breaks.

About these, I see them interconnected. Thanks for pointing out the dev version, I didn't notice it earlier.
In my ideal, I would have a multi-stage Dockerfile like:
```
From drupal:<version> as dependencies   (one for geos another one for compose)
....
FROM drupal:<version> as dev (here all the dependencies will be integrated and extra configurations added)
....
FROM drupal:<version> (Here  it would be the final copy, mainly from dev with all the sources and dependencies, without the specific dev code)
```
With this template of Dockerfile and without parameters on the build, the prod version image will be created, and using the following on docker-compose.development.yaml will stop the build at that point.
```
version: '3'
services:
  ....
  www:
    depends_on:
      - db
    build:
      context: ./
      target: dev
```

If you consider it, I can handle those changes, not sure what are the other image dependencies, but I'm willing to collaborate to get them working

docker/Dockerfile
```diff
@@ -33,31 +33,8 @@ RUN apt-get update && apt-get install -y libgeos-dev \
...
-    echo 'realpath_cache_ttl=3600'; \
-  } > /usr/local/etc/php/conf.d/realpath_cache-recommended.ini
-
-# Set recommended OPcache for maximum performance in Symfony applications.
-# See https://symfony.com/doc/current/performance.html#configure-opcache-for-maximum-performance
-# @todo
-# Remove this when https://github.com/docker-library/drupal/pull/156 is merged.
-RUN sed -i 's|opcache.memory_consumption=128|opcache.memory_consumption=256|g' /usr/local/etc/php/conf.d/opcache-recommended.ini \
-  && sed -i 's|opcache.max_accelerated_files=4000|opcache.max_accelerated_files=20000|g' /usr/local/etc/php/conf.d/opcache-recommended.ini
+#Add custom configurations
```
mstenta: Nit: Add a space between `#` and `Add`.

docker/conf.d/opcache-recommended.overrides.ini
```diff
@@ -0,0 +1,6 @@
+# Set recommended OPcache for maximum performance in Symfony applications.
```
mstenta, line 1: Since this is changing from a search and replace of `opcache-recommended.ini` to a new `opcache-recommended.overrides.ini` file, can you take a screenshot of the relevant variables in `/admin/reports/status/php` to demonstrate that it is still working as expected? For example, here is a screenshot before this change:

Screenshot from 2023-10-12 08-22-30

docker/conf.d/realpath_cache-recommended.ini
```diff
@@ -0,0 +1,5 @@
+# Set recommended realpath_cache settings.
+# See https://www.drupal.org/docs/7/managing-site-performance/tuning-phpini-for-drupal
+
```
mstenta: Nit: Remove extra newline.

docker/conf.d/farmOS-recommended.ini
```diff
@@ -0,0 +1,9 @@
+# Set recommended PHP settings for farmOS.
+# See https://farmos.org/hosting/installing/#requirements
```
mstenta: Unrelated to your changes, but I'm noticing that this link is slightly wrong. It should be https://farmos.org/hosting/install/#server-requirements

Can you fix this in a separate commit please?

docker/Dockerfile
```diff
@@ -1,3 +1,22 @@
+# Inherit from the Drupal 9 image on Docker Hub.
+FROM drupal:9.5 as geos
+# Install git and unzip (needed to build PHP GEOS and needed by Composer).
+# RUN apt-get update \
```
mstenta: Are you sure `git` and `unzip` are no longer needed? If I remember correctly we need `unzip` during `composer install` (in `build-farmOS.sh`) for unpacking the `farmOS-map` library: https://github.com/farmOS/farmOS/blob/010a895092b30423b5b66ea06aa8f552ce8b4c18/composer.libraries.json#L13

`git` is needed in the `farmos/farmos:2.x-dev` Docker image, although perhaps that can be added in that `Dockerfile` instead of in this base one.

docker/Dockerfile
```diff
@@ -10,38 +29,20 @@ ARG PROJECT_VERSION=2.x
...
-    && make \
-    && make install \
-  ) \
-  && rm -r php-geos \
-  && docker-php-ext-enable geos
-
 #Add custom configurations
 COPY conf.d/ /usr/local/etc/php/conf.d
 
-# Install postgresql-client so Drush can connect to the database.
```
mstenta, line 62: Let's leave this comment so it is clear why we are installing `postgresql-client`.

docker/Dockerfile
```diff
@@ -10,38 +29,20 @@ ARG PROJECT_VERSION=2.x
...
 RUN apt-get update \
-  # See https://stackoverflow.com/questions/51033689/how-to-fix-error-on-postgres-install-ubuntu
```
mstenta, line 64: Is this no longer an issue? This should be a separate change/commit with explanation as to why we can remove these lines.

docker/Dockerfile
```diff
@@ -1,3 +1,22 @@
+# Inherit from the Drupal 9 image on Docker Hub.
+FROM drupal:9.5 as geos
+# Install git and unzip (needed to build PHP GEOS and needed by Composer).
+# RUN apt-get update \
+#   && apt-get install -y git
+
+ARG PHP_GEOS_VERSION=e77d5a16abbf89a59d947d1fe49381a944762c9d
```
mstenta: Can we put this below the comments that follow for code block consistency?

docker/Dockerfile
```diff
@@ -10,38 +29,20 @@ ARG PROJECT_VERSION=2.x
...
 #Add custom configurations
 COPY conf.d/ /usr/local/etc/php/conf.d
 
-# Install postgresql-client so Drush can connect to the database.
 RUN apt-get update \
-  # See https://stackoverflow.com/questions/51033689/how-to-fix-error-on-postgres-install-ubuntu
-  && mkdir -p /usr/share/man/man1 \
-  && mkdir -p /usr/share/man/man7 \
-  && apt-get install -y postgresql-client
+    && apt-get install -y git unzip postgresql-client libgeos-c1v5 \
```
mstenta: Oh I just noticed that `git` and `unzip` installs moved down here. We need to preserve the reasoning for each package with comments. Perhaps separating them with newlines and adding a comment above each one? For example:

```
# Install apt dependencies.
RUN apt-get update && apt-get install -y \
        # Install git and unzip (needed by Composer).
        git unzip \
        # Install postgresql-client so Drush can connect to the database.
        postgresql-client \
        # Install libgeos-c1v5 ... ?
        libgeos-c1v5 \
    && rm -rf /var/lib/apt/lists/* \
    && apt-get clean
```

What is `libgeos-c1v5` for? We didn't have that before. Is that what you are referring to in the commit message with "Add missing dependency"? We should include more detail for future reference.

docker/Dockerfile
```diff
@@ -55,11 +55,12 @@ RUN chmod a+x /usr/local/bin/build-farmOS.sh
 # Change the ownership of the sites directory and copy the farmOS codebase into /opt/drupal.
 RUN mkdir /var/farmOS \
   && /usr/local/bin/build-farmOS.sh --no-dev \
-  && chown -R www-data:www-data /var/farmOS/web/sites \
+  && chown -R www-data:www-data /var/farmOS/web/sites /var/www/html/sites\
   && rm -r /opt/drupal && cp -rp /var/farmOS /opt/drupal
 
 # Set the entrypoint.
```
mstenta, line 64: Can we split this block up a bit more with additional comments, now that we are doing more than just "setting the entrypoint"?

Perhaps:

```
# Copy the entrypoint script and make it executable.
COPY docker-entrypoint.sh /usr/local/bin/
RUN chmod u+x /usr/local/bin/docker-entrypoint.sh

# Run as www-data user.
USER www-data

# Set the entrypoint and command.
ENTRYPOINT ["docker-entrypoint.sh"]
CMD ["apache2-foreground"]
```

docker/Dockerfile
```diff
@@ -55,11 +55,12 @@ RUN chmod a+x /usr/local/bin/build-farmOS.sh
 # Change the ownership of the sites directory and copy the farmOS codebase into /opt/drupal.
 RUN mkdir /var/farmOS \
   && /usr/local/bin/build-farmOS.sh --no-dev \
-  && chown -R www-data:www-data /var/farmOS/web/sites \
+  && chown -R www-data:www-data /var/farmOS/web/sites /var/www/html/sites\
```
mstenta: What is this change for? Why is it necessary?

mstenta (changes requested):
Thanks @AlMaVizca! I reviewed in more detail and left some questions and change requests. Most are small nitpick changes. Overall I think this is on the right track!

I'd like to get @paul12's1 and @symbioquine's thoughts on this before we merge it too.

docker/conf.d/opcache-recommended.overrides.ini
AlMaVizca: Sure, 
Screenshot from 2023-10-13 13-10-27

docker/Dockerfile
AlMaVizca: Right, On the geos stage they are not needed, they are still required to run the build-farmOS.sh script, and composer.
Eventually, changing those to dependencies stages it will take them out from the final image too

docker/Dockerfile
AlMaVizca: Good catch, thanks.
I got a message during the installation that sites was not writtable, and I thought that was the path, because it fixed the issue.
Now that I've checked again, I've found that
- `/var/www/html is a symlink to /opt/drupal/web`
- the docker-compose.development.yml file uses a volume on that path
```    
volumes:
   - './www:/opt/drupal'
```
So the change is not required.

docker/Dockerfile
AlMaVizca: Sounds good, I've made this change in an extra commit.

docker/Dockerfile
```diff
@@ -81,7 +65,11 @@ RUN mkdir /var/farmOS \
   && rm -r /opt/drupal && cp -rp /var/farmOS /opt/drupal
 
 # Set the entrypoint.
-COPY docker-entrypoint.sh /usr/local/bin/
-RUN chmod u+x /usr/local/bin/docker-entrypoint.sh
```
AlMaVizca, line 85: Changing the permissions on the file, and persisting that into git helps to reduce this extra layer on the docker image

docker/Dockerfile
```diff
@@ -62,8 +62,11 @@ RUN mkdir /var/farmOS \
...
+COPY --chown=www-data docker-entrypoint.sh /usr/local/bin/
+
+# Run as wwww-data user
 USER www-data
+
+#Set the entrypoint and command.
```
mstenta: Nit: space after `#`

docker/Dockerfile
```diff
@@ -62,8 +62,11 @@ RUN mkdir /var/farmOS \
   && rm -r /opt/drupal && cp -rp /var/farmOS /opt/drupal
 
 # Set the entrypoint.
-COPY docker-entrypoint.sh /usr/local/bin/
-RUN chmod u+x /usr/local/bin/docker-entrypoint.sh
+COPY --chown=www-data docker-entrypoint.sh /usr/local/bin/
+
+# Run as wwww-data user
```
mstenta: Nit: change from `wwww-data` to `www-data`

mstenta: Also please add a period at the end of the comment. There is one other comment that needs a period too apart from this one. Can you fix both please and make sure there aren't others I missed?

docker/Dockerfile
```diff
@@ -62,8 +62,11 @@ RUN mkdir /var/farmOS \
   && rm -r /opt/drupal && cp -rp /var/farmOS /opt/drupal
 
 # Set the entrypoint.
-COPY docker-entrypoint.sh /usr/local/bin/
-RUN chmod u+x /usr/local/bin/docker-entrypoint.sh
+COPY --chown=www-data docker-entrypoint.sh /usr/local/bin/
```
mstenta: Oh this is neat! I wasn't aware of the `COPY --chown` option! I see that you also changed the mode of the file from `0644` to `0755` in the repo itself. Makes sense! :+1:

mstenta (changes requested):
Thanks so much @AlMaVizca! I just noted a few tiny nits. I will trigger the workflow run so we can see if this works, or if it breaks anything downstream of it, like the `2.x-dev` image.

AlMaVizca:
My pleasure.
I saw the workflow failing in two not related tasks:
- No added records on the changelog
- Failing to build dev image.

Do I need to change anything there?

mstenta:
Thanks @AlMaVizca! I think this is looking good!

> I saw the workflow failing in two not related tasks:
> 
> - No added records on the changelog
> - Failing to build dev image.
> 
> Do I need to change anything there?

Don't worry about the changlog failure... we have that to remind us to add notes about new features, bug fixes, etc - but I generally don't include internal dev tooling changes like this.

We do need to fix the dev image build, however. That is a core part of our `deliver.yml` workflow, because it is used to run our tests, static code analysis, etc.

It is based on the `farmos/farmos:2.x` image, so we probably just need to make some adjustments based on the changes in this PR.

It looks like the first thing it failed on was the change from `opcache-recommended.ini` to `opcache-recommended.overrides.ini`:

```
#5 [ 2/12] RUN sed -i 's|opcache.revalidate_freq=60|opcache.revalidate_freq=0|g' /usr/local/etc/php/conf.d/opcache-recommended.ini
#5 0.266 sed: couldn't open temporary file /usr/local/etc/php/conf.d/sedHHw5A9: Permission denied
#5 ERROR: process "/bin/sh -c sed -i 's|opcache.revalidate_freq=60|opcache.revalidate_freq=0|g' /usr/local/etc/php/conf.d/opcache-recommended.ini" did not complete successfully: exit code: 4
```

https://github.com/farmOS/farmOS/actions/runs/6505581983/job/17679156363

There may be other failures after that too, but that's what caused the test to abort.

We do a funny thing with the `2.x-dev` image regarding the `www-data` user, to make local development easier, which may need to be considered. We override the user ID of the `www-data` user inside the image so that it matches the user ID of the host machine user. On Linux this is typically uid `1000`, but we supply an `ARG` for end-users/developers to build with a different uid if they need something different. The reason we do this is so that all of the files in `/opt/drupal/` are owned by `www-data` inside the container, and the host user outside the container, so that they can load the entire codebase into their IDE and do development without having to mess with permissions.

https://github.com/farmOS/farmOS/blob/010a895092b30423b5b66ea06aa8f552ce8b4c18/docker/dev/Dockerfile#L18

You'll notice that in the `2.x-dev` image build we actually change to the `www-data` user before running `build-farmOS.sh` (as well as creating a few debug config files), but then change back to `root` user at the end. Perhaps there's a better way to meet all these needs... :thinking: Curious if you have any thoughts! But if you want to just make it work for now so we can merge this as a first step that's fine too.

AlMaVizca:
Thanks, @mstenta.
I think that it's best to have only one Dockerfile with an intermediate step for the dev version and having production as a final image. From what you explained, and looking at the workflows, it will require a change there too. So to not make this a big PR my proposal is to fix the dev  image, let's make sure everything is working as you expected in this first step.

Later I'll work on other 2 PRs later:
- Adapt the workflow and images to fit in one docker image
- Standardize the project scaffolding, while there is no actual 'standard' for it. I've been working on an article to simplify the spin up of any project. The idea is simple, using Makefile and docker along with docker-compose, any project, regarding the language and framework, has named actions that are custom per project. Just to give an example:
```
make start   #Inside the makefile it will define how to run the development environment
make deploy #Deploy according to project specifics
make test #Setup and perform test
```
So the workflow could be standard, let's say, call:
```
step 1: make test
step 2: make security-check
step 3: make deploy
```
I got a little deviated, so you can consider this one or not.

docker/dev/conf.d/opcache-revalidate-freq.ini
```diff
@@ -0,0 +1,3 @@
+# Set OPcache's revalidation frequency to 0 seconds for development.
+# See https://www.php.net/manual/en/opcache.configuration.php#ini.opcache.revalidate-freq
+opcache.revalidate_freq=0
```
mstenta: Nit: needs a newline at the end of file.

docker/dev/files/phpstan.neon
```diff
@@ -0,0 +1,9 @@
+# Configure PHPStan.
+parameters:
+  level: 1
+  ignoreErrors:
+    # new static() is a best practice in Drupal, so we cannot fix that.
+    - '#^Unsafe usage of new static#'"; \
```
mstenta: Remove `"; \` left over from copy+paste.

mstenta (changes requested):
Thanks again @AlMaVizca! I left two small comments on things I noticed in the diffs, and some more thoughts below... I know I'm asking for a lot of changes - and I appreciate all the time you're contributing to this! These are good improvements, and the `Dockerfile`s are critical pieces of our development and packaging infrastructure, so it is important to get right. :-)

I'm a proponent of the "don't break `HEAD`" rule when creating commits. I like each commit to be a self-contained incremental change, which could be merged by itself (in sequence) without breaking things (not always possible, but a good discipline to try to stick to IMO).

With that in mind, can we split up some of the changes in your last two commits so that they can be merged into relevant previous commits?

For example, the commit labelled "Moving PHP configurations to separate files to improve legibility" can include the config file changes being made to the `dev` image as well, I think. So essentially we have one commit that handles splitting out files for both images. And this way, that commit by itself could in theory be merged without breaking tests.

If we can approach each commit this way, that would be great - and also make it easier for me to review them as incremental improvements. I may even be able to merge some of them into `2.x` independently, while others await further review/change requests by other maintainers.

Right now it's a bit hard for me to juggle the latest two commits in my head because they both make changes to the way `USER` is handled in the image. It seems like the first commit breaks things, and the second commit intends to fix them. If we can rearrange things a bit to make the steps more clear-cut that would be easier to wrap my head around.

Lastly, I think it makes sense to split out `phpcs.xml` and `phpstan.neon`, but I'm not sure we should deviate from the current approach of using `sed` for the `phpunit.xml` file specifically. Currently we are not maintaining that entire file in our codebase. We are simply copying from the Drupal core `phpunit.xml.dist` file, and overriding specific bits. If we move to maintaining the whole file we lose the benefit of inheriting changes that may be introduced in upstream Drupal core. So even though it is a bit ugly in our `Dockerfile`, I think it is better to leave that one for now.

mstenta:
Thanks @AlMaVizca - looking good!

I pulled your branch down locally and made a few small changes myself. Mostly small adjustments and moving some things around (the commits themselves and some changes moved/split out). My goal was to make the commit messages "tell the story" of the changes and make each commit simple and focused. Not much different from what you had originally, so hopefully it makes sense! I left your commit author credit on all of them. Please take a moment to review and let me know if you see anything wrong or confusing.

I moved the three "small" changes to the front, and put the three bigger ones (config files + multi-step + `www-data` user) last - although the last one now isn't very big at all, from a `git diff` perspective, which is nice.

The only notable deviation from your commits is that I tried to simplify the changes for the `www-data` user commit in the `dev` `Dockerfile`. So instead of moving things around it is just adding some additional `USER` commands where needed. I figure it might be worth following up to combine some of our `RUN` commands more generally, to reduce the number of layers that are created, but that can be separate, and for now this keeps these changes very simple and focused.

Let me know what you think! I'll start the tests running to see if anything fails. I tested building both images locally and they worked.

mstenta (dismissed):

mstenta:
Docker image size differences:

**farmos/farmos:2.x**:

Before: 1,023,905,894 bytes
After: 1,005,086,774 bytes
Difference: -18,819,120 bytes

**farmos/farmos:2.x-dev**:

Before: 1,446,640,084 bytes
After: 1,427,806,248 bytes
Difference: -18,833,836 bytes

So we're saving about 18 mb in size, with the multi-step build process (which now only splits out the GEOS build). So not quite as big of an improvement as @AlMaVizca's original PR, but still a bit slimmer. :-)

mstenta:
Oops looks like PHP Codesniffer step failed... it seems that the configuration file did not get used. I'll take a look and see if there's something obvious...

mstenta:
Apart from that, it also seems that the tests are getting stuck on the "Wait until www container is ready" step.

This waits for the `docker-entrypoint.sh` script to create a file to signify that everything is ready. So something is broken with that as well.

mstenta:
Pushed fixes... let's see if these pass.

To fix PHP Codesniffer and PHP Stan I moved the  `COPY ./files/ /var/farmOS/` so that it runs after `build-farmOS.sh`.

To fix tests waiting for the container to be ready, I removed the `--chown www-data` from the `COPY docker-entrypoint.sh /usr/local/bin/` command, because `docker-entrypoint.sh` needs to be owned by `root`.

mstenta:
Hmm still hanging on tests... I have to step away but I'll try to see if I can figure out what's going wrong...

docker/Dockerfile
```diff
@@ -1,4 +1,20 @@
-# Inherit from the Drupal 9 image on Docker Hub.
+# Use PHP 8 image to build GEOS PHP extension.
+FROM php:8.1-cli as geos
```
AlMaVizca: Why did you prefer to use `php:8.1-cli` image?
I don't see the downside of using `drupal:9.5`, and whenever a Drupal with a new version of PHP is required, the image definition here doesn't need to be updated.
Maybe it would be more clear when to see with the multi-stages Dockerfile for multiple environments that I'm opening the PR soon.

AlMaVizca:
> So we're saving about 18 mb in size, with the multi-step build process (which now only splits out the GEOS build). So not quite as big of an improvement as @AlMaVizca's original PR, but still a bit slimmer. :-)

Right, I've probably made the mistake of comparing it with the dev image, I don't recall pulling a stable image

I'm going to dive deep into the workflow to replicate it locally and try to find the error

AlMaVizca (commented):
> So we're saving about 18 mb in size, with the multi-step build process (which now only splits out the GEOS build). So not quite as big of an improvement as @AlMaVizca's original PR, but still a bit slimmer. :-)

Right, I've probably made the mistake of comparing it with the dev image, I don't recall pulling a stable image

I'm going to dive deep into the workflow to replicate it locally and try to find the error

docker/Dockerfile
mstenta: It just felt like the right thing to do since building GEOS itself does not require any of the additional complexity included in the `drupal` image layers. We could potentially change it to `php:8-cli` (instead of `8.1`), so that we don't need to update it again (until PHP 9 + whatever future Drupal version that supports that). I doubt that it matters if GEOS is built with `8.1` vs `8.2` vs `8.x`... I imagine it would work the same when we copy it into the `drupal` image layer. But maybe that's not correct and there is logic in the GEOS build that varies based on minor PHP version.

For what it's worth, Drupal 9 only works with PHP 8.1, and is only supported until the end of this month. We are working on farmOS 3.x on Drupal 10 right now, which uses PHP 8.2 as the basis of the `drupal:10.1` image. Drupal itself does not support PHP 8.3 yet, although that is "planned" for Drupal 10.2.

https://www.drupal.org/docs/getting-started/system-requirements/php-requirements

This is just to say: we are already quite intentional about the PHP version that we use in these images. It doesn't change often, and when it does we also need to update other things. So it won't be adding much additional maintenance to use the `php` image for this IMO.

All that said, I don't feel very strongly about this, so if there is agreement that it is simple to use the same `drupal` image for both stages, I'm fine with that too.

mstenta:
> I'm going to dive deep into the workflow to replicate it locally and try to find the error

Thanks! I'm a bit stumped on it. I tried to replicate locally as well, but could not. The key pieces are:

1. The `docker-entrypoint.sh` script is used as the image's `ENTRYPOINT`: https://github.com/AlMaVizca/farmOS/blob/c721a2ae7c1e3a7c91d36f333074955146db3897/docker/Dockerfile#L66
2. The `docker-entrypoint.sh` script will create a file to indicate that the container is "ready" for tests to run if the `FARMOS_FS_READY_SENTINEL_FILENAME` environment variable is set: https://github.com/AlMaVizca/farmOS/blob/c721a2ae7c1e3a7c91d36f333074955146db3897/docker/docker-entrypoint.sh#L23
3. The `docker-compose.testing.common.yml` configuration sets `FARMOS_FS_READY_SENTINEL_FILENAME` to `/opt/drupal/www-container-fs-ready`: https://github.com/AlMaVizca/farmOS/blob/c721a2ae7c1e3a7c91d36f333074955146db3897/docker/docker-compose.testing.common.yml#L8
4. We run the containers using `docker-compose.testing.common.yml` in our `deliver.yml` workflow here: https://github.com/AlMaVizca/farmOS/blob/c721a2ae7c1e3a7c91d36f333074955146db3897/.github/workflows/deliver.yml#L121
5. The `deliver.yml` workflow waits for that file to be created here (**this is the part that's not working**): https://github.com/AlMaVizca/farmOS/blob/c721a2ae7c1e3a7c91d36f333074955146db3897/.github/workflows/deliver.yml#L127

This is what I did to test locally:

1. Build the `farmos/farmos:2.x` and `farmos/farmos:2.x-dev` images locally:

        docker build -t farmos/farmos:2.x docker/
        docker build -t farmos/farmos:2.x-dev docker/dev/

2. Run `bash` in the `farmos/farmos:2.x-dev` image with `FARMOS_FS_READY_SENTINEL_FILENAME` environment variable set to `/opt/drupal/www-container-fs-ready`:

        docker run --rm -it -e FARMOS_FS_READY_SENTINEL_FILENAME=/opt/drupal/www-container-fs-ready farmos/farmos:2.x-dev bash

3. Check for the existence of `/opt/drupal/www-container-fs-ready`:

        www-data@43710a373680:/opt/drupal$ ls -al
        total 524
        drwxr-xr-x  1 www-data www-data   4096 Oct 19 13:25 .
        drwxr-xr-x  1 root     root       4096 Oct 18 19:38 ..
        -rw-r--r--  1 www-data www-data    357 Oct 18 19:25 .editorconfig
        drwxr-xr-x  8 www-data www-data   4096 Oct 18 19:24 .git
        -rw-r--r--  1 www-data www-data   4034 Oct 18 19:25 .gitattributes
        -rw-r--r--  1 www-data www-data   2109 Oct 18 19:25 composer.json
        -rw-r--r--  1 www-data www-data 477751 Oct 18 19:25 composer.lock
        -rw-r--r--  1 www-data www-data    996 Oct 18 19:10 phpcs.xml
        -rw-r--r--  1 www-data www-data    245 Oct 18 19:10 phpstan.neon
        -rw-r--r--  1 www-data www-data   4614 Oct 18 19:38 phpunit.xml
        drwxr-xr-x 61 www-data www-data   4096 Oct 18 19:25 vendor
        drwxr-xr-x  8 www-data www-data   4096 Oct 18 19:25 web
        -rw-r--r--  1 www-data www-data      6 Oct 19 13:25 www-container-fs-ready
        www-data@43710a373680:/opt/drupal$ cat www-container-fs-ready 
        ready

That all seems to be working locally... so I'm not sure why it isn't working in the context of the GitHub Actions workflow... :thinking:

mstenta:
There is ONE difference... although I'm not sure why it would matter...

If I run the same test with the *current* `farmos/farmos:2.x-dev` image, this is what I see:

```
root@c34c7bbf64cb:/opt/drupal# ls -al
total 524
drwxr-xr-x 1 www-data www-data   4096 Oct 19 13:28 .
drwxr-xr-x 1 root     root       4096 Oct 18 12:30 ..
-rw-r--r-- 1 www-data www-data    357 Oct 18 12:30 .editorconfig
drwxr-xr-x 1 www-data www-data   4096 Oct 18 12:30 .git
-rw-r--r-- 1 www-data www-data   4034 Oct 18 12:30 .gitattributes
-rw-r--r-- 1 www-data www-data   2105 Oct 18 12:30 composer.json
-rw-r--r-- 1 www-data www-data 477747 Oct 18 12:30 composer.lock
-rw-r--r-- 1 www-data www-data    868 Oct 18 12:30 phpcs.xml
-rw-r--r-- 1 www-data www-data    245 Oct 18 12:30 phpstan.neon
-rw-r--r-- 1 www-data www-data   4614 Oct 18 12:30 phpunit.xml
drwxr-xr-x 1 www-data www-data   4096 Oct 18 12:30 vendor
drwxr-xr-x 1 www-data www-data   4096 Oct 18 12:30 web
-rw-r--r-- 1 root     root          6 Oct 19 13:28 www-container-fs-ready
root@c34c7bbf64cb:/opt/drupal# cat www-container-fs-ready 
ready
```

Notice that `www-container-fs-ready` is owned by `root`, not `www-data`.

So perhaps the fact that it is changing to being owned by `www-data` makes it unreadable to the GitHub Actions runner host somehow? It is world readable, though (`0644`), so I'm not sure I understand why that would be an issue.

docker/Dockerfile
AlMaVizca: You are right that it doesn't require anything else than what belongs to PHP official images, although the layers that Drupal is adding to the image are not affecting the final farmOS image. 
I'm still thinking that it's better to keep Drupal image. As you mention, the changes will be required anyway, but reducing the consistent number of changes also keeps down the unaware errors.
Let's imagine that Drupal decides to compile one of the PHP modules with some specific arguments that are not available on PHP image, and that we keep the line 30:
```suggestion
COPY --from=geos /usr/local/lib/php/extensions/ /usr/local/lib/php/extensions/
```
That would mean that the module X, will be overwritten with the PHP version, without the specified argument, and might lead to a debugging time until the cause is found. 

This could be solved by coping only the `geo.so` file, yet the extension's directory has inside a folder that ends with the PHP extension value, which I don't know when that is it updated. An alternative is to add scripting to find the right path.
Screenshot from 2023-10-21 14-52-15
Screenshot from 2023-10-21 14-52-41
 
It might be an overthinking by ignorance, but those are the possible scenarios that I can imagine.
In my opinion, keeping the base image prevents all of them.

AlMaVizca:
> So perhaps the fact that it is changing to being owned by www-data makes it unreadable to the GitHub Actions runner host somehow? It is world readable, though (0644), so I'm not sure I understand why that would be an issue.

Indeed, there is a problem of permissions.
Docker compose files are mounting `./www:/opt/drupal`. If the path `./www` doesn't exist, it will be created being root the owner.
To preserve this behavior, the only action that I can think of is to add an empty folder `www` to the repository and modify the conditionals on the `docker-entrypoint.sh` file to check something more specific.
Adding an empty folder, means creating the folder and adding a git ignore, which then makes the first condition on the entry point to be ignored and not copying all the files.
Would it be ok to check for something more specific, like `/opt/drupal/web/profiles/farm/farm.info.yml`?

Another question that popped out is, How clean are the GitHub environments?
I'm asking this because before these changes, the file defined on `FARMOS_FS_READY_SENTINEL_FILENAME` was never removed and might be there owned by root, therefore failing on the line 23 of the entry point when trying to overwrite it with owner `www-data`.

docker/Dockerfile
mstenta: > I'm still thinking that it's better to keep Drupal image. As you mention, the changes will be required anyway, but reducing the consistent number of changes also keeps down the unaware errors.

OK we can keep the Drupal image. :-)

I'll change it back and force push.

mstenta:
Oh I see! I wasn't mounting a volume in my local test. This is what I see when I do:

```
$ docker run --rm -it -e FARMOS_FS_READY_SENTINEL_FILENAME=/opt/drupal/www-container-fs-ready -v "./www:/opt/drupal" farmos-docker-improvements-dev bash
farmOS codebase not detected. Copying from pre-built files in the Docker image.
cp: cannot create directory '/opt/drupal/./.git': Permission denied
cp: cannot create regular file '/opt/drupal/./composer.json': Permission denied
cp: cannot create regular file '/opt/drupal/./composer.lock': Permission denied
cp: cannot create directory '/opt/drupal/./vendor': Permission denied
cp: cannot create regular file '/opt/drupal/./.editorconfig': Permission denied
cp: cannot create regular file '/opt/drupal/./.gitattributes': Permission denied
cp: cannot create regular file '/opt/drupal/./phpcs.xml': Permission denied
cp: cannot create regular file '/opt/drupal/./phpstan.neon': Permission denied
cp: cannot create regular file '/opt/drupal/./phpunit.xml': Permission denied
cp: cannot create directory '/opt/drupal/./web': Permission denied
cp: preserving times for '/opt/drupal/.': Operation not permitted
```

If I add `-u root`, to run the container as root, it works:

```
$ docker run --rm -it -e FARMOS_FS_READY_SENTINEL_FILENAME=/opt/drupal/www-container-fs-ready -v "./www:/opt/drupal" -u root farmos-docker-improvements-dev bash
farmOS codebase not detected. Copying from pre-built files in the Docker image.
Attempting: bash
root@6a56c28e388e:/opt/drupal#
```

So the problem is that `docker-entrypoint.sh` is running as `www-data` user now instead of `root`.

Hmm that is tricky... :thinking:

Let me give this some more thought, and consider the ideas you suggested @AlMaVizca...

mstenta:
While we are thinking about this: maybe it would be worth splitting this PR into two new ones:

1. Create multi-stage Dockerfile
2. Run Docker container as www-data user

I think all of the commits leading up the last one are ready for final review. It's just this last one that may become a bit more hairy.

mstenta:
> Would it be ok to check for something more specific, like `/opt/drupal/web/profiles/farm/farm.info.yml`?
> 
> Another question that popped out is, How clean are the GitHub environments?
> I'm asking this because before these changes, the file defined on `FARMOS_FS_READY_SENTINEL_FILENAME` was never removed and might be there owned by root, therefore failing on the line 23 of the entry point when trying to overwrite it with owner `www-data`.

Just wanted to respond to this as well...

None of this really matters, because the problem runs deeper. It happens if you try to run the normal (non-dev) container as well, even without `FARMOS_FS_READY_SENTINEL_FILENAME`:

```
$ docker run --rm -it -v "./www:/opt/drupal" farmos-docker-improvements
farmOS codebase not detected. Copying from pre-built files in the Docker image.
cp: cannot create directory '/opt/drupal/./.git': Permission denied
cp: cannot create regular file '/opt/drupal/./composer.json': Permission denied
cp: cannot create regular file '/opt/drupal/./composer.lock': Permission denied
cp: cannot create directory '/opt/drupal/./vendor': Permission denied
cp: cannot create regular file '/opt/drupal/./.editorconfig': Permission denied
cp: cannot create regular file '/opt/drupal/./.gitattributes': Permission denied
cp: cannot create directory '/opt/drupal/./web': Permission denied
cp: preserving times for '/opt/drupal/.': Operation not permitted
```

So this is a core incompatibility with our current `docker-entrypoint.sh` script logic and running the container as `www-data`.

mstenta:
> Docker compose files are mounting `./www:/opt/drupal`. If the path `./www` doesn't exist, it will be created being root the owner.

This is the core issue. If the `./www` directory does not exist before `docker compose up` (or `docker run`) runs, then it will be owned by root, and `docker-entrypoint.sh` (running as `www-data`) will not be able to copy files into it.

We could potentially work around this by modifying our `deliver.yml` to create the `www` directory and `chown www-data:www-data` it before we run `docker compose up`.

But we would also have to update our docs to tell end-users who are setting up their development/production environments to do that manually as well. I think that would work, but it's a bummer because it adds manual steps to the setup (always prefer removing steps to adding them). And we need to consider whether or not this will affect existing installs.

All-in-all, this is feeling like a more complicated change than I originally anticipated. Which means it will need a lot more community review and discussion before we can merge it regardless.

All the more reason to split it out to its own PR, IMO.

mstenta:
> To preserve this behavior, the only action that I can think of is to add an empty folder `www` to the repository

This would only address the CI/CD pipeline failures. But it would not help end-users. They would still need to manually create their bind-mount directory(ies) manually and set the ownership. See the instructions we provide here, for example:

https://farmos.org/hosting/install/#farmos-in-docker

And here: https://farmos.org/hosting/composer/ (which reminds me... that page would need to be updated with this PR too).

AlMaVizca:
> This would only address the CI/CD pipeline failures. But it would not help end-users. They would still need to manually create their bind-mount directory(ies) manually and set the ownership.

Only current developers will run into the issue because the folder would be present for new developers when they clone the repository.
I'll follow your suggestion of opening 2 PR:
1. The multi-stage version, but I will add all the changes, so you review what might be a final version.
2. The change on the user container.

AlMaVizca:
As a summary, I'm removing the last two commits, and adding a small change to build `bcmath` inside the `php-dependencies` stage.

 > And here: https://farmos.org/hosting/composer/ (which reminds me... that page would need to be updated with this PR too).

Not sure which part are you refering too, in case it's the [docker](https://farmos.org/hosting/composer/#docker) section, I think it is still valid, since it's a reference for extending the image.

mstenta:
> As a summary, I'm removing the last two commits, and adding a small change to build `bcmath` inside the `php-dependencies` stage.

Thanks @AlMaVizca! I'll take a look soon.

> Not sure which part are you refering too, in case it's the [docker](https://farmos.org/hosting/composer/#docker) section, I think it is still valid, since it's a reference for extending the image.

Yes I meant we'll need to remember to test/update that in the `www-data` PR, because right now it assumes that it is running as `root`. I'll follow up in the new PR so we remember that.

mstenta:
Tests are passing! I think we can flag this for final review now.

AlMaVizca:
Just a question in case, it will reduce the amount of effort on your side.
I've created the commits for the other PR, splitting into more stages, and removing the dev/Dockerfile.
Do you prefer them apart? 
I've created the PR on my repo, so you could give it a brief look.
- [docker image multi stages](https://github.com/AlMaVizca/farmOS/pull/1/files)

docker/dev/Dockerfile
```diff
@@ -37,42 +36,8 @@ USER www-data
...
-    echo "    - '#^Unsafe usage of new static#'"; \
-    echo '  scanDirectories:'; \
-    echo '    # Tell PHPStan where to find GeoPHP classes.'; \
-    echo '    - vendor/itamair/geophp'; \
-  } > /var/farmOS/phpstan.neon
+# Add Configurartions for PHP CodeSniffer, PHPStan.
+COPY --chown=www-data ./files/ /var/farmOS/
 
 # Configure PHPUnit.
 RUN cp -p /var/farmOS/web/core/phpunit.xml.dist /var/farmOS/phpunit.xml \
```
paul121, line 43: Could we move the PHPunit config to a separate file in `files/phpunit.xml` ? I know we're copying this over from the Drupal core `phpunit.xml.dist` but I wonder how often this changes. We haven't needed to modify our code in a few years.. but I wonder how much that is just luck not having conflicting changes from Drupal core :thinking:

paul121 (commented):
This LGTM! I'm not entirely sure but think there are now a few 3.x changes we need to incorporate into this PR too?

Really really like that configuration is getting split out into separate files. Appreciate all the work on this @AlMaVizca @mstenta !

docker/dev/Dockerfile
mstenta: @AlMaVizca did separate this one out originally and I asked that we keep it the way it was, because separating it out means we are maintaining the whole file. I agree it may not change often, but if it does there's no way for us to know. Currently we get any changes automatically (or the `sed` command will fail, notifying us that we need to update it). I think that's worth keeping as-is.

mstenta:
> think there are now a few 3.x changes we need to incorporate into this PR too?

This is true. We'll need to rebase this onto 3.x. I can take a pass at that this morning since I'm familiar with the relevant 2.x->3.x changes...

mstenta:
> I've created the commits for the other PR, splitting into more stages, and removing the dev/Dockerfile.
> Do you prefer them apart?

Thanks @AlMaVizca! Looking forward to reviewing those too! (Although it might be a little bit before I have a chance.)

Yes let's keep them separate for now, since this PR is already close to being merged.

mstenta:
> We'll need to rebase this onto 3.x. I can take a pass at that this morning

Done!

mstenta:
Not sure why tests aren't running on this anymore (maybe due to base branch change and/or the fact that the PR branch doesn't start with `3.x-*`), but I pushed a copy of it to my fork and tests passed FYI: https://github.com/mstenta/farmOS/actions/runs/6707691111

docker/dev/Dockerfile
AlMaVizca: @paul121 @mstenta you will see on a following PR that, I'v created the script with all the regexp, so the Docker file is cleaner

mstenta (approved):

paul121 (approved):

mstenta:
Merged! Thanks again for all your work (and patience!) with this one @AlMaVizca! Great improvements!

PR #736, opened 2023-10-19
Refactor Move and Group actions into quick forms

mstenta:
This refactors the "Move" and "Group" bulk actions so that they redirect to the new "Movement" and "Group assignment" quick forms. This consolidates the UI for easily creating movement/group assignment logs, and allows us to remove the deprecated `farm_log_asset_names_summary()` function from the `farm_log` module.


mstenta:
I rebased onto latest `2.x`, and split this commit out to a separate branch, so that we can consider merging this into `2.x` (instead of waiting for `3.x`): https://github.com/mstenta/farmOS/tree/3.x-remove-farm_log_asset_names_summary

mstenta:
Just noticed in `/admin/reports/status` that there was a config schema issue with the two actions, because they needed `*.schema.yml` to define their `configuration` bits. Force pushed my branch to fix those (and remove the old action schemas).

I also included a commit that adds this requirement to our module developer documentation, because it was missing previously.

paul121 (changes requested):
I have not tested all of these changes myself but overall they LGTM.

I'm just noticing one thing - is this missing a `post_update` hook to create the `quick_group` action?  We have one for creating the `quick_movement` action.

mstenta:

> I'm just noticing one thing - is this missing a `post_update` hook to create the `quick_group` action? We have one for creating the `quick_movement` action.

Ah good catch! Originally this work was going to be merged at the same time as #723, so an update hook wasn't necessary. But since it is now a follow-up it needs one, you're right!

I force pushed the commits with a new `post_update` hook. Here is the diff: https://github.com/farmOS/farmOS/compare/08dd8b0f084a3896d96e19f0d5c2000922e7190a..10203fffad41b98b8af5e855e3c4b3eba68de17e

paul121 (approved):

mstenta:
Merged.

PR #707, opened 2023-08-16
Configurable quick forms

mstenta:
See https://www.drupal.org/project/farm/issues/3354935

PR #738, opened 2023-10-23
Add post update hook to install the new quick_form entity type on existing farmOS installations

wotnak:
Fixes problem reported at https://farmos.discourse.group/t/status-quick-form-entity-type/1772.

By default, new entity types are automatically installed only on module install. Since the quick_form entity type [was added](https://github.com/farmOS/farmOS/pull/707) in an existing farm_quick module that is already installed on many farmOS instances, a [hook_update_N](https://api.drupal.org/api/drupal/core!lib!Drupal!Core!Extension!module.api.php/function/hook_update_N/9) implementation should be added to install the entity type during database updates.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - [Fix planting quick form creating empty quantities #737](https://github.com/farmOS/farmOS/pull/737)
+- [Add post update hook to install the new quick_form entity type on existing farmOS installations #738](https://github.com/farmOS/farmOS/pull/738)
 
 ## [2.2.1] 2023-10-09
 
--- a/modules/core/quick/farm_quick.post_update.php
+++ b/modules/core/quick/farm_quick.post_update.php
@@ -1,0 +1,15 @@
+<?php
+
+/**
+ * @file
+ * Post update hooks for the farmOS Quick Form module.
+ */
+
+/**
+ * Install the new quick_form entity type.
+ */
+function farm_quick_post_update_install_quick_form_entity_type(&$sandbox) {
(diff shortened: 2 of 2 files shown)


mstenta:
Thank you @wotnak!!! I will review and test this ASAP.

modules/core/quick/farm_quick.install
```diff
@@ -0,0 +1,17 @@
+<?php
+
+/**
+ * @file
+ * Update functions for the farmOS Quick Form module.
+ */
+
+declare(strict_types=1);
```
mstenta: I like the idea of using `declare(strict_types=1)`, but since we do not use this anywhere else in farmOS I think we should omit it for now, just for consistency. But I would love to consider adopting it throughout farmOS and documenting what the benefits are in an issue...

modules/core/quick/farm_quick.install
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * Install the new quick_form entity type.
+ */
+function farm_quick_update_9001(): void {
```
mstenta: Can we change this to an implementation of `hook_post_update_NAME()` instead of `hook_update_N()`? We've adopted the former as our default update hook and have not used `hook_update_N()` in 2.x at all (unless I'm forgetting one).

This also means it needs to move from `farm_quick.install` to `farm_quick.post_update.php`.

mstenta (changes requested):

modules/core/quick/farm_quick.install
wotnak: Removed the strict_types declaration.

modules/core/quick/farm_quick.install
wotnak: Moved the implementation to a post_update hook.

mstenta (approved):

mstenta:
Merged. Thanks again @wotnak!

PR #741, opened 2023-10-27
farmOS 3.x

mstenta:
This PR represents the official proposal to start the `3.x` branch of farmOS. It currently does two kinds of things:

1. Changes all occurrences of `2.x` to `3.x` everywhere.
2. Implements all the "breaking changes" that we currently have lined up for `3.x`.

Notably, this does NOT include the upgrade to Drupal 10. I have that prepared as a separate branch that builds off of this one, which I will open a follow-up PR for.

The only other big thing we are working on is the Simple OAuth v6 upgrade. We are still in the process of investigating whether or not that will have any breaking changes. More to come on that in the near future...

Merging of this PR will mean that `3.x` becomes the official "main branch" of farmOS, and all new PRs must be opened against it first. Hopefully we will be able to proceed quickly and there won't be a need to maintain `2.x` separately, but we have that option if we need to. In that case, PRs that are needed in `2.x` will need to be merged into `3.x` first, and then backported to `2.x`.


paul121 (approved):

mstenta:
Merged! `3.x` is now the official default branch on https://github.com/farmOS/farmOS and https://git.drupalcode.org/project/farm! :tada:

PR #743, opened 2023-10-30
Update Simple OAuth module to v6

paul121:
Updates [simple_oauth](https://www.drupal.org/project/simple_oauth) to v6. To start we will only provide static scopes for our default roles (manager, worker and viewer).

We also will include `simple_oauth_password_grant` as a dependency but only depend on this in the `farm_fieldkit` and (new) `farm_api_default_consumer` modules.

Other TODO:

- [x] Update docs
- [x] Update consumer grant types via `consumer.grant_types` field:
  - [x]   Disable client credentials on all (this grant now requires a user, and in general is not a grant type that should be used with our default clients)
  - [x]  Field kit consumer ONLY needs password grant type
  - [x]  Move default consumer to `farm_api_default_consumer` submodule
- [x]  Test if access/refresh tokens are lost/revoked during 2.x -> 3.x upgrade
- [x]  Test FieldKit integration (shouldn't be hard using existing FK with a local 3.x instance)
- [x]  Test if our previous patch for consumers with "empty" secret is still needed: https://www.drupal.org/project/simple_oauth/issues/3322325
  - [x] If so, likely adapt to a new patch OR provide update hook ourselves (at this point very unlikely simple_oauth will help us)


paul121:
> TODO: Need to update `farm` consumer to allow password grant when simple_oauth_password_grant is enabled 🤔

> Default `farm` consumer only needs... hmm... authorization code? maybe it really only needs password grant?

Idea! Maybe we should introduce a new sub-module that provides the "default" `farm` consumer. This would solve the conundrum about enabling the password grant for the consumer only when the `simple_oauth_password_grant` is enabled. `farm_api` by itself only sets up the API.

In general, I don't think people should use the default `farm` consumer - they should really create their own consumers and use auth code or client credentials when a username/password is basically needed. But providing this `farm` consumer in a sub-module would provide a short-term solution for Farmier hosting to maintain non-BC API access. cc @mstenta

paul121:
> Idea! Maybe we should introduce a new sub-module that provides the "default" farm consumer. 

So, we considered this earlier: https://www.drupal.org/project/farm/issues/3256745 I think now is the time! Will update this PR with an optional `farm_api_default_consumer` sub-module

paul121:
>  Test FieldKit integration (shouldn't be hard using existing FK with a local 3.x instance)

Indeed this wasn't hard, and it revealed one minor issue. I am just now realizing that FK was dependent on the `grant_user_access` feature because it does not request any specific scope in its implementation of the password grant. I had thought it requested the `farm_manager` scope specifically.

I think the best solution for this is to use the "Default scopes" configuration on consumer entities set to `farm_manager` and `farm_viewer`. Sites with custom roles/scopes can modify and add these here instead. These scopes are "the default scopes when authorizing" which will allow FK to continue not specifying a specific scope, but ensure that something is granted to the token. With the new OAuth v6 changes this is safe because a consumer will never grant more access than the user has

modules/core/api/farm_api.module
```diff
@@ -34,47 +34,8 @@ function farm_api_consumers_list_alter(&$data, $context) {
 function farm_api_entity_base_field_info(EntityTypeInterface $entity_type) {
   $fields = [];
 
-  // Add bundle fields to the consumer entity.
+  // Add allowed_origin field to the consumer entity.
```
mstenta: Minor comment issue: This says `allowed_origin` but should say `allowed_origins`.

modules/core/role/modules/roles/farm_role_roles.module
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * @file
+ * Hooks implemented by the Farm Role Roles module.
+ * */
+
+/**
+ * Implements hook_oauth2_scope_info_alter().
+ */
+function farm_role_roles_oauth2_scope_info_alter(array &$scopes) {
+  // Enable the password grant for static role scopes.
```
mstenta, line 13: Nit: Add one blank line above comment.

modules/core/fieldkit/farm_fieldkit.post_update.php
```diff
@@ -0,0 +1,39 @@
...
+ * @file
+ * Post update functions for farm_fieldkit module.
+ */
+
+use Drupal\simple_oauth\Oauth2ScopeInterface;
+
+/**
+ * Enable simple oauth password grant.
+ */
+function farm_fieldkit_post_update_enable_password_grant(&$sandbox = NULL) {
```
mstenta, line 13: Do we need to make sure this runs after `farm_api_post_update_enable_static_oauth2_scopes()`? If it runs first, then the `static` scope provider won't be enabled, and the default `['farm_manager', 'farm_worker']` scopes won't load, and then won't be added to the `fieldkit` consumer. I think anyway...

mstenta (changes requested):
Looks like this is coming along nicely @paul121!

I know it's still a work-in-progress, but I gave it a quick review and left a few nitpicks/comments.

Question:

If someone added their own custom consumers, will they need to do any manual updates to them? If so, I wonder if we could automate that, or if we should just document it in the release notes and include some guidance.

One specific one that comes to mind is the consumer provided by the `farm_surveystack` module... I assume we'll need to make some changes there, along with an update hook?

https://github.com/mstenta/farm_surveystack/blob/4c43212901a5ecaefc3b41d8eb11b208cf89f958/farm_surveystack.install#L12

paul121:
> If someone added their own custom consumers, will they need to do any manual updates to them? If so, I wonder if we could automate that, or if we should just document it in the release notes and include some guidance.

Yes they likely will. The biggest thing will be if they want the password grant, they will need to add that on each consumer. Simple oauth has an update hook and tries to automate what they can - basically allow all grant types previously allowed at the global level & auto-create dynamic scopes if clients were previously configured with default roles - none of ours are. But one thing they do not do is re-enable the password grant.

> One specific one that comes to mind is the consumer provided by the farm_surveystack module... I assume we'll need to make some changes there, along with an update hook?

Yes, that will need an update hook similar to these

modules/core/fieldkit/farm_fieldkit.post_update.php
paul121: Hmm yes I think you are right. And taking a step back, we should make sure the `simple_oauth` hooks run before `farm_api` as well. Does this mean we can't use `hook_post_update_NAME` ?

paul121: Aha, but if we use `post_update`, we can ensure they run after `simple_oauth` normal update hooks :+1:

paul121: We also identified that `farm_fieldkit_post_update_*` should run after `farm_api_post_update_*` simply due to alphabetical. I will test this soon and confirm.

paul121: Here is the update list from a 2.x -> 3.x install, working as expected!

```
$ ddev drush updb
 --------------- -------------------------------- --------------- --------------------------------------------------------------------------
  Module          Update ID                        Type            Description
 --------------- -------------------------------- --------------- --------------------------------------------------------------------------
  system          10100                            hook_update_n   10100 - Remove the year 2038 date limitation.
  user            10000                            hook_update_n   10000 - Remove non-existent permissions created by migrations.
  dblog           10100                            hook_update_n   10100 - Remove the year 2038 date limitation.
  dblog           10101                            hook_update_n   10101 - Converts the 'wid' of the 'watchdog' table to a big integer.
  simple_oauth    8601                             hook_update_n   8601 - Set default value for the scope provider to dynamic (entity).
  simple_oauth    8602                             hook_update_n   8602 - Install new config entity type "oauth2_scope".
  simple_oauth    8603                             hook_update_n   8603 - Install/update/delete BaseFields for the consumer entity.
  simple_oauth    8604                             hook_update_n   8604 - Migrate roles used as scope and migrate settings to the consumer.
  simple_oauth    8605                             hook_update_n   8605 - Install/delete scopes BaseField for the oauth2_token entity.
  simple_oauth    8606                             hook_update_n   8606 - Add indexes to oauth2_token value field.
  farm_api        enable_default_consumer_module   post-update     Enable default consumer module.
  farm_api        enable_static_oauth2_scopes      post-update     Enable static oauth2 scopes.
  farm_api        remove_consumer_fields           post-update     Remove farm_api consumer bundle fields.
  farm_fieldkit   enable_password_grant            post-update     Enable simple oauth password grant.
  file            add_permissions_to_roles         post-update     Grant all non-anonymous roles the 'delete own files' permission.
  system          enable_password_compatibility    post-update     Enable the password compatibility module.
  system          linkset_settings                 post-update     Add new menu linkset endpoint setting.
  system          timestamp_formatter              post-update     Update timestamp formatter settings for entity view displays.
  text            allowed_formats                  post-update     Add allowed_formats setting to existing text fields.
  views           boolean_custom_titles            post-update     Update Views config schema to make boolean custom titles translatable.
  views           fix_revision_id_part             post-update     Fix '-revision_id' replacement token syntax.
  views           oembed_eager_load                post-update     Add eager load option to all oembed type field configurations.
  views           responsive_image_lazy_load       post-update     Add lazy load options to all responsive image type field configurations.
  views           timestamp_formatter              post-update     Update timestamp formatter settings for views.
 --------------- -------------------------------- --------------- --------------------------------------------------------------------------


 Do you wish to run the specified pending updates? (yes/no) [yes]:
 > yes

>  [notice] Update started: simple_oauth_update_8601
>  [notice] Update completed: simple_oauth_update_8601
>  [notice] Update started: simple_oauth_update_8602
>  [notice] Update completed: simple_oauth_update_8602
>  [notice] Update started: simple_oauth_update_8603
>  [notice] Update completed: simple_oauth_update_8603
>  [notice] Update started: simple_oauth_update_8604
>  [notice] Update completed: simple_oauth_update_8604
>  [notice] Update started: dblog_update_10100
>  [notice] Update completed: dblog_update_10100
>  [notice] Update started: simple_oauth_update_8605
>  [notice] Update completed: simple_oauth_update_8605
>  [notice] Update started: system_update_10100
>  [notice] Update completed: system_update_10100
>  [notice] Update started: user_update_10000
>  [notice] Update completed: user_update_10000
>  [notice] Update started: dblog_update_10101
>  [notice] Update completed: dblog_update_10101
>  [notice] Update started: simple_oauth_update_8606
>  [notice] Update completed: simple_oauth_update_8606
>  [notice] Reverted config: core.entity_view_display.asset.animal.map_popup
>  [notice] Reverted config: core.entity_view_display.asset.equipment.map_popup
>  [notice] Reverted config: core.entity_view_display.asset.land.map_popup
>  [notice] Reverted config: core.entity_view_display.asset.plant.map_popup
>  [notice] Reverted config: core.entity_view_display.asset.structure.map_popup
>  [notice] Reverted config: core.entity_view_display.taxonomy_term.plant_type.default
>  [notice] Reverted config: log.type.input
>  [notice] Reverted config: system.action.asset_assign_action
>  [notice] Reverted config: system.action.asset_parent_action
>  [notice] Reverted config: views.view.files
>  [notice] Reverted config: core.entity_view_display.asset.animal.map_popup
>  [notice] Reverted config: core.entity_view_display.asset.equipment.map_popup
>  [notice] Reverted config: core.entity_view_display.asset.land.map_popup
>  [notice] Reverted config: core.entity_view_display.asset.plant.map_popup
>  [notice] Reverted config: core.entity_view_display.asset.structure.map_popup
>  [notice] Reverted config: log.type.input
>  [notice] Reverted config: system.action.asset_assign_action
>  [notice] Reverted config: system.action.asset_parent_action
>  [notice] Update started: farm_api_post_update_enable_default_consumer_module
>  [notice] Update completed: farm_api_post_update_enable_default_consumer_module
>  [notice] Update started: farm_api_post_update_enable_static_oauth2_scopes
>  [notice] Update completed: farm_api_post_update_enable_static_oauth2_scopes
>  [notice] Update started: farm_api_post_update_remove_consumer_fields
>  [notice] Update completed: farm_api_post_update_remove_consumer_fields
>  [notice] Update started: farm_fieldkit_post_update_enable_password_grant
>  [notice] Update completed: farm_fieldkit_post_update_enable_password_grant
>  [notice] Update started: file_post_update_add_permissions_to_roles
>  [notice] Update completed: file_post_update_add_permissions_to_roles
>  [notice] Update started: system_post_update_enable_password_compatibility
>  [notice] Update completed: system_post_update_enable_password_compatibility
>  [notice] Update started: system_post_update_linkset_settings
>  [notice] Update completed: system_post_update_linkset_settings
>  [notice] Update started: system_post_update_timestamp_formatter
>  [notice] Update completed: system_post_update_timestamp_formatter
>  [notice] Update started: text_post_update_allowed_formats
>  [notice] Update completed: text_post_update_allowed_formats
>  [notice] Update started: views_post_update_boolean_custom_titles
>  [notice] Update completed: views_post_update_boolean_custom_titles
>  [notice] Update started: views_post_update_fix_revision_id_part
>  [notice] Update completed: views_post_update_fix_revision_id_part
>  [notice] Update started: views_post_update_oembed_eager_load
>  [notice] Update completed: views_post_update_oembed_eager_load
>  [notice] Update started: views_post_update_responsive_image_lazy_load
>  [notice] Update completed: views_post_update_responsive_image_lazy_load
>  [notice] Update started: views_post_update_timestamp_formatter
>  [notice] Update completed: views_post_update_timestamp_formatter
>  [notice] Reverted config: core.entity_view_display.asset.animal.map_popup
>  [notice] Reverted config: core.entity_view_display.asset.equipment.map_popup
>  [notice] Reverted config: core.entity_view_display.asset.land.map_popup
>  [notice] Reverted config: core.entity_view_display.asset.plant.map_popup
>  [notice] Reverted config: core.entity_view_display.asset.structure.map_popup
>  [notice] Reverted config: log.type.input
>  [notice] Reverted config: system.action.asset_assign_action
>  [notice] Reverted config: system.action.asset_parent_action
>  [notice] Reverted config: user.role.farm_manager
>  [notice] Reverted config: user.role.farm_viewer
>  [notice] Reverted config: user.role.farm_worker
>  [notice] Reverted config: views.view.farm_log
>  [notice] Reverted config: views.view.farm_people
>  [notice] Reverted config: views.view.farm_quantity
>  [notice] Reverted config: views.view.log_admin
 [success] Finished performing updates.
 ```

**Comment by paul121** [comment]
Okay, tests are passing. Pushed to correct the last code sniffer issue.

> Test if access/refresh tokens are lost/revoked during 2.x -> 3.x upgrade

Access tokens are revoked, but refresh tokens are not. So once an OAuth client refreshes their token they should be OK. But most OAuth clients will not attempt a refresh until the access token is expired. Currently that period is 1 hour. If this is an issue for existing integrations we could recommend they decrease the expiration time to a shorter time (5 minutes) before performing the update, then access tokens will have expired by the time the update is complete.

I did some testing with the below python script:

```python
import os

from farmOS import farmOS
#
# # Allow testing via http.
os.environ["OAUTHLIB_INSECURE_TRANSPORT"] = "1"

FARMOS_HOSTNAME = "http://3.x.ddev.site"

# Credentials for OAuth Auth
FARMOS_OAUTH_USERNAME = "manager"
FARMOS_OAUTH_PASSWORD = "manager"
FARMOS_OAUTH_CLIENT_ID = "farm"
FARMOS_OAUTH_CLIENT_SECRET = None

# Copy token value from debugger, save to variable
# token = {'token_type': 'Bearer', 'expires_in': 3601, 'access_token': long value', 'refresh_token': 'long value', 'expires_at': 1698785323.3884192}

farm = farmOS(
    hostname=FARMOS_HOSTNAME,
    client_id=FARMOS_OAUTH_CLIENT_ID,
    client_secret=FARMOS_OAUTH_CLIENT_SECRET,
    token=token,
)
# On 2.x: request a new token using credentials. Save this token, inspect in debugger and copy value to above variable.
# Will use this token after migrating to 3.x. Comment out this line when testing with 3.x so we don't get a new token.
token = farm.authorize(username=FARMOS_OAUTH_USERNAME, password=FARMOS_OAUTH_PASSWORD)

# Test request, get farm info
# Check meta.links.me to make sure request was authenticated correctly (but will fail altogether if access token is bad)
info = farm.info()

# On 3.x: try refreshing the token. Access tokens will  be revoked during the migration, but refresh token should be OK.
#farm.session.refresh_token(FARMOS_HOSTNAME + "/oauth/token")
#token= farm.session.token

# Test request, get farm info
info = farm.info()

# Debug breakpoint
test = True
```

paul121:
> Test if our previous patch for consumers with "empty" secret is still needed:

Yes, it is still needed. I provided a new patch for v6 and added it to our composer.json

paul121:
While testing the migration I realized we did not include the `refresh_token` grant on both the `farm` and `fieldkit` clients. I updated them to both have `refresh_token` and also added `authorization_code` to the `farm` client since that could have been used previously. I also updated the static scopes to work with all grant types for better BC compatibility. I can't think of any reason to exclude grant types from these scopes.

Finally, I updated the API Changes and Authentication docs with higher level info, and updated the OAuth development docs with more detailed information about consumers and scopes.

mstenta (approved):
LGTM! Thanks for all the hard work on this @paul121! This should set us up nicely for the future. :-)

PR #746, opened 2023-11-03
Fix warning message when rendering link to birth log

paul121:
If an animal asset has a birth date and a birth log, the birthdate field on the asset page is converted to be a link to that birth log. A nice feature but I think something changed with D10, and now the timestamp field does not use a `#markup` render item but `#text` instead. It produces this warning:

> Warning: Undefined array key "#markup" in farm_birth_asset_view_alter() (line 67 of profiles/farm/modules/log/birth/farm_birth.module).

You can reproduce this pretty easily by using the birth quick form and viewing the child animal asset that is created.


mstenta (approved):

mstenta:
Good eye - thanks @paul121!

PR #751, opened 2023-11-19
Single Dockerfile using multi-stages for production and dev image

AlMaVizca:
This proposal is reorganizing the instructions to get advantage of the Docker cache, and handle a single file for images.

Previous size:
```
~/R/A/farmOS ❯❯❯ docker images | /usr/bin/grep farmos
farmos/farmos                     3.x-dev              979ee3ee7cbe   12 days ago         1.36GB
farmos/farmos                     3.x                  601095e643d1   12 days ago         966MB
```

New image size:
```
~/R/A/farmOS ❯❯❯ docker images | /usr/bin/grep farmos
farmos/farmos                     3.x-dev              c6a35b8434fe   6 minutes ago       835MB
farmos/farmos                     3.x                  6c93b1019a43   2 hours ago         796MB
```


mstenta:
Thanks for starting this @AlMaVizca!

I haven't had a chance to dig in yet, but at first glance it looks like a large number of changes in a small number of commits, and not a lot of explanation or reasoning is included in the commit messages. This makes it harder to review, because it means maintainers need to keep a lot of considerations in their head, and make guesses at your intentions or reasoning. It would be great if you could break this up so that each change can be reviewed in smaller bites, with each commit focused on a single change.

The first commit is pretty good as-is ("Define baseimage with arguments required in multiple stages"), but the second and third commits ("Production image" and "Development image", respectively) are a lot more opaque and intimidating to review. They also seem to be making a few new additions (eg: introducing `COMPOSER_ALLOW_SUPERUSER=1`), which should have their own commits with summaries of why they are being added. This may mean adding more "preparation" commits before the actual changes you want to make.

The end-goal of this "many small commits" approach is that we have a more granular history of changes, and the reasoning behind them, in the Git history. I use the "blame" features of Git a lot when I'm trying to remember *why* a change was made.

To illustrate, compare these two blames:

https://github.com/farmOS/farmOS/blame/845ee7f6cb79f65525c88e85e69ddb133dba35a0/docker/Dockerfile

https://github.com/AlMaVizca/farmOS/blame/1d905ca31150eb5294d465753a1a9643f12608b6/docker/Dockerfile

In the latter, large chunks of the history just say "Production image" or "Development image". Ideally there would be more granular changes with more descriptive messages in that blame.

Hope that all makes sense! If you have the time/interest in reformatting please do! Otherwise I will get around to picking this apart myself in more detail sometime in the future. :-)

AlMaVizca:
I am not sure if it could be split a bit further. I hope this works.

I'm seeing that it failed the check of MariaDB due to connectivity issue:
```
<ul><li>Failed to connect to your database server. The server reports the following message: 
<em class="placeholder">
SQLSTATE[HY000] [2002] Connection refused [Tip: This message normally means that 
there is no MySQL server running on the system or that you are using an incorrect host name 
or port number when trying to connect to the server. You should also check that the TCP/IP port 
you are using has not been blocked by a firewall or port blocking service.] 
</em>.
<ul><li>Is the database server running?</li><li>Does the database exist or does the database user have 
sufficient privileges to create the database?</li>
<li>Have you entered the correct database name?</li><li>Have you entered the correct username and password?</li><li>Have you entered the correct database hostname and port number?</li></ul></li></ul>
```
The execution of the same actions on my account worked fine:
[Merge development image into a single Dockerfile](https://github.com/AlMaVizca/farmOS/actions/runs/7210713427) 

Is there a way to trigger this workflow again?

docker/Dockerfile
```diff
@@ -1,7 +1,20 @@
...
+# Use the official Drupal 10 as base image.
+FROM drupal:10.1 as baseimage
+
+# Define common paths.
+ENV FARMOS_PATH=/var/farmOS
+ENV DRUPAL_PATH=/opt/drupal
+
+# Set the farmOS and composer project repository URLs and versions.
+ARG FARMOS_REPO=https://github.com/farmOS/farmOS.git
+ARG FARMOS_VERSION=3.x
```
wotnak: Moving all build args to the first stage on which all others depend will mean that if any of their values will change, the entire build cache [will be invalidated](https://docs.docker.com/engine/reference/builder/#impact-on-build-caching). This is especially important for FARMOS_VERSION and FARMOS_REPO since these [can be different](https://github.com/farmOS/farmOS/blob/3.x/.github/workflows/deliver.yml#L26-L36) for each pr which will prevent using build cache to speedup github actions workflow.

Also, if you want to make the build args available in all build stages they can just be placed before first FROM, no need to create a separate stage for them. 

So to allow for efficient use of build cache I would move them to as late stage as possible, so that when they change they invalidate as little build cache as possible.

docker/Dockerfile
```diff
@@ -17,51 +30,118 @@ RUN apt-get update && apt-get install -y libgeos-dev \
...
-ARG FARMOS_REPO=https://github.com/farmOS/farmOS.git
-ARG FARMOS_VERSION=3.x
-ARG PROJECT_REPO=https://github.com/farmOS/composer-project.git
-ARG PROJECT_VERSION=3.x
+# Set the COMPOSER environment variables.
+ENV COMPOSER_HOME=tmp \
+    # Set the COMPOSER_MEMORY_LIMIT environment variable to unlimited.
+    COMPOSER_MEMORY_LIMIT=-1 \
+    # Allow root to install plugins.
+    COMPOSER_ALLOW_SUPERUSER=1
```
wotnak: It would be more readable and consistent with earlier env declarations if they would be split to separate ENV statements for each variable.

docker/Dockerfile
```diff
@@ -17,51 +30,118 @@ RUN apt-get update && apt-get install -y libgeos-dev \
...
+# Setup dependencies and sources for composer installations.
+FROM baseimage as composer-file
 
-# Set the farmOS and composer project repository URLs and versions.
-ARG FARMOS_REPO=https://github.com/farmOS/farmOS.git
-ARG FARMOS_VERSION=3.x
-ARG PROJECT_REPO=https://github.com/farmOS/composer-project.git
-ARG PROJECT_VERSION=3.x
+# Set the COMPOSER environment variables.
+ENV COMPOSER_HOME=tmp \
```
wotnak: Since composer install is executed in separate stages from which to the final image are copied only files from FARMOS_PATH, changing COMPOSER_HOME is no longer needed and can be safly removed.

docker/Dockerfile
```diff
@@ -17,51 +30,118 @@ RUN apt-get update && apt-get install -y libgeos-dev \
...
+
+##
+# Create layer with farmOS dev sources.
+FROM composer-file as farmos-dev-sources
+
+# Install sources.
+RUN composer install
+
+# Set the version in farm.info.yml
+RUN sed -i "s|version: 3.x|version: ${FARMOS_VERSION}|g" ${FARMOS_PATH}/web/profiles/farm/farm.info.yml
```
wotnak: `farmos-dev-sources` is almost the same as `farmos-sources`, maybe it could be based on `farmos-sources`. This would allow avoiding duplication of setting version in farm.info.yml and would speed up composer installation since prod dependencies would be installed only once in `farmos-sources` instead of in both stages separately.

It could look something like:
```Dockerfile
##
# Create layer with farmOS sources.
FROM composer-file as farmos-sources

# Install php dependencies.
RUN composer install --no-dev --no-interaction

# Set the version in farm.info.yml.
RUN sed -i "s|version: 3.x|version: ${FARMOS_VERSION}|g" ${FARMOS_PATH}/web/profiles/farm/farm.info.yml

##
# Create layer with farmOS dev sources.
FROM farmos-sources as farmos-dev-sources

# Install php dev dependencies.
RUN composer install --dev --no-interaction
```

docker/Dockerfile
```diff
@@ -17,51 +30,118 @@ RUN apt-get update && apt-get install -y libgeos-dev \
...
+
+# Set the version in farm.info.yml.
+RUN sed -i "s|version: 3.x|version: ${FARMOS_VERSION}|g" ${FARMOS_PATH}/web/profiles/farm/farm.info.yml
+
+##
+# Create layer with farmOS dev sources.
+FROM composer-file as farmos-dev-sources
+
+# Install sources.
+RUN composer install
```
wotnak: It would be good to include `--no-interaction` flag for all composer commands executed during image build or set [`COMPOSER_NO_INTERACTION=1`](https://getcomposer.org/doc/03-cli.md#composer-no-interaction) env variable.

docker/Dockerfile
```diff
@@ -17,51 +30,118 @@ RUN apt-get update && apt-get install -y libgeos-dev \
...
+ENV COMPOSER_HOME=tmp \
+    # Set the COMPOSER_MEMORY_LIMIT environment variable to unlimited.
+    COMPOSER_MEMORY_LIMIT=-1 \
+    # Allow root to install plugins.
+    COMPOSER_ALLOW_SUPERUSER=1
+
+# Install apt dependencies.
+RUN apt-get update && apt-get install -y \
+    # Install git and unzip (needed by Composer).
+    git unzip
```
wotnak, line 45: `git` and `unzip` are currently installed and available only in stages related to building farmOS. In my opinion that they should also be available at least in the final dev version of the image but probably also in the final prod image since they are required to install additional modules using composer which is a workflow available in current image versions, so removing it here would be a breaking change.

wotnak: Also, when installing them in the stage that will be included in the final image, it would be good to add `--no-install-recommends` option to the `apt-get install` command to skip installing additional packages that are not really needed and only take up space in the image. It is a good practice to use this option in all instances of apt-get in docker images.

docker/build-farmOS.sh
```diff
@@ -62,15 +54,5 @@ for plugin in ${allowedPlugins[@]}; do
...
-  composer install "$*"
-fi
-
-# Set the version in farm.info.yml.
-sed -i "s|version: 3.x|version: ${FARMOS_VERSION}|g" /var/farmOS/web/profiles/farm/farm.info.yml
-
-# Remove the Composer cache directory.
-rm -rf "$COMPOSER_HOME"
+# Create folder for composer installations.
+mkdir -p ${FARMOS_PATH}/web/sites/simpletest/browser_output
```
wotnak: This line has wrong description in comment. Also, it should be probably moved back to the dev docker image (maybe to the phpunit.sh script), build-farmOS.sh script is executed both for dev and prod images. For production, setup for executing automated tests is not needed.

docs/development/environment/docker.md
```diff
@@ -26,3 +28,8 @@ The `3.x-dev` image also provides the following:
    container. If your user ID is not `1000`, build the image with:
    `--build-arg WWW_DATA_ID=$(id -u)`
     - Default: `1000`
+
+To build the development image, you will have to define the target farmos-dev,
+for example:
+
+`docker build --build-arg WWW_DATA_ID=$(id -u)  -t farmos/farmos:3.x-dev --target farmos-dev docker`
```
wotnak: `--target farmos-dev`
maybe cleaner would be `--target dev`, generally in the Dockerfile I think that prefixing stages with `farmos` is a bit redundant since the entire Dockerfile is for farmOS

wotnak (changes requested):

docker/Dockerfile
AlMaVizca: > `git` and `unzip` are currently installed and available only in stages related to building farmOS. In my opinion that they should also be available at least in the final dev version of the image but probably also in the final prod image since they are required to install additional modules using composer which is a workflow available in current image versions, so removing it here would be a breaking change.

Can you, please, share a module that requires some of those packages?
I've tried with a default installation, also adding modules that are displayed during setup, and later on `/setup/modules` and they were installed without errors.

I understand that those packages might be useful for development image. Yet, for production, I think that adding any module without creating a new image for distribution, or without custom shared volumes, might create a scenario where if the container is replaced, it will lose functionality.
Since there is no documentation about shared volumes to persist modules information, I think, it is not a case to resolve in this PR.

AlMaVizca: >Also, when installing them in the stage that will be included in the final image, it would be good to add --no-install-recommends option to the apt-get install command to skip installing additional packages that are not really needed and only take up space in the image. It is a good practice to use this option in all instances of apt-get in docker images.

I've tried to minimize the number of changes that has to do with packages installations, and propose a different structural organization. 
I took the time to evaluate which packages are installed by recommendations, and only `git` and `unzip` are adding extra packages. Would it be enough to add them only to the dev image? Is it worth it?

docker/Dockerfile
AlMaVizca: Nice catch, I'm moving the arguments to the compose-file stage.
The idea of the `baseimage` stage is to facilitate the upgrade of Drupal version.

mstenta:
Thanks for reviewing @wotnak!

> I'm seeing that it failed the check of MariaDB due to connectivity issue

@AlMaVizca if you rebase your branch onto latest `3.x` and force-push then this issue should be fixed. @wotnak fixed it recently in #758.

docker/Dockerfile
AlMaVizca: Ideally, that would be the case. 
I tried that approach before, and I forgot the reason, yet I was able to replicate it again.

If you take a look at the `build-farmOS.sh` script, the installation of the sources is made using [composer project](https://github.com/farmOS/composer-project) and some manual modifications. Therefore, trying to reuse this stage is throwing and error with [this message](https://github.com/AlMaVizca/farmOS/actions/runs/7306844403/job/19911954466#step:9:115):
`This usually happens when composer files are incorrectly merged or the composer.json file is manually edited.`

You can take a look related to these manual modifications in [this thread](https://github.com/farmOS/composer-project/pull/6#issuecomment-1003836159)

Also, while the Drupal image is still using composer 2.6.5, the `--dev` argument is being [deprecated on Composer 3](https://github.com/composer/composer/commit/a0c5d519e985009fc6a50df03cffb6b886674ffa).

docker/Dockerfile
wotnak: I was thinking of community modules, not included in the farmOS itself. For example https://github.com/symbioquine/farmOS_asset_link or https://www.drupal.org/project/farm_map_custom_layers. 

On the [Building farmOS with Composer](https://farmos.org/hosting/composer/#docker) documentation page, there is described approach for adding additional community modules to the composer.json and building docker image based on official production one with these customizations. Currently, in instructions for building a custom image when it comes to installing composer dependencies only `composer install` is provided, without git and unzip in the prod image these instructions will no longer work and will need to be updated to also include installation of git and unzip. So not including git and uzip in production image may break some existing setups when the new farmOS base image version will be pulled.

I would include git and unzip in both dev and prod image to don't break existing documented functionality.

docker/Dockerfile
wotnak: It can be done using `composer update` instead of `composer install`. Since `composer install --no-dev` in `farmos-sources` stage already installed the latest versions of production dependencies, `composer update` in `farmos-dev-sources` won't do anything with them, but it will install missing dev dependencies and will update `composer.lock` to include them.
https://github.com/farmOS/farmOS/commit/dda987c43069c04e42cc5acd7a61ec367d845ebc
https://github.com/wotnak/farmOS/actions/runs/7307749361/job/19913727994

wotnak (changes requested):

docker/Dockerfile
AlMaVizca: Ok, although, I think that it would be more appropriate to add the information on the documentation, so the image is kept as slim as possible, It's the right thing to keep it working as it was.
Thanks for the explanation.

docs/development/environment/docker.md
```diff
@@ -26,3 +28,8 @@ The `3.x-dev` image also provides the following:
    container. If your user ID is not `1000`, build the image with:
    `--build-arg WWW_DATA_ID=$(id -u)`
     - Default: `1000`
+
+To build the development image, you will have to define the target farmos-dev,
```
wotnak:
```suggestion
To build the development image, you will have to define the target dev,
```

docker/Dockerfile
```diff
@@ -17,51 +24,122 @@ RUN apt-get update && apt-get install -y libgeos-dev \
...
+# owned by www-data inside the container, while keeping those files editable by
+# the developer outside of the container.
+# This defaults to 1000, based on the assumption that the developer is running
+# as UID 1000 on the host machine. It can be overridden at image build time with:
+# --build-arg WWW_DATA_ID=$(id -u)
+ARG WWW_DATA_ID=1000
+RUN usermod -u ${WWW_DATA_ID} www-data && groupmod -g ${WWW_DATA_ID} www-data
+
+# Add farmOS sources
+COPY --from=composer-file --chown=www-data:www-data ${FARMOS_PATH} ${FARMOS_PATH}
```
wotnak: This line can be removed, few lines below dev sources are added that already include the composer.json file generated in composer-file stage.

wotnak (changes requested):

wotnak (approved):
Looks good.
Merging this will allow configuring docker image build cache in GitHub actions, which will improve their execution time.

mstenta:
Thanks @AlMaVizca @wotnak! I haven't forgotten about this... I've just had some deadlines this month that needed my attention. I hope to get back to this and review it again soon. In the meantime I'll tag a few other folks in case they have some time/interest in reviewing.

AlMaVizca:
I've tested the code at [3.x branch](https://github.com/AlMaVizca/farmOS/actions/runs/8847111576/job/24294401789) and the same error is happening.
Please let us know if we can help with anything else.

AlMaVizca:
I've runned the actions on my repo and all the tests are checked, please let me know if there is anything else I can help with

AlMaVizca:
@mstenta @paul121 @symbioquine  
Would you be able to take a look at this, please?
I'll be around if there is any change required. 
I hope you have a good week.

mstenta:
@AlMaVizca Thanks for keeping this alive. :-)

It's been a while since I initially looked at this, so I had to reacquaint myself with all the changes. I'm reviewing again now and I still have the same feedback as I did in November:

> it looks like a large number of changes in a small number of commits

I always review commits one at a time. Ideally each commit makes a single atomic change, and does not depend on commits after it. As it stands now, there are many changes mixed across these 5 commits, and some add things that are not used until later commits. Since this commit history will be preserved forever, I want to make sure it is easy to follow for future readers as well, so that the reason for each change is easy to understand.

With that in mind, I've started a branch to work through these changes and clean up the commits: https://github.com/mstenta/farmOS/tree/3.x-docker-multi-stage-mstenta

So far I've only made it through the first two commits. I've made a few changes, but I retained your commit authorship and timestamps.

- ~I renamed `baseimage` to `drupal` because it is essentially just an alias and does not add anything to the Drupal image itself. I also made this clear in a comment for future readers.~ Edit: changed my mind and kept it as `baseimage`, but left the comment describing why.
- I kept the `ARG` statements in the final build stage, per @wotnak's [comment](https://github.com/farmOS/farmOS/pull/751#discussion_r1434439268).
- I renamed `FARMOS_PATH` to `BUILD_PATH` and added a comment above both `ENV` statements to make it clear what they are for. I feel that calling it `FARMOS_PATH` might cause confusion, and indeed we've seen this confusion from users in the past already, so moving this to an `ENV` var (and naming it properly) is a good improvement to help clarify its purpose.
- I removed `ENV COMPOSER_HOME`, because it has no effect. The `build-farmOS.sh` script overrides it with `export COMPOSER_HOME="$(mktemp -d)"`. If we want to expose this for users to override then let's consider that separately and not mix it in with other changes.

I know you have already made some of these changes yourself in later commits of this PR. But they are mixed in with other commits, so it is very hard to consider them independently.

I am just starting to review the 3 other commits. The next one ("Refactor composer dependencies to create the production image") is a particularly large diff, so I need to take some time to understand it all. I will follow up if I have any questions. Please don't force-push any changes to this PR in the meantime. Perhaps we can use my branch to clean up the commits and then force-push that to this PR. Or, it may make sense to merge them separately as multiple PRs. I will give that some thought as I continue to review...

Thanks again for the continued effort on this! It hasn't been a high priority to merge, but these will be good improvements! :-)

AlMaVizca:
@mstenta I understand your points. 
Yes, the third commit is the biggest one because splitting it (at least on my effort) doesn't deliver an image.
What I remember as a summary for that PR, is that the Dockerfile was delegating responsibilities to the `build-farmos` script about how to populate the code sources and install the dependencies, I mean if it was for production or development.
Don't worry, I won't force push anything, and if you find anything to delegate, please let me know.

mstenta:
Thanks @AlMaVizca. I'll continue working through it.

Quick question:

Can you help me understand why `ARG PROJECT_REPO` was changed from `https://github.com/farmOS/composer-project.git` to `https://github.com/farmOS/composer-project/raw/${PROJECT_VERSION}/composer.json`?

Was it because you were trying to remove `git` and `unzip`? Or was there another reason?

We cannot remove `git` and `unzip` because that would be a breaking change, therefore we can't consider doing that until we open a farmOS 4.x branch, according to [semantic versioning](https://semver.org/).

If that was the reason, can I change it back to `https://github.com/farmOS/composer-project.git` to simplify the changes?

mstenta:
@AlMaVizca Rather than try to tease apart all the things that are happening in the last three commits of this PR, I took a quick stab at refactoring the `dev/Dockerfile` into the main `Dockerfile` myself. Hopefully it accomplishes the main goal of this PR as described ("Single Dockerfile using multi-stages for production and dev image"). Apart from moving code around there are no other changes, so it should be easy to review.

Tests are passing, but I would like more eyes on it in case I'm overlooking anything. I have not tested building/using this locally, so I'm just relying on the automated tests.

See my commits here: https://github.com/farmOS/farmOS/compare/3.x...mstenta:farmOS:3.x-docker-multi-stage-mstenta

However, since it is taking a very simple approach of moving code around, the image size may not be reduced at all. I think it will be easier to focus on merging the `Dockerfile`s as a first step, and then open separate PRs for next steps. Does that make sense?

One thing to note with this change is that you MUST set `--target production` or `--target development` when building the image. Otherwise you will end up with the `development` image, since that is built last. So you must specify `--target production` to tell Docker to stop building once that step is done. I don't know if there is a way to avoid that.

If you agree with this approach, perhaps we can preserve this PR for future reference, but close it, and open a new one from my branch. That way, if we want to pick through this PR to find other improvements it is still available. What do you think?

AlMaVizca:
@mstenta 
> Was it because you were trying to remove git and unzip? Or was there another reason?

We are talking about 2 different topics here.

- The change in how to fetch composer.json file. 
I couldn't find any reason to clone the repo to later delete everything, and my thought was to make it simpler.
What would be the case for this to be a breaking change?

- git and zip
These packages were kept on the farmos final image (installed on farmos-baseimage) after the conversation with [wotnak](https://github.com/farmOS/farmOS/pull/751/files#r1435760610)

>One thing to note with this change is that you MUST set --target production or --target development when building the image. Otherwise you will end up with the development image, since that is built last. So you must specify --target production to tell Docker to stop building once that step is done. I don't know if there is a way to avoid that.

The way to avoid this is to have as a final stage without a name. Meaning that it is only required to build development images [link](https://github.com/farmOS/farmOS/blob/6f2ecb7f0392f765e07a74ceccfc5f20238bcaab/docker/Dockerfile#L138)

> If you agree with this approach, perhaps we can preserve this PR for future reference, but close it, and open a new one from my branch. That way, if we want to pick through this PR to find other improvements it is still available. What do you think?

Now, I understand better your process as an organization, and I see there might be small chunks of code that might simplify your review.
With the current changes on your branch, the update to the documentation about how to build the image is missing [link](https://github.com/farmOS/farmOS/pull/751/files#diff-22fa2688ae60fa3f309bfd72d3c737104de47ad85d797dc330d825ffd329e644).
I'm seeing only 2 additional modifications to have the full PR.
1. Extract phpunit regexp to a script, line 126 [link](https://github.com/farmOS/farmOS/compare/3.x...mstenta:farmOS:3.x-docker-multi-stage-mstenta#diff-f34da55ca08f1a30591d8b0b3e885bcc678537b2a9a4aadea4f190806b374ddcR88#L88) 
2. Refactor build-farmos.sh
- Remove the conditional installation command and locate them on the corresponding stage of the Dockerfile
- Check `COMPOSER_{MEMORY_LIMIT,ALLOW_SUPERUSER,NO_INTERACTION}` variables, at the time they were required to build the image successfully.
- The folders cleanup, at line 88 [link](https://github.com/farmOS/farmOS/compare/3.x...mstenta:farmOS:3.x-docker-multi-stage-mstenta#diff-f34da55ca08f1a30591d8b0b3e885bcc678537b2a9a4aadea4f190806b374ddcR88#L88) would be no longer necessary because the source is set on top of the docker image.
 
I would rather be 100% honest than politically correct. Leaving the effort for future possible PRs is pushing the effort forward, whenever it comes. Which personally generates a feeling of wasting time.
I'm not against this method, but I prefer to wrap the issue at hand, instead of piling up technical debts.
Yet, I understand it's not my decision to make. I'll understand whatever path you choose and try to collaborate if I have the energy.

mstenta:
@AlMaVizca Thanks for bearing with me. I know you've put a lot of time into these changes and I appreciate it.

> I couldn't find any reason to clone the repo to later delete everything, and my thought was to make it simpler.

I am not necessarily opposed to this. It just needs to be isolated and explained.

Every change needs consideration by myself and others, so keeping things in separate, small, well-described commits allows us to review each change independently, and decide if there are any considerations or unwanted side-effects with each. It also allows us to triage/prioritize changes individually. For example we can say: "these changes (commits) are simple and we all agree on them, so we can merge them now... these other changes (commits) require more thought and review, so let's move them to the end of the branch (or to another PR), so they don't hold up the easy ones."

And as this PR and the previous one(s) demonstrate, sometimes it takes a while for maintainers to review things, so the simpler the changes are, the more inclined we'll be to pick it up. I'll admit that I wanted to pick this up a few times but felt that I didn't have the time that would be required to bring myself back up to speed and understand all the changes. Yesterday I finally did. :-)

I hope that makes sense.

> What would be the case for this to be a breaking change?

Changing the format of `PROJECT_REPO` from a Git repo URL to a JSON file URL will break build processes for any downstream users who are overriding that `ARG`. Imagine someone who is building their own derivative image of farmOS, but they are setting `PROJECT_REPO` to `https://github.com/user123/my-farmOS-project-fork.git`. This will no longer work for them, forcing them to update their code. We can't do that in farmOS 3.x. It would need to be considered for farmOS 4.x.

> The way to avoid this is to have as a final stage without a name. Meaning that it is only required to build development images [link](https://github.com/farmOS/farmOS/blob/6f2ecb7f0392f765e07a74ceccfc5f20238bcaab/docker/Dockerfile#L138)

Ah good to know! I am new to these `--target` options. So I appreciate the help learning. I will give it try. :-)

> With the current changes on your branch, the update to the documentation about how to build the image is missing [link](https://github.com/farmOS/farmOS/pull/751/files#diff-22fa2688ae60fa3f309bfd72d3c737104de47ad85d797dc330d825ffd329e644).

Ah yes! I'll add that too - I'll copy it from your commit and keep your authorship.

> Refactor build-farmos.sh

Thank you for outlining the other changes. It was hard for me to separate these out in my mind by looking at the combined diff (this also emphasizes the utility of separate commits for review).

> Remove the conditional installation command and locate them on the corresponding stage of the Dockerfile

Can you point out what you mean by this? I'm still not sure I understand what you mean and why it is necessary.

> Check `COMPOSER_{MEMORY_LIMIT,ALLOW_SUPERUSER,NO_INTERACTION}` variables, at the time they were required to build the image successfully.

Maybe I'm misunderstanding. Are these necessary to build the image? The image appears to be building successfully in my branch. https://github.com/mstenta/farmOS/actions/runs/9978908148/job/27576827864

If we are adding new `ENV` vars, or moving them around, this should be done in isolated commits that describe WHY.

> The folders cleanup, at line 88 [link](https://github.com/farmOS/farmOS/compare/3.x...mstenta:farmOS:3.x-docker-multi-stage-mstenta#diff-f34da55ca08f1a30591d8b0b3e885bcc678537b2a9a4aadea4f190806b374ddcR88#L88) would be no longer necessary because the source is set on top of the docker image.

Now that I have a branch started with the "merge Dockerfile" change completed, I'd love to wrap my head around the other cleanup changes you made. I just want them to be in separate commits so that we can document (in the commit message) exactly why the change is being made.

> Leaving the effort for future possible PRs is pushing the effort forward, whenever it comes. Which personally generates a feeling of wasting time.
> I'm not against this method, but I prefer to wrap the issue at hand, instead of piling up technical debts.

Please don't misunderstand. If the commits are related to the intention of this PR (merging the dev `Dockerfile` into the main `Dockerfile`), then we can include them (including cleanup). If they have a different intention, then they should be in a separate PR. I am still trying to understand all of the different changes that were made, so that I can help to decide if that is warranted.

mstenta:
>> The way to avoid this is to have as a final stage without a name. Meaning that it is only required to build development images [link](https://github.com/farmOS/farmOS/blob/6f2ecb7f0392f765e07a74ceccfc5f20238bcaab/docker/Dockerfile#L138)

> Ah good to know! I am new to these `--target` options. So I appreciate the help learning. I will give it try. :-)

Done: https://github.com/mstenta/farmOS/commit/d77a74771865fc448ffea12add3877d36883e408

Appears to work as expected! :+1: https://github.com/mstenta/farmOS/actions/runs/9996017684/job/27629420372

>> With the current changes on your branch, the update to the documentation about how to build the image is missing [link](https://github.com/farmOS/farmOS/pull/751/files#diff-22fa2688ae60fa3f309bfd72d3c737104de47ad85d797dc330d825ffd329e644).

> Ah yes! I'll add that too - I'll copy it from your commit and keep your authorship.

Done: https://github.com/mstenta/farmOS/commit/d3035707c6222a2dd00810dd764c67bfdd8cec54

docker/build-farmOS.sh
```diff
@@ -61,16 +53,3 @@ allowedPlugins=(
 for plugin in ${allowedPlugins[@]}; do
   composer config --no-plugins allow-plugins.$plugin true
 done
-
-# Run composer install with optional arguments passed into this script.
-if [ $# -eq 0 ]; then
-  composer install
-else
-  composer install "$*"
-fi
```
paul121, line 70: Remind me why we support passing optional arguments here?

paul121 (commented):
Thanks for starting this @AlMaVizca, sorry I haven't taken much time to review thus far. I didn't know about the docker build `--target` options, this is nice. Really supportive of these changes!

The commits in @mstenta branch are looking good, agree we should continue migrating to there. @mstenta  could you open a PR soon so it's easier to review?

Some comments on next steps:

- :+1: to extracting phpunit to a build script, this seems like a simple add
- :+1: to moving the `composer install` out of build-farmos.sh into the docker file. (although see my in-line comment below)
- :+1: to consolidating the cleanup steps around "Install apt dependencies and clean up"
- Regarding git + unzip, I agree we should keep these in, even the base image. It seems like a common thing others might use when extending these images. Regardless, I'm glad we can remove/cleanup the "recommended" dependencies which seemed to mostly be attributed to git + unzip. In some ways removing these "recommended" dependencies could be a breaking change as well but we can only do so much. If we really want, we could hold off on this change for a future release, but it seems like low risk.

- Composer ENVs: These all sound good to me. Even if they aren't "required" now, I see all of these as "best practices" and steps to improving the DX with our docker image which is a major part of this PR. Agree they could be set in separate commits but really a single commit with all of these seems sufficient.
  
  - [`COMPOSER_ALLOW_SUPERUSER`](https://getcomposer.org/doc/03-cli.md#composer-allow-superuser): Seems fine to add this, removes warnings and as docs mention "should really only set this if you use Composer as a super use at all times like in docker containers"
  
  - [`COMPOSER_MEMORY_LIMIT`](https://getcomposer.org/doc/03-cli.md#composer-memory-limit): If not set it will default to PHP's memory limit. I wonder if we could encounter issues where users set PHP memory limit to a lower production value and thus affect composer. That should be separate from the development/build process and perhaps this is a "user error" but seems like a simple add to set this to `-1`, even if it isn't "necessary".
  
  - [`COMPOSER_NO_INTERACTION`](https://getcomposer.org/doc/03-cli.md#composer-no-interaction): This seems normal to set in build/CI processes. Even if it isn't necessary now it might prevent future build errors.

paul121:
Regarding commits/process... personally, I feel like I'm constantly learning ever-evolving best practices of how to manage commits, PRs, etc as I contribute and work with new people. I don't have any real big revelations but I'm increasingly of the opinion that PRs should be a super low-friction environment for creating short, simple and descriptive commits to address feedback and minimize contributor burden. I think this improves the review process and makes it easier to see changes as things have been addressed, even if it results in too much granularity for the final commit history. When jumping in to an existing long review it makes it easier to follow the changes over time, right now I'm kinda lost from how this PR originally started to where it is now.

On final merge these some commits could be squashed, modified and renamed as necessary, but I feel there is so much subjectivity in this, its "never perfect", it should be left as responsibility when merging. That said, obviously creating as clean of a commit history from the start will always help!!

docker/build-farmOS.sh
paul121, line 1: Oh, I guess I'm also a little curious if we could remove `build-farmOS.sh` entirely. This logic for the selecting the FARMOS_VERSION feels like it should really be in our github workflows. We could simply have a 3 arguments that this image takes: `PROJECT_REPO` + `PROJECT_VERSION` for git, and `FARMOS_COMPOSER_VERSION`. We could consider a `PROJECT_JSON_URL` as well to support both git + simpler direct download of json files.

With this change and moving the `composer install` steps out there wouldn't be much left in this file.

mstenta:

> @mstenta could you open a PR soon so it's easier to review?

Done: https://github.com/farmOS/farmOS/pull/855

I'll go through all the comments you left and work them into commits in the new PR when I have a chance. :+1:

mstenta:
Closing this PR, but we can still use this for comments that are more relevant here. The new PR to work from is: https://github.com/farmOS/farmOS/pull/855

docker/build-farmOS.sh
mstenta: It is only used for passing `--no-dev` in production builds: https://github.com/farmOS/farmOS/blob/1bcc2942168e390797ee474c5bb914d22c66fc63/docker/Dockerfile#L63

docker/build-farmOS.sh
mstenta: Sure I'm open to considering that. I prefer to do that in a separate PR with a singular purpose, so that it is easy to review. It does not need to block #855 IMO, which I believe is ready to merge right now, since the declared intention of it is accomplished.

mstenta:
> Regarding commits/process... personally, I feel like I'm constantly learning ever-evolving best practices of how to manage commits, PRs, etc as I contribute and work with new people. I don't have any real big revelations but I'm increasingly of the opinion that PRs should be a super low-friction environment for creating short, simple and descriptive commits to address feedback and minimize contributor burden.

Agreed. I hope everything I've said has been in the service of supporting a better process. We need the barrier to contribution to be low. And we need processes and policies that ensure stable, secure, maintainable, and debug-able code.

The rule of "only change one thing at a time" makes it easier for contributors and maintainers alike. :-)

(Of course, how you define "one thing" will always be subjective.)

mstenta:
> 👍 to moving the composer install out of build-farmos.sh into the docker file. (although see my in-line comment below)

I like the idea of dissolving `build-farmOS.sh` altogether, if it makes sense to. But let's just merge #855 (I'm happy to right now if it gets approvals!) and just base this effort off of that. I see no reason to hold that up.

The nice thing about merging changes like this in chunks, in my opinion, is that it allows me to dump the merged chunk completely out of my memory. I don't need to think about it anymore, because it is no longer a branch. It is `HEAD` at that point, and all that matters is what comes next. If that makes sense. :-)

The longer a PR becomes, the more effort it takes to hold the whole thing in one's head. Even if we've all decided the first X commits are "good to go", if a few months go by before we end up merging them, then that may no longer be the case. And when there are multiple efforts happening in parallel it becomes even harder to keep things in "bRAM" (brain random access memory). :-P

It helps for me to think of each commit is a little self-contained box, and ideally each box can be merged into `HEAD` without breaking it. Once some boxes are ready... get em in! Then focus on the next boxes. :-)

So if "merge Dockerfiles" is done, let's merge it. Then let's "dissolve build-farmOS.sh". If you want to get started on that right now go for it! We can merge it quickly if it's simple and clear.

(Sorry for going off on so many dev process tangents in this thread! :laughing:)

mstenta:
Update: #855 has been merged! :tada:

We can make additional PRs for the next changes at any time.

Thanks again for all your work and advocating for these improvements @AlMaVizca! :raised_hands:

PR #755, opened 2023-11-30
Correct alter hook to add password grant to static scopes

paul121:
Simple fix - a little typo meant we were not allowing the `password` grant for any of static scopes we provide.

Right now you get the following error as reported here: https://farmos.discourse.group/t/invalid-scope-when-trying-to-login-to-field-kit/1809

```
{
    "error": "invalid_scope",
    "error_description": "The requested scope is invalid, unknown, or malformed",
    "hint": "Check the `farm_manager` scope",
    "message": "The requested scope is invalid, unknown, or malformed"
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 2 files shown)


mstenta:
Thanks @paul121! Merged.

PR #756, opened 2023-11-30
Use strict identical operator when checking geometry format

paul121:
Right now the `ContentEntityGeometryNormalizer` ends up being used for all normalization of content entities - no bueno!

This doesn't break any core farmOS features, but would break any other modules/features that use the default normalization of content entities. A simple example is using `restws` module and configuring formats to export entities like `/asset/1?_format=json`. Right now this results in an empty array.


mstenta:
Thanks @paul121!

PR #762, opened 2024-01-01
Use paratest to speedup tests in Github actions workflow

wotnak:
Reverts https://github.com/farmOS/farmOS/commit/6a9c872e9b04a5893e92f5e764aef1904303e368 and adds fix for the issue because of which paratest was removed.

> PostgreSQL tests fail because Drupal core patch is trying to install pg_trgm
extension in parallel multiple times.
> Failed to run installer database tasks: Drupal could not check for the pg_trgm extension: SQLSTATE[23505]: Unique violation: 7 ERROR:  duplicate key value violates unique constraint &quot;pg_extension_name_index&quot;
DETAIL:  Key (extname)=(pg_trgm) already exists.: CREATE EXTENSION IF NOT EXISTS pg_trgm;

To fix this, I added a step executed only when testing with Postgres that installs pg_trgm before starting tests so that automatic extension installation provided by Drupal is not executed during tests.

### DIFF
--- a/.github/workflows/deliver.yml
+++ b/.github/workflows/deliver.yml
@@ -125,6 +125,10 @@ jobs:
       # The www-container-fs-ready file is only created once we expect the containers to be online
       # so waiting for that lets us know it is safe to start the tests
         run: until [ -f ./www/www-container-fs-ready ]; do sleep 0.1; done
+      - name: Install pg_trgm PostgreSQL extension
+      # This avoids race conditions when trying to automatically install it in concurrently run tests.
+        if: matrix.dbms == 'pgsql'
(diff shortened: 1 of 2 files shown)


mstenta:
Oh this is great! Thanks @wotnak! Nice and simple. :-)

I'll run tests a few times to confirm this works, but it LGTM!

mstenta:
I've run tests 4 times (started a 5th just now), and they've all passed. Before this they failed every time with `paratest`, so I think it's safe to say this works! :-)

I'll tag @paul121 and @symbioquine for their reviews too, but I think this is good to go. Thanks again @wotnak!

mstenta (approved):

paul121 (approved):
LGTM!

PR #763, opened 2024-01-01
Use phpcs and phpstan result cache in github actions

wotnak:
Adds several optimizations to how phpcs and phpstan are run in github actions:
- run phpcs checks in parallel
  - phpcs by default runs all checks in series but provides cli argument that allows to execute checks in parallel
- combine PHP compatibility checks of contrib modules and themes into single phpcs run
  - passing two paths at the same time allows to skip additional phpcs bootstrapping time
- use phpcs results cache and persist it between workflow runs
  - using results cache can significantly speedup performing checks since checks are run only for files that changed or are missing from cache
 - persist phpstan result cache between workflow runs
   - phpstan uses results cache by default but in github actions it is discarded after container stops, this change allows for it to be shared between workflow runs

### DIFF
--- a/.github/workflows/deliver.yml
+++ b/.github/workflows/deliver.yml
@@ -71,13 +71,38 @@ jobs:
           key: farmos-dev-${{ github.run_id }}
       - name: Load farmos/farmos:3.x-dev image
         run: docker load < /tmp/farmos-dev.tar
+      - name: Create cache directory
+        run: mkdir -p ./cache
+      - name: Load phpcs cache
+        uses: actions/cache/restore@v3
+        with:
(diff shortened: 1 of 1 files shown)


paul121 (commented):
This looks great, thanks @wotnak.

The one thing I'm hung up on is caching... I see how this could be nice but it feels like it *could be* a lot of complexity to maintain here. I'm really just not too familiar with github actions caching.. kinda wonder if it could cause any false positive "passing" tests, but in theory it shouldn't?

I assume the phpcs in parallel would still be the largest improvement anyways? Maybe we just start with that?

PR #765, opened 2024-01-04
Update to Drupal 10.2

paul121:
Updated a Drupal core patch [2339235](https://www.drupal.org/project/drupal/issues/2339235#comment-15382658) and got tests to pass, but have not tested this locally.

10.2.0 release notes: https://www.drupal.org/project/drupal/releases/10.2.0

I don't think there are any major changes to be worried about here....

> [Form elements in the stable9 base theme now have form-type-[form-element-type] class](https://www.drupal.org/node/3385635).

This could potentially affect some UI/CSS changes we have made, but seems unlikely.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - [Increase weight of Asset and Log tasks on canonical user route #757](https://github.com/farmOS/farmOS/pull/757)
+- Update Drupal core to 10.2.0.
 
 ### Deprecated
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - [Increase weight of Asset and Log tasks on canonical user route #757](https://github.com/farmOS/farmOS/pull/757)
-- Update Drupal core to 10.2.0.
+- [Update Drupal core to 10.2](https://github.com/farmOS/farmOS/pull/765)
 
 ### Deprecated
 
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
     "require": {
         "cweagans/composer-patches": "^1.6",
         "drupal/admin_toolbar": "^3.3",
-        "drupal/core": "10.1.6",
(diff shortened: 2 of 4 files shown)


mstenta:
This LGTM! Can I merge it? Or did you leave it in draft for a reason?

> Remove time format from description 

Interesting. Does this change also fix https://github.com/farmOS/farmOS/issues/625, perhaps?

mstenta:
> Does this change also fix https://github.com/farmOS/farmOS/issues/625, perhaps?

Nevermind, I don't think it fixes that issue, but removing the format is a good change because as we can see in those screenshots it is wrong most of the time anyway.

paul121:
@mstenta and I chatted about this today and we feel pretty good. Should be fine to merge this and we could tag 3.0.0 stable soon after. If there are any issues, they can be fixed as we go.

mstenta (approved):
LGTM!

Also, 10.2.1 was released today, so I added a commit to update to that. And I found the 10.2.x patch you made here @paul121: https://www.drupal.org/project/drupal/issues/2339235#comment-15382708

Thanks!

mstenta (changes requested):
Dang. Now tests are failing. :thinking:

```
There was 1 failure:

1) Drupal\Tests\farm_inventory\Functional\InventoryTest::testApiInventory
Failed asserting that an array is not empty.

/opt/drupal/vendor/phpunit/phpunit/src/Framework/Constraint/Constraint.php:121
/opt/drupal/vendor/phpunit/phpunit/src/Framework/Constraint/Constraint.php:55
/opt/drupal/web/profiles/farm/modules/core/inventory/tests/src/Functional/InventoryTest.php:111
/opt/drupal/vendor/phpunit/phpunit/src/Framework/TestResult.php:728
```

mstenta:
Huh. My test passed locally. But I'm running PostgreSQL and the failure was on SQLite3... :thinking:

paul121:
Huh, I think I saw this same error when I first made this PR but re-ran tests and it passed.. maybe there is a new race condition of some kind. Not sure.

mstenta:
Wow I think you're right @paul121. I ran tests 4 times, and it failed 3 times, but just now it passed once. So that suggests it is not 10.2.1 that's causing the problem. Huh. Well I guess we can merge this and tag 3.0.0 after all! We can figure out the test issue as a follow up...

PR #766, opened 2024-01-04
Inventory quick form

mstenta:
This adds an Inventory quick form. For more details on the design and discussion, refer to the following links:

- https://farmos.discourse.group/t/interest-in-an-inventory-adjustment-quick-form/1096
- https://gitlab.com/OpenTEAMAg/feedback/support_requests/-/issues/181

### DIFF
--- a/modules/quick/inventory/src/Plugin/QuickForm/Inventory.php
+++ b/modules/quick/inventory/src/Plugin/QuickForm/Inventory.php
@@ -13,6 +13,7 @@ use Drupal\farm_quick\Plugin\QuickForm\QuickFormBase;
 use Drupal\farm_quick\Plugin\QuickForm\QuickFormInterface;
 use Drupal\farm_quick\Traits\QuickFormElementsTrait;
 use Drupal\farm_quick\Traits\QuickLogTrait;
+use Drupal\log\Entity\Log;
 use Psr\Container\ContainerInterface;
 
 /**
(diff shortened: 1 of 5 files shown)


paul121 (changes requested):
I'm getting an error when I try to configure the `units` on the quick form with a new term that does not exist. It gets a 500 error with this log message:

> InvalidArgumentException: The configuration property settings.units.entity doesn't exist. in Drupal\Core\Config\Schema\ArrayElement->get() (line 76 of /var/www/html/web/core/lib/Drupal/Core/Config/Schema/ArrayElement.php).

mstenta:
> I'm getting an error when I try to configure the `units` on the quick form with a new term that does not exist. 

Oh good catch @paul121!

This made me think of something related: maybe we should store the term label instead of the term ID in the configuration entity for inventory quick forms. This would allow the config YML to be shared between farmOS instances without any hard-coded term IDs. The term can be generated on-demand using our `QuickTermTrait::createOrLoadTerm()` method.

I've pushed a commit to that effect. Requesting your review...

I also rebased onto latest `3.x` and added a dependency on `farm_quick` that was missing before.

modules/quick/inventory/src/Plugin/QuickForm/Inventory.php
```diff
@@ -121,7 +123,7 @@ public function defaultConfiguration() {
     return [
       'asset' => NULL,
       'measure' => NULL,
-      'units' => NULL,
+      'units' => '',
```
paul121: Is there any reason this can't be `NULL` by default?

paul121 (approved):

modules/quick/inventory/src/Plugin/QuickForm/Inventory.php
mstenta: Oh sure it could be. I'll fixup that.

modules/quick/inventory/src/Plugin/QuickForm/Inventory.php
```diff
@@ -494,7 +496,10 @@ public function buildConfigurationForm(array $form, FormStateInterface $form_sta
    */
   public function submitConfigurationForm(array &$form, FormStateInterface $form_state) {
     $this->configuration['asset'] = $form_state->getValue('asset');
-    $this->configuration['units'] = $form_state->getValue('units');
+    $this->configuration['units'] = '';
```
paul121: I believe this would end up saving an empty string if nothing was provided?

mstenta: True, I suppose it should explicitly set it to `NULL` if that's the new default. I'll change it to `NULL`.

We do need to explicitly set it here otherwise there's no way to "unset" units in configuration.

PR #768, opened 2024-01-11
Add min/max options to integer fields in farm_field.factory

mstenta:
This is a simple PR that just adds `min` and `max` options to the `integer` field type in the `farm_field.factory` service (https://farmos.org/development/module/services/#field-factory-service).

### DIFF
--- a/modules/core/field/src/FarmFieldFactory.php
+++ b/modules/core/field/src/FarmFieldFactory.php
@@ -781,6 +781,14 @@ class FarmFieldFactory implements FarmFieldFactoryInterface {
       $field->setSetting('size', $options['size']);
     }
 
+    // Set the min/max constraints, if specified.
+    if (isset($options['min'])) {
+      $field->setSetting('min', $options['min']);
+    }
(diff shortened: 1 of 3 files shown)


modules/core/field/src/FarmFieldFactory.php
```diff
@@ -781,6 +781,14 @@ protected function modifyIntegerField(BaseFieldDefinition &$field, array $option
       $field->setSetting('size', $options['size']);
     }
 
+    // Set the min/max constraints, if specified.
+    if (!empty($options['min'])) {
```
paul121: `!empty()` will not work for a min/max value of `0`. I think we should use `isset()`.

paul121 (changes requested):

modules/core/field/src/FarmFieldFactory.php
mstenta: Oh good point @paul121 thanks for reviewing! I'll make this change and re-request review.

paul121 (approved):

PR #770, opened 2024-01-16
Edit form UI improvements

paul121:
Implements the gin content form and moves fields to vertical tabs within the form and sidebar. Follow up from #720 

More discussion here: https://farmos.discourse.group/t/ui-improvements-editing-forms/1743?u=paul121

### DIFF
--- a/modules/core/ui/theme/farm_ui_theme.api.php
+++ b/modules/core/ui/theme/farm_ui_theme.api.php
@@ -15,6 +15,27 @@
  */
 
 /**
+ * Add entity form field groups.
+ *
+ * @param string $entity_type
+ *   The entity type.
+ * @param string $bundle
+ *   The bundle.
+ */
+function hook_farm_ui_theme_field_groups(string $entity_type, string $bundle) {
+  if ($entity_type == 'asset' && $bundle == 'animal') {
+    return [
+      'custom' => [
+        'location' => 'main',
+        'title' => t('Custom'),
+        'weight' => 50,
+      ],
+    ];
+  }
+  return [];
+}
+
+/**
  * Specify the field groups to place entity fields in.
  *
  * @param string $entity_type
--- a/modules/core/ui/theme/farm_ui_theme.api.php
+++ b/modules/core/ui/theme/farm_ui_theme.api.php
@@ -15,6 +15,27 @@
  */
 
 /**
+ * Specify the field groups to place entity fields in.
+ *
+ * @param string $entity_type
+ *   The entity type.
+ * @param string $bundle
+ *   The bundle.
+ *
+ * @return string[]
+ *   An array keyed by field ID mapping to field group.
+ */
+function hook_farm_ui_theme_field_group_items(string $entity_type, string $bundle) {
+  if ($entity_type == 'asset' && $bundle == 'animal') {
+    return [
+      'nickname' => 'bundle',
+      'sex' => 'bundle',
+    ];
+  }
+  return [];
+}
+
+/**
  * Specify the regions that asset, log, and plan content items should be in.
  *
  * @param string $entity_type
--- a/modules/core/ui/theme/farm_ui_theme.module
+++ b/modules/core/ui/theme/farm_ui_theme.module
@@ -25,6 +25,11 @@ function farm_ui_theme_theme($existing, $type, $theme, $path) {
     'page__asset__map_popup' => [
       'base hook' => 'page',
     ],
+    // Implement the node edit form theme hook.
+    // See https://www.drupal.org/project/gin/issues/3342164
+    'node_edit_form' => [
+      'render element' => 'form',
+    ],
   ];
 }
 
@@ -45,6 +50,19 @@ function farm_ui_theme_theme_suggestions_menu_local_task(array $variables) {
  */
 function farm_ui_theme_theme_suggestions_menu_local_tasks(array $variables) {
   return ['menu_local_tasks__farm'];
+}
+
+/**
+ * Implements hook_gin_content_form_routes().
+ */
+function farm_ui_theme_gin_content_form_routes() {
+  $routes = [];
+  $entity_types = ['asset', 'log', 'plan'];
+  foreach ($entity_types as $entity_type) {
+    $routes[] = "entity.$entity_type.add_form";
+    $routes[] = "entity.$entity_type.edit_form";
+  }
+  return $routes;
 }
 
 /**
--- a/modules/core/ui/theme/farm_ui_theme.module
+++ b/modules/core/ui/theme/farm_ui_theme.module
@@ -79,6 +79,46 @@ function farm_ui_theme_entity_form_display_alter(EntityFormDisplayInterface $for
 }
 
 /**
+ * Implements hook_farm_ui_theme_field_group_items().
+ */
+function farm_ui_theme_farm_ui_theme_field_group_items(string $entity_type, string $bundle) {
+
+  // Define base fields for asset, log, and plans on behalf of core modules.
+  $fields = [
+    'name' => 'entity',
+    'status' => 'entity',
+    'flag' => 'meta',
+    'file' => 'file',
+    'image' => 'file',
+    'revision' => 'revision',
+    'revision_log_message' => 'revision',
+  ];
+  switch ($entity_type) {
+    case 'asset':
+      $fields['owner'] = 'meta';
+      $fields['intrinsic_geometry'] = 'location';
+      $fields['is_location'] = 'location';
+      $fields['is_fixed'] = 'location';
+      break;
+
+    case 'log':
+      $fields['timestamp'] = 'entity';
+      $fields['category'] = 'meta';
+      $fields['owner'] = 'meta';
+      $fields['geometry'] = 'location';
+      $fields['location'] = 'location';
+      break;
+
+    case 'plan':
+      break;
+
+    default:
+      $fields = [];
+  }
+  return $fields;
+}
+
+/**
  * Implements hook_gin_content_form_routes().
  */
 function farm_ui_theme_gin_content_form_routes() {
--- a/modules/core/ui/theme/farm_ui_theme.module
+++ b/modules/core/ui/theme/farm_ui_theme.module
@@ -8,7 +8,9 @@
 use Drupal\Core\Block\BlockPluginInterface;
 use Drupal\Core\Entity\Display\EntityFormDisplayInterface;
 use Drupal\Core\Form\FormStateInterface;
-use Drupal\farm_ui_theme\Form\GinContentFormBase;
+use Drupal\farm_ui_theme\Form\AssetForm;
+use Drupal\farm_ui_theme\Form\LogForm;
+use Drupal\farm_ui_theme\Form\PlanForm;
 
 /**
  * Implements hook_theme().
@@ -97,18 +99,22 @@ function farm_ui_theme_farm_ui_theme_field_group_items(string $entity_type, stri
   switch ($entity_type) {
     case 'asset':
       $fields['owner'] = 'meta';
+      $fields['parent'] = 'parent';
       $fields['intrinsic_geometry'] = 'location';
       $fields['is_location'] = 'location';
       $fields['is_fixed'] = 'location';
+      $fields['id_tag'] = 'id_tag';
       break;
 
     case 'log':
       $fields['timestamp'] = 'default';
       $fields['category'] = 'meta';
       $fields['owner'] = 'meta';
+      $fields['asset'] = 'asset';
       $fields['geometry'] = 'location';
       $fields['location'] = 'location';
       $fields['is_movement'] = 'location';
+      $fields['quantity'] = 'quantity';
       break;
 
     case 'plan':
@@ -140,9 +146,9 @@ function farm_ui_theme_entity_type_build(array &$entity_types) {
 
   // Override the default add and edit form class.
   $target_entity_types = [
-    'asset' => GinContentFormBase::class,
-    'log' => GinContentFormBase::class,
-    'plan' => GinContentFormBase::class,
+    'asset' => AssetForm::class,
+    'log' => LogForm::class,
+    'plan' => PlanForm::class,
   ];
   foreach ($target_entity_types as $entity_type => $form_class) {
     if (isset($entity_types[$entity_type])) {
--- a/modules/core/ui/theme/farm_ui_theme.module
+++ b/modules/core/ui/theme/farm_ui_theme.module
@@ -8,6 +8,7 @@
 use Drupal\Core\Block\BlockPluginInterface;
 use Drupal\Core\Entity\Display\EntityFormDisplayInterface;
 use Drupal\Core\Form\FormStateInterface;
+use Drupal\farm_ui_theme\Form\GinContentFormBase;
 
 /**
  * Implements hook_theme().
@@ -129,6 +130,25 @@ function farm_ui_theme_gin_content_form_routes() {
     $routes[] = "entity.$entity_type.edit_form";
   }
   return $routes;
+}
+
+/**
+ * Implements hook_entity_type_build().
+ */
+function farm_ui_theme_entity_type_build(array &$entity_types) {
+
+  // Override the default add and edit form class.
+  $target_entity_types = [
+    'asset' => GinContentFormBase::class,
+    'log' => GinContentFormBase::class,
+    'plan' => GinContentFormBase::class,
+  ];
+  foreach ($target_entity_types as $entity_type => $form_class) {
+    if (isset($entity_types[$entity_type])) {
+      $entity_types[$entity_type]->setFormClass('add', $form_class);
+      $entity_types[$entity_type]->setFormClass('edit', $form_class);
+    }
+  }
 }
 
 /**
--- a/modules/core/ui/theme/farm_ui_theme.module
+++ b/modules/core/ui/theme/farm_ui_theme.module
@@ -86,8 +86,8 @@ function farm_ui_theme_farm_ui_theme_field_group_items(string $entity_type, stri
 
   // Define base fields for asset, log, and plans on behalf of core modules.
   $fields = [
-    'name' => 'entity',
-    'status' => 'entity',
+    'name' => 'default',
+    'status' => 'meta',
     'flag' => 'meta',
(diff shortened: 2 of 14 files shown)


modules/core/ui/theme/farm_ui_theme.module
```diff
@@ -25,6 +25,9 @@ function farm_ui_theme_theme($existing, $type, $theme, $path) {
     'page__asset__map_popup' => [
       'base hook' => 'page',
     ],
+    'node_edit_form' => [
```
mstenta, line 34: Can we add a comment here so it's obvious to future readers why this is here? Someone looking at this `hook_theme()` for the first time might wonder why there is something related to the `node` module, when farmOS doesn't enable `node`.

modules/core/ui/theme/farm_ui_theme.module
```diff
@@ -50,6 +51,32 @@ function farm_ui_theme_theme_suggestions_menu_local_tasks(array $variables) {
...
+  $entity_types = ['asset', 'log', 'plan'];
+  if (!in_array($context['entity_type'], $entity_types)) {
+    return;
+  }
+
+  // Ask modules for a list of region items.
+  $field_map = \Drupal::moduleHandler()->invokeAll(
+    'farm_ui_theme_field_group_items',
+    [$context['entity_type'], $context['bundle'],
+    ],
```
mstenta: Nit: this newline seems a bit weird. Codesniffer didn't complain, but could we change it to `[$context['entity_type'], $context['bundle']],` on one line?

modules/core/ui/theme/farm_ui_theme.api.php
```diff
@@ -14,6 +14,27 @@
...
+ * @param string $entity_type
+ *   The entity type.
+ * @param string $bundle
+ *   The bundle.
+ *
+ * @return string[]
+ *   An array keyed by field ID mapping to field group.
+ */
+function hook_farm_ui_theme_field_group_items(string $entity_type, string $bundle) {
+  if ($entity_type == 'asset' && $bundle = 'animal') {
```
mstenta: Should be `$bundle == 'animal'` (two `=`s). Inconsequential, because it's just an example, but might as well fix.

mstenta (changes requested):
Nice this is amazing @paul121!

I noted a few small things in code review, mostly nits.

I'm testing it out now in my local instance, and noticed one small issue on assets: when the sidebar is open, the ID Tags field gets cut off, and the "Remove" buttons get covered. I don't think this needs to block the PR, because the workaround is to collapse the sidebar. But maybe there's an easy CSS fix? Seems like it's related to inline entity forms.

Screenshots:

Screenshot from 2024-01-16 11-31-14
Screenshot from 2024-01-16 11-31-07

mstenta:
> Seems like it's related to inline entity forms.

Ah nevermind... the ID tag fields provide their own widget. No relationship to inline entity form (they are fields not entities). And quantities (which do use inline entity form) look fine on log forms.

mstenta:
@paul121 I took a quick stab at moving the "Group" and "Equipment" fields around a bit. I pushed to my own branch, but if you agree with the approach feel free to cherry-pick them over to this PR. It basically adds a new hook to allow modules to alter the field groups, then adds hooks to both the `farm_group` and `farm_equipment` modules to move their fields. The `farm_group` module adds a `group` field group and moves the `group` and `is_group_assignment` fields to it. The `farm_equipment` module just moves the `equipment` field to the `asset` field group. Here are the commits:

- https://github.com/mstenta/farmOS/commit/7a1de146e5ea974cf61539e390b599da2ec4581c
- https://github.com/mstenta/farmOS/commit/a08fcde0eb1a0347f87041507053f403dc0d18bf
- https://github.com/mstenta/farmOS/commit/533898637ebe5775bd8e92b0881a5af9a0b1acf0

paul121:
> But maybe there's an easy CSS fix?

Oooh yeah. I'll take a look and see if it might be an easy fix. I think this is an existing issue anyways, but only noticeable if you make your window quite narrow. But still these content form changes would make it a bit less usable.

> It basically adds a new hook to allow modules to alter the field groups, then adds hooks to both the farm_group and farm_equipment modules to move their fields.

Nice! I hadn't thought about making it an alter hook but that is a simple approach. It does allow others to fully alter these groups.... hopefully not too dangerous. Whereas a separate hook to only add new groups would prevent that. Likely not a big deal.

Also reminds me that at some point we should probably move on from hooks and use an event subscriber (iirc?) for this. I know we have various other custom hooks, but would it be worth exploring those options now to prevent changes in the future? At some point we should decide on a new approach and stick to that for any "new features". I haven't looked at these ideas for while, I wonder if D10/D11 has brought any new efforts.

mstenta:
> Nice! I hadn't thought about making it an alter hook but that is a simple approach. It does allow others to fully alter these groups.... hopefully not too dangerous. Whereas a separate hook to only add new groups would prevent that. Likely not a big deal.

Yea a regular hook (non-alter) would also work, and would actually be better... I think typical practice in Drupal is (or at least was) to always provide both, so downstream devs have lots of options. The alter hook sort of achieves both, but it means that you need to use alter to add field groups too... which then makes it harder for other modules to alter those, because of the order of operations module hooks run... although there are ugly workarounds for that too. So maybe we should actually provide both as a first step. I can take a pass at that and update my commits accordingly...

> Also reminds me that at some point we should probably move on from hooks and use an event subscriber (iirc?) for this. I know we have various other custom hooks, but would it be worth exploring those options now to prevent changes in the future? At some point we should decide on a new approach and stick to that for any "new features". I haven't looked at these ideas for while, I wonder if D10/D11 has brought any new efforts.

Yes definitely. Hooks are on the way out, although I think it will be a long tail and I doubt they will ever go away completely. In this particular case, I think it's OK for us to add a few more to *this* module (`farm_ui_theme`), since it already provides `hook_farm_ui_theme_region_items()`. But I would love to take a broader look at deprecating these and using event subscribers instead where we can. Perhaps we can open an issue to look into doing that in 3.x, and plan to remove them in 4.x. In either case, for this particular case I think it's OK (and arguably better) to be consistent and use hooks for field groups like we already do for regions provided by `farm_ui_theme`.

mstenta:
> But I would love to take a broader look at deprecating these and using event subscribers instead where we can.

To clarify this, I meant to imply we can look into deprecating ALL hooks that farmOS provides... not just in `farm_ui_theme`. I forget what other ones are out there but we might as well do them all at once as part of a broad stroke effort separate from this PR.

mstenta:
>  So maybe we should actually provide both as a first step. I can take a pass at that and update my commits accordingly...

I updated my commits and force-pushed my branch. Here they are for review:

https://github.com/mstenta/farmOS/commit/55e8fa01813300071e30dd15e7c0530adc58c31f
https://github.com/mstenta/farmOS/commit/63ca8e0c820347936b4d387cb865ce8d23f6417f
https://github.com/mstenta/farmOS/commit/bf0e95e08fe0fafe87953a6e4f5d9d25f6f9d8c3

modules/core/ui/theme/src/Form/GinContentFormBase.php
```diff
@@ -0,0 +1,223 @@
...
+    $form = parent::form($form, $form_state);
+
+    // Add process callback to alter form after GinContentFormHelper.
+    $form['#process'][] = '::processContentForm';
+
+    // Only alter the form display if farm_ui_theme.use_field_group is TRUE
+    // or if the form display is new and not saved.
+    /** @var \Drupal\Core\Entity\Display\EntityFormDisplayInterface $form_display */
+    $form_display = $form_state->get('form_display');
+    if (!$form_display || !$form_display->getThirdPartySetting('farm_ui_theme', 'use_field_group', $form_display->isNew())) {
```
mstenta, line 127: I just encountered a bug caused by this logic. I was experimenting with adding a new field to Equipment assets through the UI, which created a `core.entity_form_display.asset.equipment.default` config entity. After I did this the content forms were broken on Equipment assets.

Screenshot:

Screenshot from 2024-01-17 12-13-39

paul121:
> To clarify this, I meant to imply we can look into deprecating ALL hooks that farmOS provides... not just in farm_ui_theme. I forget what other ones are out there but we might as well do them all at once as part of a broad stroke effort separate from this PR.

I agree we should likely change all existing ones at once... but IMO it would be good to identify a new solution sooner than later and start adopting it for any new features we add. Otherwise we are implementing things now knowing they will need to change in the future. We have a chance to save ourselves some time in the future, how rare is that :smile: 

And we do already have some event subscribers so it's not a totally new concept for developers to use. I believe that event subscribers also have a feature where the single event subscriber can serve to provide + alter the data in question. Just to say, I think we should get familiar with this sooner than later, and this could be an opportunity!

>  In either case, for this particular case I think it's OK (and arguably better) to be consistent and use hooks for field groups like we already do for regions provided by farm_ui_theme.

Sure, consistency is kinda nice in this case. I'd be OK using hooks for now. It shouldn't be used a whole lot anyways, otherwise I might be more persistent.

modules/core/ui/theme/src/Form/GinContentFormBase.php
paul121: Ah yeah. It's really a feature tbh, so as to not break any existing form displays someone might have configured ;-)

We should be able to alter the form-display form to include this third party setting boolean. I didn't do this for now considering that modifying these view/form displays is a more advanced thing and we can expect the `use_field_group` third party setting to be configured manually on the config if needed. If you disagree we could try modifying the form here.

mstenta: I see what you mean! That makes sense.

However, I think we should hide the blank sidebar that appears in those cases. That's what "feels broken".

Is that easy to do?

mstenta:
> I agree we should likely change all existing ones at once... but IMO it would be good to identify a new solution sooner than later and start adopting it for any new features we add.

Agreed!

I created a drupal.org issue for us to use for this: https://www.drupal.org/project/farm/issues/3415475

> And we do already have some event subscribers so it's not a totally new concept for developers to use. I believe that event subscribers also have a feature where the single event subscriber can serve to provide + alter the data in question.

Yup! `MapRenderEvent` in the `farm_map` module is a good example: https://github.com/farmOS/farmOS/blob/3.x/modules/core/map/src/Event/MapRenderEvent.php

We also implemented `AssetEvent`, `QuantityEvent`, and `DataStreamEvent` which get called by the Drupal core `presave`, `insert`, `update`, and `delete` hooks, while we wait for core to start using Events for those itself. (We did this first in the the Log module w/ `LogEvent` and then replicated it in farmOS core entities. Although it looks like we're missing `Plan` and maybe others.)

Here is the core issue: https://www.drupal.org/project/drupal/issues/2551893

And here is our old issue from the 2.x dev cycle: https://www.drupal.org/project/farm/issues/3306227

> Sure, consistency is kinda nice in this case. I'd be OK using hooks for now. It shouldn't be used a whole lot anyways, otherwise I might be more persistent.

Yea in this case it's six of one half dozen of the other. I'm inclined to say let's just add these two hooks for now, and then add Events for all *three* hooks `farm_ui_theme` provides at the same time. Otherwise I feel like we should add an Event for `hook_farm_ui_theme_region_items()` and deprecate that *before* we add anything new here. Otherwise we're not being consistent within this one module, and if years go by (hope not!) before we get around to cleaning things up then it's just another inconsistency sitting around. And in this particular case I don't think it will add much extra work in the future.

All that said, I'm optimistic that the new issue I opened will be pretty simple! I'll take the first step of documenting all the hooks right now, and who knows maybe it will move fast! (If it's really easy maybe it will end up winning the PR race and getting in before this one! In which case the new hook decision might be moot! :smile:)

mstenta:
OK! I did a pretty detailed review of our current Events and hooks here: https://www.drupal.org/project/farm/issues/3415475

Tl;dr: I think the `farm_ui_theme` hooks will actually be some of the easiest ones to convert to Events.

But that said, it's still work... and most of that work is just taking the first step. There isn't going to be much of a difference between converting one and converting three, IMO. And I stand by this conviction:

> I feel like we should add an Event for `hook_farm_ui_theme_region_items()` and deprecate that before we add anything new here.

So, I think we have two choices:

1. Block this PR on deprecating `hook_farm_ui_theme_region_items()` and work on that first.
2. Add these two new hooks to `farm_ui_theme` and deprecate all three together in a follow-up.

I'm more inclined towards 2 just because it doesn't feel like a big deal and I'm eager to get this PR merged. But I'm not opposed to 1! Heck maybe I'll get some more motivation soon and tackle it myself. :-)

modules/core/ui/theme/src/Form/GinContentFormBase.php
paul121: Oh the blank sidebar, hmm. This might be tricky because we tell Gin to use the content editing form on our `entity.{type}.add_form` and `entity.{type}.edit_form` routes so it might always try to apply these styling/changes regardless of the form mode.

But I have an idea. We can unset some of the changes that gin makes to this page, specifically the sidebar. And in the process we can still keep the "Save" button in the top right which is still a nice improvement for a standard form. It is a little hacky but works well. I'll push a commit up with this:

```php
    // In GinContentFormBase::processContentForm
    // Remove the sidebar if the display is not using field groups.
    /** @var \Drupal\Core\Entity\Display\EntityFormDisplayInterface $form_display */
    $form_display = $form_state->get('form_display');
    if (!$form_display || !$form_display->getThirdPartySetting('farm_ui_theme', 'use_field_group', $form_display->isNew())) {
      if (($index = array_search('gin/sidebar', $form['#attached']['library'])) !== false) {
        unset($form['#attached']['library'][$index]);
      }
      unset($form['gin_sidebar']);
      unset($form['gin_sidebar_overlay']);
      unset($form['gin_actions']['gin_sidebar_toggle']);
    }
```

paul121:
Question: is it actually a requirement to allow altering the provided field groups? I don't think we have a clear use-case for renaming an existing field group. I guess you could also use the field groups alter to move them to the sidebar or change their weight. But really in this scenario it would be better to just provide a new field group and define it's behavior as you'd like.

More importantly, I think we're missing the ability to alter `hook_farm_ui_theme_field_group_items`.

I was about to say I'm fine with choice 2, just do hooks. Buttttt there is something really nice about just using events and avoiding the need to communicate the `alter` concept altogether. Especially for this particular use-case I think it is more advanced than most modules will need, but events still allow that as an option. This feels like the "correct" way.

modules/core/ui/theme/src/Form/GinContentFormBase.php
paul121: Example:
Screenshot from 2024-01-18 18-54-10

mstenta:
Update: I spent a little time yesterday experimenting with converting `hook_farm_ui_theme_region_items()` to an event, and I'm not sure it's worth the effort. It gave me a chance to refresh on how that two-column stacked layout is applied, and I think our time would be better spent improving how that works overall, rather than further standardizing it with a new event.

This is just to say that I'm softening on my previous statement:

> I feel like we should add an Event for `hook_farm_ui_theme_region_items()` and deprecate that before we add anything new here. Otherwise we're not being consistent within this one module

I don't think that needs to be a determining factor. If we want to use events for field groups, but leave the hook for region items, I can live with it. :-)

Or, if we want to just add *two* hooks for field groups now (and omit the `_alter()` hook), and then deprecate them in the near future (https://www.drupal.org/project/farm/issues/3415475) I'm fine with that too. It would get this feature merged sooner - we're already done! :-)

> Question: is it actually a requirement to allow altering the provided field groups?

No it's not. So if we want to consider merging this with two hooks, and no alters, we can.

I've updated my commits to remove the alter hook. So it's up to you @paul121 if you want to cherry-pick those and call this done or refactor the hooks into an event now.

https://github.com/farmOS/farmOS/commit/e30cd1ead3fbeda542b80dd8d2dfa568ad74c4a7
https://github.com/farmOS/farmOS/commit/89da1d14b876444f1238c6e2bbf17315727bf94f
https://github.com/farmOS/farmOS/commit/d6c085273533c3176ee5a282b85861f385d4d4d1

mstenta:
> if you want to cherry-pick those and call this done

With the exception of the blank sidebar issue... see my latest comment above...

modules/core/ui/theme/src/Form/GinContentFormBase.php
mstenta: Hmm I'm still seeing a blank sidebar with this change. But it did remove the button for opening and closing it.

Screenshot from 2024-01-20 12-22-59

paul121: Nice catch. I thought removing the `gin/sidebar` js would solve this, but looks like Gin's global `init.js` includes logic for the sidebar that will still open this when the `#gin_sidebar` element is present. This element is provided by Gin's `node_edit_form` theme/template so completely removing the theme solves the problem:

```php

// Do not use the node_edit_form theme.
// The template includes elements for the sidebar.
unset($form['#theme']);
```
      
But unfortunately completely removing the `node_edit_form` means that we won't get the "Save" button at the top. That would have been nice, but I guess we can't have everything :shrug:

paul121:
I found and fixed a couple other issues:
- Delete button was removed from the sidebar in content forms. Fixed by including Gin's `meta` group at the bottom of the sidebar, which also has some dedicated styling which is nice to distinguish it as an "actions" part of the sidebar.
- Without the sidebar enabled the revision information was not being included. Fixed by keeping the default revision information behavior (vertical tabs + details element) and slightly decreasing the `Notes` field weight from 100 to 95 so it is above the revision area.

> I've updated my commits to remove the alter hook. So it's up to you @paul121 if you want to cherry-pick those and call this done or refactor the hooks into an event now.

Done. This is good, we've simplified things down as much as we can and identified ways to improve & potentially consolidate all of this in the future :100:

modules/core/ui/theme/src/Form/GinContentFormBase.php
mstenta: This looks like it's working better now - thanks @paul121!

Yea too bad about the "Save" button at the top, but anyone who has custom entity displays can use the `use_field_group` settings.

mstenta (approved):

mstenta:
Rebased onto 3.x. This is ready to merge!

PR #771, opened 2024-01-16
Patch drupal/core for issue 3414883

paul121:
The datetime_timestamp widget does not use default field value. This means that the timestamp field in the log form does not have a default value and requires the user to input a value before submitting the form


mstenta:
Related issue in Log module, for reference: https://www.drupal.org/project/log/issues/3396419

mstenta:
Link to Drupal core issue: https://www.drupal.org/project/drupal/issues/3414883

PR #773, opened 2024-01-16
Fix duplicated revision tab on entities

### LINKED ISSUE #772: Revisions tab is duplicated on entities
The "Revisions" tab appears twice on asset, log, plan, and taxonomy term entities.

Screenshot from 2024-01-16 13-21-50

@paul121 and I discussed this briefly on a dev call and we suspect that it is due to that fact that we were getting Revisions UI from the Entity API module originally, and then Drupal core added their own in 10.1.x.

- https://www.drupal.org/node/2350939
- https://www.drupal.org/node/3160443

We moved the original "Revisions" tab to the secondary tabs in #634, but it appears that the Drupal core tab is still being added to the primary tabs.

So there are two issues:

1) Both Drupal core and Entity API modules are adding a "Revisions" tab.
2) The Drupal core tab is not being moved to the secondary tabs.

mstenta:
Fixes https://github.com/farmOS/farmOS/issues/772 via the method described in this comment: https://github.com/farmOS/farmOS/issues/772#issuecomment-1894305024

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - [Patch drupal/core to fix Issue #3414883: datetime_timestamp widget does not use default field value #771](https://github.com/farmOS/farmOS/pull/771)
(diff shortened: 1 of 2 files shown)


paul121 (approved):
Looks good, thanks for the great summary in the issue. I agree I think this is our best option.

PR #774, opened 2024-01-16
Override the lab_test log timestamp label and description

mstenta:
I would like to propose that we override the `timestamp` field label and description on `lab_test` logs.

From: "Timestamp" / "Timestamp of the event being logged."
To: "Date sampled" / "The date when the sample was collected."

Before:

Screenshot from 2024-01-16 15-08-26

After:

Screenshot from 2024-01-16 15-08-36

Reasoning:

This idea came up originally back when I was working on the first draft convention for soil tests, which states that:

> The "Timestamp" (`timestamp`) field of the log MUST be the date when the sample(s) were collected.

Source: https://github.com/mstenta/farmOS-conventions/blob/wip/conventions/farmOS/soil-test/1.0/index.md

It came up again recently in work that I'm doing for the [Cornell Soil Health Lab](https://soilhealthlab.cals.cornell.edu/). I decided to go ahead and make the change in their system, and then figured it was worth proposing as a change in farmOS core.

The other reason I like this is it make all three date fields on lab test logs more consistent (especially when they are displayed together):

Screenshot from 2024-01-16 15-17-17

We don't currently display them together in farmOS core, but I think that would be a worthwhile next step to consider... especially once #770 is merged.

This is achieved using a little-known mechanism that Drupal core provides, which lets you provide a single config entity to override base field labels and descriptions. There's a good writeup here: https://www.previousnext.com.au/blog/overriding-base-field-labels-and-descriptions-of-drupal-entities-without-custom-code

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [Set the minimum value of maturity_days and transplant_days to 1 #794](https://github.com/farmOS/farmOS/pull/794)
 - [Move transplant_days field to farm_transplant module #795](https://github.com/farmOS/farmOS/pull/795)
 - [Use content form for taxonomy terms #810](https://github.com/farmOS/farmOS/pull/810)
+- [Override the lab_test log timestamp label and description #774](https://github.com/farmOS/farmOS/pull/774)
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


paul121:
Hmm. I'm going to think on it some! Overall supportive of this. At minimum the description should be updated.

> This is achieved using a little-known mechanism that Drupal core provides, which lets you provide a single config entity to override base field labels and descriptions.

+1 Nice! We should implement this for the Birth log's asset field being renamed to "Child".

mstenta:
> +1 Nice! We should implement this for the Birth log's asset field being renamed to "Child".

For future readers, here is the PR for that... :-)

- #824

mstenta:
I tacked on one additional commit, which shifts the `weight`s of `date_received` and `date_processed` fields so they appear directly under the `timestamp` field. This is what it looks like:

Screenshot from 2024-04-10 16-23-04

paul121 (approved):

PR #776, opened 2024-01-16
Remove "All" from "Items per page" options in Views

mstenta:
We currently provide the option to show "All" items in our Views. On a site with a lot of records, this is a guaranteed `PHP Fatal error: Allowed memory size of ... bytes exhausted`. I've seen this in my logs on Farmier when the "All" filter is used.

As nice as it is to offer this option, I think we need to acknowledge that it's not realistic.

Aside from "All" we currently offer these options in the "Items per page" dropdown: 25, 50, 100, 250, 500.

In some cases, even 500 could cause an OOM error, but I think we can leave that for now.


paul121 (approved):

mstenta:
> In some cases, even 500 could cause an OOM error, but I think we can leave that for now.

Now I'm wondering if we should just remove the 500 option at the same time... :speak_no_evil: :thinking:

symbioquine (approved):

PR #782, opened 2024-01-26
Make "done" the default log status

mstenta:
Currently log's default to "Pending" if their status is not explicitly set. This is not an intentional decision anywhere, it is simply the default behavior of a `state` field when no default value is set on the base field definition. It defaults to whatever the "first" status is in the workflow. This makes sense from the perspective of the `state_machine` module, where workflows are typically used to enforce a step-by-step procession through states. In farmOS, however, a lot of the time you just want to record something after it has happened, so you don't need it to start in `pending` and then move to `done`. You want it to just start in `done` and no more changes are needed after that.

FWIW, the "first" status of assets is `active`, so they naturally default to `active` in the UI/API. If the first state were `archived`, we would essentially have the same issue with assets and we probably would have fixed both of these sooner.

There is also some discussion about adding client-side JavaScript logic that would automatically set the status to done/pending based on the timestamp (if timestamp > now, set it to `pending`). I think that would be a good next step after making this change. I think a default log status of `done` makes sense most of the time.

This would affect both UI form default values as well as API defaults, so logs created without an explicit status would default to `done` everywhere.

I'm proposing we do this in farmOS, and not in the Log module itself, because that module can be used outside of farmOS as well, and I don't want to make assumptions about requirements in other contexts. In farmOS I think we can safely assume that `done` is a good default status.

See forum discussions for context:

- https://farmos.discourse.group/t/pending-done-now-option-for-logs/1854
- https://farmos.discourse.group/t/log-category-defaults-setting-or-avoiding-them/1730/5


symbioquine (approved):

paul121 (commented):
> I think a default log status of done makes sense most of the time.

I agree :+1: 

> FWIW, the "first" status of assets is active, so they naturally default to active in the UI/API. If the first state were archived, we would essentially have the same issue with assets and we probably would have fixed both of these sooner.

I feel like the proper solution here is to define a new state_machine workflow that has `done` listed as the first state. The default log workflow is here: https://git.drupalcode.org/project/log/-/blob/2.x/log.workflows.yml?ref_type=heads#L5-9

It's possible for a log bundle to provide its own workflow that doesn't even use the `done` status, in which case setting the default value to `done` could cause a bug. And more generally I think using the state_machine workflow system as designed is better! It could make sense for some log types to default to different states. Eventually this could even be configurable on each farmOS instance.

paul121:
>I feel like the proper solution here is to define a new state_machine workflow that has done listed as the first state.

This is an assumption I'm making. Just checking the [state_machine docs](https://git.drupalcode.org/project/state_machine/-/blob/8.x-1.x/README.md?ref_type=heads) to see if there is a more formal concept of "default" but I don't see anything. It seems that the logic to use the first defined state as default is implemented here: https://git.drupalcode.org/project/state_machine/-/blob/8.x-1.x/src/Plugin/Field/FieldType/StateItem.php?ref_type=heads#L128-138

mstenta:
Thanks for looking deeper at `state_machine` @paul121!

> It's possible for a log bundle to provide its own workflow that doesn't even use the `done` status, in which case setting the default value to `done` could cause a bug.

That's a really good point. This PR is too heavy handed then.

> I feel like the proper solution here is to define a new `state_machine` workflow that has `done` listed as the first state.

This feels weird to me, but it may be the best way to achieve it.

It's weird that we would define a workflow that "starts" with done, because that implies that logs go from "done" to "pending". What we really want to be able to say is "most of the time logs are done when they are created. **sometimes** they are pending when they are created and then move to done". Putting "done" before "pending" feels weird because of that, IMO... but it's really just semantic hair-splitting.

Ultimately, it doesn't sound like `state_machine` has a concept of "default" state, which seems like the "proper" way to define this semantically. But oh well.

I'll convert this PR back to draft status and we can explore a custom `log` workflow for farmOS in 3.2.0.

paul121:
> Ultimately, it doesn't sound like state_machine has a concept of "default" state, which seems like the "proper" way to define this semantically. But oh well.

Yes yes, this makes a lot of sense. I think that would be a nice feature addition. cc @pcambra ? :-)

mstenta:
> we can explore a custom `log` workflow for farmOS in 3.2.0.

Maybe this would be a good time to explore adding an `abandoned` status too.

Relevant discussion: https://farmos.discourse.group/t/ability-to-create-abandoned-logs/1636

mstenta:
I've refactored this PR so that it adds a new `farm_log_workflow` via the `farm_log` module, in a new `farm_log.workflows.yml` file. This is identical to the `log_default` workflow, except `done` is before `pending`, which makes it the default.

The PR also updates all the `config/install/log.type.*.yml` files accordingly, and adds an update hook to the `farm_log` module which will update all existing log types that currently use the `log_default` workflow. Worth noting: this will also update all contrib log types, so I will also update those separately once this gets merged (eg: https://www.drupal.org/project/farm_ledger and any others I maintain).

I also updated the documentation page for adding log types so that it recommends the `farm_log_workflow` instead of `log_default`.

I'll leave this in draft status while the tests run...

paul121:
> Worth noting: this will also update all contrib log types, so I will also update those separately once this gets merged (eg: https://www.drupal.org/project/farm_ledger and any others I maintain).

This sounds problematic. Can't we make the core update hook only update core log types? Contrib modules should implement their own update hook if/when they make this change

mstenta:
> Can't we make the core update hook only update core log types? Contrib modules should implement their own update hook if/when they make this change

Sure, I can make that change.

For what it's worth, the update hook only touches log types that use the `log_default`workflow, so I don't think there's much harm.

That said, if the `farm_update` module is installed, and a contrib module is still using `log_default` in their `config/install/log.type.*.yml`, then it will be reverted to that when the cache is rebuilt. So it's sort of pointless to update all log types anyway.

Tests are also failing now... I think because I changed the workflow in a bunch of kernel test modules, which don't depend on `farm_log`. So I may change some of those back to `log_default` for simplicity. I'll force push fixes for all of these things today...

mstenta:
> Sure, I can make that change.

Done and force pushed. It only updates core log types now.

I think the only contrib module that needs an update is `farm_ledger`, so I've got some commits ready for that too. I'll push out a new release after we tag 3.2.0, so it can depend on `farmos/farmos ^3.2`.

I changed all the log types in testing modules back to `workflow: log_default` - hopefully that fixes the tests.

Alongside that I also added a new kernel test to the `farm_log` module to confirm that `done` is the default status when `farm_log_workflow` is used, and `pending` is default when `log_default` is used.

mstenta:
Made a small adjustment to fix tests: the `farm_entity` module depends on `farm_log_asset`, but not on `farm_log`. I changed it so it depends directly on `farm_log` (which already depends on `farm_log_asset`. This will ensure that all modules that provide log types (which should depend on `farm_entity`) will have the workflow available.

paul121 (approved):
> That said, if the farm_update module is installed, and a contrib module is still using log_default in their config/install/log.type.*.yml, then it will be reverted to that when the cache is rebuilt. So it's sort of pointless to update all log types anyway.

Cool, I assumed this was the case but wasn't sure. So we really only need the update hook for sites without `farm_update`, but of course should include one more generally. LGTM!

PR #785, opened 2024-01-30
Add UI for creating instances of quick forms

paul121:
See https://gitlab.com/OpenTEAMAg/feedback/support_requests/-/issues/182

This PR currently includes commits from #766 

I believe everything is working... except disabling the quick form does not disable associated actions :-( I'm not sure if there will be an easy way to do accomplish this. I tried using a config override service but can only override the action config entity's `label`, not the `status`. TIL... it seems to be due to the fact that not all of Drupal config entities actually uses the `status` key. https://www.drupal.org/project/drupal/issues/2872149

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [Announce new releases on farmOS.discourse.group #780](https://github.com/farmOS/farmOS/pull/780)
 - [Add a Product asset type and Product type taxonomy #787](https://github.com/farmOS/farmOS/pull/787)
 - [Inventory quick form #766](https://github.com/farmOS/farmOS/pull/766)
+- [Add UI for creating instances of quick forms #785](https://github.com/farmOS/farmOS/pull/785)
 
 ### Changed
 


paul121:
>  I tried using a config override service but can only override the action config entity's label, not the status

My attempt is here: https://github.com/paul121/farmOS/commit/2267854f6622f8dde1270039d770249b4d8ec30d

It gets close, I can alter the labels to be "disabled" when the corresponding config entity is disabled, but can't completely disable the action config items.

Screenshot from 2024-01-29 23-54-54

modules/core/quick/farm_quick.links.menu.yml
```diff
@@ -1,3 +1,8 @@
 farm.quick:
   class: Drupal\Core\Menu\MenuLinkDefault
   deriver: Drupal\farm_quick\Plugin\Derivative\QuickFormMenuLink
+farm.quick_setup:
+  title: Quick Forms
+  description: Quick forms make it easy to record common activities.
+  parent: farm.setup
```
mstenta: We need to add a dependency on `farm_setup` with this commit.

modules/core/quick/tests/src/Functional/QuickFormTest.php
```diff
@@ -91,8 +91,8 @@ public function testQuickForms() {
     $this->assertSession()->statusCodeEquals(200);
     $this->assertSession()->responseContains('value="100"');
 
-    // Go to the test configuration form and confirm that access is denied.
-    $this->drupalGet('quick/configurable_test/configure');
+    // Go to the test edit form and confirm that access is denied.
+    $this->drupalGet('setup/quick/configurable_test/edit');
     $this->assertSession()->statusCodeEquals(403);
```
mstenta: Looks like this test is failing because the `setup/quick/configurable_test/edit` path is returning a 404.

mstenta (changes requested):
This looks and works great @paul121! I reviewed the commits and noted two small things. Manual testing all looks good!

I think we need to sanitize the quick form label, description, and help text where they are printed to the page in a few places. I'll do some testing with this. I think I've identified all the places that need it.

As for disabling the action, maybe the only way to do it is with a `hook_form_alter()` :shrug:. I'm not aware of any internal Drupal APIs for hiding bulk actions... but maybe there's something! Looking at the form that creates the action dropdown might be a good place to start if you want to investigate that. Maybe there's some logic there about which actions are included/excluded that would provide a clue.

mstenta:
Oh one small thing I just noticed... changing the help text of a quick form required a cache clear for that change to take effect.

**Edit**: Hmm actually when I tried it a second time it did not happen... I wonder if it behaves differently the first time? I saw it happen when I first overrode the planting quick form help text.

mstenta:
> As for disabling the action, maybe the only way to do it is with a `hook_form_alter()` 🤷. I'm not aware of any internal Drupal APIs for hiding bulk actions... but maybe there's something!

The `entity_reference_integrity` module implements `hook_action_info_alter()` to modify actions... I wonder if we could leverage that?

It's weird because `hook_action_info_alter()` does not seem to be documented in Drupal 8+, and [this change notice](https://www.drupal.org/node/2020549) indicates that it was removed, but it's still working for `entity_reference_integrity` so... :shrug: Hard to say if it will work to solve our problem.

modules/core/quick/farm_quick.links.menu.yml
mstenta: I pushed a fixup commit for this: https://github.com/farmOS/farmOS/pull/785/commits/205b88502027621364abcfd85bc3368c72d14a10

mstenta:
> I think we need to sanitize the quick form label, description, and help text where they are printed to the page in a few places. I'll do some testing with this. I think I've identified all the places that need it.

I pushed a commit that sanitizes all the labels, descriptions, and help text of quick forms wherever we print them to HTML with `Html::escape()`.

In one case (the `title` passed to `admin_block_content`) I found that it was already being sanitized by downstream code. But I felt it was better to be consistent in our own code and sanitize it ourselves anyway, just in case anything changes in the future (eg: we stop using `admin_block_content` but forget to add `Html::sanitize()` to the `title`), so I wrapped that one in `Markup::create()` which works to tell the downstream code that it's already safe, so it doesn't get double-sanitized.

I tested the following strings in label, description, and help text:

- `<script>alert(123);</script>` (to make sure XSS exploits were sanitized)
- `Mike's Inventory & "Stuff"` (to make sure quotes and ampersands were not being double-escaped)

I checked the output in these places where quick form info is shown:

- `/quick`
- `/quick/planting`
- `/setup/quick`
- the main menu

All looks good and safe! :+1:

modules/core/quick/src/QuickFormListBuilder.php
```diff
@@ -120,13 +121,13 @@ public function buildRow(EntityInterface $entity) {
       '#markup' => $quick_form->getPlugin()->getLabel(),
     ];
     $row['label'] = [
-      '#markup' => $quick_form->getLabel(),
+      '#markup' => Html::escape($quick_form->getLabel()),
     ];
```
paul121: TIL I learned more about this!

But looks like we might not need `Html::escape` here and instead just use `#plain_text`. Reading https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Component%21Utility%21Html.php/function/Html%3A%3Aescape/8.2.x:

> When constructing [render arrays](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Render%21theme.api.php/group/theme_render/10) passing the output of [Html::escape](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Component%21Utility%21Html.php/function/Html%3A%3Aescape/10)() to '#markup' is not recommended. Use the '#plain_text' key instead and the renderer will autoescape the text.

paul121: I think this QuickFormListBuilder is the only place that needs `#plain_text`, all the other places changed in this commit are not explicitly render arrays so `Html::escape` LGTM

paul121:
> Oh one small thing I just noticed... changing the help text of a quick form required a cache clear for that change to take effect.
> Edit: Hmm actually when I tried it a second time it did not happen... I wonder if it behaves differently the first time? I saw it happen when I first overrode the planting quick form help text.

This might have happened if you had rendered a quick form page before checking out the changes that add the cache tag. In this case the help text would have been rendered without caching, and would require a cache rebuild for it to be rendered.

modules/core/quick/src/QuickFormListBuilder.php
mstenta: > But looks like we might not need `Html::escape` here and instead just use `#plain_text`. 

TIL about `#plain_text`! :smile:

modules/core/quick/farm_quick.links.menu.yml
paul121: Squashed

modules/core/quick/tests/src/Functional/QuickFormTest.php
paul121: Aha. This is because the quick form config entity had not yet been created in these tests. Later on it does create an config entity and do similar tests so I moved this logic there.

modules/core/quick/src/QuickFormListBuilder.php
paul121: I pushed this change

paul121:
Okay, we have passing tests.

mstenta (approved):
Everything looks great @paul121! Approved! :tada:

mstenta:
Rebased onto latest 3.x and added a changelog line. This is ready to merge into 3.1.0!

mstenta:
Merged!

Issue #653, opened 2023-03-11
Composer merge plugin dependencies are not correctly updated


paul121:
**Describe the bug**

Following the [Updating dependencies with composer](https://farmos.org/hosting/composer/#updating-dependencies) documentation produces the following bug where the `composer.project.json` dependencies included by the composer merge plugin are not correctly updated.

This is not currently a large issue because there are only two non-dev dependencies in `composer.project.json`. https://github.com/farmOS/farmOS/blob/93e13d6f8384b6026110754dc2fc7a403f82847e/composer.project.json#L1-L5

This only becomes an issue when a new release of farmOS comes with changes to either of these additional project dependencies. Unfortunately because we pin `drupal/core-composer-scaffold` we typically change this dependency with each release of farmOS, so in practice this is an issue with every release of farmOS.

**To Reproduce**
1. Pin farmOS dependency to `2.0.0-rc1` in composer.json.
2. Install project for the first time: `composer install --no-dev`. Observe that `drupal/core` and `drupal/core-composer-scaffold` are both installed and locked to `9.4.9`
3. Pin farmOS dependency to `2.0.0` in composer.json.
4. Update dependencies: `composer update --no-dev`. Observe that `drupal/core` is updated to `9.5.0` but `drupal/core-composer-scaffold` is not updated.
```shell
$ ddev composer update --no-dev
Xdebug: [Step Debug] Could not connect to debugging client. Tried: host.docker.internal:9003 (through xdebug.client_host/xdebug.client_port) :-(
Loading composer repositories with package information                                                                                                                                                                                                             Updating dependencies                                 
Lock file operations: 0 installs, 2 updates, 0 removals
  - Upgrading drupal/core (9.4.9 => 9.5.0)
  - Upgrading farmos/farmos (2.0.0-rc1 => 2.0.0)
Writing lock file
```
5. Pin farmOS dependency to `2.0.1` in composer.json.
6. Update dependencies: `composer update --no-dev`. Observe that `drupal/core` is updated to `9.5.3` but `drupal/core-composer-scaffold` is only updated to `9.5.0`.
```shell
$ ddev composer update --no-dev
Xdebug: [Step Debug] Could not connect to debugging client. Tried: host.docker.internal:9003 (through xdebug.client_host/xdebug.client_port) :-(
Loading composer repositories with package information                                                                                                                                                                                                             Updating dependencies                                 
Lock file operations: 0 installs, 3 updates, 0 removals
[...]

PR #786, opened 2024-01-30
Recommend running composer update twice #653

mstenta:
This is documentation update as an interim step towards fixing #653.


mstenta:
@paul121 recommended adding `composer update --no-dev` twice too, in addition to the note, to make it extra extra clear. I force pushed my commit with that change included.

paul121 (approved):

PR #787, opened 2024-01-30
Add a Product asset type and Product type taxonomy

mstenta:
I originally started this in a separate module (https://github.com/mstenta/farm_product), but I think we should consider adding it to farmOS core as a standard asset type.

Inspired by this forum topic today: https://farmos.discourse.group/t/asset-type-product/1864

For more discussion/context see: https://farmos.discourse.group/t/tracking-products-and-inventory-product-quantity/1613


paul121 (approved):
I'm approving given that it is inevitable we will have a product asset type.

I share some of the concerns/issues raised in the forum topic linked with more context/discussion around how effectively this can be used without more dedicated UIs for managing products and inventory. I hope the product asset leads to more solutions than conceptual issues/questions for users but alas, power to the people.

PR #788, opened 2024-01-30
Remove decreased page top margin

paul121:
We previously implemented this style change to decrease some white space that is included at the top of pages. Since then I think something has changed as now some of the page content is slightly cut-off at the very top. I think this indicates that some previous extra space has since been removed. Let's just remove this style change.

page-margin-top-before
Screenshot from 2024-01-30 15-11-58


mstenta (approved):
LGTM!

PR #794, opened 2024-02-08
Add "Days of harvest" field to Plant type terms

mstenta:
This adds a `harvest_days` field to `plant_type` taxonomy terms, alongside the existing `transplant_days` and `maturity_days`.

Along with the other two field, this will provide core data storage at the plant type level, which can then be used in other places. In particular, we would like to use this field in the [Crop planning](drupal.org/project/farm_crop_plan) module.

### DIFF
--- a/modules/taxonomy/plant_type/config/install/field.field.taxonomy_term.plant_type.harvest_days.yml
+++ b/modules/taxonomy/plant_type/config/install/field.field.taxonomy_term.plant_type.harvest_days.yml
@@ -1,0 +1,25 @@
+langcode: en
+status: true
+dependencies:
+  config:
+    - field.storage.taxonomy_term.harvest_days
+    - taxonomy.vocabulary.plant_type
+  enforced:
+    module:
+      - farm_plant_type
+id: taxonomy_term.plant_type.harvest_days
+field_name: harvest_days
+entity_type: taxonomy_term
+bundle: plant_type
+label: 'Days of harvest'
+description: ''
+required: false
+translatable: false
+default_value: {  }
+default_value_callback: ''
+settings:
+  min: 1
+  max: null
+  prefix: ''
+  suffix: ' day| days'
+field_type: integer
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [Update patch for Issue #3397275 to fix boolean field JSON Schema #819](https://github.com/farmOS/farmOS/pull/819)
 - [Do not trim whitespace from quantity field item content #820](https://github.com/farmOS/farmOS/pull/820)
 - [Do not install base modules when --existing-config is used #821](https://github.com/farmOS/farmOS/pull/821)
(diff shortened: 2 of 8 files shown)


modules/taxonomy/plant_type/config/install/field.field.taxonomy_term.plant_type.harvest_days.yml
```diff
@@ -0,0 +1,25 @@
...
+entity_type: taxonomy_term
+bundle: plant_type
+label: 'Days of harvest'
+description: ''
+required: false
+translatable: false
+default_value: {  }
+default_value_callback: ''
+settings:
+  min: null
```
wotnak: For this field, only positive integers make sense, so the min setting could be set to 0 or 1 to prevent inputting negative numbers.

mstenta: That's a good idea @wotnak.

If we do that to this field we should also do that to `transplant_days` and `maturity_days` too IMO. The question there is: what do we do if there is already "invalid" data in the database (values <= 1)?

Out of principle, I don't think we should touch any of it, but just adding a `min` would cause a validation error if someone tries to edit a term with invalid data in the future. Maybe that's the best we can do.

mstenta: Oh I just remembered... these fields are already set as `unsiged: true` in their field storage definitions, which means negative values can't be stored. So maybe we don't need to do anything?

wotnak: `usigned: true` prevents passing negative numbers but allows 0. If `min` setting would be set to `1`, 0 would be no longer allowed and frontend validation would be added to the field.
image

mstenta: Makes sense, thanks @wotnak.

Should we apply the same `min` to `transplant_days` and `maturity_days`? Is it OK to leave potential zero values in the database? Or maybe we should update any `0` values to `1`?

wotnak: Yes, I think these fields should all have the same validation logic.
If existing 0 values would be updated to 1 (or maybe cleared? 0 days basically means the field don't have a value?), then any code using these values could skip handling 0 and only check if the field has value and be sure it will always be a positive integer.

mstenta: OK! I force pushed with an additional commit that changes `min` to `1` in the `maturity_days` and `transplant_days` fields, with an update hook to make that change and also to delete `0` values from the database. Tested locally and it all seems to work as intended.

wotnak (approved):

mstenta:
Rebased this onto #806, to leverage the benefits of that PR.

I also removed the commit that adds a dependency on the `image` module to `farm_plant_type`. I would like to instead make another PR that adds `file` and `image` fields to *all* taxonomy terms (see #536). This would remove the need to declare that dependency on this module specifically.

Other than that, this PR is functionally unchanged.

paul121 (approved):

PR #797, opened 2024-02-09
Merge multiple functional tests together to reduce test startup time

mstenta:
I was looking into ways of improving the performance of our functional tests, and found this suggestion by @Berdir:

> Another thing to consider is to avoid having many test methods. Each runs in a completely new, isolated environment. A trick that is often used is to have a single testAll() method and many doTestSomething() methods, then they are faster by almost factor count(test methods), since set up is usually by far the slowest part.

https://drupal.stackexchange.com/a/213561

This PR tests out this suggestion. It seems to shave off about 30 seconds from our tests overall, based on the few times that I've run them so far.

### DIFF
--- a/modules/asset/sensor/tests/src/Functional/SensorDataApiTest.php
+++ b/modules/asset/sensor/tests/src/Functional/SensorDataApiTest.php
@@ -44,9 +44,17 @@ class SensorDataApiTest extends FarmBrowserTestBase {
   }
 
   /**
+   * Run all tests.
+   */
+  public function testAll() {
+    $this->doTestApiGet();
+    $this->doTestApiPost();
+  }
+
+  /**
    * Test API GET requests.
    */
-  public function testApiGet() {
+  public function doTestApiGet() {
 
     // Build the path.
     $uri = $this->buildPath($this->asset);
@@ -82,7 +90,7 @@ class SensorDataApiTest extends FarmBrowserTestBase {
   /**
    * Test API POST requests.
    */
-  public function testApiPost() {
+  public function doTestApiPost() {
 
     // Build the path.
     $uri = $this->buildPath($this->asset);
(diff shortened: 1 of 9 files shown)


Berdir:
I have been summoned  ;)

FWIW, consider whether this is actually worth. It also reduces test isolation, which might cause weird bugs in tests and makes them harder/slower to run when debugging a specific test, you need to comment out things or so or you also always need to run a bulk of tests.

mstenta:
Thank you for your input and insight @Berdir, and for taking the time to offer it here! :smile: :pray: 

These are great considerations. I agree we should weigh the costs of doing something like this. I was curious how big a difference it would make. In the process I also discovered some tests that clearly did not need to be separate, so it was a good exercise. Perhaps that would be a good next step: look at the tests that were modified and decide on a case by case basis if there is value in keeping them separate.

mstenta:
Looked at this together on the dev call yesterday. We might wait and see if this has any more significant impact after #864 is merged.

In the meantime I rebased onto latest 3.x and force-pushed so we can get a more updated benchmark.

mstenta:
It looks like the results are a bit more significant than they were in the past. Here are the past three *scheduled* test runs where PostgreSQL tests passed (without these changes):

- 362s (https://github.com/farmOS/farmOS/actions/runs/10769337570)
- 333s (https://github.com/farmOS/farmOS/actions/runs/10788558086)
- 352s (https://github.com/farmOS/farmOS/actions/runs/10826936842)

Average: 349s

I ran the tests on this pull request three times right now, and these were the results:

- 5m 9s (https://github.com/farmOS/farmOS/actions/runs/10737245165/attempts/1?pr=797)
- 5m 5s (https://github.com/farmOS/farmOS/actions/runs/10737245165/attempts/2?pr=797)
- 4m 57s (https://github.com/farmOS/farmOS/actions/runs/10737245165?pr=797)

Average: ~304s

So with these changes we save an average of 45 seconds, based on these tests.

mstenta:
I'm going to open this for review, because it does feel worthwhile to me.

paul121 (approved):
Sounds good to me.

PR #798, opened 2024-02-12

mstenta:
Draft PR for https://www.drupal.org/project/farm/issues/3352183 - see that issue for more details.

This started as PR #668 against the 2.x branch. I've closed that and reopened this new PR against the 3.x branch.

This also includes an additional `farm_comment_plan` module, which enables commenting on plans.


paul121 (commented):
Doing some initial testing and this is working well!

One issue is that the comment "preview" button is not working for me... the page loads indefinitely and runs out of memory. I'm not seeing any other error messages. Maybe we just disable this button for now? I think the preview behavior might be an option on the comment field (disabled/optional/required)

I've also taken a stab at styling the comments and comment thread a bit more. I've come up with two options, the first with a simple border around the comment, and the second with a single left border bar, as inspired by this post [Styling Comment Threads](https://css-tricks.com/styling-comment-threads/). I think I am starting to prefer the second option. Available in my 3.x-comment branch here: https://github.com/paul121/farmOS/commit/ec254eba75e3846fc2b8509ae057d78e0c54dde2

I'm curious if we could also improve the Edit/Reply links so they appear more like small buttons. I'm not sure if it would be better to do this in CSS or alter the links to have classes eg like `button button--small button--primary`. I think adding these classes is better but not sure where is best to do so..

Otherwise, I'm curious about trying to "condense" some this spacing in the "Comments" wrapper section layer. This would be similar to how we did this for the views tables. I'm curious if there is a way we could standardize this and make it easier to adopt a more compact section layer.

comments-grouped

comment-bars

paul121:
Oh also @mstenta, I'm curious if we could manage view/form displays without including the `default` config. This would make it much easier for other modules to add fields to comments.

I think we could basically do the other approach I've shown you. Rather and doing it within each asset/log/plan comment sub-module I wonder if we could implement this "once" in the `farm_comment` module since all our entity types will have the same comment fields/behavior by default.

mstenta:
> One issue is that the comment "preview" button is not working for me... the page loads indefinitely and runs out of memory.

Huh. I'm not experiencing that. :thinking: 

> Maybe we just disable this button for now? I think the preview behavior might be an option on the comment field (disabled/optional/required)

But I'm not opposed to disabling it. I don't think it adds much value, personally. I'll look into this...

> I've also taken a stab at styling the comments and comment thread a bit more. I've come up with two options, the first with a simple border around the comment, and the second with a single left border bar

I love this! And agree the second one is my preference too.

> I'm curious if we could also improve the Edit/Reply links so they appear more like small buttons. I'm not sure if it would be better to do this in CSS or alter the links to have classes eg like `button button--small button--primary`. I think adding these classes is better but not sure where is best to do so..

Yea that might be nice... I don't mind the default links either, so it doesn't need to block this IMO. But if you want to take a swing at it @paul121 go for it!

> Otherwise, I'm curious about trying to "condense" some this spacing in the "Comments" wrapper section layer.

Likewise for this... I don't mind the default spacing, but I wouldn't argue if you want to take a pass at it before we merge this. :-)

> Oh also @mstenta, I'm curious if we could manage view/form displays without including the `default` config. This would make it much easier for other modules to add fields to comments.

Yea! I'll work on that and push a commit...

mstenta:
> I love this! And agree the second one is my preference too.

Actually just realized that what I'm seeing looks like a combination of both the options you described. But I like it!

Screenshot from 2024-03-19 10-53-36

mstenta:
Ah man, I accidentally ran `git push mstenta :3.x-comment` :facepalm:

mstenta:
Opened a new PR: #804

PR #804, opened 2024-03-19

mstenta:
Reopening this as a new pull request because I accidentally deleted the branch from the original PR... :facepalm:

- #798

See all the comments in #798 for more details...


modules/core/ui/theme/css/comment.css
```diff
@@ -0,0 +1,30 @@
...
+}
+/* Alter border style when comment is rendered on entity page. */
+/* TODO: choose a border style. */
+section.comment-wrapper article.comment {
+  /* Wrap comments in border. */
+  border: 1px solid var(--gin-border-color);
+  border-radius: var(--gin-border-m);
+  /* OR add border as left bar. */
+  border-left: 5px solid var(--gin-border-color-secondary);
+}
```
paul121: > Actually just realized that what I'm seeing looks like a combination of both the options you described. But I like it!



> > I love this! And agree the second one is my preference too.
> 
> Actually just realized that what I'm seeing looks like a combination of both the options you described. But I like it!
> 
> Screenshot from 2024-03-19 10-53-36

Ha... so actually I had a TODO to choose one or the other. I included both approaches in my  WIP commit for simplicity.

I think I still prefer the second option with only the left border. It's a bit simpler but serves the purpose of highlighting each comment in the threaded hierarchy quite well.

Wrapping each comment with a full border seems a bit unnecessary and accentuates the random whitespace to the right of the comment. It might make more sense if we could wrap an entire thread in a border and make the thread collapsible. But with the default structure of the HTML this would be hard to accomplish.

mstenta: > Ha... so actually I had a TODO to choose one or the other. I included both approaches in my WIP commit for simplicity.

That explains it. :-)

I'm OK either way. Do you want to just amend your commit to use the second approach and force push?

modules/core/comment/farm_comment.module
```diff
@@ -69,6 +71,49 @@ function farm_comment_base_field_definition(string $entity_type) {
   return $field;
 }
 
+/**
+ * Implements hook_entity_form_display_alter().
+ */
+function farm_comment_entity_form_display_alter(EntityFormDisplayInterface $form_display, array $context) {
+  if ($form_display->getTargetEntityTypeId() == 'comment' && $form_display->getMode() == 'default' && $form_display->isNew()) {
```
paul121: I think we should only alter form/view displays for the `asset`, `log` and `plan` comment bundles we provide. I'll work on a change for this

modules/core/ui/theme/css/comment.css
paul121: Done. I also decreased the space a bit around the comments and cleaned up how the comment renders when on the "reply" page.
comment-spacing-after

modules/core/comment/farm_comment.module
paul121: Done

paul121: Oh, we also need to alter the `full` view mode which is used when viewing comments on the "Reply" page.

paul121 (approved):

mstenta:
Awesome thanks @paul121! LGTM!

PR #806, opened 2024-03-20
Set plant_type fields form/view display settings via alter hook instead of config

mstenta:
This changes the way we configure the entity form/view display settings for fields that are added to `plant_type` taxonomy terms. Previously, farmOS core provided two configuration entities via `config/install` YML files:

- `plant_type/config/install/core.entity_form_display.taxonomy_term.plant_type.default.yml`
- `plant_type/config/install/core.entity_view_display.taxonomy_term.plant_type.default.yml`

The problem with providing these config entities in our module is that it makes it difficult for downstream code to add more fields to `plant_type` terms. It isn't easy for them to modify those configuration entities. And if the `farm_update` module is enabled, any changes that are made through the UI are automatically reverted when the cache is rebuilt.

This PR removes those two configuration entity YML files from the `farm_plant_type` module's `config/install` directory, and instead uses `hook_entity_form_display_alter()` and `hook_entity_view_display_alter()` to make the additional fields visible in the entity form/view displays. (Credit to @paul121 for proposing this alternative approach.)

This makes it a lot easier for other module to add fields to `plant_type` terms. We can leverage this to simplify the other PRs I opened recently:

- #794
- #795


paul121 (approved):
LGTM!

PR #807, opened 2024-03-20
Add file and image base fields to all taxonomy terms

mstenta:
This adds `file` and `image` base fields to all taxonomy terms.

This PR supersedes #536, and resolves all the issues @paul121 mentioned there.

It also closes [Issue #3244880: Add file/image fields to Material Type terms](https://www.drupal.org/project/farm/issues/3244880)

Note: this is built on top of the #806 PR branch.


modules/taxonomy/plant_type/farm_plant_type.module
```diff
@@ -0,0 +1,120 @@
...
+    ]);
+    $form_display->setComponent('image', [
+      'type' => 'image_image',
+      'settings' => [
+        'progress_indicator' => 'throbber',
+        'preview_image_style' => 'thumbnail',
+      ],
+      'region' => 'content',
+      'weight' => 50,
+    ]);
```
paul121: With this PR adding `image` as a base field we shouldn't need to set the `image` component for `plant_type` terms, yes?

modules/taxonomy/plant_type/farm_plant_type.module
```diff
@@ -0,0 +1,120 @@
...
+    $display->setComponent('image', [
+      'type' => 'image',
+      'label' => 'above',
+      'settings' => [
+        'image_style' => '',
+        'image_link' => '',
+      ],
+      'region' => 'content',
+      'weight' => 50,
+    ]);
```
paul121: Ditto

paul121 (changes requested):
This is awesome

modules/taxonomy/plant_type/farm_plant_type.module
paul121: I think this will fix a minor issue where the ontology field is rendered between images+files, only seeing this on `plant_type` terms

Screenshot from 2024-03-20 18-02-31

mstenta: Oh good catch @paul121! I meant to remove that. :-)

mstenta: Force pushed.

paul121 (approved):

PR #808, opened 2024-03-20
Issue #3390486: Add an External URI field to all taxonomy terms

mstenta:
See issue: [Issue #3390486: Add an Ontology URI field to all taxonomy terms](https://www.drupal.org/project/farm/issues/3390486)

This builds on top of PRs #806 and #807


mstenta:
Oops accidentally pushed an old version of the branch. Force pushed - it should be correct now. :-)

paul121 (changes requested):
Not trying to be overly pedantic but want to clarify `URI` vs `URL`, where `URL` is a subset of `URI`: https://en.wikipedia.org/wiki/Uniform_Resource_Identifier

> A Uniform Resource Identifier (URI) is a unique sequence of characters that identifies an abstract or physical resource,[[1]](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#cite_note-FOOTNOTEBerners-Lee,_Tim;_Fielding,_Roy_T.;_Masinter,_Larry20051-1) such as resources on a webpage, mail address, phone number,[[2]](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#cite_note-FOOTNOTEBerners-Lee,_Tim;_Fielding,_Roy_T.;_Masinter,_Larry20057-2) books, real-world objects such as people and places, concepts.

> URIs which provide a means of locating and retrieving information resources on a network (either on the Internet or on another private network, such as a computer filesystem or an [Intranet](https://en.wikipedia.org/wiki/Intranet)) are [Uniform Resource Locators](https://en.wikipedia.org/wiki/Uniform_Resource_Locator) (URLs).

This seems to be adding full `URL` validation for our `ontology_uri` field. If this is our desired behavior I would suggest we rename the field `ontology_url`.

But I think there may actually be value in allowing `URI` more broadly. As an example, given a Crop Ontology there may exist an identifier for "carrot". Within plant_type terms it would be simplest to save this reference as the identifier `crop-ontology:carrot` rather than a full URL. While we couldn't render this identifier as a link it would allow separate modules that know about `crop-ontology` to implement one of potentially multiple methods of locating and retrieving this resource as necessary.

Screenshot from 2024-03-20 16-47-59


All this said.... unfortunately it seems that the Drupal core `url` form element (used by the `UriWidget`) always enforces a complete `URL` as documented in this open issue: https://www.drupal.org/project/drupal/issues/3186651

> Consider however the url form element calls UrlHelper::isValid() for validation, which checks only a subset of RFC 3986. (It actually also broadens the validation to restirct to only http/https/feed.)

Perhaps the easiest solution would be to make this a `string` field with some light regex validation for `URI`s?

paul121:
This just brings up more thoughts concerning "how this field should be used". In some ways it is nice to have a generic purpose field to store things as needed but it also makes this much harder/unreliable to use this field in a programmatic way (eg: maybe you cannot expect this to always be a full URL/URI that will successfully resolve as a link).

It also makes me wonder about "ontology" vs "vocabulary". Doing some searching online and I see many quotes to this w3 link though it seems to be broken now: http://www.w3.org/standards/semanticweb/ontology

> There is no clear division between what is referred to as “vocabularies” and “ontologies”. The trend is to use the word “ontology” for more complex, and possibly quite formal collection of terms, whereas “vocabulary” is used when such strict formalism is not necessarily used or only in a very loose sense. Vocabularies are the basic building blocks > for inference techniques on the Semantic Web.

With this definition I feel like a "Vocabulary URI" is maybe more correct. It seems more accurate to say "all ontologies are (or have) vocabularies" rather than "all vocabularies are ontologies" - thus with this field we're always referencing an external vocabulary identifier. I know we already have "vocabularies" in Drupal (of the loosest sense) but I don't think that should prevent us from using the "Vocabulary" language again for this field, we're just linking our simple vocabulary to something else slightly more formal.

This would save the concept of "ontologies" for future work/implementations within farmOS. It might make sense to have a distinct "configuration" or "registry" of more complex ontologies that can be integrated into farmOS which use a more strict system fields/etc for linking terms to the ontology vocabularies.

paul121:
Oh also, as implemented this is a base field added to all taxonomy terms. I kinda thought it might be an optional feature via a separate module. Although making it a separate module is really just a simple fix for the larger issue I'm groking: **I worry about this change in combination with image/file fields "exploding" the UX complexity of taxonomy terms.**

Taxonomy terms are already one of the more challenging and confusing things for users and Ontology/Vocabulary is a particularly confusing concept on top of that. Maybe we should adopt the Gin content form for taxonomy terms and add a `Vocabulary/Ontology`, `Files` tab.. move `Relationships` and `Published` to the sidebar.. it wouldn't be so bad. Hides some of the complexity and brings some general form improvements, too?

As an example of these issues, adding the "Unit" term you are presented with Ontology, Images and File fields, making the form much much longer than before:

Screenshot from 2024-03-20 17-44-20

mstenta:
> Not trying to be overly pedantic but want to clarify `URI` vs `URL`, where `URL` is a subset of `URI`: https://en.wikipedia.org/wiki/Uniform_Resource_Identifier

I greatly appreciate your pedantry in this case @paul121! :smile:

> I think there may actually be value in allowing `URI` more broadly.

Agreed. We want `URI`s for referencing ontology items, not `URL`s.

The original issue I created on drupal.org proposed `URL` but I changed that to `URI` yesterday: https://www.drupal.org/node/3390486/revisions/view/13289988/13493233

> This seems to be adding full `URL` validation for our `ontology_uri` field.

And this is why your reviews are so astute @paul121 - you always find the shortcuts I take! :laughing:

I wanted to ensure we had constraint validation in this first pass (so we didn't end up with invalid URIs in the database), but was disappointed to find that neither Drupal nor Symfony included a `URI` constraint validator. Symfony does, however, have a `URL` constraint validator: https://github.com/symfony/symfony/blob/7.0/src/Symfony/Component/Validator/Constraints/UrlValidator.php

So, I went with that... knowing that it wasn't *exactly* what we wanted... but at least would be a step in the right direction, and wouldn't limit us from expanding to `URI` validation (as `URL`s are just more strict `URI`s).

> Perhaps the easiest solution would be to make this a string field with some light regex validation for `URI`s?

Perhaps... and I think your next thought is where my mind goes next too... (I'll respond in a separate comment to keep the topics roughly separated).

mstenta:
> This just brings up more thoughts concerning "how this field should be used". In some ways it is nice to have a generic purpose field to store things as needed but it also makes this much harder/unreliable to use this field in a programmatic way (eg: maybe you cannot expect this to always be a full URL/URI that will successfully resolve as a link).

Yea agreed. I think most of the use-cases that led to the creation of this feature request were for "links to external ontology URLs" so being able to render them as a clickable link is important.

But not all `URI`s are `URL`s, so not all of them would be "clickable", so it brings us back to the "should this be a `URL`?" question perhaps.

I wonder if regex can be used to distinguish a `URL` from a `URI`? We could conditionally render links perhaps... :thinking: 

> It also makes me wonder about "ontology" vs "vocabulary".
> With this definition I feel like a "Vocabulary URI" is maybe more correct.

I was worried that the field naming would spin out... but it's good. :sweat_smile:

I don't have very strong opinions on what we call it... I generally try to balance brevity (keeping the field name itself short) with intuition (ensuring the purpose of it is clear to the user).

I went back and forth myself on whether or not we should call it `ontology_item_uri` - because technically it would be referencing an *item* in an ontology, not an ontology itself. But I think the context (this is a field on an individual Drupal vocabulary term) makes that somewhat intuitive, and therefore unnecessary, so simply calling it `ontology_uri` is OK. (IMO :shrug:) But that's a separate point...

Considering whether to call it "ontology" or "vocabulary" is both a) interesting (because I love picking these things apart and being as accurate as possible), and b) frightening (because it feels like we could end up in an endless bikeshedding vortex :cyclone:).

> we already have "vocabularies" in Drupal (of the loosest sense)

This was my first concern as well.

> The trend is to use the word “ontology” for more complex, and possibly quite formal collection of terms, whereas “vocabulary” is used when such strict formalism is not necessarily used or only in a very loose sense.

My interpretation of this makes me think "ontology" is what we want, no?

The purpose of this field, in my mind, is to make our "vocabulary terms" more strict, by linking them to a *specific* item in an external ontology.

I think you made the same point:

> we're just linking our simple vocabulary to something else slightly more formal.

For example (the user story mentioned in the d.o issue):

> I grow 5 crops, and I have one "Plant type" term for each of them. I would like those terms to link to the crops in the USDA Plants Database: https://plants.usda.gov/home

Doesn't all of this suggest "ontology" is a better choice for this field name than "vocabulary"? :thinking: 

> With this definition I feel like a "Vocabulary URI" is maybe more correct. It seems more accurate to say "all ontologies are (or have) vocabularies" rather than "all vocabularies are ontologies" - thus with this field we're always referencing an external vocabulary identifier.

I'm still wrapping my head around this point.

But maybe all of this circles back to the first question... maybe we should actually call this `URL` instead of `URI`? So that it requires the value to be a location, not just a reference.

I don't know... this is an important decision... but I feel like I'm coming up against the limits of how finely I can split a hair. And I'm sure we wouldn't be alone in this. The first sentence of your quote, after all, was:

> There is no clear division between what is referred to as “vocabularies” and “ontologies”.

mstenta:
Let's phone a friend...

Help us @julietnpn, you're our only hope! :laughing:

mstenta:
> Maybe we should adopt the Gin content form for taxonomy terms and add a `Vocabulary/Ontology`, `Files` tab.. move `Relationships` and `Published` to the sidebar.. it wouldn't be so bad. Hides some of the complexity and brings some general form improvements, too?

:100: I had the same thought. That's a separate commit/PR in my mind, but if you want to take a pass at that right now so it can all get merged together please do!

mstenta:
@aislinnpearson, @symbioquine, @Fosten, @gbathree, @paul121, and I spent some time talking about this on the dev call today. I'll try to summarize some of the points, but feel free to correct/amend:

- The purpose of the field is to reference the *definition* of an item.
- It should support URIs (not just URLs).
- The name of the field is where there's the most back-and-forth. Ideas that came up included: `ontology_uri`, `vocabulary_uri`, `definition_uri`, `external_uri`, as well as all variations of that with `_item_` in the middle (eg: `ontology_item_uri`), to avoid any potential confusion that it's a link to an ontology as a whole.
- "Ontology" is not a perfect name for it, but it might be the "least ambiguous".
  - "Users will either immediately understand what it's for, or they won't care." (acceptable) vs "Users won't understand what it's for, or they won't care." (bad)
- The field description is important to clarify any confusion, and the current description does a good job of that: "Link this term to one or more external ontology item URIs."

mstenta:
After some more discussion in chat, @paul121 and I are starting to lean more towards naming this `external_uri` (and adding proper URI validation logic to fix the issue discovered above). The field description could be:

> Link this term to one or more external URLs or ontology item URIs.

`external_uri` seems like a nice balance of being intuitive, supporting the cases that have come up, and not being overly/inaccurately specific (which "ontology" runs the risk of).

Does that seem like a good compromise?

aislinnpearson:
Hi @mstenta, @paul121,

> The name of the field is where there's the most back-and-forth. Ideas that came up included: ontology_uri, vocabulary_uri, definition_uri, external_uri, as well as all variations of that with _item_ in the middle (eg: ontology_item_uri), to avoid any potential confusion that it's a link to an ontology as a whole.

I was hoping this might provide some idea of structure but I am not sure if it does: https://isa-specs.readthedocs.io/en/latest/isamodel.html#ontology-annotation (and here for the JSON format: https://isa-specs.readthedocs.io/en/latest/isajson.html)

> external_uri seems like a nice balance of being intuitive, supporting the cases that have come up, and not being overly/inaccurately specific (which "ontology" runs the risk of).

I like this because it is less specific, and so more flexible - it could be an ontology term, a publication DOI, an IGSN number ([International Generic Sample Number Service (IGSN) | ARDC](https://ardc.edu.au/services/ardc-identifier-services/international-geo-sample-number-service/#:~:text=An%20International%20Generic%20Sample%20Number,persistent%20identifier%20for%20physical%20samples.))

My only comment is that you'd need to be very clear about what this was in the documentation and the description - if we were a bit confused about different between a url and a uri, it makes sense people who are even less conversant in these types of issues will be as well (but reading what you have written above, you are obviously already aware of this).

The data field will get used and abused a little bit but that flexibility is what we love about FarmOS - things might not go in perfectly, but they do go in! For example, if I wanted to tag Samurai (a herbicide containing glyphosate) I'd put in this ontology term: http://purl.obolibrary.org/obo/CHEBI_24527. Notice how the link breaks though because OBO isn't working at the moment, which I assume in a URI (is there any way to check this?) - here is the URL: https://www.ebi.ac.uk/ols4/ontologies/agro/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FCHEBI_24527

However, I might also link to this which might not be a URI but describes the legislation around that product in the UK (again I suspect this is a URL): 
https://secure.pesticides.gov.uk/pestreg/getfullproduct.asp?productid=28771&pageno=1&origin=prodsearch

On top of that, we would continue to use the MAPP number in the name, so we can cross reference even if the HSE link above breaks:

image

This flexibility is important to us as users, which is where I think a validation layer might become more useful later on.

In my head, we have always needed a 'taxonomy' of ontology terms which can be applied to any entity (including another taxonomy term) but I can see that is quite complicated. So I could tag a tractor as a tractor based on the list in my system, but also tag the term 'wheat' in my plant taxonomy.

I agree this is complicated. To be honest, I'm glad it's not just me who struggles with it :)

mstenta:
Pushed a `fixup` commit that changes the name and description of the field to `external_uri` + "Link this term to one or more external URLs or ontology item URIs."

mstenta:
I think the last remaining todo here is to fix URI validation, both at the widget level (for UI use) and the entity field constraint level (for API use).

mstenta:
OK I moved some stuff around and squashed my commits, so this is ready for final review I think.

The first commit adds general support for a `uri` field in the `farm_field.factory` service, along with a custom `uri` field widget and formatter to work around issues with Drupal core's confusion of URL and URI. It also adds a constraint validator that uses the same regex as Brad's in the core Drupal issue (https://www.drupal.org/project/drupal/issues/3186651). The second commit is where the `external_uri` field is added, so it's pretty simple at that point.

paul121 (approved):

PR #810, opened 2024-03-23
Use content form for taxonomy terms

paul121:
With upcoming changes/additions to taxonomy terms #807 #808 it seems it may be time we use the content form for taxonomy terms so keep the editing forms simple and also standardize with our other editing forms. This PR will add content forms for taxonomy terms and make a few other general improvements to content forms.

Adding this for taxonomy terms is not quite as straight forward though... will comment with some questions for us to figure out. Here is a screenshot of the basic changes this makes:

Screenshot from 2024-03-22 18-27-05

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [Make "done" the default log status #782](https://github.com/farmOS/farmOS/pull/782)
 - [Set the minimum value of maturity_days and transplant_days to 1 #794](https://github.com/farmOS/farmOS/pull/794)
 - [Move transplant_days field to farm_transplant module #795](https://github.com/farmOS/farmOS/pull/795)
+- [Use content form for taxonomy terms #810](https://github.com/farmOS/farmOS/pull/810)
 
 ### Fixed
 
--- a/modules/core/ui/theme/farm_ui_theme.module
+++ b/modules/core/ui/theme/farm_ui_theme.module
@@ -11,6 +11,7 @@ use Drupal\Core\Form\FormStateInterface;
 use Drupal\farm_ui_theme\Form\AssetForm;
 use Drupal\farm_ui_theme\Form\LogForm;
 use Drupal\farm_ui_theme\Form\PlanForm;
+use Drupal\farm_ui_theme\Form\TaxonomyTermForm;
 
 /**
  * Implements hook_theme().
@@ -69,7 +70,7 @@ function farm_ui_theme_theme_suggestions_menu_local_tasks(array $variables) {
 function farm_ui_theme_entity_form_display_alter(EntityFormDisplayInterface $form_display, array $context) {
 
   // Only alter farm entity types.
-  $entity_types = ['asset', 'log', 'plan'];
+  $entity_types = ['asset', 'log', 'plan', 'taxonomy_term'];
   if (!in_array($context['entity_type'], $entity_types)) {
     return;
   }
@@ -127,6 +128,9 @@ function farm_ui_theme_farm_ui_theme_field_group_items(string $entity_type, stri
     case 'plan':
       break;
 
+    case 'taxonomy_term':
+      break;
+
     default:
       $fields = [];
   }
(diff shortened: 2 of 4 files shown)


paul121:
So one issue here is that by using the content form we do not get the "Relations" field group for the `term.parent` and `term.weight` field options. I was hoping the "Relations" would be a simple add into the sidebar . This happens because the relations are custom logic coded into the core `TermForm` form that we lose this when creating our own term form class.

The simple option would be to basically copy/paste the Relations logic from Drupal core. But this also gives us the opportunity to improve the `term.parent` widget in particular. Any ideas?

Screenshot from 2024-03-22 18-29-01

paul121:
Separately, I've discovered that our "Revision Information" sidebar group isn't entirely the true revision information... we're actually showing the original `author`, `created` timestamp, and last `changed` timestamp. None of this actually comes from the "revision" data on entities.

Instead of the `changed` timestamp we could use the latest revision timestamp and latest revision user (important because this could be different from the original author). Combined with original created/author this could be summarized to the following:

> Revision Information:
> - Created 2024-01-01 by Jane
> - Last saved 2024-03-03 by Bob

I became aware of this because taxonomy terms don't have `author` or `created` timestamps, but do have  a `changed` timestamp and additional revision information. But it's a little messy because there is currently no way to specify or view taxonomy revision information via the UI... I'm not sure if we want to open this up right now :sweat_smile:

SO right now with this PR taxonomy terms would only have `Last saved` revision information in the sidebar

paul121:
Also adding a small change to only open the first (default) and sidebar tabs by default. This should make navigating easier on mobile, thanks to @ Pat on the forum! https://farmos.discourse.group/t/farmos-3-1-1-has-been-released/1871/2?u=paul121

mstenta:
> The simple option would be to basically copy/paste the Relations logic from Drupal core. But this also gives us the opportunity to improve the `term.parent` widget in particular. Any ideas?

Just to capture something @paul121 and I discussed in chat on this point in particular...

> @paul121: Realizing that we basically need to replace the core `TermForm`, but that core form is what provides the `Relations` logic for building the `term.parent` form field... so basically we would need to copy/paste that, I think?
> @paul121: But it's also an opportunity to reconsider how that relations + parent thing works. I do feel like we need to provide something, hard to completely remove it
> @mstenta: Yeaaa
> @mstenta: Maybe we *should* just hide/remove the parent field
> @mstenta: It allows multiple parents 👎👎
> @mstenta: The hierarchy drag and drop (although not perfect) works
> @paul121: But if we remove then existing data might be stuck
> @paul121: :-/
> @mstenta: Ah you mean users wouldn't be able to undo multi parent?
> @paul121: Yeah
> @mstenta: True

I would still like to consider hiding/removing the `parent` + `weight` fields from the form. It's true that this could create a new issue for users who have terms with multiple parents, but perhaps we can address that separately. I would hope that this would be very rare... but I agree we don't want to leave those users high and dry.

Assuming for a moment we *can* come up with a solution to that problem, I think there are some good reasons for hiding/removing these fields from the term form.

For one, it would help to *prevent* users from assigning multiple parents to terms. IMO this is one of the worst flaws in Drupal core's taxonomy module. I recall working with a client on Drupal 6 who used multiple term parents and it actually broke the core hierarchy drag and drop form (they also had a LOT of terms). I'm not sure if that core bug has been fixed since then, but the bottom line for us is: multi-parent term hierarchy creates more potential problems than advantages. I can't think of any reason to use it which can't be achieved in a different way. Perhaps this needs a separate/broader discussion. This issue raised the question, though, so those are my feelings. :laughing:

Secondly, these `parent` and `weight` fields are not intuitive. On many many occasions, when I am walking someone through farmOS UI they ask what they are for, and I explain it but tell them to use the hierarchy drag and drop instead. Editing hierarchy on a term-by-term basis just isn't that useful or intuitive in practice, IMO. It's a "nice to have" at best. We have the option to remove it and simplify UX here.

Perhaps we could add a link to the hierarchy drag-and-drop form in their place, with some more descriptive text to help guide the user?

mstenta:
> Combined with original created/author this could be summarized to the following:

I like this! :+1: 

> SO right now with this PR taxonomy terms would only have Last saved revision information in the sidebar

That's fine IMO.

paul121:
Still dependent on #808 but ready for review.

I made a "Relations" tab in the sidebar and copied/replicated all logic from the core form.

I also moved the `external_uri` field into a "Reference" tab. But maybe we could combine this field with the "Relations" tab, too? This could live in the sidebar or the main tab, I'm not sure what would be better. In some ways I think we should "hide" both of these, basically so the tab isn't open by default, which would mean putting them in the `main` region.

Screenshot from 2024-03-28 17-25-47

mstenta:
Reviewed, tested, and added a `CHANGELOG.md` line. This is ready to merge!

mstenta (approved):

PR #814, opened 2024-03-27
Fix Client error when passing asset parameter to forms

### LINKED ISSUE #813: Client error using the "Add log" bulk action
It seems that the "Add log" bulk action is broken. To reproduce, go to `/assets`, select an asset, select "Add log" from the actions dropdown, and then select "Observation". After submitting, you will see "Client error / A client error happened":

Originally reported here (with screenshot): https://github.com/Rothamsted-Ecoinformatics/farm_rothamsted/issues/644

mstenta:
Fixes #813

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - [Set plant_type fields form/view display settings via alter hook instead of config #806](https://github.com/farmOS/farmOS/pull/806)
 
+### Fixed
+
+- [Fix Client error when passing asset parameter to forms #814](https://github.com/farmOS/farmOS/pull/814)
+
 ## [3.1.2] 2024-02-26
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


paul121 (approved):
Interesting. This error does not happen if you use a query param like `&asset=25` as when using the "Add observation log" action button from a single asset page.

But this error does happen when appending brackets to potentially indicate multiple asset IDs `&asset[]=25`. I hadn't heard of this deprecation, just wonder where else this may apply...

mstenta:
> query param like `&asset=25` as when using the "Add observation log" action button from a single asset page.

Oh good point @paul121.

I wonder if my fix breaks that case?

mstenta:
> I wonder if my fix breaks that case?

Dang. Yes it does. Now I get "Client error" in that context.

mstenta:
OK, I reworked the logic so that it works with both single value `asset` parameters as well as arrays.

It's a bit annoying because `$query->all('asset')` errors when it isn't an array and `$query->get('asset')` errors when it is an array. But `$query->all()` gives me a keyed array of ALL parameters, which I can use to get `asset` in both cases. 🙄

mstenta:
This also needs to be fixed here: https://github.com/farmOS/farmOS/blob/3.x/modules/core/quick/src/Traits/QuickPrepopulateTrait.php#L77

paul121 (approved):

mstenta:
Merged.

PR #815, opened 2024-03-27
Add geometry/location fields to CSV importers

mstenta:
The goal of this is to add support for the following fields in CSV importers:

- On Logs:
  - `geometry`
  - `is_movement`
- On Assets:
  - `intrinsic_geometry`
  - `is_location`
  - `is_fixed`

We talked about this in some recent dev calls and I started sketching it up, so I wanted to get a draft PR opened up that we can work collaboratively on.

I have not tested any of this code yet, but I think it's functional.

Remaining todos:

- [x] Automated tests
- [x] Allow `assets`, `locations`, `categories`, and `parents` columns to be omitted from CSVs.

### DIFF
--- a/modules/core/import/modules/csv/migrations/csv_asset.yml
+++ b/modules/core/import/modules/csv/migrations/csv_asset.yml
@@ -25,6 +25,17 @@ process:
       delimiter: ,
     - plugin: asset_lookup
 
+  # Is location/fixed booleans.
+  is_location:
+    plugin: boolean
+    source: is location
+  is_fixed:
+    plugin: boolean
+    source: is fixed
+
+  # Intrinsic geometry.
+  intrinsic_geometry: intrinsic geometry
+
   # Asset status.
   status:
     - plugin: get
@@ -42,6 +53,12 @@ third_party_settings:
         description: Parents of the asset. Accepts asset names, ID tags, UUIDs, and IDs. Multiple assets can be separated by commas with the whole cell wrapped in quotes.
       - name: notes
         description: Notes about the asset.
+      - name: is location
+        description: Whether this asset is a location. Accepts most boolean values. Leave this blank to use the default for this asset type.
+      - name: is fixed
+        description: Whether this asset has a fixed location. Accepts most boolean values. Leave this blank to use the default for this asset type.
+      - name: intrinsic geometry
+        description: The intrinsic geometry of the asset in WKT format. This is only used if the asset has a fixed location.
       - name: status
         description: 'Status of the asset. Allowed values: active, archived. Defaults to active.'
 
--- a/modules/core/import/modules/csv/migrations/csv_asset.yml
+++ b/modules/core/import/modules/csv/migrations/csv_asset.yml
@@ -18,9 +18,11 @@ process:
 
   # Parent asset reference lookup.
   parent:
+    - plugin: skip_on_empty
+      method: process
+      source: parents
     - plugin: explode
       delimiter: ,
-      source: parents
     - plugin: asset_lookup
 
   # Asset status.
--- a/modules/core/import/modules/csv/migrations/csv_log.yml
+++ b/modules/core/import/modules/csv/migrations/csv_log.yml
@@ -84,6 +84,14 @@ process:
     - plugin: term_lookup
       bundle: log_category
 
+  # Geometry.
+  geometry: geometry
+
+  # Is movement boolean.
+  is_movement:
+    plugin: boolean
+    source: is movement
+
   # Log status.
   status:
     - plugin: get
@@ -115,6 +123,10 @@ third_party_settings:
         description: Notes about the log.
       - name: categories
         description: Log category taxonomy terms. Multiple terms can be separated by commas with the whole cell wrapped in quotes.
+      - name: geometry
+        description: The geometry of the log in WKT format.
+      - name: is movement
+        description: Whether this log represents an asset movement. Accepts most boolean values. Leave this blank to use the default for this log type.
       - name: status
         description: 'Status of the log. Allowed values: pending, done. Defaults to done.'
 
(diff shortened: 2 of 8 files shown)


mstenta:
One other big question is: do we want `intrinsic_geometry`, `is_fixed`, and `is_location` to show on all asset types?

Technically they all have those fields, and technically it's possible for an animal to be a fixed location asset with intrinsic geometry (:stuck_out_tongue_closed_eyes:), but in practice that will almost never be the case, and including them in the CSV importers is very confusing IMO.

As it is written now, this PR adds those fields to ALL asset types. That's just the easiest way to do it. If we want to restrict it to certain types (perhaps only asset types that are locations by default, like `land`, `structure`, and `water`), then we would need to move the mapping definitions out of `csv_asset.yml` and into conditional PHP logic.

We could also punt on the question entirely and remove those fields from this PR. Then it would only add `geometry` and `is_movement` to logs, which is arguably the more widely useful feature addition. Although, I think the benefit of adding `intrinsic_geometry` to assets is that users could import CSVs of locations with their geometry more easily.

So, if we do need to change those to use conditional PHP logic, should we do that all in this PR, or merge the log fields first and split the asset fields out to a separate PR?

modules/core/import/modules/csv/migrations/csv_log.yml
```diff
@@ -97,6 +105,8 @@ third_party_settings:
         description: Assets referenced by the log. Accepts asset names, ID tags, UUIDs, and IDs. Multiple assets can be separated by commas with the whole cell wrapped in quotes.
       - name: locations
         description: Location assets where the log took place. Accepts asset names, ID tags, UUIDs, and IDs. Multiple assets can be separated by commas with the whole cell wrapped in quotes.
+      - name: geometry
```
mstenta: Remove duplicate.

modules/core/import/modules/csv/migrations/csv_asset.yml
```diff
@@ -36,10 +47,18 @@ third_party_settings:
     columns:
       - name: name
         description: Name of the asset. Required.
+      - name: intrinsic_geometry
```
mstenta: Remove duplicate.

mstenta:
Added automated tests.

I also made a small change to allow assets, locations, categories, and parents columns to be omitted from CSVs. I noticed this was an issue because I wanted to create simple automated tests that only included a few columns, but they weren't working because the `explode` process plugin doesn't work when the value it receives is `null`. So I simply added a `skip_on_empty` plugin in front of it. Most other columns are already optional, it was just these ones that use `explode` that weren't.

mstenta:
Ready for review.

modules/core/migrate/src/Plugin/migrate/process/Boolean.php
```diff
@@ -0,0 +1,47 @@
...
+ * process:
+ *   is_movement:
+ *     plugin: boolean
+ *     source: is_movement
+ * @endcode
+
+ * @codingStandardsIgnoreEnd
+ *
+ * @MigrateProcessPlugin(
+ *   id = "boolean"
```
paul121, line 27: Curious if this ID is too generic. Did a quick search but didn't see anything that is conflicting. This was the closest boolean plugin, interesting that it takes a different approach: https://git.drupalcode.org/project/migration_tools/-/blob/8.x-2.x/src/Plugin/migrate/process/ConvertBoolean.php

I like the logic that is implemented here as a more generic approach but it does make we wonder if having an option to provide a "mapping" would be useful.

paul121 (approved):

modules/core/migrate/src/Plugin/migrate/process/Boolean.php
mstenta: > Curious if this ID is too generic.

It is indeed generic. I was surprised that this was not already implemented in core. If we want to go to the trouble, we could consider turning it into a core patch and proposing it.

I found the `migration_tools` module and its `convert_boolean` plugin in my research. I felt that PHP's native `filter_var($value, FILTER_VALIDATE_BOOLEAN)` was a more elegant approach.

Sources:

- https://www.php.net/manual/en/filter.filters.validate.php
- https://stackoverflow.com/questions/4775294/parsing-a-string-into-a-boolean-value-in-php

I'm not too worried about potential conflicts with future core/contrib plugin IDs. If that happens, then we can cross that bridge... but I would assume that it won't be very difficult to transition.

> I like the logic that is implemented here as a more generic approach but it does make we wonder if having an option to provide a "mapping" would be useful.

I suppose if there's a need to explicitly restrict the available options, then a mapping would be useful. In that case, I think a developer can use the existing `static_map` process plugin. So I don't think we need to provide that option, personally. But if it comes up in the future as a valid need, then I suppose we could append configuration options to this `boolean` plugin that allow an optional mapping to be specified.

paul121: > Sources:
> https://www.php.net/manual/en/filter.filters.validate.php

This first link is now a 404... but this seems to work: https://www.php.net/manual/en/filter.constants.php#constant.filter-validate-bool

> [FILTER_VALIDATE_BOOL](https://www.php.net/manual/en/filter.constants.php#constant.filter-validate-bool) ([int](https://www.php.net/manual/en/language.types.integer.php))
Returns [true](https://www.php.net/manual/en/reserved.constants.php#constant.true) for "1", "true", "on", and "yes". Returns [false](https://www.php.net/manual/en/reserved.constants.php#constant.false) for "0", "false", "off", "no", and "". The return value for non-boolean values depends on the [FILTER_NULL_ON_FAILURE](https://www.php.net/manual/en/filter.constants.php#constant.filter-null-on-failure). If it is set, [null](https://www.php.net/manual/en/reserved.constants.php#constant.null) is returned, otherwise [false](https://www.php.net/manual/en/reserved.constants.php#constant.false) is returned.

PR #818, opened 2024-03-28
Plan record views integration

paul121:
This PR makes two changes so that `plan_record` entities can be used in views.

I was surprised that the `plan_record_data` table actually isn't even being created right now... I know I suggested we add that here: https://github.com/farmOS/farmOS/pull/781#pullrequestreview-1850491589

It seems this table might only be created if you have a `translatable` entity? Anyways, I think we should remove the `data_table` attribute from the entity definition. The presence of the attribute without existence of the table breaks the core `EntityViewsData` integration.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [Add file and image base fields to all taxonomy terms #807](https://github.com/farmOS/farmOS/pull/807)
 - [Issue #3390486: Add an Ontology URI field to all taxonomy terms](https://www.drupal.org/project/farm/issues/3390486)
 - [Add "Days of harvest" field to Plant type terms #794](https://github.com/farmOS/farmOS/pull/794)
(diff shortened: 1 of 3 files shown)


paul121:
> It seems this table might only be created if you have a translatable entity? 

I'm also curious what changes we could make so that we would utilize `plan_record_data`.. right now we don't have any metadata or revision information about these entities. It would be nice to have simple changed/updated timestamps. Maybe an initial author.

... If not just full-blown revisions? Would there be any issues with that? I don't think `entity_reference_revisions` would be a factor for these `plan_record` entities?

mstenta:
> It would be nice to have simple changed/updated timestamps. Maybe an initial author.

:+1:

> If not just full-blown revisions? Would there be any issues with that?

I'm not opposed. My thought was we could enable revisions as a next step after #781 if they seemed useful.

> I don't think `entity_reference_revisions` would be a factor for these `plan_record` entities?

Yea I don't think `entity_reference_revisions` would even come into play, because nothing references these entities. They do all the referencing. So that keeps things nice and simple. :-)

mstenta (approved):

PR #820, opened 2024-03-29
Do not trim whitespace from quantity field item content

### LINKED ISSUE #817: no commas in list of multiple materials, material quantity
image

mstenta:
This fixes #817.

I'm not sure if it's the "best" solution, but it was very simple, so I'm throwing it at the wall to see if it sticks. :-)

Curious your thoughts @paul121 since you were the original author of these lines. Were they copied from a core template?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - [Fix Client error when passing asset parameter to forms #814](https://github.com/farmOS/farmOS/pull/814)
 - [Update patch for Issue #3397275 to fix boolean field JSON Schema #819](https://github.com/farmOS/farmOS/pull/819)
+- [Do not trim whitespace from quantity field item content #820](https://github.com/farmOS/farmOS/pull/820)
 
 ## [3.1.2] 2024-02-26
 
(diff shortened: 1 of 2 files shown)


mstenta:
Before/after screenshots:

Screenshot from 2024-03-29 14-29-18

Screenshot from 2024-03-29 14-29-29

paul121 (commented):
Interesting, this twig syntax is new to me. I believe I did copy this from somewhere in core.

I guess it works... but agree doesn't seem "best". What would it look like to clearly add a space within this loop? Add a "prefix" whenever it is not the 0th item?

I kind of like just using spaces but see how commas could be preferred

mstenta:
> Interesting, this twig syntax is new to me. I believe I did copy this from somewhere in core.

Me too. I almost didn't notice it at first, and then had to look up what it meant.

> What would it look like to clearly add a space within this loop? Add a "prefix" whenever it is not the 0th item?

I'm not sure if there's any benefit to that over just allowing whitespace like this? :thinking: 

> I kind of like just using spaces but see how commas could be preferred

I thought about figuring out how to use commas, but this begs the question: do we use commas in other places?

All of those potential next steps vs this simple fix makes me wonder if we should just commit this to fix the "bug" and worry about "improvements" separately (and maybe more broadly).

paul121 (approved):
> All of those potential next steps vs this simple fix makes me wonder if we should just commit this to fix the "bug" and worry about "improvements" separately (and maybe more broadly).

Sounds good to me

Issue #816, opened 2024-03-27
Add module self-dependency to CSV Importers docs


Fosten:
Maybe worth noting in the documentation here:
https://github.com/farmOS/farmOS/blob/3.x/docs/development/module/csv.md

Adding your own module as a dependency allows for the CSV import configuration to be removed when the module is uninstalled.

<code>dependencies:
&nbsp;&nbsp;module:
&nbsp;&nbsp;&nbsp;&nbsp;- your_module
&nbsp;&nbsp;enforced:
&nbsp;&nbsp;&nbsp;&nbsp;module:
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;- your_module</code>

Otherwise you’ll need to remove this manually before re-install:
<code>drush config-delete migrate_plus.migration.{{ UNIQUE_ID }}</code>

mstenta:
Yeaaa good point @Fosten.

On the dev call after we recorded https://farmos.org/blog/2024/how-to-build-custom-csv-importer-video/, I mentioned that there's another way to do this that would simplify some of these things. Basically: if you just put the YML in `[module]/migrations` (and rename it and make a few small tweaks) instead of `config/install`, then you don't need to worry about this "uninstall" issue, and updating it is just a matter of clearing the cache. So part of me really wants to change our recommendation to that.

But it wouldn't be a bad idea to add the `enforced` bit to dependencies in the docs in the meantime (not you only need the `enforced` bit, but not the other top-level `module` bit).

Do you want to take a pass at a PR?

PR #823, opened 2024-04-04
Recommend enforced module self-dependency in CSV doc

Fosten:
Solves issue #816


mstenta (changes requested):
Thanks @Fosten - this is great!

My only minor request is to break the line at 80 characters (for consistency with all our docs), and use a single space between sentences. Just nitpicks. I can make these changes myself before merging if you don't have a chance. :-)

mstenta:
Thanks @Fosten! Merged!

PR #824, opened 2024-04-09
Override the birth log asset label and description

mstenta:
This overrides the "Asset" field on Birth logs to change the label and description as follows.

- Label: Children
- Description: Which child assets were born?

See @paul121's comment here for the inspiration: https://github.com/farmOS/farmOS/pull/774#issuecomment-1894813820

This is nearly identical in implementation to #774 

I'm open to suggestions on the label/description. I thought "Children" would be better than "Child" in this case, even though it's inconsistent with our normally singular labels. I don't feel too strongly though, so if others prefer "Child" I can change it to that.

### DIFF
--- a/modules/log/birth/farm_birth.post_update.php
+++ b/modules/log/birth/farm_birth.post_update.php
@@ -1,0 +1,23 @@
+<?php
+
+/**
+ * @file
+ * Post update hooks for the farm_birth module.
+ */
+
+/**
+ * Override the birth log asset label and description.
+ */
+function farm_birth_post_update_override_birth_asset_label_description(&$sandbox) {
+
+  // Override the asset field label and description on birth logs to make it
+  // clear that it should be used to reference the child assets that were born.
+  // This also checks to make sure they haven't been overridden already first.
+  /** @var \Drupal\Core\Field\Entity\BaseFieldOverride $config */
+  $config = \Drupal::service('entity_field.manager')->getBaseFieldDefinitions('log')['asset']->getConfig('birth');
+  if ($config->isNew()) {
+    $config->set('label', 'Children');
+    $config->set('description', 'Which child assets were born?');
+    $config->save();
+  }
+}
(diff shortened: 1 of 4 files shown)


mstenta:
Oops one of the birth log tests is failing... I'm guessing the additional config requires something special in the kernel test. Standby for a fix...

mstenta:
Oh just realized we can also remove all of this: https://github.com/farmOS/farmOS/blob/0903f96b2bf111c9c0f2624af28e4a2e00b47d19/modules/log/birth/farm_birth.module#L14-L42

modules/log/birth/farm_birth.post_update.php
```diff
@@ -0,0 +1,19 @@
...
+ * Override the birth log asset label and description.
+ */
+function farm_birth_post_update_override_birth_asset_label_description(&$sandbox) {
+
+  // Override the asset field label and description on birth logs to make it
+  // clear that it should be used to reference the child assets that were born.
+  $config = \Drupal::service('entity_field.manager')->getBaseFieldDefinitions('log')['asset']->getConfig('birth');
+  $config->set('label', 'Children');
+  $config->set('description', 'Which child assets were born?');
+  $config->save();
```
paul121: Oh hmm I would have guessed that we create a new `base_field_override` config entity. This must be happening behind the scenes. But is it possible to do it more directly?

mstenta: Ah good question. That may be possible too. :thinking: 

My code was based on this: https://www.previousnext.com.au/blog/overriding-base-field-labels-and-descriptions-of-drupal-entities-without-custom-code

I'm not sure what the "right" / "best" way to do it is. Perhaps there's an argument that going through `entity_field.manager` has some benefits because it is doing it from a higher level? It probably doesn't make a big difference either way. The nice thing about this is that we only need to set `label` and `description`. Constructing a `BaseFieldOverride` entity would require defining *all* the properties of that entity, which we get for free via `getConfig('birth')`.

mstenta:
> Oops one of the birth log tests is failing... I'm guessing the additional config requires something special in the kernel test. Standby for a fix...

Interesting... I traced this to a Drupal core issue: https://www.drupal.org/project/drupal/issues/3193351

Tl;dr: If you have a base field that also has a validation constraint associated with it, and then you provide a base field override for it, the constraint is no longer applied.

There is a simple workaround: use `hook_entity_bundle_field_info()` instead of `hook_entity_field_info_alter()` to add the constraint, and load the `BaseFieldOverride` first.

I've updated our code and included a comment that points to the core issue as the reasoning. Force pushed... tests should pass now.

mstenta:
Force pushed to check `$config->isNew()` in the update hook, in case someone is already overriding this config.

paul121 (approved):

PR #825, opened 2024-04-10
Add a Notes filter to the logs views

mstenta:
We created this on the April 10th farmOS Monthly Call. It adds a "Notes" filter to the Logs Views.


mstenta:
I'm going to consider this approved, based on a poll conducted in the farmOS chat room, which currently sits at 5 approve / 0 disapprove. (If that changes drastically before 3.2.0 is released I'll update this.)

PR #829, opened 2024-04-12
Remove data_table from existing plan_record entity type definition

paul121:
This fixes a bug introduced with #818 . We need to remove `data_table` from the existing `plan_record` entity type definition.

Screenshot from 2024-04-12 12-02-37


mstenta (approved):
Tested this and it works. Thanks @paul121!

PR #832, opened 2024-04-23
Entity layout and minor UI changes

paul121:
My original motivation for this was to improve the display of the "Add new comment" field on entities. Ultimately I think the best solution is to make each of our layout regions use the "gin-layer-wrapper" styles so they get a background color/shadow and are lifted on the page to match the existing comment styles. Then decreasing the size of the comment text headers (h2 is pretty big) makes things look nice.

Last, some minor changes to make other css in farm_ui_theme consistent.

### DIFF
--- a/modules/core/location/farm_location.base_fields.inc
+++ b/modules/core/location/farm_location.base_fields.inc
@@ -42,8 +42,14 @@ function farm_location_asset_base_fields() {
     'label' => t('Current geometry'),
     'computed' => AssetGeometryItemList::class,
     'hidden' => 'form',
-    'weight' => [
-      'view' => 95,
+    'view_display_options' => [
+      'label' => 'hidden',
+      'type' => 'farm_map_geofield',
+      'settings' => [
+        'output_format' => 'wkt',
+        'output_escape' => TRUE,
+      ],
+      'weight' => 95,
     ],
   ];
   $fields['geometry'] = \Drupal::service('farm_field.factory')->baseFieldDefinition($options);
@@ -154,6 +160,15 @@ function farm_location_log_base_fields() {
       'view' => 95,
     ],
     'populate_file_field' => 'file',
+    'view_display_options' => [
+      'label' => 'hidden',
+      'type' => 'farm_map_geofield',
+      'settings' => [
+        'output_format' => 'wkt',
+        'output_escape' => TRUE,
+      ],
+      'weight' => 95,
+    ],
   ];
   $fields['geometry'] = \Drupal::service('farm_field.factory')->bundleFieldDefinition($options);
 
--- a/modules/core/ui/theme/css/layout.css
+++ b/modules/core/ui/theme/css/layout.css
@@ -12,6 +12,11 @@
   }
 }
 
+/* Hide layout regions if there is no content. */
+.layout--twocol .layout__region:not(:has(div)){
+  display: none;
+}
+
 /* Style entity layout regions similar to gin-layer-wrapper. */
 .layout--twocol .layout__region {
   padding: var(--gin-spacing-s);
--- a/modules/core/ui/theme/css/layout.css
+++ b/modules/core/ui/theme/css/layout.css
@@ -1,4 +1,10 @@
-/* Add some vertical padding to layout regions. */
+/* Style entity layout regions similar to gin-layer-wrapper. */
 .layout--twocol .layout__region {
   margin-bottom: 1em;
+  padding: var(--gin-spacing-s);
+  background: var(--gin-bg-layer);
+  border: 1px solid var(--gin-border-color-layer);
+  border-radius: var(--gin-border-l);
+  box-sizing: border-box;
+  box-shadow: var(--gin-shadow-l1);
 }
--- a/modules/core/ui/theme/css/layout.css
+++ b/modules/core/ui/theme/css/layout.css
@@ -21,3 +21,8 @@
   box-sizing: border-box;
   box-shadow: var(--gin-shadow-l1);
 }
+
+/* Remove padding from top region if it only has a map. */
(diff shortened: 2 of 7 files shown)


paul121:
Some before/after examples...

<details>
<summary>Assets</summary>

asset-before
asset-after

</details>

<details>
<summary>Logs</summary>

log-before
log-after

</details>

modules/core/ui/theme/css/layout.css
```diff
@@ -1,10 +1,33 @@
...
+  }
 }
 
-/* Hide geometry field label in asset and log page top region. */
-body.path-asset .layout__region--top .field--name-geometry .field__label,
-body.path-log .layout__region--top .field--name-geometry .field__label {
+/* Hide layout regions if there is no content. */
+.layout--twocol .layout__region:not(:has(div)){
   display: none;
 }
```
paul121, line 18: Hmm finding some situation where this might not work.

Screenshot from 2024-04-23 12-53-30

paul121: This seemed to be caused by an empty `<div>` that seems like it would have displayed error/warning messages but it didn't render any and I haven't been able to reproduce. I was in-between switching branches and clearing cache so it's possible it was a fluke but I do wonder if this is an issue.

paul121:
I will try and get some more screenshots (mobile!) with more data on these pages. Most notably log quantities and files that are attached to the page.

paul121:
A couple more screenshots with more data... definitely a few things to clean up, but the mobile LGTM!

Anyone have thoughts on the below? @mstenta @Farmer-Eds-Shed @Fosten @symbioquine 

1. Images are dropped in the `right` dashboard region, sometimes above other fields (`is_location`, `is_fixed`, see asset example).
    - I think we should consider adding in the `Images` field label and try to make images appear last in this column.
    - OR should we move images to the bottom with files? Only display the first image above?
1. Files are dropped in the `bottom` dashboard region above comments, inconsistent styling.
    - We should add a `Files` field label to make it clear what these are. Also need to make sure there is some margin/spacing between the Files + Comments.
    - Also, because the Files are rendered in another `gin-layer-wrapper`, this makes the comments look a little weird... hmm. We could make comments behave similarly but if other modules add fields to this bottom dashboard region they would have some inconsistency as well. 
    - Maybe consider making the files render in a collapsed `details`?
1. Consider moving more "meta" fields to the `right` dashboard region?
    - I first noticed this for the "Land Type" field (see asset example). It seems this might make more sense under the "Asset Type" field in the `right` region. BUT.... the Land Type edit form is in the main region, not the sidebar... I don't know, the `right` sidebar doesn't need to only be "meta" fields...
    - Maybe move "Flags", "Log Category" to the `right` ?
    - Move "Land Type" above "Flags"
  

<details>

<summary>Asset</summary>

more-data
more-data-2

</details>

<details>

<summary>Log</summary>

log-more-data
log-more-data-2

</details>

<details>

<summary>Log mobile</summary>

mobile-log
mobile-log-2

</details>

paul121:
I've pulled out some of the items from this PR into #846  and #847  - let's pick up the styling aspects of this issue after those are reviewed/merged

paul121 (commented):
> definitely a few things to clean up, but the mobile LGTM!

Just revisiting this, on mobile the Log type + Status should probably at the top of the screen. As is, these are in the "right" column which gets moved down below the "left" column. Not sure the best solution... either move Log type + Status, or move the right column above the left on smaller screens? I think this will be better with #847

modules/core/location/farm_location.base_fields.inc
```diff
@@ -42,8 +42,14 @@ function farm_location_asset_base_fields() {
...
-    'weight' => [
-      'view' => 95,
+    'view_display_options' => [
+      'label' => 'hidden',
+      'type' => 'farm_map_geofield',
+      'settings' => [
+        'output_format' => 'wkt',
+        'output_escape' => TRUE,
+      ],
+      'weight' => 95,
```
mstenta, line 52: Instead of explicitly setting `view_display_options` here (which requires setting all the other bits and sub-settings), can we leave this as it was, but put a line after the call to `\Drupal::service('farm_field.factory')->baseFieldDefinition($options);` which simply sets the `label` to `hidden`?

For example:

```
$fields['geometry'] = \Drupal::service('farm_field.factory')->baseFieldDefinition($options);

// Hide the geometry field label in the view display.
$view_display_options = $fields['geometry']->getDisplayOptions('view');
$view_display_options['label'] = 'hidden';
$fields['geometry']->setDisplayOptions('view', $view_display_options);
```

This makes it very explicit what we're doing, and means that if any of those other settings change in `farm.field_factory` in the future we don't need to remember to update them here as well.

modules/core/location/farm_location.base_fields.inc
```diff
@@ -154,6 +160,15 @@ function farm_location_log_base_fields() {
...
     'populate_file_field' => 'file',
+    'view_display_options' => [
+      'label' => 'hidden',
+      'type' => 'farm_map_geofield',
+      'settings' => [
+        'output_format' => 'wkt',
+        'output_escape' => TRUE,
+      ],
+      'weight' => 95,
+    ],
```
mstenta, line 171: (Same as above)

mstenta (changes requested):

mstenta:
It seems like there are still some bigger questions to answer / decisions to make with this PR generally (although I love where we're headed with it)! I'm inclined to wait until after the 3.3.0 release so we have more time to work through it.

mstenta:
We decided on the dev call today to merge #847 first for farmOS v3.3.0, and then rebase this and pick up where we left off with it.

PR #837, opened 2024-05-02
Include SMTP module

mstenta:
This adds the SMTP module to farmOS's `composer.json`, and updates the documentation to note that it doesn't need to be downloaded separately anymore.

This makes it easier for people who are self-hosting farmOS in a Docker container to configure email.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- [Include SMTP module #837](https://github.com/farmOS/farmOS/pull/837)
+
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


pcambra:
Is there any reason to include this rather than leveraging Symfony Mailer which is partially in core now? https://www.drupal.org/project/drupal/issues/1803948

mstenta:
Oh that's interesting - I wasn't aware of that! At quick glance it looks like we won't need this after all. I need to test to see what's possible (and update farmOS documentation accordingly), but this is very promising. Thanks for the heads up @pcambra!

mstenta:
Closing this - I will test out the `symfony_mailer` plugin and see if it accomplishes what we need. If so I'll open a new PR with updated documentation.

mstenta:
I tested `symfony/mailer` and it works great! I opened a new PR for review here: #844

Thanks again @pcambra!

PR #838, opened 2024-05-02
Display a warning if there are outstanding database updates

mstenta:
This adds a warning message that is displayed on every page if/when there are outstanding database updates that need to be run.

We've seen in the community forum/chat that sometimes people who are self-hosting don't know that they need to run database updates and rebuild the cache. This is dangerous because it means they could be running code that is not compatible with their database/configuration. This warning serves as a simple safety check to help prevent that.

Notably, this copies the logic and message text directly from the Drupal core `system` module's `hook_requirements()`, which shows a message on the admin status report page. All we are doing here is showing that same message on all pages, so it can't be missed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- [Display a warning if there are outstanding database updates #838](https://github.com/farmOS/farmOS/pull/838)
+
 ### Fixed
 
 - [Remove data_table from existing plan_record entity type definition #829](https://github.com/farmOS/farmOS/pull/829)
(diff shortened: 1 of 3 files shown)


pcambra:
IMHO rather than doing the whole update check on every request, we could use cron to update a state, or at least restrict running this once an hour?

mstenta:
@pcambra +1 to making this less "heavy" :-)

Thinking... a cron job that updates a state variable seems like the right way to do it. But I worry that the same users who are most likely not running `update.php` also have not set up cron, so that might miss the target audience of this PR.

> or at least restrict running this once an hour?

Perhaps this is the next-best option, and will ensure that the people who need to see it will.

The only downside, obviously, is that someone may end up running updated code for up to an hour before realizing they need to update their database. Depending on the updates, this could be more/less dangerous... but I suppose anything is better than nothing. :shrug:

PR #839, opened 2024-05-06
Allow multiple locations to be referenced in the planting quick form

### LINKED ISSUE #749: Planting quick form does not support multiple locations
I could have sworn we already fixed this, but I ran into it myself today when I was recording a planting into two beds...

**Describe the bug**

In the Planting quick form, within the Seeding/Transplanting sections, it is not currently possible to reference multiple locations.

Or rather, to be more specific, and to highlight why this is a "bug": the form does not clearly state that it only supports a single location, so if you try to add multiple (comma separated) locations, like you can do in other forms, it will accept it, but only the LAST location will be referenced in the log that gets created. I wouldn't have even noticed it, if I hadn't looked closely at the asset that was created and noticed that it's "Current location" only showed the second bed and not the first.

mstenta:
Fixes #749 

I made this commit last year but completely forgot about it. It still needed tests, so I added some.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- [Allow multiple locations to be referenced in the planting quick form #839](https://github.com/farmOS/farmOS/pull/839)
+
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


mstenta:
I filed this under "Changed" in `CHANGELOG.md`, rather than under "Added" or "Fixed", because it's sort of on the line between a feature and a bug fix. But I want to get it out in a minor release so users who are still doing spring plantings can benefit from it ASAP.

paul121 (approved):

PR #843, opened 2024-05-08
Update simple_oauth to 6.0.0-beta6

paul121:
Before we merge this I want to make sure my updated patch gets reviewed either by us or other simple_oauth maintainers. I found a new (minor) issue and slightly refactored the client validation logic in the process. The simple_oauth tests are all still passing and I feel confident in the changes but just something that should get more eyes on it.

https://www.drupal.org/project/simple_oauth/issues/3322325#comment-15586095

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -84,7 +84,7 @@
                 "Issue #3390505: Error: uri is not a valid type for a JSON document": "https://www.drupal.org/files/issues/2024-04-11/3390505-3.patch"
             },
             "drupal/simple_oauth": {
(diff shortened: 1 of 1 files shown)


mstenta:
I gave this a quick look, and although I haven't followed this thread closely your changes and explanations make sense. Looking at the old patch and the new patch, and the differences between them, I don't see anything that concerns me about including the new patch in farmOS. We've been using the old one for over a year without any issue, and the new changes don't look like they are significantly different enough to cause issues. But I'm admittedly not an OAuth2 expert, so there may be cases I'm overlooking.

I personally think this looks safe to merge. :+1:

mstenta (approved):

PR #847, opened 2024-05-16
Reorder and improve entity field rendering

paul121:
This PR changes field weights for consistency and to improve their rendering in entity view pages. Even though the fields' form weights are changing, this PR should not be changing the order of any fields in forms because many fields are grouped into tabs of similar fields.

This PR also moves images to the `bottom` region and renders them in a simple `flex` container so that multiple images can be displayed horizontally in the same row and wrap to new rows as necessary (adjusts to screensize). I think this is a simple short-term optimization to displaying a potentially long vertical list of images all in the `second` region. It keeps images and files together and adds a field `label` for each.

<details>
<summary>Land asset before:</summary>

land-asset-before-1
land-asset-before-2

</details>

<details>
<summary>Land asset after:</summary>

land-asset-after-1
land-asset-after-2

</details>

<details>
<summary>Log before:</summary>

log-before-1

</details>

<details>
<summary>Log after:</summary>

log-after-1

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [Update Drupal core to 10.3 #872](https://github.com/farmOS/farmOS/pull/872)
 - [Move farm_quantity View to farm_log_quantities and require log relationship #858](https://github.com/farmOS/farmOS/pull/858)
 - [Quantity CSV export improvements #861](https://github.com/farmOS/farmOS/pull/861)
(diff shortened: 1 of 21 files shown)


paul121:
cc @Farmer-Eds-Shed @Fosten @mstenta hopefully this covers most of the simple field re-orderings we discussed on the dev call today...

mstenta (approved):
LGTM! Thanks @paul121! I added two additional commits to move the `is_movement` and `is_group_assignment` fields to the second region, consistent with `is_fixed` and `is_location` on assets. This is ready to merge IMO.

PR #850, opened 2024-06-14
Add an asset.logs service for retrieving logs that reference an asset

mstenta:
This adds a new `asset.logs` service for retrieving logs that reference an asset.

See the included developer documentation for more info.

### DIFF
--- a/modules/core/log/src/AssetLogs.php
+++ b/modules/core/log/src/AssetLogs.php
@@ -1,0 +1,93 @@
+<?php
+
+namespace Drupal\farm_log;
+
+use Drupal\asset\Entity\AssetInterface;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+
+/**
+ * Service for loading logs that reference assets.
+ */
+class AssetLogs implements AssetLogsInterface {
+
+  /**
+   * Entity type manager.
+   *
+   * @var \Drupal\Core\Entity\EntityTypeManagerInterface
+   */
+  protected EntityTypeManagerInterface $entityTypeManager;
+
+  /**
+   * Log query factory.
+   *
+   * @var \Drupal\farm_log\LogQueryFactoryInterface
+   */
+  protected LogQueryFactoryInterface $logQueryFactory;
+
+  /**
+   * Class constructor.
+   *
+   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_type_manager
+   *   Entity type manager.
+   * @param \Drupal\farm_log\LogQueryFactoryInterface $log_query_factory
+   *   Log query factory.
+   */
+  public function __construct(EntityTypeManagerInterface $entity_type_manager, LogQueryFactoryInterface $log_query_factory) {
+    $this->entityTypeManager = $entity_type_manager;
+    $this->logQueryFactory = $log_query_factory;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getLogs(AssetInterface $asset, string $log_type = NULL, bool $access_check = TRUE): array {
+    $log_ids = $this->query($asset, $log_type, $access_check)->execute();
+    if (empty($log_ids)) {
+      return [];
+    }
+    return $this->entityTypeManager->getStorage('log')->loadMultiple($log_ids);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getFirstLog(AssetInterface $asset, string $log_type = NULL, bool $access_check = TRUE) {
+    $log_ids = $this->query($asset, $log_type, $access_check, 1)->execute();
+    if (empty($log_ids)) {
+      return NULL;
+    }
+    return $this->entityTypeManager->getStorage('log')->load(reset($log_ids));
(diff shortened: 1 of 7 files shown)


mstenta:
Context: this code was originally written for the `farm_crop_plan` module, to make it easier to load an asset's logs and display them as dots on the timeline.

See: https://github.com/mstenta/farm_crop_plan/blob/fa02e9dc3ae80b45b4b0b16f4d613307d91d927b/src/CropPlan.php#L97

By moving this upstream to farmOS core, it can be leveraged by other modules (like the `farm_grazing_plan` module), as well as by core itself. This is a prerequisite for showing a timeline on individual asset pages with dots for all of the asset's logs.

mstenta:
Amended the last commit to fix codesniffer complaints.

modules/core/log/src/AssetLogs.php
```diff
@@ -0,0 +1,89 @@
...
+   *   Optionally filter by log type.
+   * @param bool $access_check
+   *   Whether to check log entity access.
+   * @param int|null $limit
+   *   The number of logs to return.
+   *
+   * @return \Drupal\Core\Entity\Query\QueryInterface
+   *   A query object.
+   */
+  protected function query(AssetInterface $asset, string $log_type = NULL, bool $access_check = TRUE, int $limit = NULL) {
```
paul121, line 77: Just wondering if we should make this a `public` function so you could further customize the query. It wouldn't add a whole lot... you could just use the log query factory and replicate this anyways. :shrug:

modules/core/log/src/AssetLogs.php
```diff
@@ -0,0 +1,89 @@
...
+    $this->entityTypeManager = $entity_type_manager;
+    $this->logQueryFactory = $log_query_factory;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getLogs(AssetInterface $asset, string $log_type = NULL, bool $access_check = TRUE): array {
+    $log_ids = $this->query($asset, $log_type, $access_check)->execute();
+    return $this->entityTypeManager->getStorage('log')->loadMultiple($log_ids);
```
paul121: I know `loadMutliple()` will load all entities if `NULL` is passed instead of an array of IDs. I'm not quite sure what the behavior is if an empty array is passed... I assume it won't load any? But maybe we could add a quick check and only `loadMultiple()` if we have some IDs?

modules/core/log/src/AssetLogs.php
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getFirstLog(AssetInterface $asset, string $log_type = NULL, bool $access_check = TRUE) {
+    $log_ids = $this->query($asset, $log_type, $access_check, 1)->execute();
+    return $this->entityTypeManager->getStorage('log')->load(reset($log_ids));
```
paul121: Ditto, wouldn't hurt to make sure we have at least one log ID

modules/core/log/src/AssetLogs.php
mstenta: The method can still be overridden by a downstream class (`protected` can, `private` can't) if that's what you mean? I don't think it makes sense to expose this method publicly in the `AssetLogsInterface`, since it is basically just a helper for reducing code duplication within the other two methods.

modules/core/log/src/AssetLogs.php
mstenta: fixup commit: https://github.com/farmOS/farmOS/pull/850/commits/18172abd9099d65ffb5b09f1cecd0db0e439ac62

modules/core/log/src/AssetLogs.php
mstenta: fixup commit: https://github.com/farmOS/farmOS/pull/850/commits/18172abd9099d65ffb5b09f1cecd0db0e439ac62

modules/core/log/src/AssetLogs.php
paul121: Agree, no need to make it public then!

paul121 (approved):

mstenta:
Merged my fixup commit and rebased onto 3.x.

One additional change I made to this PR is I removed `@internal` from `class AssetLogs` and `interface AssetLogsInterface`. This was copied and pasted directly from `LogQueryFactory(Interface)`, which we consider an internal service. But this new `asset.logs` service is meant to be a public-facing API service, and we document it in `docs/development/module/services`. Notably, the other services documented there are NOT marked `@internal`, so this is consistent with those.

I also fixed the docblock of the `AssetLogsInterface` (it was copied/pasted from `LogQueryFactoryInterface`.

mstenta:
Hmm tests are failing now for some reason. :thinking:

mstenta:
Oh this seems to be happening on the `3.x` branch generally as of last night: https://github.com/farmOS/farmOS/actions/runs/10055315529

Setting this back to "ready to merge".

mstenta:
Rebased onto latest 3.x.

PR #854, opened 2024-07-15
Issue #3314741: EntityFieldManager::getFieldMap() doesn't show farmOS bundle fields

mstenta:
This fixes [Issue #3314741: EntityFieldManager::getFieldMap() doesn't show farmOS bundle fields](https://www.drupal.org/project/farm/issues/3314741).

This PR replaces an earlier one that targeted the 2.x branch of farmOS: #583

It also greatly simplifies the changes by leveraging a patch to Drupal core (thanks for finding this @paul121! :raised_hands:):

- [Issue #3129179: Provide some way to rebuild the persistent bundle field map](https://www.drupal.org/project/drupal/issues/3129179).

The patch adds a new method to Drupal core's `entity_field.manager` service, which provides a way to rebuild the bundle field map:

`\Drupal::service('entity_field.manager')->rebuildBundleFieldMap();`

However, the patch itself does not call this method when new modules are installed/uninstalled, so it does not fix [#3314741](https://www.drupal.org/project/farm/issues/3314741) by itself, so this PR also adds simple implementations of `hook_modules_installed()` and `hook_modules_uninstalled()` which run the method. This serves to ensure bundle fields are correctly added to (or removed from) the field map when a module that provides them is installed/uninstalled.

An update hook is also provided that runs the same method. This serves to ensure that any bundle fields that were installed previously (and are currently broken) are correctly represented in the field map.

I copied the functional test that I wrote for #583 to add test coverage and tests are passing.

One other notable difference from #583 is that this PR is NOT changing the `quick` or `equipment` fields to base fields. We may still want to consider doing that, but there are other considerations there, and they don't need to block this fix.

### DIFF
--- a/modules/core/entity/tests/src/Functional/FarmEntityBundleFieldTest.php
+++ b/modules/core/entity/tests/src/Functional/FarmEntityBundleFieldTest.php
@@ -60,6 +60,49 @@ class FarmEntityBundleFieldTest extends FarmBrowserTestBase {
   }
 
   /**
+   * Test that bundle field maps are updated on install/uninstall.
+   */
+  public function testBundleFieldMapUpdates() {
+
+    // Get the entity field map.
+    $field_map = $this->entityFieldManager->getFieldMap();
+
+    // Confirm that the 'test_default_bundle_field' exists in the log field map.
+    $this->assertArrayHasKey('test_default_bundle_field', $field_map['log']);
+
+    // Confirm that the 'test_contrib_hook_bundle_field' does NOT exist (yet).
+    $this->assertArrayNotHasKey('test_contrib_hook_bundle_field', $field_map['log']);
+
+    // Install the farm_entity_contrib_test module.
+    $result = $this->moduleInstaller->install(['farm_entity_contrib_test']);
+    $this->assertTrue($result);
+
+    // Reload the entity field map.
+    $this->entityFieldManager->setFieldMap([]);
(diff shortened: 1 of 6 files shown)


modules/core/entity/tests/src/Functional/FarmEntityBundleFieldTest.php
```diff
@@ -59,6 +59,51 @@ protected function setUp():void {
...
+
+    // Uninstall the farm_entity_contrib_test module.
+    $result = $this->moduleInstaller->uninstall(['farm_entity_contrib_test']);
+    $this->assertTrue($result);
+
+    // Reload the entity field map. We need to get a new instance of the
+    // entity_field.manager service from the container without old state.
+    $this->container->set('entity_field.manager', NULL);
+    $this->entityFieldManager = $this->container->get('entity_field.manager');
+    $field_map = $this->entityFieldManager->getFieldMap();
```
paul121, line 100: Updated the patch for the core issue so that the ::rebuildBundleFieldMap() helper function resets the cache. In the tests we still need to reset the internal state of the `entity_field.manager` service though, otherwise we won't get new values. I think its safer to get a new instance from the container rather than call the `->setMap([])` helper method that seems like a weird thing to exist....

mstenta: LGTM thanks @paul121!

mstenta:
Merged.

PR #858, opened 2024-07-26
Move farm_quantity View to farm_log_quantities and require log relationship

mstenta:
As discussed on the dev call yesterday, this PR moves the `farm_quantity` View to `farm_log_quantity` and makes the `log` relationship required. This serves to make this a View of "Log Quantities" specifically, instead of "Quantities" generally. This was implicitly true before, because the View joins in data from the `log` entities that reference the `quantity` entities, and adds log-specific columns and filters to the View. So it has always been specific to Log Quantities.

And while we don't have any plans in farmOS core to associate quantities with other entities, it's possible that downstream modules may do that. And if they do, then currently their quantities would appear in this View, but would have empty columns because they are not associated with logs. So this PR is mostly a defensive change to avoid potential issues in those cases, by making the intention of this View more explicit.

We also talked on the dev call about possible ways to create a general Quantities View that joined in ANY entity that references them, but there are some tricky technical questions/hurdles there. So this is a simple first step that at least avoids problems in the short term.

Notably, this also helps with (but doesn't solve) #775, by ensuring that orphaned quantities will not appear in the View anymore.

### DIFF
--- a/modules/core/ui/views/src/Plugin/Derivative/FarmViewsMenuLink.php
+++ b/modules/core/ui/views/src/Plugin/Derivative/FarmViewsMenuLink.php
@@ -27,6 +27,18 @@ class FarmViewsMenuLink extends ViewsMenuLink {
   protected string $entityType;
 
   /**
+   * Specify the View ID. This should be set in child classes.
+   *
+   * @var string
+   *
+   * @see \Drupal\farm_ui_views\Plugin\Derivative\FarmAssetViewsMenuLink
+   * @see \Drupal\farm_ui_views\Plugin\Derivative\FarmLogViewsMenuLink
+   * @see \Drupal\farm_ui_views\Plugin\Derivative\FarmPlanViewsMenuLink
+   * @see \Drupal\farm_ui_views\Plugin\Derivative\FarmQuantityViewsMenuLink
+   */
+  protected string $viewId;
+
+  /**
    * The entity type manager service.
    *
    * @var \Drupal\Core\Entity\EntityTypeManagerInterface
@@ -78,8 +90,8 @@ class FarmViewsMenuLink extends ViewsMenuLink {
     foreach ($bundles as $type => $bundle) {
       $links['farm.' . $this->entityType . '.' . $type] = [
         'title' => $bundle->label(),
-        'parent' => 'views_view:views.farm_' . $this->entityType . '.page',
-        'route_name' => 'view.farm_' . $this->entityType . '.page_type',
+        'parent' => 'views_view:views.' . $this->viewId . '.page',
+        'route_name' => 'view.' . $this->viewId . '.page_type',
         'route_parameters' => ['arg_0' => $type],
       ] + $base_plugin_definition;
     }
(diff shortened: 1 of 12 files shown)


mstenta:
Note: this moves `/quantities` to `/log-quantities`. So it also adds a redirect for backwards compatibility.

mstenta:
One thing I had to change here is the way that we automatically derive sub-menu links for Asset/Log/Plan/Quantity Views. I added a `$viewId` attribute to the `FarmViewsMenuLink` class (and child classes) to explicitly specify the View ID for each entity type. Previously this was inferred to be "`farm_` + entity type". But since the quantity View is changing from `farm_quantity` to `farm_log_quantity`, that wouldn't work, so the quantity type menu items broke.

mstenta:
Adding this to the v3.3.0 milestone since it is a change that should be reserved for a minor release IMO.

modules/core/ui/views/src/Plugin/Derivative/FarmViewsMenuLink.php
```diff
@@ -15,16 +15,29 @@
...
 
+  /**
+   * Specify the View ID. This should be set in child classes.
+   *
+   * @var string
+   *
+   * @see \Drupal\farm_ui_views\Plugin\Derivative\FarmAssetViewsMenuLink
+   * @see \Drupal\farm_ui_views\Plugin\Derivative\FarmLogViewsMenuLink
+   * @see \Drupal\farm_ui_views\Plugin\Derivative\FarmPlanViewsMenuLink
+   * @see \Drupal\farm_ui_views\Plugin\Derivative\FarmQuantityViewsMenuLink
```
paul121: I don't think these @see are necessary. It should be trivial to find child classes with a modern IDE, especially these because they are all in the same directory. These strings are just another thing to maintain if they ever change in the future.

paul121 (changes requested):
I tested the update hook locally and it worked great! Just a nit on function comments but otherwise looks good to me.

One thing to mention is that if any other modules reference the `view.farm_quantity.page` route then this will be a breaking change & could result in page crashes. They might do so if they were cross-linking to this page for documentation or help. For example, in `farm_ui_views_help` we render a link to `/logs` in the help text for this Log Quantities page. *I don't think any of our contrib modules do this so it is unlikely to cause an issue - but just want to mention*

mstenta:
> _I don't think any of our contrib modules do this so it is unlikely to cause an issue - but just want to mention_

Good point @paul121. I suppose our other option is to keep the original View ID of `farm_quantity`, and plan to change it during 4.x. However, the risk feels very low for this particular change, so I'm inclined to just do it.

modules/core/ui/views/src/Plugin/Derivative/FarmViewsMenuLink.php
mstenta: Agreed. We can remove those. I'll add a commit to the PR.

paul121 (approved):

PR #859, opened 2024-07-26
When a quantity is deleted, clean up log references to it

mstenta:
Currently, the [Entity Reference Integrity](https://www.drupal.org/project/entity_reference_integrity) module prevents entities from being deleted if they are referenced by other entities. This does NOT currently protect `quantity` entities, however, because `entity_reference_integrity` only works with `entity_reference` fields, but logs reference quantities via an `entity_reference_revisions` field.

This means that when a quantity is deleted, references to it in the `quantity` field on `log` entities do not get cleaned up.

This PR takes the approach of adding an event subscriber that will clean up those references on `log` entities when a quantity is deleted. This serves to solve both problems, and also enables us to add a "Bulk delete" action for quantities as a next step (this will be a separate PR).

This is done in the `farm_log_quantity` module's event subscriber, so the logic is contained within the same module that adds the `quantity` field to `log` entities in the first place. Notably, this module also provides event subscriber logic for cloning quantities when logs are cloned, and deleting quantities when a log is deleted. We didn't have automated tests for these subscribers, so I added some, in addition to a new test for this PR's logic.

### DIFF
--- a/modules/core/log/modules/quantity/src/EventSubscriber/LogQuantityEventSubscriber.php
+++ b/modules/core/log/modules/quantity/src/EventSubscriber/LogQuantityEventSubscriber.php
@@ -7,9 +7,9 @@ use Drupal\log\Event\LogEvent;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 
 /**
- * Perform actions on log presave.
+ * Subscribe to events related to log quantities.
  */
-class LogEventSubscriber implements EventSubscriberInterface {
+class LogQuantityEventSubscriber implements EventSubscriberInterface {
 
   /**
    * Entity type manager.
@@ -70,7 +70,7 @@ class LogEventSubscriber implements EventSubscriberInterface {
    * Perform actions on log delete.
    *
    * @param \Drupal\log\Event\LogEvent $event
-   *   Config crud event.
+   *   The log event.
    */
   public function logDelete(LogEvent $event) {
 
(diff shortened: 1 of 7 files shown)


modules/core/log/modules/quantity/src/EventSubscriber/LogQuantityEventSubscriber.php
```diff
@@ -91,4 +96,41 @@ public function logDelete(LogEvent $event) {
...
+    foreach ($logs as $log) {
+      $log->set('quantity', array_filter($log->get('quantity')->getValue(), function ($value) use ($quantity) {
+        if (!empty($value['target_id']) && $value['target_id'] == $quantity->id()) {
+          return FALSE;
+        }
+        return TRUE;
+      }));
+      $log->setNewRevision(TRUE);
+      $log->setRevisionLogMessage($this->t('Removed reference to deleted quantity %uuid.', ['%uuid' => $quantity->uuid()]));
+      $log->save();
```
paul121, line 132: Just was thinking it would be nice if there weren't multiple revision messages if you delete multiple quantities at the same time. But that might be hard to implement since this event subscriber reacts to a single quantity being deleted, not multiple.

paul121 (approved):

modules/core/log/modules/quantity/src/EventSubscriber/LogQuantityEventSubscriber.php
mstenta: Yea I had the same thought. I don't see any way around it, but I think it's good to have a revision log message.

PR #860, opened 2024-07-26
Add bulk delete action to farm_log_quantity View

mstenta:
This PR adds a "Delete" action for quantity entities, and adds a "Bulk update" column to the `farm_log_quantity` View, allowing users to delete quantities in bulk across multiple logs.

Note that this PR builds on top of #858 (so that needs to be merged first). We should also merge #859 before this, to ensure references to deleted quantities are cleaned up.

This is also a pre-step towards adding a new "Export CSV" bulk action to the `farm_log_quantity` View. We need to add the "Bulk update" column to the `farm_log_quantity` View in order for that new "Export CSV" action to show up (without explicitly depending on the `farm_export_csv` module), but the Drupal core Views module does not make a "Bulk update" column available unless there are bulk actions available for the entity type. Right now, `quantity` entities do not have any bulk actions available. So by adding a "delete" action, we both get a new feature (ability to delete quantities in bulk), and we make it easier to add new actions in next steps.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- [Add bulk delete action to farm_log_quantity View #860](https://github.com/farmOS/farmOS/pull/860)
+
 ### Changed
 
 - [Update Drupal core to 10.3 #872](https://github.com/farmOS/farmOS/pull/872)
--- a/modules/core/quantity/config/optional/system.action.quantity_delete_action.yml
+++ b/modules/core/quantity/config/optional/system.action.quantity_delete_action.yml
@@ -1,0 +1,10 @@
+langcode: en
+status: true
+dependencies:
+  module:
+    - quantity
+id: quantity_delete_action
+label: 'Delete quantity'
+type: quantity
+plugin: entity:delete_action:quantity
+configuration: {  }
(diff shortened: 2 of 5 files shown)


mstenta:
Adding this to the v3.3.0 milestone since it is a new feature.

paul121 (changes requested):
This works great! One thing to note is that the delete confirmation form displays `$quantity->label()` but this only returns the quantity ID. I think we should update the `$quantity->label()` method to return the 'rendered quantity' in the default template where we've already implemented the logic for displaying base and bundle quantity fields.

```php
  /**
   * {@inheritdoc}
   */
  public function label() {
    $view_builder = \Drupal::entityTypeManager()->getViewBuilder($this->entityTypeId);
    $renderable = $view_builder->view($this);
    return \Drupal::service('renderer')->renderPlain($renderable);
  }
```

If we don't want to modify the label function I think we could accomplish this same thing by implementing a custom `QuantityDeleteForm`. With a custom form it could be nice to display some info about the log the quantity is associated with, too. Maybe even group multiple quantities in a nested list under their respective log. 

Screenshots for reference:
quantity-id-delete

quantity-label-delete

mstenta:
> I think we should update the `$quantity->label()` method to return the 'rendered quantity' in the default template 

This seems sensible to me. I wonder if that label gets used anywhere else. I can't think of anything.

@paul121 Could you add that as a separate commit in this branch and push so we can see if it breaks any tests?

paul121:
> @paul121 Could you add that as a separate commit in this branch and push so we can see if it breaks any tests?

Done and looks like tests passed

mstenta:
Thanks @paul121! I think this is ready to merge.

paul121 (approved):

PR #842, opened 2024-05-07
Fix incorrect CSV headers when exporting multiple record types

mstenta:
Fixes #805

This omits bundle-specific columns from CSV exports that includes entities of multiple bundles. It adds a warning message to the user to let them know this is happening.

It also adds the ability to select which bundles are included in the export via checkboxes.

**Original description for posterity:**

~~This is a *quick and dirty* fix. It builds the list of CSV column headers by iterating over *every* row to build a combined list. So the more rows you are exporting the less performant it may be.~~

~~So it fixes a bug at the expense of a new performance consideration, which is the lesser of two evils IMO. And it buys us some time to work on a better overall fix (along with some of the other *feature* ideas we talked about in #805).~~

~~Worth noting, one of the effects of this is that the order of columns will depend on the data, so if you export two sets of data, or if you apply different sorts to your data, you may get different column orders. But all the data will be there in the proper place.~~

Issue #805, opened 2024-03-19
Export CSV on all assets/logs chooses header columns from the first asset/log in the list


Fosten:
When using the Export CSV button from all /assets or all /logs page, only the header columns from the first asset in the list are chosen.

From 3.1.0 release blog post... "In previous versions, only the columns that are visible on the page were included in the CSV. Now, the columns include all text, reference, timestamp, and boolean data by default."

This works fine for exporting similar asset/log types.  But not for "all" pages that include multiple types.

**Expected behavior**
- Perhaps "all" pages could only export the columns visible on the all /assets and all /logs pages.
  - assets: id, asset_type, status
  - logs: status, id, timestamp, name, log_type, owner
- Or "all" pages could specify all fields that are common to all assets/logs
  - assets: id,uuid,type,uid,name,status,created,changed,archived,notes,flag,location,is_location,is_fixed,owner,parent,quick
  - logs: id,uuid,type,uid,name,timestamp,status,created,changed,notes,flag,location,is_movement,asset,category,owner,equipment,quick
- Or "all" pages could hide the option to export from the dropdown altogether.

**Example assets export**
id,uuid,type,uid,name,status,created,changed,archived,notes,flag,location,is_location,is_fixed,owner,parent,quick,animal_type,birthdate,is_castrated,nickname,sex
2,952f0dac-a9ba-4ecb-b042-269d298a797e,animal,admin,"test animal",active,2024-03-19T20:57:35+00:00,2024-03-19T20:57:35+00:00,,,,,,,,,,"test breed",,,,
3,1cdcfa41-6f0b-4bf6-8981-7cf694a529df,equipment,admin,"test equipment",active,2024-03-19T20:58:02+00:00,2024-03-19T20:58:02+00:00,,,,,,,,,,,,
4,44e55b63-8fa4-477f-b26b-0f815946ddb3,land,admin,"test land",active,2024-03-19T20:58:20+00:00,2024-03-19T20:58:20+00:00,,,,,1,1,,,,bed
5,310948da-b398-4efe-8516-38da54bed309,structure,admin,"test structure",active,2024-03-19T20:58:38+00:00,2024-03-19T20:58:38+00:00,,,,,1,1,,,,building
1,e79d294b-2de2-4ee2-b848-482323a52a01,plant,admin,"test tomato",active,2024-03-01T23:01:54+00:00,2024-03-01T23:01:54+00:00,,,,,,,,,,tomato,
[...]

PR #861, opened 2024-07-26
Quantity CSV export improvements 

mstenta:
The goal of this PR is to improve the Quantity CSV export functionality in farmOS.

- [x] Add support for Fraction fields in CSV exports.
- [x] Add an "Export CSV" bulk action to the `farm_log_quantity` View (depends on, and includes commits from, #860)
- [x] Remove the existing "Export CSV" link and display from the View.
- [x] Append log fields to Quantity CSV exports: `id`, `status`, `timestamp`, `type`, `name`, `assets`
- [x] Add an "Export Quantity CSV" bulk action to the `farm_log` View, so that it's possible to export quantities directly from Views of logs.

### DIFF
--- a/modules/core/export/modules/csv/src/Plugin/Action/LogQuantityCsv.php
+++ b/modules/core/export/modules/csv/src/Plugin/Action/LogQuantityCsv.php
@@ -1,0 +1,99 @@
+<?php
+
+namespace Drupal\farm_export_csv\Plugin\Action;
+
+use Drupal\Core\Action\Plugin\Action\EntityActionBase;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\Session\AccountInterface;
+use Drupal\Core\TempStore\PrivateTempStoreFactory;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Action that exports a CSV file of log quantities.
+ *
+ * @Action(
+ *   id = "farm_export_csv:log_quantity",
+ *   action_label = @Translation("Export Quantities CSV"),
+ *   type = "log",
+ *   confirm_form_route_name = "entity.quantity.csv_form",
+ * )
+ */
+class LogQuantityCsv extends EntityActionBase {
+
+  /**
+   * The tempstore object.
+   *
+   * @var \Drupal\Core\TempStore\SharedTempStore
+   */
+  protected $tempStore;
+
+  /**
+   * The current user.
+   *
+   * @var \Drupal\Core\Session\AccountInterface
+   */
+  protected $currentUser;
+
+  /**
+   * Constructs a new LogQuantityCsv object.
+   *
+   * @param array $configuration
+   *   A configuration array containing information about the plugin instance.
+   * @param string $plugin_id
+   *   The plugin ID for the plugin instance.
+   * @param mixed $plugin_definition
+   *   The plugin implementation definition.
+   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_type_manager
+   *   The entity type manager service.
+   * @param \Drupal\Core\TempStore\PrivateTempStoreFactory $temp_store_factory
+   *   The tempstore factory.
+   * @param \Drupal\Core\Session\AccountInterface $current_user
+   *   Current user.
+   */
+  public function __construct(array $configuration, $plugin_id, $plugin_definition, EntityTypeManagerInterface $entity_type_manager, PrivateTempStoreFactory $temp_store_factory, AccountInterface $current_user) {
+    parent::__construct($configuration, $plugin_id, $plugin_definition, $entity_type_manager);
+    $this->tempStore = $temp_store_factory->get('entity_csv_confirm');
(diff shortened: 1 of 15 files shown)


modules/core/log/modules/quantity/src/Normalizer/QuantityNormalizer.php
paul121, line 1: @mstenta I implemented what you started here. Note how this prepends the log fields to the normalized data array and does a recursive normalize on the `log.timestamp`, using the same `$context` to maintain the human-readable timestamps. Works well!

I wasn't sure about 'bailing' if the quantity didn't reference a log.. returning an empty `$data` array might cause other issues? But really we should be able to expect that quantities will have a log so I don't think this is an issue.

One issue with this approach is that if you had quantities associated with other entity types eg "asset quantities", then it would be hard for modules to provide their own normalizers to handle the quantity -> csv normalization correctly and include "asset" data instead of logs. Not sure if this is something we can solve. Perhaps quantities could have a `serialization` event and allow modules to tack on additional metadata as needed?

modules/core/export/modules/csv/src/Plugin/Action/LogQuantityCsv.php
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * Action that exports a CSV file of log quantities.
+ *
+ * @Action(
+ *   id = "farm_export_csv:log_quantity",
+ *   action_label = @Translation("Export Quantities CSV"),
+ *   type = "log",
+ *   confirm_form_route_name = "entity.quantity.csv_form",
+ * )
```
paul121, line 19: This is a bit of a clever way to implement this Log Quantity CSV export... I found that we could reuse the `entity.quantity.csv_form` entirely, all we need to do is save the `log.quantities` into the tempstore instead of the `log` itself. This means we only need to implement this single Action plugin (not using a deriver) and don't need to duplicate almost all of the logic from `EntityCsvActionForm`.

This could create a little of bit of weird UX though... When the user selects/checks logs and clicks the "Export Quantities CSV" action, it will take them to the same "Export a CSV of [X] quantities?" confirmation form as you would get when using "Export CSV" action from a quantities view. It will not mention anything of the logs they selected. It's a little odd, but really it's just elegant and might not be too confusing to users anyways. If you do click 'Cancel' it takes you back to the correct destination Log page, not Log Quantities.

The major downside I foresee with this approach is that it just might not scale and exporting many logs with many quantities might crash, even if we are able to add a Batch API export to the general `EntityCsvActionForm`. This is simply because loading all the log quantities during the action plugin's `executeMultiple()` step can't be included in a Batch operation. Ideally, the Batch export for Log Quantities would handle a few or even a single _log_ (not quantity) in each batch operation. In this case I think the solution is providing a new Export Action Form that handles this Log + Quantity export more elegantly. I think this could be added/changed in the future without breaking change to the Action config entity added here.

modules/core/log/modules/quantity/src/Normalizer/QuantityNormalizer.php
mstenta: Looks good @paul121!

> I wasn't sure about 'bailing' if the quantity didn't reference a log.. returning an empty $data array might cause other issues?

I like that it will only modify `$data` if it finds a log associated with it. Otherwise the original `$data` will be returned. That feels like the right behavior, and means that if a log is associated with another entity (or is orphaned), the export will still work.

> Perhaps quantities could have a `serialization` event and allow modules to tack on additional metadata as needed?

Yea this seems like the only way to handle it. That can be a future task.

modules/core/export/modules/csv/src/Plugin/Action/LogQuantityCsv.php
mstenta: > This is a bit of a clever way to implement this Log Quantity CSV export...

Brilliant! That works well!

> This could create a little of bit of weird UX though...

I don't think that's too weird. If we find that it causes too much confusion we can consider ways to make that more clear. But I think this is fine for now.

> The major downside I foresee with this approach is that it just might not scale

Hmm this is a good point. I agree we can merge this for now and address that potential issue in the future.

mstenta:
Force pushed to fix two things:

- PHPCS error.
- Fixed config dependency name (`farm_log_quantity` instead of `log_quantity`).

mstenta:
Also added `enforced` dependency to the log quantity action config for consistency (not sure if this is actually necessary for `optional` configs, but that's a separate question IMO).

mstenta:
Realized on Rothamsted dev call today that this implementation misses something that is critical for their use-case. It traces back to issue #805 and PR #842.

Tl;dr: when multiple quantities are exported and they are different type (eg: `standard` and `material`), then all the bundle-specific fields are omitted from the export.

Rothamsted needs the `material_type` to be included in exports.

mstenta:
Figured out the failing tests. See the last commit message for details @paul121 - this is ready for final review.

paul121 (approved):
Interesting. This seems like a fine workaround!

PR #862, opened 2024-07-29
Add farmOS-timeline library

mstenta:
This adds the [farmOS-timeline](https://github.com/farmOS/farmOS-timeline) library to farmOS, via a new `farm_timeline` module with supporting code and render element.

This depends on #850, and therefore includes its commits.


composer.libraries.json
```diff
@@ -17,6 +18,21 @@
                   "installer-name": "farmOS-map"
                 }
             }
+        },
+        "farmos-timeline": {
+            "type": "package",
+            "package": {
+                "name": "farmos/farmos-timeline",
+                "version": "0.0.1",
```
paul121: I'd like to make a new release of the timeline library - we got a new release of the gantt library that should make a few things easier!

paul121:
Okay this should be ready for review again. I updated farmos-timeline to v0.0.2 and fixed a bug with the parent/child row processing that was implemented in `farm_timeline.js`. I think this might have been related to what you were seeing @mstenta 

I think something got goofed while I was originally working on this and making commits/rebasing.. not entirely sure but  I noticed there is still a fixup commit in this PR. So I've squashed that as I originally intended and then added 3 new follow-up commits. Available in this branch: https://github.com/paul121/farmOS/tree/3.x-timeline-fixed

composer.libraries.json
paul121: Done - https://github.com/farmOS/farmOS-timeline/releases/tag/v0.0.2

paul121:
@mstenta I should say that I tried testing this with the crop plan and it was working well. The current branch of the grazing plan doesn't seem to work with this timeline branch from farmOS core, it is dependent on some other features maybe? I get an error when trying to install the grazing plan that I'd rather not worry about debugging right now... :-)

mstenta:
Rebased onto latest 3.x and merged @paul121's `3.x-timeline-fixed` branch.

paul121 (approved):

PR #867, opened 2024-08-12
Update drupal/entity to ^1.5

mstenta:
The upstream Entity API module merged @paul121's patch: https://www.drupal.org/project/entity/issues/3190436

This was the only patch we were applying for this package, so this PR updates our version constraint to `^1.5` (unpinned) and removes the patch.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
         "drupal/consumers": "^1.19",
         "drupal/csv_serialization": "^4.0",
         "drupal/date_popup": "^1.3",
-        "drupal/entity": "1.4",
+        "drupal/entity": "1.5",
         "drupal/entity_browser": "^2.10",
         "drupal/entity_reference_integrity": "1.2",
         "drupal/entity_reference_revisions": "1.11",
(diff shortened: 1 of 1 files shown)


symbioquine:
In case anybody else is curious, the diff for 1.4 vs 1.5 can be viewed here: https://git.drupalcode.org/project/entity/-/compare/8.x-1.4...8.x-1.5?from_project_id=58581&straight=false

Not sure if it is critical or not, but the diff that got merged isn't the same. https://git.drupalcode.org/project/entity/-/commit/a64d88644639a580cff7bc7e1de4e852b8e617aa vs https://www.drupal.org/files/issues/2022-05-11/3206703-10.patch

The highlighted part below isn't present in what got included in 8.x-1.5;

image

https://git.drupalcode.org/project/entity/-/blob/8.x-1.5/src/EntityViewsData.php?ref_type=tags#L68

mstenta:
Oops! My mistake... I confused these two issues:

- https://www.drupal.org/project/entity/issues/3190436
- https://www.drupal.org/node/3206703

The former was merged into `entity` 1.5, but the patch from the latter is the one we're using in farmOS.

It looks like there is a fair bit of overlap in the two, though, so this probably means we'll need to reroll our patch.

Oh well.

mstenta:
Ahh @paul121 already explained this overlap in his comment on the latter issue:

> Adding a patch that includes the change from #3190436: EntityViewsData Column information not available for computed entity reference fields. The two patches are conflicting and we need both of them for farmOS

https://www.drupal.org/project/entity/issues/3206703#comment-14056032

So we were basically getting a two-for-one patch. :-)

mstenta:
Converted back to draft. It might make sense to close and open a new PR, but I'll leave this here for now as a reminder that we need to reroll our patch.

paul121:
I've rerolled our patch! https://www.drupal.org/project/entity/issues/3206703#comment-15751768

Force pushed here to run tests

mstenta:
Force pushed to give you commit credit @paul121 - thanks for this!

PR #869, opened 2024-08-27
Do not strip html tags when encoding csv exports

paul121:
This fixes the "notes text truncated in export" issue identified here: https://github.com/Rothamsted-Ecoinformatics/farm_rothamsted/issues/733#issuecomment-2313106176

At first I thought it was truncating based on text length. Turns out it's actually because there is a single `<` in the notes text this is considered an "open" html tag that is never closed, and thus all text following the `<` gets truncated. Other special characters seem to work, it's just these html tags.

This also means that all valid or intentional HTML tags will get stripped from the export. This seems like a significant bug, too?

Turns out there is a simple fix, passing the `'strip_tags' = FALSE` option via `csv_settings` (it defaults to TRUE). However... this then means that the export is wrapped in HTML paragraphs by default, as documented by the default text format's help text:

> Lines and paragraphs break automatically.

text_format

### Example

Notes test
notes

Strip tags = TRUE
html_strip

Strip tags = FALSE
keep_html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [Update drupal/consumers to ^1.19 and fix kernel API tests](https://github.com/farmOS/farmOS/pull/857)
 - [When a quantity is deleted, clean up log references to it #859](https://github.com/farmOS/farmOS/pull/859)
 
+### Security
+
+- [Sanitize CSV exports against formula injection #871](https://github.com/farmOS/farmOS/pull/871)
+
 ## [3.2.3] 2024-07-21
 
 ### Fixed
--- a/modules/core/csv/src/Normalizer/TextLongFieldItemNormalizer.php
+++ b/modules/core/csv/src/Normalizer/TextLongFieldItemNormalizer.php
@@ -21,9 +21,9 @@ class TextLongFieldItemNormalizer extends FieldItemNormalizer {
   public function normalize($field_item, $format = NULL, array $context = []): array|string|int|float|bool|\ArrayObject|NULL {
     /** @var \Drupal\text\Plugin\Field\FieldType\TextLongItem $field_item */
 
-    // Return processed text, if desired.
-    if (isset($context['processed_text']) && $context['processed_text'] === TRUE) {
-      return $field_item->get('processed')->getValue();
+    // Return raw text, if desired.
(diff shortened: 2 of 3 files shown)


paul121:
> However... this then means that the export is wrapped in HTML paragraphs by default, as documented by the default text format's help text:
Lines and paragraphs break automatically.

Wow, I didn't realize all the complexity that comes with the text formats. All of this behavior is configurable:

Screenshot from 2024-08-27 10-58-06
Screenshot from 2024-08-27 10-58-15

paul121:
Maybe a more generic solution would be to "export raw/unformatted text values". I assume we would need to override the field normalization for text/text long/etc fields to achieve this behavior. Unless maybe this already has a setting...

paul121:
> Unless maybe this already has a setting...

Aha. We introduce a similar setting and set this to TRUE:
```php
      // Return processed text from long text fields.
      'processed_text' => TRUE,
```

And I see why, trying `'process_text' = FALSE` and `'strip_tags' = FALSE` isn't great (I think we get both the processed and unprocessed?)
Screenshot from 2024-08-27 11-19-22

paul121:
Pushing a commit... I think we actually want to only serialize the raw `value` from `text_long` fields (and maybe other text fields that are formattable, too?), but also set `'strip_tags' = FALSE` to preserve raw HTML

Screenshot from 2024-08-27 11-31-23

mstenta:
Rebased this on top of #871, and added an option to "Export processed text", but defaults exports to raw user input.

@paul121 and I discussed the various options here and this seems like the most sane default. Notably, the JSON:API offers both processed and raw text values for long text fields, so this option gives users the choice. But it assumes that they want raw values. I explained in the form description that processed text may be desired if the exported data is going to be embedded directly in HTML (without any additional filtering).

Screenshot from 2024-08-31 08-52-21

The other big change here is that this PR is disabling `strip_tags` everywhere now. This is probably what we want, but I'm curious if this will cause any surprises. All of our testing up until now has had `strip_tags` enabled... so I wonder if there are places where we were benefiting from that without realizing it. It would be worth doing a bunch of test exports of various data types to see if any tags are sneaking in that we don't want.

symbioquine (approved):
No tests? Otherwise looks good to me.

mstenta:
> No tests?

https://github.com/farmOS/farmOS/pull/871#issuecomment-2353710606

PR #870, opened 2024-08-29
Simplify recommendation for XDEBUG_CONFIG in local development environments

mstenta:
We currently provide some guidance on how to set up XDebug in local development environments: https://farmos.org/development/environment/debug/

For browser-based debugging, it works fine to set the `XDEBUG_CONFIG: discover_client_host=yes` environment variable, which tells XDebug to automatically figure out how to connect to a client host outside of the Docker container for debugging purposes. However, this does not work for debugging code that is run via command-line (eg: `drush`), so we provide additional guidance on how to find the IP address of the client host manually and set `XDEBUG_CONFIG: client_host=192.168.128.1`.

This is not ideal, because the `client_host` IP address changes whenever Docker containers are rebuilt (every time you run `docker compose down` and then `docker compose up` again).

Docker provides a `host.docker.internal` hostname that resolves to the host machine's IP address, but it is only available on Mac and Windows hosts, and not on Linux. So we haven't recommended that as a solution, and instead opted for documenting how to update the IP address manually (for consistency across platforms).

I recently learned that we can add this `extra_hosts` configuration to set `host.docker.internal` on Linux hosts:

```
extra_hosts:
  - host.docker.internal:host-gateway
```

Adding the above to our `docker-compose.development.yml` file allows us to use `XDEBUG_CONFIG: client_host=host.docker.internal` regardless of the host operating system, and thus our XDebug documentation can be simplified!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - [Single Dockerfile using multi-stages for production and dev image #855](https://github.com/farmOS/farmOS/pull/855)
+- [Simplify recommendation for XDEBUG_CONFIG in local development environments #870](https://github.com/farmOS/farmOS/pull/870)
 
 ### Fixed
 
--- a/docker/docker-compose.development.yml
+++ b/docker/docker-compose.development.yml
@@ -18,9 +18,11 @@ services:
       - './www:/opt/drupal'
     ports:
       - '80:80'
+    extra_hosts:
+      - host.docker.internal:host-gateway
     environment:
       XDEBUG_MODE: debug
-      XDEBUG_CONFIG: discover_client_host=yes
+      XDEBUG_CONFIG: client_host=host.docker.internal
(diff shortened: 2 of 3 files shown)


paul121 (approved):
It seems that DDEV uses this same approach: https://ddev.readthedocs.io/en/stable/users/debugging-profiling/step-debugging/#troubleshooting-xdebug

> ddev logs may show you something like Xdebug: [Step Debug] Could not connect to debugging client. Tried: host.docker.internal:9003 (fallback through xdebug.client_host/xdebug.client_port) :-(. If it does, it may mean that your firewall is blocking the connection, or in a small number of cases that host.docker.internal is not figured out successfully by DDEV or Docker.

symbioquine:
Found a blog post explaining this work-around: https://medium.com/@TimvanBaarsen/how-to-connect-to-the-docker-host-from-inside-a-docker-container-112b4c71bc66

symbioquine (approved):
LGTM!

mstenta:
Thanks for the reviews @paul121 and @symbioquine!

PR #871, opened 2024-08-30
Sanitize CSV exports against formula injection

mstenta:
This adds some basic protection against CSV injection exploits when exporting CSVs from farmOS containing data from untrusted/malicious users.

See:

- https://owasp.org/www-community/attacks/CSV_Injection
- https://georgemauer.net/2017/10/07/csv-injection.html

Note that this is NOT considered a security vulnerability of farmOS itself. It is an existing attack vector in spreadsheet applications like Microsoft Excel. Similar issues have been filed against Drupal modules in the past, and the Drupal security team has come to that conclusion. Reference: https://www.drupal.org/project/webform/issues/2788591

> Spreadsheets are vulnerable to hidden formulas in CSV files. If webform data is submitted by untrusted users, and the submissions are downloaded in CSV format, maliciously-formatted submission data may be used to create a security vulnerability. The Drupal security team has concluded that this vulnerability does not rest in webform, but rather in the spreadsheets. Nonetheless, a warning about opening downloaded submissions in CSV format with spreadsheets should be added to the webform download page.

This pull request take a pro-active step by prepending cells that begin with suspicious characters (=, @, +, -) with a tab character, as described in https://georgemauer.net/2017/10/07/csv-injection.html

The plan is to merge this as "default behavior" in farmOS first, and then add an option (with a visible warning) that allows users to disable it. Thus we'll be able to offer a base level of protection by default, but also allow raw export of text data when necessary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - [Update Drupal core to 10.3 #872](https://github.com/farmOS/farmOS/pull/872)
 
+### Security
+
+- [Sanitize CSV exports against formula injection #871](https://github.com/farmOS/farmOS/pull/871)
+
 ## [3.2.4] 2024-09-18
 
 ### Changed
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         "drupal/core": "10.3.5",
         "drupal/config_update": "^2.0@alpha",
         "drupal/consumers": "^1.19",
-        "drupal/csv_serialization": "^4.0",
+        "drupal/csv_serialization": "^4.0.1",
         "drupal/date_popup": "^1.3",
         "drupal/entity": "1.5",
         "drupal/entity_browser": "^2.10",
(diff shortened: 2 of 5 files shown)


mstenta:
> The plan is to merge this as "default behavior" in farmOS first, and then add an option (with a visible warning) that allows users to disable it. Thus we'll be able to offer a base level of protection by default, but also allow raw export of text data when necessary.

It was very easy to add this option as a checkbox during export, so I went ahead and did that too. Force-pushed my commit. See screenshot below for what it looks like.

Screenshot from 2024-08-30 14-45-15

paul121 (approved):

mstenta:
Force pushed to set our `composer.json` version constraint for `"drupal/csv_serialization": "^4.0.1",`, since the ability to override CSV settings was introduced in 4.0.1 (via https://www.drupal.org/project/csv_serialization/issues/3192772).

symbioquine (approved):
No tests? Otherwise looks good to me.

mstenta:
> No tests?

Fair point. Unfortunately, we don't currently have any tests of the CSV exporters. So adding tests for this is not trivial. IMO we should take a pass at adding basic tests for CSV exports generally, and include tests for this functionality at the same time. But realistically that 's a non-trivial lift and won't happen before v3.3 is released. I don't want to wait to add this, though. Admittedly not ideal. :-/

PR #872, opened 2024-08-31
Update Drupal core to 10.3

mstenta:
This PR makes all the necessary changes to upgrade farmOS to Drupal 10.3.x. Some notable changes include:

**Use new `permission_checker` service for `farm_role` permission logic**

Some tests related to our `farm_role` module were failing, and it turned out that none of the "managed roles" were receiving the permissions they normally would have. I didn't end up figuring out the exact cause of this, but in the process I discovered that Drupal 10.2 introduced a new `permission_checker` service (see https://www.drupal.org/node/3348054). So I converted our `farm_role` module to use that (instead of overriding the `user_role` entity definition's storage class), and that fixed the tests, so I left it at that. Aside from changing from an entity storage class to a service class, the actual PHP logic is not much different.

**Views Entity Reference filters patch**

Drupal core has merged [Issue #3347343: Add Views EntityReference filter to support better UX for exposed filters](https://www.drupal.org/project/drupal/issues/3347343) into Drupal 11 upstream, and they are working on a backport to Drupal 10.4. The merged changes differ slightly from the earlier patch we were using, but the end result is the same. The core maintainers decided that it would be a breaking change if they took the approach from the patch(es) we were previously using. The existing (and still default) Views entity reference filter only works with entity IDs (not selects or autocompletes), and the previous patches replaced this with select/autocomplete. This means that any existing sites that depended on entity ID filtering would break. So, the core maintainers decided to add the select/autocomplete filter as an optional plugin, which will be selectable in the Views UI once some other core issues are merged (specifically [#3458099: Views handler loading should respect configuration](https://www.drupal.org/project/drupal/issues/3458099) and [#3438054: Configurable views filters to allow for different widgets](https://www.drupal.org/project/drupal/issues/3438054)). In the meantime, a contrib module has been published which provides the ability for site administrators to opt-in to using the new filter ([Views Core Entity Reference](https://www.drupal.org/project/views_core_entity_reference)). The alternative to using this module is to perform the same logic it does manually via `hook_views_data_alter()`. This is the approach that this PR takes via our existing `hook_views_data_alter()` implementation in the `farm_ui_views` module. Once the remaining two core issues are merged, we will be able to remove that and configure the filter directly in individual Views configuration YML.

Notably, the patch we are applying for 10.3 is being included directly in the farmOS codebase (in a new `patches` directory), because there isn't one available in the issue itself, and I didn't want to add more noise to an already very long core issue. I produced the patch by saving the diff of the [3347343-10-3-x--comment-214-onwards--do-not-merge](https://git.drupalcode.org/issue/drupal-3347343/-/tree/3347343-10-3-x--comment-214-onwards--do-not-merge) MR branch: https://git.drupalcode.org/project/drupal/-/merge_requests/9036.diff

**`EntityTaxonomyTermReferenceArgument` changes**

(copied directly from the commit message)

> Originally we were extending Drupal\taxonomy\Plugin\views\argument\Taxonomy,
which was a pass-through extension of Drupal\views\Plugin\views\argument\NumericArgument
in Drupal 10.2.x. In Drupal 10.3.x, the Taxonomy class now extends from a new
Drupal\taxonomy\Plugin\views\argument\EntityArgument class. EntityArgument extends from
NumericArgument, but injects new dependencies that we don't need. So this simply
changes our EntityTaxonomyTermReferenceArgument class to extend from NumericArgument
directly.

**Deprecation fixes**

A few changes were made to avoid using deprecated base classes and services. See commits for more details.

**Fix failing tests**

There were a few minor changes that caused tests to fail. Nothing too big. See commits for more details.

**Config entity YML updates**

I took the opportunity to re-export a bunch of our core configuration entity YML files, since a lot of them have been overridden for a while. My process was basically to install farmOS with all modules, enable the `config_update_ui` module, and then review and export YMLs from the list of changed config.


modules/core/role/src/ManagedRolePermissionChecker.php
```diff
@@ -0,0 +1,69 @@
...
+   * {@inheritdoc}
+   */
+  public function hasPermission(string $permission, AccountInterface $account): bool {
+    $has_permission = parent::hasPermission($permission, $account);
+
+    // Check if the permission is included via farm_role rules.
+    if (!$has_permission) {
+      foreach ($account->getRoles() as $role_id) {
+        $managed_roles = $this->managedRolePermissionsManager->getMangedRoles();
+        if (in_array($role_id, array_keys($managed_roles))) {
```
paul121: Only load the `$managed_roles` once:

```suggestion
    if (!$has_permission) {
      $managed_roles = $this->managedRolePermissionsManager->getMangedRoles();
      foreach ($account->getRoles() as $role_id) {
        if (in_array($role_id, array_keys($managed_roles))) {
```

modules/taxonomy/animal_type/config/install/taxonomy.vocabulary.animal_type.yml
```diff
@@ -4,7 +4,8 @@ dependencies:
   enforced:
     module:
       - farm_animal_type
-name: Animal type
+name: 'Animal type'
 vid: animal_type
 description: 'A list of animal species/breeds.'
 weight: 0
+new_revision: false
```
paul121, line 11: IIRC taxonomy terms have revisions, it's just that the UI doesn't fully support or show them. But in line with our other entity types, would it be good to enable revisions by default for taxonomy terms?

Perhaps this would require an update hook and should happen in a separate PR too.

paul121 (changes requested):

modules/core/role/src/ManagedRolePermissionChecker.php
mstenta: Good call. I'll push a commit and give you credit.

mstenta: https://github.com/farmOS/farmOS/pull/872/commits/ae7080ab0e24544cb4a06fafcb2016d3c8182615

modules/taxonomy/animal_type/config/install/taxonomy.vocabulary.animal_type.yml
mstenta: Good point! I agree we should enable revisions by default. This does seem like a separate PR, though. The change made in this PR is simply exporting the current configuration to YML, to make explicit what (I assume) is currently the default (no revisions).

IMO it doesn't need to block this PR, and we can decide separately if we want to try to prioritize that change for v3.3.0.

mstenta: (Also curious: do term revisions capture the author of the change? AFAIK terms do not have any authorship information associated with them. That is arguably one of the most important things about revisions: knowing *who* made the change. Knowing *when* is also useful though... :thinking:)

symbioquine (approved):
I haven't wrapped my head around that huge patch file's diff, but everything else seems reasonable (as long as it works :stuck_out_tongue_closed_eyes:).

mstenta:
> I haven't wrapped my head around that huge patch file's diff

Yea I don't love including that in the repo, to be honest... it's big and we shouldn't have to bloat our repo with upstream code.

But for what it's worth, the core patch is already merged into Drupal 11, so this is simply back-porting the functionality to 10.3. We were already applying/updating the core patch for this while the community was ironing out the final changes and tests. The only difference now is that this PR is including the patch file directly in the farmOS codebase, rather than linking to an external URL for the patch file like we normally do. I did this because there isn't a linkable patch file in the core issue for Drupal 10.3. They committed it to 11.x, and *might* officially backport it to 10.4, but for 10.3 we still need a patch. We could technically link to the patch file generated by GitLab but I don't trust those because anyone with push access to the merge request on drupal.org can change it at any time.

mstenta:
Rebased onto `3.x` and added a commit to update to Drupal core 10.3.5.

Release notes for the three Drupal core releases that happened since 10.3.2:

- https://www.drupal.org/project/drupal/releases/10.3.3
- https://www.drupal.org/project/drupal/releases/10.3.4
- https://www.drupal.org/project/drupal/releases/10.3.5

paul121 (approved):

mstenta:
This was mistakenly merged ahead of schedule. I rolled it back and force pushed 3.x. The *plan* is to merge it the way that I did... but I didn't mean to do that right now. I'll leave this closed though... hopefully will be merging everything very soon.

paul121 (commented):
I think I might have found an issue with our refactor to the new `permission_checker` service: https://github.com/farmOS/farmOS/pull/872/commits/f7e123ac764b443b971a54282ba68ca098d9912d

While working on #875 with these changes (in the WIP 3.3.x branch) I was experiencing some weird behavior where I didn't have any of the permissions I expected to have (my API user is not root and has the `farm_manager` role).

Doing a little debugging I added a simple test to check if the account had a permission:

```php
$test = $account->hasPermission('administer plans');
```

and realized that the new `ManagedRolePermissionChecker::hasPermission()` function wasn't being called.

Looking closer into the call stack I came to Simple OAuth... I'm afraid what's happening is that Simple OAuth still uses the `user_role` storage class to check `$storage->isPermissionInRoles($permission, $roles)`: https://git.drupalcode.org/project/simple_oauth/-/blob/72d1960624a9b05ac743e1223a821b5690eb4fb0/src/Oauth2ScopeProvider.php#L75-108

After copy pasting `FarmRoleStorage.php` back into my 3.3.x environment & enabling it for the `user_role` entity type, the permissions started working as expected via API w/ Simple OAuth authorization (I'm using CURL fwiw).

---

Sooo.....
1) We need to figure out why our existing tests didn't catch this. I assume we must not have any tests that combine the `farm_api` module + a user authorized w/ simple OAuth using a "static scope" connected to one of our managed roles like `farm_manager` ?
1) We might still need our `FarmRoleStorage` class
1) Eventually Simple oauth should probably be updated to use the new `permission_checker` from 10.3? I didn't see any open issues considering this

mstenta:
Wow good find @paul121 - that's a bummer. It makes me wonder if there are other places where `user_role` storage class is still being used.

> We need to figure out why our existing tests didn't catch this. I assume we must not have any tests that combine the `farm_api` module + a user authorized w/ simple OAuth using a "static scope" connected to one of our managed roles like `farm_manager` ?

Agreed. We should add some generic tests for API access control that would catch this sort of thing.

> We might still need our `FarmRoleStorage` class

Yea. Maybe we should just add it back as a shim, but make it use the new `permission_checker` service?

> Eventually Simple oauth should probably be updated to use the new `permission_checker` from 10.3? I didn't see any open issues considering this

Yea we should open an issue at the very least. If we can contribute (and start using) a patch, that would be the best.

mstenta:
Notes from investigating this more this morning...

Drupal core still provides a `RoleStorage::isPermissionInRoles()` function, but it doesn't seem that it is called by any core code. The only thing that uses it in the farmOS codebase (outside of Drupal core) is the `simple_oauth` module's `OAuth2ScopeProvider`. This explains why a) tests started failing until we refactored to override the new `permission_checker` service, and why @paul121 discovered that OAuth2 scopes are not receiving the permission.

It looks like [Issue #3347873: Centralize permission checks in a service](https://www.drupal.org/project/drupal/issues/3347873) (introduced in 10.2) originally kept the call to `isPermissionInRoles()` in the new `PermissionChecker::hasPermission()` method (see diff: https://git.drupalcode.org/project/drupal/-/commit/085ae2c2fb36c0d0b91645c0e886b7aa191b216b).

But then [Issue #3376846: Implement the new access policy API ](https://www.drupal.org/project/drupal/issues/3376846) (introduced in 10.3) refactored `PermissionChecker::hasPermission()` and removed the call to `RoleStorage::isPermissionInRoles()` (see diff: https://git.drupalcode.org/project/drupal/-/commit/9bb46296cee3aa33a9750633b211055caee316d1).

I'm curious how Drupal core is checking if a permission is in a role, if it isn't using `RoleStorage::isPermissionInRoles()` anymore. I need to dig deeper into the new access policy frramework code... :thinking: 

> Eventually Simple oauth should probably be updated to use the new `permission_checker` from 10.3? I didn't see any open issues considering this

This may be a little conceptually problematic, now that I look at it...

The new `permission_checker` service only provides a single method for checking permissions, and it requires that the user account be provided: `hasPermission(string $permission, AccountInterface $account)`

However, in the context of `Oauth2ScopeProvider::scopeHasPermission(string $permission, Oauth2ScopeInterface $scope)`, there is no user account... only `string $permission` and `Oauth2ScopeInterface $scope`.

This gets at a larger question... `simple_oauth` is concerned with checking if a *role* has a permission, but Drupal core's new `permission_checker` is concerned with checking if a *user* has a permission. This is a subtle distinction, but an important one in this case.

It may mean `simple_oauth` shouldn't actually use the `permission_checker` service... but maybe it should use some lower level access policy methods... :thinking:

mstenta:
> I'm curious how Drupal core is checking if a permission is in a role, if it isn't using `RoleStorage::isPermissionInRoles()` anymore. I need to dig deeper into the new access policy frramework code... 🤔

Drupal core now uses an `access_policy.user_roles` service (class: `Drupal\Core\Session\UserRolesAccessPolicy`) with a `calculatePermissions()` method that runs `$user_role->getPermissions()`.

See: https://git.drupalcode.org/project/drupal/-/blob/694cb802f42f08b44a791fc7d94827802a82c955/core/lib/Drupal/Core/Session/UserRolesAccessPolicy.php#L27

> Drupal core still provides a `RoleStorage::isPermissionInRoles()` function, but it doesn't seem that it is called by any core code.

I opened a core issue to consider deprecating/removing `RoleStorage::isPermissionInRoles()`: https://www.drupal.org/project/drupal/issues/3476166

It seems to me that `isPermissionInRoles()` is vestigial code now that core access policy logic is implemented. Any contrib code that is still using it (like `simple_oauth`) should probably be refactored. We'll see if the core maintainers agree... if so, then `simple_oauth` should probably use similar logic to `UserRolesAccessPolicy::calculatePermissions()` (eg: `$role->hasPermission()`).

Given those assumptions, how should `farm_role` work, ideally?

Maybe overriding the `permission_checker` service isn't the right approach. But neither is using `RoleStorage::isPermissionInRoles()` (since the core `permission_checker` doesn't use it). 

What if we overrode `RoleStorage`, but used the `doLoadMultiple()` method to add permissions to the `Role::$permissions` array whenever a role entity is loaded? :thinking: 

In theory, these permissions would be picked up by both the new `permission_checker` service, and the existing `RoleStorage::isPermissionInRoles()` (which uses `$role->hasPermission($permission)`). So in theory, if `RoleStorage::isPermissionInRoles()` goes away in the future, and `simple_oauth` is refactored accordingly, it *should* still work.

Would this cause any issues? It would essentially mean that the permissions are stored in each role's config entity itself, at least during runtime. That means they would also get saved to the role YML if it were exported. I don't think that's desirable, but curious your thoughts @paul121.

mstenta:
> Would this cause any issues?

I gave this a quick try, just to see if it would work. It did work to grant permissions when roles are loaded. So after reverting the `permission_checker` commit, and replacing `FarmRoleStorage::isPermissionInRole()` with `FarmRoleStorage::doLoadMultiple()`, I was able to use farmOS as a Viewer role and everything worked as expected.

However, there's logic in the `farm_role_account_admin` module that causes an infinite loop under certain conditions, because it attempts to load all roles here: https://github.com/farmOS/farmOS/blob/46e497f3c980197870496a5d65ba6c0af9963ecf/modules/core/role/modules/account_admin/src/AccountAdminPermissions.php#L72

This combined with the fact that this approach also adds considerations around exporting role config makes me think it's not a good path to continue down...

mstenta:
I tested a slightly different approach. Rather than replacing the `storage` handler class for `user_role` entities, I tested replacing the entity `class` itself (providing our own `Drupal\farm_role\ManagedRole` class as a replacement for the core `Drupal\user\Entity\Role` class). The `ManagedRole` class simply overrides the parent's `getPermissions()` and `hasPermissions()` methods:

```
class ManagedRole extends Role {

  /**
   * {@inheritdoc}
   */
  public function getPermissions() {
    return $this->permissions + \Drupal::service('plugin.manager.managed_role_permissions')->getManagedPermissionsForRole($this);
  }

  /**
   * {@inheritdoc}
   */
  public function hasPermission($permission) {
    return parent::hasPermission($permission) || \Drupal::service('plugin.manager.managed_role_permissions')->isPermissionInRole($permission, $this);
  }

}
```

With this approach, we don't need to override the `permission_checker` service, and `simple_oauth` automatically works because it ends up using Drupal core's `RoleStorage::isPermissionInRoles()`, which uses `$role->hasPermission()`.

This feels like an elegant solution... and should continue to work even if `simple_oauth` moves away from using `RoleStorage::isPermissionInRoles()`. Presumably they would use `$role->hasPermission()`, like core's new `UserRolesAccessPolicy`.

There's one thing I don't like about it, though: our `ManagedRole` class needs to copy all of the class annotations from core's `Role` class. Those aren't inherited. This is a big chunk of duplicate code, and if something changes in the upstream annotations we won't necessarily know about it. So it's a code maintenance consideration...

mstenta:
Discussed all of this with @paul121 today. We decided to restore the <code>FarmRoleStorage</code> class to fix the issue, and we made a plan for a more complete refactoring as a next step.

Here is the fix PR: #878

And here is the next step: https://www.drupal.org/project/farm/issues/3476250

modules/core/ui/views/farm_ui_views.views_execution.inc
paul121, line 1: Note that this is somewhat a breaking change.... In the Rothamsted modules we provide some views with an entity reference filter with specific handler/handler settings and after these changes they default back to default handler settings. @mstenta might be worth mentioning in the changelog/release notes?

I was alerted to this because I got schema errors that break the automated tests:

> Drupal\Core\Config\Schema\SchemaIncompleteException: Schema errors for views.view.rothamsted_uncategorized_logs with the following errors: views.view.rothamsted_uncategorized_logs:display.default.display_options.filters.owner_target_id.handler missing schema,

paul121: Had to make the same views changes as the commit included in this core PR: https://github.com/farmOS/farmOS/pull/872/commits/8beda88c1c75812e58686ca7dde6c4acaa07a500

Once I got all of the `entity_reference` filters fixed in my contrib module tests are passing again: https://github.com/Rothamsted-Ecoinformatics/farm_rothamsted/pull/758/commits/f897317205cbc7bd5e78f767a143533be1ac147f

Without tests it would be hard to know if a site is affected. It would be nice to make a script to check if there are any affected views on a given install or code base?

paul121: Maybe we could include a 'status report' check for this?

PR #874, opened 2024-09-13
Update authentication.md - Refresh Tokens

Farmer-Eds-Shed:
Bearer Header and client secret not needed for refresh token

### DIFF
--- a/docs/development/api/authentication.md
+++ b/docs/development/api/authentication.md
@@ -161,10 +161,10 @@ has expired.
 It is a one step process:
 
 The client sends an authenticated request to the `/oauth/token`endpoint with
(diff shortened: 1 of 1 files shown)


mstenta:
Thanks @Farmer-Eds-Shed! Tagging @paul121 for review since he wrote these docs.

paul121 (approved):
Sorry for the delay here @Farmer-Eds-Shed ... I haven't tested, but seems correct and trust you if you tested!

https://datatracker.ietf.org/doc/html/rfc6749#section-6

The spec seems to be a little generic on these points tho, especially re: `Authorization` header. It makes sense that this isn't necessary though. I wish `simple_oauth` had better docs on this but they just point to the spec.

PR #877, opened 2024-09-19
Configure Inline Entity Form widget to automatically delete entities

mstenta:
I noticed recently that when you delete a quantity through the farmOS UI, you have to explicitly tell it to delete the quantity entity and not just the reference to it.

Screenshot from 2024-09-19 10-47-32

This feature was added to Inline Entity Form in the [8.x-1.0-rc15](https://www.drupal.org/project/inline_entity_form/releases/8.x-1.0-rc15) release: https://www.drupal.org/project/inline_entity_form/issues/2875716

In farmOS, we always want to delete quantity entities when the reference to them is deleted, otherwise it results in orphaned quantities (see related: #775). This PR configures the new Inline Entity Form widget setting to ensure that quantities are deleted.

It also makes the same change for `data_stream` reference widgets (the only other Inline Entity Form widget we provide default configuration for in our `farm.field_factory` logic. @paul121 does this make sense? I forget what all the possible use-cases of data stream references might be... If not, we can set that one to `REMOVED_OPTIONAL`.

### DIFF
--- a/modules/core/field/src/FarmFieldFactory.php
+++ b/modules/core/field/src/FarmFieldFactory.php
@@ -6,6 +6,7 @@ use Drupal\Core\Field\BaseFieldDefinition;
 use Drupal\Core\Field\FieldException;
 use Drupal\Core\Field\FieldStorageDefinitionInterface;
 use Drupal\entity\BundleFieldDefinition;
+use Drupal\inline_entity_form\Plugin\Field\FieldWidget\InlineEntityFormComplex;
 
 /**
  * Factory for generating farmOS field definitions.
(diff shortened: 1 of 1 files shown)


mstenta:
Here is what it looks like with this change:

Screenshot from 2024-09-19 10-48-51

paul121 (approved):
This a great addition! It should be fine to do this for data streams as well. Right now you can't manually create data streams from the UI, they are always associated with a sensor asset, and can't really be navigated to without using the field listing from the sensor asset page. In the future if they were to operate "independently" from sensor assets then we might want to revisit this.

paul121:
Oh minor thing, but maybe we should increase the IEF dependency to rc15?

mstenta:
> Oh minor thing, but maybe we should increase the IEF dependency to rc15?

Currently we aren't pinning IEF and have the version constraint set to `^1.0@RC`, and I'm not sure Composer supports `^1.0-rc15` syntax. I think I recall trying to do that in the past and it didn't work. I think it's OK to leave it at `^1.0@RC` for this. The setting will either work for installations with `1.0-rc15` or greater, or have no effect.

On a related note, it looks like IEF has a newer 3.x branch, which we should look at converting to (separate topic/PR).

PR #881, opened 2024-09-26
Remove dependency on admin_toolbar:admin_toolbar_links_access_filter

mstenta:
I discovered an issue with the upgrade path to farmOS 3.3.0. I'm still figuring out exactly why this is happening (and why it didn't happen in local testing), but basically the `admin_toolbar:admin_toolbar_links_access_filter` has been deprecated, and our `farm_ui_menu` module depends on it. For some reason, during the update, that module gets uninstalled automatically, which then uninstalls `farm_ui_menu` (which depends on it), and `farm_ui` (which depends on `farm_ui_menu`, along with some other `farm_ui_*` modules (which also depend on `farm_ui_menu`). After that , the upgrade throws an error when it is trying to revert Views configuration that no longer exists.

Removing the dependency on `admin_toolbar:admin_toolbar_links_access_filter` seems to fix the issue. I tested two site upgrades with the dependency, and they both broke. I tested one without the dependency and it worked.

More details to come... just wanted to get this PR opened to share with people ASAP.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- [Remove dependency on admin_toolbar:admin_toolbar_links_access_filter #881](https://github.com/farmOS/farmOS/pull/881)
+
 ## [3.3.0] 2024-09-24
 
 ### Added
--- a/modules/core/ui/menu/farm_ui_menu.info.yml
+++ b/modules/core/ui/menu/farm_ui_menu.info.yml
@@ -5,5 +5,4 @@ package: farmOS UI
 core_version_requirement: ^10
 dependencies:
   - admin_toolbar:admin_toolbar
-  - admin_toolbar:admin_toolbar_links_access_filter
   - drupal:toolbar


mstenta:
Apparently the `admin_toolbar` module added an update hook that uninstalls the module :facepalm: : https://git.drupalcode.org/project/admin_toolbar/-/merge_requests/85/diffs

paul121:
FWIW when upgrading to 3.3 my site got in a broken state, specifically while running updates via `drush updb`. I was able to resolve by:

```shell
# Complete unattempted updates
drush updb -y

# Uninstall farm_ui_dashboard (for some reason this module wasn't uninstalled when farm_ui was uninstalled?)
drush pmu farm_ui_dashboard

# Re-install all farm_ui modules
drush en farm_ui
```

mstenta:
I understand why this did not come up in local testing...

The way I tested was:

1. I checked out the old version of farmOS.
2. I ran `composer update`.
3. I installed farmOS via Drush.
4. I checked out the new version of farmOS.
5. I ran `composer update`.
6. I ran database updates via `drush updb`.

In step 3, farmOS was installed with the `admin_toolbar` module and the deprecated `admin_toolbar_links_access_filter` module, because both are dependencies of `farm_ui_menu`. It was the new version of `admin_toolbar`, with the additional `admin_toolbar_update_8003()` update hook... but because it was installed from scratch that update hook never ran, and Drupal considers it unnecessary in future updates. So in steps 5 and 6, `admin_toolbar` didn't get updated, and the update hook did not run. Thus the issue did not occur in my local testing.

paul121 (approved):

PR #888, opened 2024-11-18
Fix group location logic

### LINKED ISSUE #886: Creating old movement logs disrupts current location logic
Open an asset that has a location. Create a movement log in the past. The current location is updated to this old log location not the actual location. 
Screen recording for clarity. https://drive.google.com/file/d/11pesiLqE-I-nr9-K-mEzbNcthuNNu1n5/view?usp=sharing

version 3.2.1

mstenta:
Fixes #886 (included a regression test)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - [Fix config export via UI #883](https://github.com/farmOS/farmOS/pull/883)
+- [Fix group location logic #888](https://github.com/farmOS/farmOS/pull/888)
 
 ### Security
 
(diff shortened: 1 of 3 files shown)


symbioquine (approved):
LGTM - kind of amazing we didn't run into this before now

mstenta:
> kind of amazing we didn't run into this before now

Yea it's actually a very specific state that causes it, which is kind of unlikely, so not too surprising.

It only happens if you have an asset movement log with a larger log ID (but not necessarily a larger timestamp) than the latest group movement log. But if there is another more recent asset movement log, regardless of it's ID, then it doesn't happen.

This is because the logic basically finds the "latest asset movement log" as a first step (using the `farm_location` module's default logic). Then it loads all the "latest group movement logs" for each group that the asset is in, and iterates over them to check to see if they should take precedence over the "latest asset movement log". The "should take precedence" logic is what was broken, because it was checking if the group movement log's ID was greater than the asset movement log's ID, which in hindsight doesn't make any sense at all... but works most of the time, so it slipped through. :shrug:

PR #892, opened 2024-11-19
Update Drupal core to 10.3.9

mstenta:
Multiple security advisories fixed by 10.3.9. None of them are directly relevant to farmOS, because they require conditions not present in farmOS. In theory a user who has customized farmOS may be affected, and should therefore review the advisories.

More information: https://www.drupal.org/project/drupal/releases/10.3.9


mstenta:
A 10.3.9 security release just came out, so updating this and adding a changelog line...

PR #894, opened 2024-11-19
Add Equipment type field to Equipment assets

mstenta:
This adds a new "Equipment type" taxonomy, with a new `equipment_type` term reference field on `equipment` assets. This allows categorizing equipment. It follows the same pattern as `animal_type`, `material_type`, `plant_type`, and `product_type`.

Forum topic for general discussion (reserve this PR for technical review): https://farmos.discourse.group/t/equipment-types/2158

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- [Add Equipment type field to Equipment assets #894](https://github.com/farmOS/farmOS/pull/894)
+
 ### Changed
 
 - [Pin PHP version to 8.3 in Dockerfile #923](https://github.com/farmOS/farmOS/pull/923)
--- a/docs/model/type/asset.md
+++ b/docs/model/type/asset.md
@@ -254,6 +254,10 @@ Equipment Assets have the following additional attributes:
 - Model (string)
 - Serial number (string)
 
+And the following additional relationships:
+
+- Equipment type (References a Term in the "Equipment type" vocabulary)
+
 #### Land Assets
 
 Land Assets have the following additional attributes:
--- a/docs/model/type/term.md
+++ b/docs/model/type/term.md
@@ -17,6 +17,7 @@ in. Vocabularies are defined by modules, and are only available if their module
 is enabled. The modules included with farmOS define the following vocabularies:
 
 - Animal type
+- Equipment type
 - Log category
 - Material type
 - Plant type
--- a/modules/asset/equipment/farm_equipment.info.yml
+++ b/modules/asset/equipment/farm_equipment.info.yml
@@ -6,3 +6,4 @@ core_version_requirement: ^10
 dependencies:
   - farm:asset
   - farm:farm_entity
+  - farm:farm_equipment_type
--- a/modules/asset/equipment/farm_equipment.post_update.php
+++ b/modules/asset/equipment/farm_equipment.post_update.php
@@ -1,0 +1,35 @@
+<?php
+
+/**
+ * @file
+ * Post update hooks for the farm_equipment module.
+ */
+
+declare(strict_types=1);
+
+/**
+ * Add "Equipment type" field to Equipment assets.
+ */
+function farm_equipment_post_update_install_equipment_type(&$sandbox) {
+
+  // Install the farm_equipment_type taxonomy module.
+  if (!\Drupal::service('module_handler')->moduleExists('farm_equipment_type')) {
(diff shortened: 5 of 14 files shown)


Fosten:
Looks great!

This was discussed a few months ago on a dev call.
https://github.com/Fosten/farm_equipment_asset_extended

At that time, Paul brought up adding
      'multiple' => TRUE,

Otherwise, looks similar.  Thank you!

mstenta:
Oh wow thanks @Fosten I completely forgot about your module!

If we do decide to merge this PR I'd be happy to help figure out a graceful upgrade path for `farm_equipment_asset_extended`. It looks like it could just be deprecated? Just want to make sure the handoff off the taxonomy and field definitions to farmOS core is smooth.

> At that time, Paul brought up adding 'multiple' => TRUE,

Ah great point. We should definitely allow multiple equipment types. I copied/modified from the [Animal Type field](https://github.com/farmOS/farmOS/blob/9b9d979f15fddbc9cdcf793c49503d500fe183b3/modules/asset/animal/src/Plugin/Asset/AssetType/Animal.php#L23) (which does not allow multiple), so I missed that.

I'll push a fixup commit to add that...

mstenta:
> Just want to make sure the handoff off the taxonomy and field definitions to farmOS core is smooth.

So actually, I think all you would need to do is delete these two files:

- https://github.com/Fosten/farm_equipment_asset_extended/blob/main/config/install/taxonomy.vocabulary.equipment_type.yml
- https://github.com/Fosten/farm_equipment_asset_extended/blob/main/farm_equipment_asset_extended.module

Then deploy that change at the same time as the new version of farmOS that includes this PR. Then you could just uninstall `farm_equipment_asset_extended` and farmOS would take over the responsibilities.

mstenta:
Tests are failing because of #895 - but it makes me realize that the added `farm_equipment_type` dependency may need to be added to other kernel tests as well. Even if they pass without it now, it could cause future issues/confusion. Setting this back to draft until that's done...

mstenta:
> So actually, I think all you would need to do is delete these two files...

@Fosten and I talked about this on the dev call and realized that there's still a potential conflict if `farm_equipment_asset_extended` has been installed, because the taxonomy and field will both already exist, so when the update hook (in this PR) runs, it may cause an error.

@Fosten expressed that he can just uninstall the module and manually re-enter the Equipment types... which isn't that bad as a "worst case scenario", but I would like to take the opportunity to figure out the "right" approach for "absorbing" contrib functionality like this into core, because I want to be able to encourage folks to experiment in modules like this and still have a clear pathway to bring features into core when it makes sense to.

mstenta:
> there's still a potential conflict if `farm_equipment_asset_extended` has been installed, because the taxonomy and field will both already exist, so when the update hook (in this PR) runs, it may cause an error.

Confirmed. In my local environment, I installed `farm_equipment_asset_extended`, then checked out this PR and ran `drush updb` and this is the error:

```
 ---------------- --------------- ------------- ------------------------------- 
  Module           Update ID       Type          Description                    
 ---------------- --------------- ------------- ------------------------------- 
  farm_equipment   install_equip   post-update   Add "Equipment type" field to  
                   ment_type                     Equipment assets.              
 ---------------- --------------- ------------- ------------------------------- 


 Do you wish to run the specified pending updates? (yes/no) [yes]:
 > yes

>  [notice] Update started: farm_equipment_post_update_install_equipment_type
>  [error]  Configuration objects (taxonomy.vocabulary.equipment_type) provided by farm_equipment_type already exist in active configuration 
>  [error]  Update failed: farm_equipment_post_update_install_equipment_type 
 [error]  Update aborted by: farm_equipment_post_update_install_equipment_type 
 [error]  Finished performing updates. 
```

I don't know if there's a standard way around this... Drupal core will throw that error anytime you try to install a module that contains config in `config/install` that already exists in the site.

mstenta:
We may have to spin off a separate issue/discussion for this conundrum specifically. It probably doesn't need to hold up this pull request.

mstenta:
https://farmos.discourse.group/t/documenting-best-practices-for-community-module-development/2180

mstenta:
Sorry, forgot there's still more to finish here... back to draft and canceled review requests.

mstenta:
Ready for review! I added/updated tests to include the new dependency. All green.

Fosten:
I have added DEPRECATION warnings to [Fosten/farm_equipment_asset_extended](https://github.com/Fosten/farm_equipment_asset_extended/) via this commit https://github.com/Fosten/farm_equipment_asset_extended/commit/021cab2f78c1f6cde4a8e6d52d098fbbf25bbab3 in advance of this PR being merged.

paul121 (approved):

PR #897, opened 2024-12-03
Fix Cache Invalidation on /people Page

### LINKED ISSUE #885: /people page not updating last logged in time
The people page shows a list of users and when they last logged in. This page does not always show the correct information. Discussion offline with @mstenta
>I think it's just a minor bug with caching of that /people page
>It doesn't reset when users log in, so it shows cached values

hannahkow:
This pull request addresses an issue where the /people page, which displays a list of users and their last login times, does not always show the correct information. The problem occurs because the page is showing cached values and not updating the data when a user logs in. The /people page in FarmOS is powered by Drupal Views and often caches data for better performance. However, this caching can result in outdated information being displayed, especially when a user's last login time changes. The cache does not reset automatically when a user logs in, which leads to the issue of showing old login data. To resolve this issue, We implemented an event subscriber that listens for the UserLoginEvent in Drupal. When a user logs in, the subscriber is triggered, and it invalidates the cache tags associated with the /people page. This ensures that the cache is cleared every time a user logs in, so the /people page will always display the most current data, including the "last logged in" time. The caching for the /people page View was disabled to ensure that it always retrieves the most up-to-date data. By removing or adjusting the caching settings programmatically, we ensure that the page does not rely on outdated cached values. Overall these changes ensure that the /people page always displays up-to-date information about users, particularly their last login time. By disabling unnecessary caching and implementing cache invalidation on user login, we prevent the page from showing outdated information. This fix improves the reliability of user-related data on the /people page in FarmOS.


mstenta:
Thanks for the PR @hannahkow! I linked it to the related bug report here: #885 

I'll review soon!

mstenta:
@hannahkow I'm going to convert this to a draft because I assume you are still working on it. As it is now, this can't work because the event subscriber is placed in the root of the repository. It needs to be in a module's `src/EventSubscriber` directory (the filename also needs to match the class name, and end in `.php`).

(There's also an unrelated change made to `[farm.info.yml](https://github.com/farmOS/farmOS/pull/897/files#diff-ead8764f2f8e9a466aa6c2381dcd449492868e530f496a95dce608e814e7e46d)` which needs to be removed from the PR.)

Happy to help guide you through the process of testing/updating this if you want. Ping me in the farmOS chat room: https://app.element.io/#/room/#farmOS:matrix.org

hannahkow:
Thank you for getting back to me! Yes I looks like I sent the wrong thing as it is missing two whole files which should fix that build error, I will send it your way.

mstenta:
Closing this due to lack of activity.

PR #898, opened 2024-12-03
Fixed: value active in the status element is not allowed

### LINKED ISSUE #856: Error: The submitted value active in the status element is not allowed
If the `farm_plan` module is enabled without any plan types, the following error is displayed on the `/plans` View page:

> The submitted value active in the status element is not allowed.

Screenshot from 2024-07-24 08-47-53

This is because the View is configured to show a list of `status` options. Apparently this only gets populated if there are `plan` entity bundles.

I've seen this happen to users on Farmier hosted instances when they enable the `farm_comment_plan` module in the `/setup/modules` form, which has a dependency on `farm_plan`. I wonder if we should hide that module (and any others that depend on `farm_plan`) in `/setup/modules` unless there are `plan` entity bundles present.

Or, maybe there's a way to fix the View itself so that it doesn't create an error when there are no plan bundles. :thinking:

alishapatel7:
The updated farm_comment_plan.module file introduces a new function, farm_comment_plan_form_alter(), which dynamically modifies the module installation form (/setup/modules) to hide the farm_comment_plan module if no plan entity bundles are present. This enhancement leverages the entityTypeManager service to check for existing plan bundles, ensuring that the module cannot be enabled prematurely, which could lead to errors. If no bundles are found, the module is hidden, and a warning message is displayed, guiding users to create a plan bundle first. This change improves the user experience by preventing misconfigurations, aligns with observed issues on Farmier-hosted instances, and ensures the module is only activated when its dependencies are met.

### DIFF
--- a/modules/core/comment/modules/plan/farm_comment_plan.module
+++ b/modules/core/comment/modules/plan/farm_comment_plan.module
@@ -20,3 +20,22 @@ function farm_comment_plan_entity_base_field_info(EntityTypeInterface $entity_ty
 
   return $fields;
 }
+
+/**
+ * Implements hook_form_alter().
+ */
+function farm_comment_plan_form_alter(&$form, \Drupal\Core\Form\FormStateInterface $form_state, $form_id) {
+  // Target the module installation form.
+  if ($form_id === 'system_modules') {
+    // Check if there are any plan bundles.
+    $plan_bundles = \Drupal::entityTypeManager()->getStorage('plan_type')->loadMultiple();
+
+    // If no plan bundles exist, hide farm_comment_plan and its dependencies.
+    if (empty($plan_bundles)) {
+      if (isset($form['modules']['farm_comment_plan'])) {
+        $form['modules']['farm_comment_plan']['#access'] = FALSE;
+        drupal_set_message(t('The farm_comment_plan module is hidden because there are no plan entity bundles. Please create a plan type before enabling this module.'), 'warning');
+      }
+    }
+  }
+}


mstenta:
Thanks for the PR @alishapatel7 ! I linked it to the related bug report here: #856

I'll review soon!

modules/core/comment/modules/plan/farm_comment_plan.module
```diff
@@ -20,3 +20,22 @@ function farm_comment_plan_entity_base_field_info(EntityTypeInterface $entity_ty
 
   return $fields;
 }
+
+/**
+ * Implements hook_form_alter().
+ */
+function farm_comment_plan_form_alter(&$form, \Drupal\Core\Form\FormStateInterface $form_state, $form_id) {
+  // Target the module installation form.
+  if ($form_id === 'system_modules') {
```
mstenta, line 29: This is targeting the Drupal core module form (at `/admin/modules`) not the farmOS-specific modules form (at `/setup/modules`), but the PR description seems to suggest this is a fix for `/setup/modules`.

modules/core/comment/modules/plan/farm_comment_plan.module
```diff
@@ -20,3 +20,22 @@ function farm_comment_plan_entity_base_field_info(EntityTypeInterface $entity_ty
...
+    // Check if there are any plan bundles.
+    $plan_bundles = \Drupal::entityTypeManager()->getStorage('plan_type')->loadMultiple();
+
+    // If no plan bundles exist, hide farm_comment_plan and its dependencies.
+    if (empty($plan_bundles)) {
+      if (isset($form['modules']['farm_comment_plan'])) {
+        $form['modules']['farm_comment_plan']['#access'] = FALSE;
+        drupal_set_message(t('The farm_comment_plan module is hidden because there are no plan entity bundles. Please create a plan type before enabling this module.'), 'warning');
```
mstenta, line 37: It isn't possible for a user to create a plan type through the UI, so this might not be the best guidance to provide. Plan types can only be added via a module. See https://farmos.org/development/module/entities/

mstenta (changes requested):
@alishapatel7 I'm going to convert this to a draft because I assume you're still working on it. The logic can't work right now because it's inside the `farm_comment_plan` module... but that logic can only run if the `farm_comment_plan` module is enabled... and it's trying to prevent that module from being enabled.

Generally speaking, you might be on the right track... the two big questions are:

1. Where does the logic belong?

It might make sense to put it in this existing function, which is used to generate the form at `/setup/modules`: https://github.com/farmOS/farmOS/blob/a18dff81390b82b8665cb25a86267bc3686af032/farm.profile#L19

2. More broadly, I wonder if this the best way to solve #856...

Technically speaking, #856 arises because of the combination of a) the `plan` module being enabled without any plan types (which isn't necessarily bad), and b) the `farm_plan` View has a default filter applied that wants to only show plans with a status of `active`.

This feels like a tricky issue to solve, and may have deeper upstream dependency implications (eg: in the `views` module and/or the `state_machine` module). So, it might require some deeper digging to understand. :thinking: 

All that said, it may still be worth putting a temporary simple fix into place that just prevents the issue by hiding the module. I'm happy to help polish this PR up so we can at least consider it... feel free to ping me in the farmOS chat if you want to talk about it: https://app.element.io/#/room/#farmOS:matrix.org

mstenta:
#856 was fixed in 4.x via #986

PR #900, opened 2024-12-18
Update Drupal core to 10.4

mstenta:
Drupal 10.4.0 was released yesterday: https://www.drupal.org/project/drupal/releases/10.4.0

Drafting this PR to run tests, but we need to wait for the official Drupal Docker image to tag 10.4: https://github.com/docker-library/drupal/issues/276

This PR also removes the patch for https://www.drupal.org/project/drupal/issues/3347343, which was included in 10.4.


mstenta:
> we need to wait for the official Drupal Docker image to tag 10.4: https://github.com/docker-library/drupal/issues/276

Added a `TEMP` commit to run tests with the 10.3 image in the meantime. Shouldn't be much different.

PR #902, opened 2024-12-19
Add and enforce declare(strict_types=1) everywhere

mstenta:
This adds `declare(strict_types=1)` to all PHP files in farmOS. It includes a number of commits that prepare for this change, and the final commit implements the new requirement. As far as I can tell nothing in this branch is a breaking change, but welcome review and feedback.


modules/core/quick/src/Traits/QuickStringTrait.php
```diff
@@ -121,7 +121,7 @@ protected function prioritizedString(array $strings = [], array $priority_keys =
           $parts[] = $value;
         }
         else {
-          $parts[] = $this->trimString($value, $non_priority_max_length);
+          $parts[] = $this->trimString($value, (int) $non_priority_max_length);
```
paul121, line 133: > As far as I can tell nothing in this branch is a breaking change, but welcome review and feedback.

Hmm yeah just pondering this. Is this an example where contrib modules/external code might experience errors after this change if they were passing strings into `QuickStringTrait::trimString()`?

mstenta: No. `declare(strict_types=1)` only affects code that is called from the file that it is declared in. So, if a contrib module uses `QuickStringTrait::trimString()` with the wrong types, AND they do not  include `declare(strict_types=1)` in the file that calls it, it will not throw an error.

This is why it's necessary to `declare(strict_types=1)` in every file, rather than a system-wide `php.ini` setting. As I understand it, the PHP community made this decision after a lot of discussion specifically about this potential for downstream breakage.

I confirmed this (and you can to) by adding this to the Planting quick form's `buildForm()` method:

```
    $form['test'] = [
      '#type' => 'markup',
      '#markup' => $this->trimString('this sting is too dang long', '2'),
    ];
```

If `Planting.php` has `declare(strict_types=1);` at the top, an error is thrown and the quick form doesn't load (and PHPStorm highlights the incorrect type parameter as a "incompatible with the declaration"). If you remove `declare(strict_types=1);` then it works (and PHPStorm does not highlight it).

So I think we can safely add `declare(strict_types=1);` without worrying about backwards compatibility. The main thing we need to be careful of is *changing* parameter types. In this PR I only *added* parameter types, but the original type was always kept as well. The only thing I did change was the default value of `$units` in some methods. Previously it was `int $units = 0`, and now it is `string|int|null $units = NULL`. I think that's safe though, because `0` and `NULL` have the same effect within the method logic.

paul121: Great, thanks for the explanation and confirmation. That makes sense!

paul121 (approved):
I believe this looks good. I'm little worried that we'll introduce a bug/errors but the checks should cover this :crossed_fingers: 

Just in case I feel like it would be great to do more local testing before including this in a release. I think the easiest way would be to get this merged ASAP so it can be somewhat naturally tested while devs are working on other things.

mstenta:
> I feel like it would be great to do more local testing before including this in a release.

Yea good idea. Well I'm hoping to put out a 3.3.3 soon with Drupal 10.4 etc, so maybe I'll do that before merging this. That way it can get some more testing before 3.4.0 / 3.3.4.

PR #852, opened 2024-06-28
Update to gin 3.0-rc16

paul121:
Opening this issue to track any changes we need to support gin [3.0-rc12](https://www.drupal.org/project/gin/releases/8.x-3.0-rc12).

Since we are currently on 3.0-rc10 this also includes various improvements and bugfixes from [3.0-rc11](https://www.drupal.org/project/gin/releases/8.x-3.0-rc11)

But most notably a new feature that moves all form action buttons to a top sticky header: https://www.drupal.org/project/gin/issues/3356717
- So far I've found one bug related to this change and submitted a fix: https://www.drupal.org/project/gin/issues/3440148
- The feature is turned off by default but can be turned on in user overrides, so we should make sure it doesn't break anything too much. It seems that it might default to "on" in an upcoming release so it might be good to embrace the change.
- I have mixed feelings but am coming around to it. It's a nice improvement and would create some consistency across all forms but certainly is a significant change for existing users.

PR #903, opened 2024-12-20
Update to stable gin 4.0.0 release

paul121:
Gin has a stable release :partying_face: 

Also a couple commits to fix a minor thing with the views exposed filter details element.

I had thought the gin stable release would enable sticky action buttons by default but looks like this was only upgraded form "experimental" to "beta" and new users will not get this enabled by default. See [gin.settings.yml](https://git.drupalcode.org/project/gin/-/blob/4.0.0/config/install/gin.settings.yml?ref_type=tags#L12). So I don't think there is a need to do the update hook just yet - we had discussed: https://github.com/farmOS/farmOS/pull/852#issuecomment-2228571744

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - [Update Drupal core to 10.4.0 #900](https://github.com/farmOS/farmOS/pull/900)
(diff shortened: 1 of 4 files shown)


paul121:
> Also a couple commits to fix a minor thing with the views exposed filter details element.

Before: 
Screenshot from 2024-12-20 12-42-34

After:
Screenshot from 2024-12-20 12-42-56

mstenta:
Wow great news!

mstenta (approved):

PR #904, opened 2024-12-21
Contextual API

paul121:
See https://gitlab.com/OpenTEAMAg/feedback/support_requests/-/issues/208

I've just dusted off my previous [2.x-context branch](https://github.com/paul121/farmOS/tree/2.x-context) and rebased on latest 3.x. Opening a draft PR so this doesn't get lost going forward.

I want to review some of the implementation details, but here is the current status and quick notes:
- A new `FarmContext` plugin that is `ContextAware` to detect Asset, Log, Plan, etc entity types. Facilitate by `FarmContextManager` service.
  - Context aware plugins are a somewhat magical part of this. Initially, this allows us to easily get an entity based of the entity set in the route parameters/context + works nicely with Block plugin contexts.
  - But via the manager, this also provides a nice API of sorts to gather contextual information about an entity in other contexts, just set the necessary context and ask away. Could be useful for contextual information via API or other locations in UI
- Context plugins can provide an array of messages with: a type (info/warning/etc), short text, long text, array of related links and a weight.
- A new FarmContextBlock plugin that renders contextual info a `details` element. An instance of this block is configured to render in the `help` region (below task links and status messages).

It's been almost 2 years but all of the above was still working with minimal changes :D It definitely needs a through review and could possible use new Drupal core APIs/PHP language features (attributes vs annotation, etc) BUT I think I do still feel good about this general architecture. It feels quite robust and can be adapted to our needs, but should evaluate this again as well.

### DIFF
--- a/modules/core/asset/src/ContextProvider/AssetRouteContext.php
+++ b/modules/core/asset/src/ContextProvider/AssetRouteContext.php
@@ -1,0 +1,78 @@
+<?php
+
+namespace Drupal\asset\ContextProvider;
+
+use Drupal\asset\Entity\Asset;
+use Drupal\Core\Cache\CacheableMetadata;
+use Drupal\Core\Plugin\Context\Context;
+use Drupal\Core\Plugin\Context\ContextProviderInterface;
+use Drupal\Core\Plugin\Context\EntityContext;
+use Drupal\Core\Plugin\Context\EntityContextDefinition;
+use Drupal\Core\Routing\RouteMatchInterface;
+use Drupal\Core\StringTranslation\StringTranslationTrait;
+
+/**
+ * Sets the current asset as a context on asset routes.
+ */
+class AssetRouteContext implements ContextProviderInterface {
+
+  use StringTranslationTrait;
+
+  /**
+   * The route match object.
+   *
+   * @var \Drupal\Core\Routing\RouteMatchInterface
+   */
+  protected $routeMatch;
+
+  /**
+   * Constructs a new AssetRouteContext.
+   *
+   * @param \Drupal\Core\Routing\RouteMatchInterface $route_match
+   *   The route match object.
+   */
+  public function __construct(RouteMatchInterface $route_match) {
+    $this->routeMatch = $route_match;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getRuntimeContexts(array $unqualified_context_ids) {
+    $result = [];
+    $context_definition = EntityContextDefinition::create('asset')->setRequired(FALSE);
+    $value = NULL;
+    if (($route_object = $this->routeMatch->getRouteObject())) {
+      $route_contexts = $route_object->getOption('parameters');
+      // Check for an asset revision parameter first.
+      if (isset($route_contexts['asset_revision']) && $revision = $this->routeMatch->getParameter('asset_revision')) {
+        $value = $revision;
+      }
+      elseif (isset($route_contexts['asset']) && $asset = $this->routeMatch->getParameter('asset')) {
+        $value = $asset;
+      }
+      elseif ($this->routeMatch->getRouteName() == 'asset.add') {
+        $asset_type = $this->routeMatch->getParameter('asset_type');
+        $value = Asset::create(['type' => $asset_type->id()]);
+      }
+    }
+
+    $cacheability = new CacheableMetadata();
+    $cacheability->setCacheContexts(['route']);
+
+    $context = new Context($context_definition, $value);
+    $context->addCacheableDependency($cacheability);
+    $result['asset'] = $context;
+
+    return $result;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getAvailableContexts() {
+    $context = EntityContext::fromEntityTypeId('asset', $this->t('Asset from URL'));
+    return ['asset' => $context];
+  }
+
+}
--- a/modules/core/plan/src/Plugin/FarmContext/PlanAssetFarmContext.php
+++ b/modules/core/plan/src/Plugin/FarmContext/PlanAssetFarmContext.php
@@ -1,0 +1,54 @@
+<?php
+
+namespace Drupal\plan\Plugin\FarmContext;
+
+use Drupal\farm_ui_context\Plugin\FarmContext\FarmContextBase;
+
+/**
+ * Provides context when an asset in a part of a plan.
+ *
+ * @todo Move to a module with a better dependency on the plan.asset field.
+ *
+ * @FarmContext(
+ *   id = "plan_asset_farm_context",
+ *   admin_label = @Translation("Plan asset farm context"),
+ *   context_definitions = {
+ *     "asset" = @ContextDefinition("entity:asset", label = @Translation("Asset")),
+ *   },
+ * )
+ */
+class PlanAssetFarmContext extends FarmContextBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getMessages(): array {
+    $messages = [];
+
+    /** @var \Drupal\asset\Entity\AssetInterface $asset */
+    if ($asset = $this->getContextValue('asset')) {
+
+      // Query for plans the asset might be a part of.
+      $plans = \Drupal::entityTypeManager()->getStorage('plan')->loadByProperties([
+        'asset' => $asset->id(),
+      ]);
+
+      // Build a message for each plan the asset is a part of.
+      foreach ($plans as $plan) {
+        $messages[] = [
+          'type' => 'info',
+          'message' => $this->t(
+            'This asset is a part of the plan: <a href="@plan_uri">@plan_label</a>',
+            [
+              '@plan_uri' => $plan->toUrl()->setAbsolute()->toString(),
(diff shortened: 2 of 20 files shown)


paul121:
Otherwise, today I was mostly playing with how the context block renders the messages into a `details` element, working on how to format lists of short/long messages + links into the details title + summary. Here is a quick preview:

Screenshot from 2024-12-20 16-35-10

Some simple ideas for follow up:
- Reduce padding/size of details element, see collapsible views filters/changes in #903
- Add some color/icons/list icon styles for different message types (info, warning, etc...)
- Consider removing/drastically reducing the size of this on entity edit forms?

mstenta:
Rebased onto 3.x. Let's see if anything needs to be updated for the new PHPStan level, coding standards, etc.

modules/core/ui/farm_ui.info.yml
```diff
@@ -6,6 +6,7 @@ core_version_requirement: ^10
 dependencies:
   - farm:farm_ui_action
   - farm:farm_ui_breadcrumb
+  - farm:farm_ui_context
```
mstenta, line 9: We'll need an update hook to enable this on existing sites.

modules/core/asset/src/ContextProvider/AssetRouteContext.php
```diff
@@ -0,0 +1,78 @@
...
+   */
+  public function __construct(RouteMatchInterface $route_match) {
+    $this->routeMatch = $route_match;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getRuntimeContexts(array $unqualified_context_ids) {
+    $result = [];
```
mstenta, line 42: Looks like a lot of this is copied from core, yea? Maybe it would be worth including a comment here to that effect, so future devs can refer back to the core code if updates are necessary etc (in plan and log contexts too).

modules/core/ui/context/src/Plugin/Block/FarmContextBlock.php
```diff
@@ -0,0 +1,155 @@
...
+use Drupal\Core\Block\BlockBase;
+use Drupal\Core\Extension\ModuleHandlerInterface;
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Drupal\farm_ui_context\FarmContextManagerInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Provides a farm context block.
+ *
+ * @Block(
```
mstenta, line 14: This needs to be changed from an annotation to an attribute.

modules/core/ui/context/config/optional/block.block.farm_context.yml
```diff
@@ -0,0 +1,23 @@
+langcode: en
+status: true
+dependencies:
```
mstenta, line 3: Needs an enforced dependency so that it gets removed when `farm_ui_context` is uninstalled.

modules/core/ui/context/config/optional/block.block.farm_context.yml
```diff
@@ -0,0 +1,23 @@
...
+  module:
+    - farm_ui_context
+  theme:
+    - gin
+id: farm_context
+theme: gin
```
mstenta, line 9: We'll need to figure out how to do this without an explicit dependency on Gin. We maintain the ability for downstream users to develop/use their own theme by encapsulating all Gin-specific stuff in `farm_ui_theme`.

Perhaps `farm_ui_theme` could install this block, and depend on `farm_ui_context`? :thinking: 

Then it would be the responsibility of downstream users with custom themes to add the block themselves, but at least all the other functionality of contexts would exist.

modules/core/ui/views/src/Routing/RouteSubscriber.php
```diff
@@ -25,6 +25,16 @@ public function alterRoutes(RouteCollection $collection) {
 
       // Set default log_type to mark primary tab as active.
       $route->setDefault('log_type', 'all');
+
+      // @todo Fix so that an asset context message displays on views tabs.
+      // This is needed so that the asset route parameter can pick up the route.
+      // General issue: https://www.drupal.org/project/drupal/issues/2528166
```
mstenta, line 31: I'd like to understand this a bit more. It doesn't look like the core issue is getting much action recently, but I only skimmed it.

I see this commit is marked as `TEMP: ...` - what is the path forward here?

modules/core/plan/src/Plugin/FarmContext/PlanAssetFarmContext.php
```diff
@@ -0,0 +1,54 @@
...
+   */
+  public function getMessages(): array {
+    $messages = [];
+
+    /** @var \Drupal\asset\Entity\AssetInterface $asset */
+    if ($asset = $this->getContextValue('asset')) {
+
+      // Query for plans the asset might be a part of.
+      $plans = \Drupal::entityTypeManager()->getStorage('plan')->loadByProperties([
+        'asset' => $asset->id(),
```
mstenta, line 35: This is probably fine as-is, but we should confirm that this doesn't break on plans that do NOT include the default `asset` or `log` reference fields. It's possible for downstream plan types to omit those if they want to take their own approach (or use `plan_record` entities).

(I thought the `crop` and `grazing` plan modules both did that, but it seems that they are still inheriting the `asset` and `log` fields. That's a separate issue... those should be removed since they are not used by those plan types. I'll make note to follow up on that separately...)

mstenta (changes requested):
Overall this looks good and makes sense @paul121! This will be a great addition!

I reviewed the commits to understand the overall architectural decisions, but I haven't tested or reviewed final UI yet. I'll plan to do that once everything else is ironed out.

We should add some automated tests. I'm thinking some kernel tests for the actual context plugin discovery logic, and a functional test for the block.

It would also be nice to include some documentation for module developers. Any developers building plan types that use `plan_record` entities instead of the `plan` and `log` fields will need to write a plugin. Or maybe we can add a generic plugin for `plan_record` use-cases? :thinking: 

Lastly, I wonder if we should consider splitting the "core" bits of `farm_ui_context` out to a `farm_context` module, so that it could potentially still be used on a headless farmOS instance. I know this PR doesn't currently expose contextual information to the API (I see you mentioned that in your comments above though), but that could be useful in the future, and isn't "UI".

mstenta:
This feels like a good feature to include in 4.0.0. I'm going to flag it with that milestone for now, but if we want to get it merged into 3.x I'm fine with that too. It isn't a breaking change, but I'm starting to feel like we're approaching an unofficial feature freeze on the 3.x branch (we've already decided to move #849 to the 4.0.0 milestone).

modules/core/plan/src/Plugin/FarmContext/PlanAssetFarmContext.php
paul121: Ah interesting. As long as there is one plan bundle that has a `plan.asset` field I do not think this would cause an issue - but yes, if there were no plan bundles that had this `plan.asset` field, I can see how this might break. IIRC `loadByProperties` builds a simple entity API query behind the scenes which I do seem to remember failing when specified fields do not exist...

modules/core/ui/views/src/Routing/RouteSubscriber.php
paul121: Hmm yeah, afraid I don't remember if this change was enough to display the contextual info load on the "logs" tab for any given asset, or if this change also needs the core path... My hunch is that this change is enough but the comment is confusing.. but I should test.

I think this comment is a good summary of the issue: https://www.drupal.org/project/drupal/issues/2528166#comment-14855963 but yea, doesn't look like that will merge anytime soon because it could cause some breaking change (existing code that expects the `asset` parameter to only be an ID will fail if this is changed to start returning an entity object)

---

Tho there is also a larger question if we even want the asset's contextual information available on these pages? I see how this could be an issue if the contextual info UI takes up excessive space on the page but I hope that we can make the contextual information UI minimal enough so that there is little to no reason to ever not include on any "asset page"... otherwise we might as well isolate contextual information to a dedicated page and try to show it in other places... but IMO the ability to bubble up contextual info into more places/pages is the driving factor of this PR :sweat_smile: 

Anyways - adding this additional `asset` route parameter to this `view.farm_log.page_asset` route definition makes it possible for the `ContextProvider/AssetRouteContext` class to provide the asset into the route's context on these pages without having to know the specific details about this `view.farm_log.page_asset` route specifically. This is all important because the route context is used by the `FarmContextBlock` itself. When rendering an instance of the `FarmContextBlock` on a given entity page, assetl/log/etc, it means this context will be mapped directly from the route context. _BUT_ the block's context can also be set manually, allowing for the ability to potentially render one or more `FarmContextBlock` on demand for one or more entities in a given context (eg: maybe when interacting with the asset timeline?)

mstenta: Thanks I think I understand a bit better now.

> Tho there is also a larger question if we even want the asset's contextual information available on these pages?

Personally, I think the MVP only requires showing the contextual information on `/asset/%`. If it simplifies this PR to just do that as a first step, and then make next step decisions in follow-ups, I would be OK with that.

PR #905, opened 2025-01-02
Update simple_oauth to 6.0.0-beta8 and unpin version constraint

paul121:
With this [6.0.0-beta8 release](https://www.drupal.org/project/simple_oauth/releases/6.0.0-beta8) we can remove our patch for [issue #3322325](https://www.drupal.org/project/simple_oauth/issues/3322325)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - [Update Drupal core to 10.4.0 #900](https://github.com/farmOS/farmOS/pull/900)
 - [Update to stable Gin 4.0.0 #903](https://github.com/farmOS/farmOS/pull/903)
-- [Update simple_oauth to 6.0.0-beta7 #893](https://github.com/farmOS/farmOS/pull/`893`)
+- [Update simple_oauth to 6.0.0-beta7 #893](https://github.com/farmOS/farmOS/pull/893)
+- [Update simple_oauth to 6.0.0-beta8 #905](https://github.com/farmOS/farmOS/pull/905)
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [Update Drupal core to 10.4.0 #900](https://github.com/farmOS/farmOS/pull/900)
 - [Update to stable Gin 4.0.0 #903](https://github.com/farmOS/farmOS/pull/903)
 - [Update simple_oauth to 6.0.0-beta7 #893](https://github.com/farmOS/farmOS/pull/893)
-- [Update simple_oauth to 6.0.0-beta8 #905](https://github.com/farmOS/farmOS/pull/905)
+- [Update simple_oauth to 6.0.0-beta8 and unpin version constraint #905](https://github.com/farmOS/farmOS/pull/905)
 
 ### Fixed
 
--- a/composer.json
+++ b/composer.json
@@ -42,7 +42,7 @@
         "drupal/migrate_source_ui": "^1.0",
         "drupal/migrate_tools": "^6.0.3",
         "drupal/role_delegation": "^1.2",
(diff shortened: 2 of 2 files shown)


mstenta (changes requested):
> we can remove our patch for [issue #3322325](https://www.drupal.org/project/simple_oauth/issues/3322325)

Great!! That was the last patch we were applying to `simple_oauth`! :tada: 

In theory this also means we can stop pinning `simple_oauth` and change the version constraint to `"drupal/simple_oauth": "^6.0@beta",`.

Our unofficial policy is that we only pin dependencies that we are patching (to ensure that the patch still applies and doesn't break builds). So I think in order to merge this and remain consistent with that policy we'll need to unpin.

OR, we should discuss why it might be a good idea to continue pinning certain modules and amend our pinning policy, and add a comment to `composer.json` explaining why we are pinning `simple_oauth` (and maybe document the pinning policy somewhere at the same time to make this all more official).

What do you think @paul121? Is it risky to unpin `simple_oauth`? Up until now we have been very careful with updates to it, since we've run into bugs in the past IIRC, although recent updates have been stable for a while it seems.

paul121:
> What do you think @paul121? Is it risky to unpin simple_oauth? Up until now we have been very careful with updates to it, since we've run into bugs in the past IIRC, although recent updates have been stable for a while it seems.

Good questions. I don't have a strong opinion but would say that removing the pin would be OK. I agree that recent updates have been more stable, and one good thing is that we at least have some additional tests that use the API + OAuth. So if something fairly significant were to break then we would be notified via our daily test runs. But of course it is possible that other bugs could be introduced in new versions of `simple_oauth` and go unnoticed.

> OR, we should discuss why it might be a good idea to continue pinning certain modules and amend our pinning policy, and add a comment to composer.json explaining why we are pinning simple_oauth (and maybe document the pinning policy somewhere at the same time to make this all more official).

Somewhat related... I'm curious what we should consider wrt our policy for updating the minimum pinned versions of dependencies. For `simple_oauth` it might be good to update the minimum version more frequently than we have been doing for other modules. In practice this means that users would get `simple_oauth` updates even when they only request to update FarmOS core. This could be considered important because it would help keep some consistency in the OAuth authorization protocol that applications external to FarmOS might be dependent on.

mstenta:
> But of course it is possible that other bugs could be introduced in new versions of `simple_oauth` and go unnoticed.

Yep. And this is true of all other unpinned modules too, so sort of a separate concern IMO. Committing `composer.lock` to the https://github.com/farmOS/composer-project repo is the right solution to that I think. So we can continue that discussion separately...

> Somewhat related... I'm curious what we should consider wrt our policy for updating the minimum pinned versions of dependencies. For `simple_oauth` it might be good to update the minimum version more frequently than we have been doing for other modules. In practice this means that users would get `simple_oauth` updates even when they only request to update FarmOS core. This could be considered important because it would help keep some consistency in the OAuth authorization protocol that applications external to FarmOS might be dependent on.

Interesting... Not opposed! But still trying to understand. So you're saying that if we keep bumping the version constraint for `simple_oauth`, for example from `^6.0` to `^6.1` to `^6.2` as they are released, rather than just leaving it at `^6.0`, and then a user updates JUST farmOS (eg: by running `composer require farmos/farmos:3.3.3`), then because `farmos/farmos` requires a higher minimum dependency of `simple_oauth` than the user's `composer.lock`, it will force an update. That makes sense, and I like it!

Although... this might be a moot point because `simple_oauth` is still in beta. If we unpin then the `simple_oauth` version constraint will be `^6.0@beta`. Last I checked, Composer doesn't support version constraints like `^6.0@beta8`, `^6.0@beta9`, etc. So we can't unpin AND specify a specific minimum beta. :-/

So if we wanted to do that we would need to keep pinning I think. I say we unpin in this PR, and then consider whether or not we want to start re-pinning, and update/document our policy accordingly, in a follow up, since these are separate concerns/considerations.

With regard to our minimum versions policy generally, currently we only update it when we know it's necessary for some reason. For example, in the Drupal 11 branch I started, I've updated the version constraints of all modules to ensure they are at least the versions that are compatible with Drupal 11. Other than that, we don't really have much else in terms of policy.

mstenta (approved):

PR #901, opened 2024-12-18
Use Rector to scan for code upgrades

mstenta:
This PR adds Rector to the farmOS Docker dev image and runs `rector process --dry-run` on the farmOS codebase during our `delivery.yml` workflow on GitHub Actions. Rector identifies code that can be upgraded/refactored for new versions of Drupal and Symfony, and also helps find other general PHP coding improvements (alongside our current PHP CodeSniffer and PHPStan analysis).

We can also use it to automatically upgrade code for Drupal 11 (see https://www.drupal.org/project/farm/issues/3488916).

This PR also implements the changes recommended by Rector, which allows the new step in `deliver.yml` to pass.

PR #906, opened 2025-01-15
Upgrade to PHPStan level 2

paul121:
Working to (hopefully) get our codebase to PHPStan level 2 which really just starts checking two things: (see [PHPStan rule levels](https://phpstan.org/user-guide/rule-levels)

> unknown methods checked on all expressions (not just $this), validating PHPDocs

Some notes:

- I discovered that the [phpstan-drupal extension](https://github.com/mglaman/phpstan-drupal) has a `entity_mappings` helper to map entity types to entity class + storage classes that drastically reduced the number of errors when first turning on level 2.
  - See related PR for Log module: https://www.drupal.org/project/log/issues/3499826
  - I think we could also make this same PR for the consumers module but I was able to fix by adding a doc bloc
- In various places we just need to reference the more specific entity type class and/or storage class instead of just using `EntityInterface`. A common one was `Call to an undefined method Drupal\Core\Entity\EntityInterface::get()` which is caused when loading field values from an entity. This happens because the `::get()` method is defined on the `FieldableEntityInterface` implemented by `ContentEntityInterface`, which is finally implemented by our entity classes/interfaces like `AssetInterface`
  - I believe this is working in combination with phpstan-drupal `entity_mappings` to help determine the proper class (instead of defaulting to something like `EntityInterface` from core)


paul121 (commented):
One PHPStan error we are getting is related to `getWorkflowId` and `setWorkflowId` methods on our Asset type, log type and plan type entities: (the respective content entities have a `getWorkflowId` method that delegates the config entity type's method)

```
 ------ -------------------------------------------------------------------------------------- 
  Line   modules/core/asset/src/Form/AssetTypeForm.php                                         
 ------ -------------------------------------------------------------------------------------- 
  87     Call to an undefined method Drupal\asset\Entity\AssetTypeInterface::getWorkflowId().  
 ------ -------------------------------------------------------------------------------------- 
```

These methods exist, but the issue is that they have an `@inheritdoc` but are not inheriting the method from any parent class or implemented interface: https://github.com/farmOS/farmOS/blob/06ef0d82eef3fc1e61ebd441fe67e2fa46004b2b/modules/core/asset/src/Entity/AssetType.php#L137-L150

It seems logical that there would be a shared interface for these methods but it doesn't seem to exist. SO we should declare these methods in their respective interface eg `AssetTypeInterface`, which would then fix the PHPStan error message...

But... I'm curious if we even need these helper methods. I see very few uses of the getter method, and only one usage of a setter method from the `farm_log.post_update.php` file. IMO these few uses could implement the logic themselves and save us unnecessary complexity in our entity definitions. BUT... doing so would be a breaking change. So maybe best that we fix this, deprecate, and then remove at a later time...

This example makes me very happy about getting PHPStan level 2+ implemented, there will be many slips like this that simply never make it in :sun_with_face:

paul121:
Found some good information on this topic here: https://www.midcamp.org/2024/topic-proposal/tighten-your-drupal-code-using-phpstan

docker/dev/files/phpstan.neon
```diff
@@ -1,8 +1,20 @@
...
+  - baseline.neon
 parameters:
-  level: 1
+  level: 2
+  reportUnmatchedIgnoredErrors: true
   ignoreErrors:
     # new static() is a best practice in Drupal, so we cannot fix that.
     - '#^Unsafe usage of new static#'
+  excludePaths:
+    - web/profiles/farm/docker/*
```
paul121: In my local env I was getting errors related to the `docker/dev/files/rector.php` file when running `phpstan analyze web/profiles/farm` so I put this exclude in to ignore it. We might be able to configure PHPStan to load the rector classes it didn't recognize but I don't think it's very important we check this so I chose to exclude.

I don't think we're getting this error in the Github actions because we do not copy these docker files to the final web root (don't remember specific details)?

modules/asset/group/src/Plugin/views/argument/AssetGroup.php
```diff
@@ -72,6 +73,11 @@ public static function create(ContainerInterface $container, array $configuratio
    */
   public function query($group_by = FALSE) {
 
+    // Bail if not a SQL query.
+    if (!$this->query instanceof Sql) {
+      return;
+    }
+
```
paul121, line 80: This will be a common pattern in going forward in most Views plugins! The Drupal views API is abstracted to work with not only SQL databases, but we have been assuming we are *always making SQL queries. By adding `instanceof` checks like this it fixes potential errors in the event it is not a SQL query AND provides additional type guarantees to PHPStan.

modules/asset/group/tests/src/Functional/GroupTest.php
```diff
@@ -66,7 +66,7 @@ public function testGroupFieldVisibility() {
     $this->assertFalse($group_field->isVisible());
 
     // Make the log a group assignment.
-    $log->is_group_assignment = TRUE;
+    $log->set('is_group_assignment', TRUE);
```
paul121, line 69: Use `set()` method to modify entity field values

modules/core/asset/src/AssetStorage.php
```diff
@@ -80,14 +81,17 @@ public static function createInstance(ContainerInterface $container, EntityTypeI
...
     $id = parent::doPreSave($entity);
 
     // If there is no original entity, bail.
     if (empty($entity->original)) {
       return $id;
     }
 
+    // Assert field type for PHPStan checks.
+    assert($entity->get('status')->first() instanceof StateItemInterface);
+
```
paul121, line 94: In some of our test classes I found this to be the lowest impact/most efficient way to provide additional type checking to PHPStan for custom field types that have additional properties/methods we use

modules/core/data_stream/src/Plugin/migrate/destination/DataStream.php
```diff
@@ -52,7 +52,7 @@ protected function getEntity(Row $row, array $old_destination_id_values) {
       // Update the assets data_stream field if the asset was found
       // and the asset type has the field.
       if (!is_null($asset) && $asset->hasField('data_stream')) {
-        $asset->data_stream[] = $entity->id();
+        $asset->get('data_stream')->appendItem($entity->id());
```
paul121, line 54: Use `FieldItemListInterface::appendItem()` to append values to a field

modules/core/id_tag/src/Plugin/Field/FieldWidget/IdTagWidget.php
```diff
@@ -38,7 +38,7 @@ public function formElement(FieldItemListInterface $items, $delta, array $elemen
     $tag_type = $items[$delta]->type ?? NULL;
 
     // Get the current asset bundle.
-    $bundle = $items->getParent()->getEntity()->bundle();
+    $bundle = $items->getEntity()->bundle();
```
paul121, line 40: Note to self: need to do some manual testing to ensure this change doesn't break anything. _I think_ this will work, but the types were very dynamic and complex here. If this breaks something we can revert and add this error to our baseline.

modules/core/log/modules/quantity/src/EventSubscriber/LogQuantityEventSubscriber.php
```diff
@@ -90,7 +90,7 @@ public function logDelete(LogEvent $event) {
     }
 
     // Get any quantities the log references.
-    $quantities = $log->quantity->referencedEntities();
+    $quantities = $log->get('quantity')->referencedEntities();
```
paul121, line 93: Use `get()` method to get entity field values

modules/core/plan/src/Form/PlanTypeForm.php
```diff
@@ -16,6 +16,13 @@
  */
 class PlanTypeForm extends EntityForm {
 
+  /**
+   * The plan type entity.
+   *
+   * @var \Drupal\plan\Entity\PlanTypeInterface
+   */
+  protected $entity;
+
```
paul121, line 25: I have always noticed that it was common for core/contrib to override this `$entity` parameter in forms extending `EntityForm`. This is an easy way to provide additional type checking for our custom entity types across all usages in this form file.

modules/core/quick/tests/src/Kernel/QuickFormTestBase.php
```diff
@@ -29,21 +29,21 @@ abstract class QuickFormTestBase extends KernelTestBase {
   /**
    * Asset entity storage.
    *
-   * @var \Drupal\Core\Entity\EntityStorageInterface
+   * @var \Drupal\asset\AssetStorage
    */
   protected $assetStorage;
```
paul121, line 34: Using the more specific entity type storage interface, in combination with `phpstan-drupal` entity mapping, helps PHPStand understand types for all entities that are created/loaded from the entity storage without additional type hints in our code, eg `AssetStorage` -> `AssetInterface`

paul121 (commented):
Leaving some inline comments highlighting general patterns and changes that we will need to adopt in our "coding standards/style" going forward to be compatible with PHPStan level 2

paul121 (commented):
Interesting, just found two things we had been missing in our tests. @mstenta check out these commits, I think you'll find them interesting:

> Fix broken quick form test after updating code to use field getters and setters
https://github.com/farmOS/farmOS/pull/906/commits/a000e4d93a564710db76e2b522785ea60dbadd9c#diff-b8115338570401e526fc7d26cb62e159a075bf574ba6d398ae034130d5ced670
https://github.com/farmOS/farmOS/pull/906/commits/ba0da3300073d16fe452268b05a771c52a0e75fb

> Fix IdTagTypeConstraintValidator to work on multiple fields with multiple values
https://github.com/farmOS/farmOS/pull/906/commits/9b0cbdeebb4fc6bd770e9bce9dab1b74c9b8b3d1

In both cases I thought I had *accidentally broken the tests while making changes to appease PHPStan, but in reality the tests are correct and it revealed gaps in our logic :-)

Now there is just 1 remaining odd & vague PHP unit test failure but I think this PR is nearly done...

modules/core/id_tag/src/Plugin/Field/FieldWidget/IdTagWidget.php
paul121: Just did some testing and this seems to be working just fine. The ID tag options that are displayed in the entity form are correct. Animal assets have many options, land assets only have a few.

paul121:
> Now there is just 1 remaining odd & vague PHP unit test failure but I think this PR is nearly done...

The failure is in `MapFormTest::testMapForm` on line 52. It's a test that uses a simple quick form to test things. The test was crashing because no logs were being saved after submitting the form. After running in my local debugger I found it was crashing when `QuickLogTrait` was trying to write to the `log.quick` field that didn't exist... (more info below)

TLDR; after adding the missing `farm_quick` dependency on `farm_entity` this issue is fixed! But iirc we had some weird dependency issues in the past regarding `farm_entity` so want to make sure @mstenta can grok this. But as is, this change isn't breaking anything else, and it makes sense to me that `farm_quick` needs to depend on `farm_entity` to ensure it's `log/asset.quick` bundle field exists.

___

Some background info.. this was caused by this commit:
- See `QuickLogTrait` / `QuickAssetTrait`: https://github.com/farmOS/farmOS/pull/906/commits/69ddc9033a78dc8a53413bf71870d845f874ac48#diff-3a8d1af4cf0d0f2f2dc7aaed684ae54db8e1d9aa17c0203d8780d964a82e8be0

Specifically:

```diff
-    $entity->quick[] = $this->getQuickId();
+    $entity->get('quick')->appendItem($this->getQuickId());
```

I had fixed another test that broke with this change, too. It was essentially the same fix, but a kernel test and different dependencies to grok: https://github.com/farmOS/farmOS/pull/906/commits/30e827dc5d9da493df105817f67ee37c12a9b910

> Currently the tests are relying on a $entity->quick dynamic property
to be created and saved on the entity object which is then passed back
to the tests via form state storage. Instead, we need to make sure we are using
the real entity.quick field we provide. This required adding `farm_entity` and
`entity` module dependencies and implementing plugins for the entity bundles.

> Fixes Drupal\Tests\farm_quick\Kernel\QuickFormTest::testQuickFormSubmission
InvalidArgumentException: Field quick is unknown.

docker/dev/files/phpstan.neon
mstenta: Interesting... I found this comment because I am testing locally and didn't see any difference with this `excludePaths` present/absent. I'm not seeing errors related to `rector.php`. :thinking:

Could you test again without it @paul121 and paste the errors you're seeing? And if they're gone now, maybe we can revert this bit?

modules/asset/group/src/Plugin/views/argument/AssetGroup.php
mstenta: Agreed this is a good practice to get into. Officially speaking we don't support non-SQL databases in farmOS, but that's just been implicit. This makes it explicit, which is a good thing. :+1:

modules/core/asset/src/AssetStorage.php
mstenta: Very interesting! So simply including an `assert()` is enough to tell PHPStan what we're expecting. Makes sense!

docker/dev/files/phpstan.neon
paul121: I get these errors when removing the `excludePaths`:

```
 ------ -------------------------------------------------------------------------------------------- 
  Line   docker/dev/files/rector.php                                                                 
 ------ -------------------------------------------------------------------------------------------- 
  15     Parameter $rectorConfig of anonymous function has invalid type Rector\Config\RectorConfig.  
  18     Call to method sets() on an unknown class Rector\Config\RectorConfig.                       
         💡 Learn more at https://phpstan.org/user-guide/discovering-symbols                         
  19     Access to constant DRUPAL_10 on an unknown class DrupalRector\Set\Drupal10SetList.          
         💡 Learn more at https://phpstan.org/user-guide/discovering-symbols                         
  24     Call to method autoloadPaths() on an unknown class Rector\Config\RectorConfig.              
         💡 Learn more at https://phpstan.org/user-guide/discovering-symbols                         
  31     Call to method fileExtensions() on an unknown class Rector\Config\RectorConfig.             
         💡 Learn more at https://phpstan.org/user-guide/discovering-symbols                         
  32     Call to method importNames() on an unknown class Rector\Config\RectorConfig.                
         💡 Learn more at https://phpstan.org/user-guide/discovering-symbols                         
  33     Call to method importShortClasses() on an unknown class Rector\Config\RectorConfig.         
         💡 Learn more at https://phpstan.org/user-guide/discovering-symbols                         
  38     Call to method skip() on an unknown class Rector\Config\RectorConfig.                       
         💡 Learn more at https://phpstan.org/user-guide/discovering-symbols                         
  39     Class DrupalRector\Rector\PHPUnit\ShouldCallParentMethodsRector not found.                  
         💡 Learn more at https://phpstan.org/user-guide/discovering-symbols    
         ```

mstenta: This file was added very recently in https://github.com/farmOS/farmOS/pull/901

Is it possible that your dev environment hasn't been updated yet to include `palantirnet/drupal-rector` (the new dev dependency)?

paul121: Aha, yes! This was the issue. Problem solved after running a composer update.

mstenta: Awesome! I'll remove that commit then.

mstenta:
Reviewed, rebased, and added a few commits (in discussion with @paul121):

- https://github.com/farmOS/farmOS/pull/906/commits/8206e565458d7306765f11218ab21810de3b9e74 - I tagged a 2.3.1 release of Log module that includes `entity_mapping.neon`, so these are no longer necessary in farmOS.
- https://github.com/farmOS/farmOS/pull/906/commits/1e7aeb82e9eb837347de33d5f631b269dd5db1e2 - This is just a follow-up to a previous commit. Not necessary for PHPStan, but it removes our last call to `isConfigurable()` and replaces it with an `instanceof ConfigurableQuickFormInterface` check for consistency.
- https://github.com/farmOS/farmOS/pull/906/commits/375697a757e45154816cd391295cf14d96c0ccf0 - Another follow-up... All modules that implement hook_farm_entity_bundle_field_info()
should depend on farm_entity.
- https://github.com/farmOS/farmOS/pull/906/commits/376d39ea163793d3e5894f5907e576d0b60e3a36 - This fixes the PHPStan errors related to `getWorkflowId()` and `setWorkflowId()`.
- https://github.com/farmOS/farmOS/pull/906/commits/78934195b3a4da5b4b70deaa08628690650f06aa and https://github.com/farmOS/farmOS/pull/906/commits/ef5028b2a3598731d44970969c45e0dabe05de76 - Removes the `baseline.neon` file and moves the remaining errors to inline `@phpstan-ignore` comments.
- https://github.com/farmOS/farmOS/pull/906/commits/fe29d3418a8deaef5723e19926483d325315446c - Adds comments to *all* `@phpstan-ignore` declarations to describe why they are being ignored.

I think this is ready to merge, if you're OK with that @paul121! Thanks again for all the work on this! :rocket:

mstenta (approved):

paul121:
Great, yep this LGTM once we see tests pass :+1:

mstenta:
CodeSniffer didn't like the indentation I used in the `@phpstan-ignore` comments. https://github.com/farmOS/farmOS/actions/runs/13037875472/job/36377880732?pr=906

Fixed and force-pushed.

PR #908, opened 2025-01-16
Convert Drupal core annotations to attributes

mstenta:
This PR is a first step towards the work described in https://www.drupal.org/project/farm/issues/3461548

It converts all of the Drupal core plugins that we are implementing to use native PHP attributes instead of annotations, where available.

There are a few Drupal core plugin types that still use annotations in Drupal 10, so we won't be able to convert them until Drupal 11. This includes `ConfigEntityType`, `ContentEntityType`, and `MigrateSource`. I have started a `4.x-attributes` branch that makes these changes, which we can rebase onto my work-in-progress `4.x` branch after this PR is merged.

https://github.com/mstenta/farmOS/tree/4.x-attributes

We can also start to add support for attributes to the plugin types that farmOS provides. I started a `3.x-farm-attributes` branch for that purpose. It currently adds enforcement for them, but does not convert any yet.

https://github.com/mstenta/farmOS/tree/3.x-farm-attributes

### DIFF
--- a/docker/dev/files/rector.php
+++ b/docker/dev/files/rector.php
@@ -8,6 +8,8 @@
 declare(strict_types=1);
 
 use DrupalFinder\DrupalFinderComposerRuntime;
+use DrupalRector\Drupal10\Rector\Deprecation\AnnotationToAttributeRector;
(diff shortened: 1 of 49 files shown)


paul121 (approved):
LGTM! There had been one test failure but re-ran and all is good

PR #909, opened 2025-01-18
Run phpcs, phpstan and rector in separate jobs

paul121:
I don't think there is any reason that a failed phpcs needs to prevent phpstan and rector from running. With it taking longer to run our tests (mostly building the docker images) it can save lots of debugging time to run all these checks at the same time.

I think this will become especially useful for contributors as we start leveraging stricter PHPStan rules, see #906 - In this PR I've been battling some phpcs while really wanting to just test the PHPStan test :-)

### DIFF
--- a/.github/workflows/deliver.yml
+++ b/.github/workflows/deliver.yml
@@ -88,74 +88,56 @@ jobs:
     name: PHP Codesniffer
     runs-on: ubuntu-latest
     needs: build
-    strategy:
-      matrix:
-        platform:
-        # The PHP code we're scanning is the same for all platforms so scan the copy
-        # from the amd64 Docker image since that one will start up fastest
-         - amd64
     steps:
       - name: Checkout the repository
         uses: actions/checkout@v4
       - name: Restore the farmOS dev Docker image from cache
         uses: actions/cache@v4
         with:
-          path: /tmp/farmos-dev-${{ matrix.platform }}.tar
-          key: farmos-dev-${{ matrix.platform }}-${{ github.run_id }}
+          path: /tmp/farmos-dev-amd64.tar
+          key: farmos-dev-amd64-${{ github.run_id }}
       - name: Load the Docker dev image
-        run: docker load < /tmp/farmos-dev-${{ matrix.platform }}.tar
+        run: docker load < /tmp/farmos-dev-amd64.tar
       - name: Run PHP CodeSniffer
-        run: docker run $DOCKER_IMG_PREFIX:3.x-dev-${{ matrix.platform }} phpcs /opt/drupal/web/profiles/farm
+        run: docker run $DOCKER_IMG_PREFIX:3.x-dev-amd64 phpcs /opt/drupal/web/profiles/farm
       - name: Check PHP compatibility of contrib modules and themes (ignore warnings)
         run: |
-          docker run $DOCKER_IMG_PREFIX:3.x-dev-${{ matrix.platform }} phpcs --standard=PHPCompatibility --runtime-set testVersion 8.3- --warning-severity=0 /opt/drupal/web/modules
-          docker run $DOCKER_IMG_PREFIX:3.x-dev-${{ matrix.platform }} phpcs --standard=PHPCompatibility --runtime-set testVersion 8.3- --warning-severity=0 /opt/drupal/web/themes
+          docker run $DOCKER_IMG_PREFIX:3.x-dev-amd64 phpcs --standard=PHPCompatibility --runtime-set testVersion 8.3- --warning-severity=0 /opt/drupal/web/modules
+          docker run $DOCKER_IMG_PREFIX:3.x-dev-amd64 phpcs --standard=PHPCompatibility --runtime-set testVersion 8.3- --warning-severity=0 /opt/drupal/web/themes
 
   phpstan:
     name: PHPStan
     runs-on: ubuntu-latest
     needs: build
-    strategy:
-      matrix:
-        platform:
-        # The PHP code we're scanning is the same for all platforms so scan the copy
-        # from the amd64 Docker image since that one will start up fastest
-         - amd64
     steps:
       - name: Checkout the repository
         uses: actions/checkout@v4
       - name: Restore the farmOS dev Docker image from cache
         uses: actions/cache@v4
         with:
-          path: /tmp/farmos-dev-${{ matrix.platform }}.tar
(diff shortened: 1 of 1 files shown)


paul121:
Okay, got tired of waiting for *all* the docker (arm) images to build before tests ran... :-)

So I added one more commit separating `build` into `build-amd` and `build-arm`, and changed so that tests only depend on `build-amd` to run. To keep things simple I copied all of the `build` job but only changed the `matrix` values to include that platform(s) that are needed. This way we don't need to worry about hard-coded platform values in `build-amd` or `build-arm` when we need to make changes to this in the future.

I also streamlined the other dependencies (publish, create-release, announce, etc) to be chained on one another, rather than list everything out. I don't think this should create any issues but worthy of your review @mstenta @symbioquine !

fast-tests

mstenta:
In general I like the ideas of a) not waiting for ARM builds, and b) running sniffs in parallel seems fine. I think my only hesitation is this YML file is getting to be pretty crazy, and I don't love duplicating steps a lot. It's probably fine for this purpose now, but I wonder if we can give some thought to taming this workflow (or breaking it up into multiple pieces that somehow are cross-dependent on one another? maybe we need to write our own actions?)

paul121:
> I think my only hesitation is this YML file is getting to be pretty crazy, and I don't love duplicating steps a lot.

Agreed. It's not necessarily very complex, but is a growing list of steps with repeated patters (mostly docker) within each job.

Fwiw I did do a little research to see if there was an easier option instead of separating the build steps. I was really hoping we could add a dependency on a single job from the matrix but [it seems that isn't possible](https://github.com/orgs/community/discussions/42335).

That discussion mentioned the ability to "[reuse workflows](https://docs.github.com/en/actions/sharing-automations/reusing-workflows)". I think there are many reasons and ways you might want to reuse workflows so I'm not sure which method might best solve our problem. Copy/pasting our build steps was simpler than going down that rabbit hole... but if we want to look further I'm happy to investigate. I think the rough structure and "dependency" chain I've outlined here is still a good goal though.

If we add a comment in the duplicated build jobs reading like "Make sure you make changes to both build jobs".. would that be sufficient?

mstenta:
> If we add a comment in the duplicated build jobs reading like "Make sure you make changes to both build jobs".. would that be sufficient?

I suppose so... for now at least.

> That discussion mentioned the ability to "[reuse workflows](https://docs.github.com/en/actions/sharing-automations/reusing-workflows)". I think there are many reasons and ways you might want to reuse workflows so I'm not sure which method might best solve our problem. Copy/pasting our build steps was simpler than going down that rabbit hole...

Yea I agree this is probably a larger task, and can be pursued after this.

.github/workflows/deliver.yml
```diff
@@ -17,18 +17,74 @@ env:
 
 jobs:
 
-  build:
-    name: Build Docker images
+  build-amd:
+    name: Build AMD Docker images
```
mstenta: nit: Change "Build AMD Docker images" to "Build AMD Docker image"

.github/workflows/deliver.yml
```diff
@@ -221,9 +321,7 @@ jobs:
     if: github.repository == 'farmOS/farmOS' && github.event_name == 'push' && (github.ref_type == 'tag' || (github.ref_type == 'branch' && needs.build.outputs.farmos_version == '3.x'))
     runs-on: ubuntu-latest
     needs:
-      - build
-      - sniff
-      - test
+      - release
```
mstenta: I think we want to keep this as it was... this is responsible for pushing the `3.x` images to Docker Hub, so we don't want it to only run on `release` (which is for tagged releases).

mstenta (dismissed):

.github/workflows/deliver.yml
paul121: Good catch, that is correct

paul121:
Okay, I removed the commit separating `build` into `build-amd` and `build-arm`. We can tackle that in a separate PR. It is still available at this branch: https://github.com/paul121/farmOS/tree/3.x-separate-build-arch

Added a few follow up commits to further decrease how much code/lines are in the linting steps. They are quite slim now!

.github/workflows/deliver.yml
```diff
@@ -82,36 +82,53 @@ jobs:
...
+      - name: Restore the farmOS dev Docker image from cache
+        uses: actions/cache@v4
+        with:
+          path: /tmp/farmos-dev-amd64.tar
+          key: farmos-dev-amd64-${{ github.run_id }}
+      - name: Load the Docker dev image
+        run: docker load < /tmp/farmos-dev-amd64.tar
+      - name: Run PHPStan
+        run: docker run $DOCKER_IMG_PREFIX:3.x-dev-amd64 phpstan analyze --memory-limit 1G /opt/drupal/web/profiles/farm
+  rector:
```
mstenta, line 121: nit: Add a line break above this (to separate jobs) for consistency.

mstenta (changes requested):
Only one small whitespace nit, but otherwise I think this looks good.

mstenta (approved):

mstenta:
I'll give @symbioquine a chance to review before I merge, but this LGTM!

.github/workflows/deliver.yml
```diff
@@ -143,8 +161,6 @@ jobs:
       # https://github.com/docker/setup-qemu-action
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v3
-      - name: Checkout the repository
-        uses: actions/checkout@v4
```
symbioquine: Curious if/how the `cp` command on line 177 still works?

symbioquine (commented):
LGTM - other than the question about `cp` and the removed checkout steps.

.github/workflows/deliver.yml
paul121: Good catch! This is failing. After rebasing on 3.x so that our builds complete, it failed with: https://github.com/paul121/farmOS/actions/runs/13184528286/job/36804425740

> cp: cannot stat 'docker/docker-compose.testing.*': No such file or directory

paul121:
Okay, rebased on 3.x and force-pushed to fix the issue @symbioquine found. We can't remove the `actions/checkout` step from the PHPUnit test job.

This should be good now!

symbioquine (approved):
LGTM

mstenta (approved):
LGTM too! Thanks @paul121!

PR #911, opened 2025-01-25
PHPStan level 5

paul121:
This PR builds off #906  adds fixes for PHPStan levels 3 and 4. Brief summary of what these level check for:

> 3. return types, types assigned to properties
> 4. basic dead code checking - always false instanceof and other type checks, dead else branches, unreachable code after return; etc.

Level 3 required minimal changes, the first ~15 commits.

Level 4 seemed a bit more daunting at first. But once I turned off [`treatPhpDocTypesAsCertain`](https://phpstan.org/config-reference#treatphpdoctypesascertain) there were way fewer errors. I'm not sure yet if we will want this on or off.

### DIFF
--- a/docker/dev/files/phpstan.neon
+++ b/docker/dev/files/phpstan.neon
@@ -1,5 +1,5 @@
 parameters:
-  level: 4
+  level: 5
   reportUnmatchedIgnoredErrors: true
   ignoreErrors:
     # new static() is a best practice in Drupal, so we cannot fix that.
--- a/docker/dev/files/phpstan.neon
+++ b/docker/dev/files/phpstan.neon
@@ -1,5 +1,5 @@
 parameters:
-  level: 3
+  level: 4
   reportUnmatchedIgnoredErrors: true
   ignoreErrors:
     # new static() is a best practice in Drupal, so we cannot fix that.
--- a/docker/dev/files/phpstan.neon
+++ b/docker/dev/files/phpstan.neon
@@ -1,5 +1,5 @@
 parameters:
-  level: 2
+  level: 3
   reportUnmatchedIgnoredErrors: true
   ignoreErrors:
     # new static() is a best practice in Drupal, so we cannot fix that.
--- a/modules/asset/equipment/tests/src/Functional/EquipmentFieldTest.php
+++ b/modules/asset/equipment/tests/src/Functional/EquipmentFieldTest.php
@@ -67,7 +67,7 @@ class EquipmentFieldTest extends FarmBrowserTestBase {
 
     // Create a log that references the equipment.
     $log = $log_storage->create(['type' => 'test']);
-    $log->equipment[] = ['target_id' => $asset->id()];
+    $log->get('equipment')->appendItem($asset->id());
     $log->save();
 
     // Go to the log view page.
--- a/modules/asset/group/src/Field/AssetGroupItemList.php
+++ b/modules/asset/group/src/Field/AssetGroupItemList.php
@@ -6,6 +6,7 @@ namespace Drupal\farm_group\Field;
 
 use Drupal\Core\Field\EntityReferenceFieldItemList;
 use Drupal\Core\TypedData\ComputedItemListTrait;
+use Drupal\asset\Entity\AssetInterface;
 
 /**
  * Computes the current group value for assets.
@@ -23,6 +24,7 @@ class AssetGroupItemList extends EntityReferenceFieldItemList {
     $entity = $this->getEntity();
 
     // Get the asset's current groups.
+    assert($entity instanceof AssetInterface);
     $groups = \Drupal::service('group.membership')->getGroup($entity);
 
     // Update the assets current group values to match.
--- a/modules/asset/group/src/GroupMembership.php
+++ b/modules/asset/group/src/GroupMembership.php
@@ -133,17 +133,8 @@ class GroupMembership implements GroupMembershipInterface {
       return NULL;
     }
 
-    // Load the first log.
-    /** @var \Drupal\log\Entity\LogInterface $log */
-    $log = $this->entityTypeManager->getStorage('log')->load(reset($log_ids));
-
-    // Return the log, if available.
-    if (!is_null($log)) {
-      return $log;
-    }
-
-    // Otherwise, return NULL.
-    return NULL;
+    // Load the first log or return NULL.
+    return $this->entityTypeManager->getStorage('log')->load(reset($log_ids));
   }
 
   /**
--- a/modules/asset/group/src/Plugin/Validation/Constraint/CircularGroupMembershipConstraintValidator.php
+++ b/modules/asset/group/src/Plugin/Validation/Constraint/CircularGroupMembershipConstraintValidator.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Drupal\farm_group\Plugin\Validation\Constraint;
 
 use Drupal\Core\DependencyInjection\ContainerInjectionInterface;
+use Drupal\asset\Entity\AssetInterface;
 use Drupal\farm_group\GroupMembershipInterface;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\Validator\Constraint;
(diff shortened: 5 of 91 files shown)


paul121:
There are a couple level 3 & 4 errors (unrelated to `treatePhpDocTypeAsCertain`) that I'm stumped on the best way to fix. We get these same two errors in the `buildForm()` method of many of our `**EntityActionForm` forms because we use a similar pattern for checking existing of entities in the private temp store before potentially returning a redirect response.

**Update** I've resolved both of these errors now:
- https://github.com/farmOS/farmOS/pull/911/commits/642b06c8561c87f4e9ea0a64c9fb23f35703a7c8
- https://github.com/farmOS/farmOS/pull/911/commits/393fddae8019cf91361397ddcfd76cf08a2298c0

<details>
Here is the `LogCategorizeActionForm` for example (PHPStan line numbers are a little off but you can see):

```
 ------ -------------------------------------------------------------------- 
  Line   modules/taxonomy/log_category/src/Form/LogCategorizeActionForm.php  
 ------ -------------------------------------------------------------------- 
  126    Property                                                            
         Drupal\farm_log_category\Form\LogCategorizeActionForm::$entityType  
         (Drupal\Core\Entity\EntityTypeInterface) in empty() is not falsy.   
  127    Method                                                              
         Drupal\farm_log_category\Form\LogCategorizeActionForm::buildForm()  
         should return array but returns                                     
         Symfony\Component\HttpFoundation\RedirectResponse. 
```
https://github.com/farmOS/farmOS/blob/c83888b83e464d5c5956d3b379bf8e58d4592ac1/modules/taxonomy/log_category/src/Form/LogCategorizeActionForm.php#L125-L135

Regarding the `$entityType` property error, I think we have this pattern because some of our action forms are generalized to support multiple entity types. See `AssignActionForm`:

https://github.com/farmOS/farmOS/blob/c83888b83e464d5c5956d3b379bf8e58d4592ac1/modules/core/owner/src/Form/AssignActionForm.php#L137-L159

Thus, we *should* check that the `$entity_type` passed to the `buildForm()` method is valid. But for other forms I think we could even load the hard-coded entity type and populate the `$entityType` property in the `__construct()` method. Just an observation, this wouldn't solve the error by itself though...

An easy fix for the error would be to make the type for `$entityType` be `EntityTypeInterface|NULL`. Then we are allowed to do an `empty($this->entityType)` check.

The other option would be to check that the entity type is loaded (not "empty") before assigning the property. This feels more robust. That way you don't need to check that the property exists in every context that you want to access it. Like this (this pattern could be used in the `__construct()` method too):

```php
    // Only allow asset and log entities.
    if (
      !in_array($entity_type, ['asset', 'log'])
      || !$loaded_entity_type = $this->entityTypeManager->getDefinition($entity_type)
    ) {
      throw new PluginException('Unsupported entity type given when building form to assign entity');
    }
    $this->entityType = $loaded_entity_type;
```

___

Somewhat separately is the incorrect return type of `Symfony\Component\HttpFoundation\RedirectResponse`. I'm tempted to ignore this with inline comments (or create a rule to ignore since we have multiple instances of this). But also curious if there is a better way. In some searching I found there is a `RedirectResponse::send()` method but also read that manually sending a response can cause issues, which makes sense.
- https://drupal.stackexchange.com/questions/317149/how-to-do-do-a-redirect-in-buildform
- https://drupalsun.com/philipnorton42/2022/05/22/drupal-9-correct-way-redirect-page

</details>

paul121:
I've rebased this PR on #919 while these PRs are awaiting merge... but otherwise this is ready for review!

Also bumped this to PHPStan level 5 "checking types of arguments passed to methods and functions" as there were only ~40 more errors and pretty simple fixes !

Level 6 is "report missing typehints" and finds 1000+ more errors.... lots of functions missing return types, parameters without types, etc. This includes hook functions in `farm_module.module` files and would be a lot of work to clean up.

mstenta:
Rebased onto 3.x - let's see if any of the recently merged 3.4.x PR code requires any changes for this.

mstenta:
> Level 4 seemed a bit more daunting at first. But once I turned off [treatPhpDocTypesAsCertain](https://phpstan.org/config-reference#treatphpdoctypesascertain) there were way fewer errors. I'm not sure yet if we will want this on or off.

@paul121 I'm not seeing this anywhere in the commits. Did you end up avoiding the need to change this config?

mstenta:
I went through this commit by commit, and it looks good to me! I made a few small stylistic changes (nothing functionally different):

- Moved the "Bump to PHPStan level X" commits to the *end* of each group of commits relevant to that level. This way, we are never breaking `HEAD`... so we fix all the issues and then bump the level. This is just more in line with how I expect to see commit history, although it makes perfect sense why it was committed that way during development.
- Moved the "Add comments to ignore PHPStan error for incorrect return type" before the "Bump to PHPStan level 3" commit, since it is necessary for level 3.
- Added a commit to the end that changes type `NULL` to `null` everywhere (notably just for *type* hints and docblocks, but not where we use the `NULL` constant for assignment). This is just for consistency, but I did find that it's part of the PHP style guide, too: https://www.php-fig.org/psr/psr-12/ (section 2.5) - Maybe we can consider changing the `NULL` constants to `null` as well, but that's a separate question IMO.

Lastly, I'd like to go through all the `@phpstan-ignore` comments and make their description comments consistent with the format we established in #906:

```
// PHPStan level [level]+ throws the following error on the next line:
// [error]
// We ignore this because [reason].
// @phpstan-ignore [error.type]
```

I'll work on that... and also take the opportunity to review the ignored code in more detail to understand the considerations better.

mstenta:
> Somewhat separately is the incorrect return type of `Symfony\Component\HttpFoundation\RedirectResponse`. I'm tempted to ignore this with inline comments (or create a rule to ignore since we have multiple instances of this). But also curious if there is a better way.

I may have found a way to fix these without `@phpstan-ignore return.type`, by adding a type hint of `array|RedirectResponse` to the `buildForm()` methods.

I'm not sure if this is a good practice or not... the upstream `buildForm()` methods do not provide a type-hint, only a docblock, so presumably that's what PHPStan is looking at. I don't think there's anything wrong with adding a type hint in downstream methods. And if Drupal core ever adds them itself, and it conflicts with ours, we'll catch that in tests and be able to reassess. Presumably they will also need to figure this out upstream, so maybe they will take the same approach. :sweat_smile: 

I pushed a commit for review. What do you think @paul121?

mstenta:
Pushed three more commits. The first two reformat some code to remove the need for 3 `@phpstan-ignore` comments, and the last one just reformats remaining ones for consistency.

I'm happy with all of this, and feel like it's ready to merge!

What do you think about the changes I made @paul121? Any concerns?

mstenta (approved):

mstenta:
Fixed minor whitespace complaint from PHP Codesniffer.

paul121:
> Lastly, I'd like to go through all the @phpstan-ignore comments and make their description comments consistent with the format we established in https://github.com/farmOS/farmOS/pull/906:

I agree we could standardize this but could we simplify it? I don't think there is a need to include the level or specific error message. Providing is a reason is good, but the `error.type` identifier should be fairly clear and can also be looked up here if more context is needed? https://phpstan.org/error-identifiers

---

> I may have found a way to fix these without @phpstan-ignore return.type, by adding a type hint of array|RedirectResponse to the buildForm() methods.

Oh interesting... hmm. Well it does seem a little hacky. But also convenient! I don't have a strong opinion. If/when Drupal core fixes this then I think we would no longer need the inline `@phpstan-ignore`, which itself would cause the PHPStan tests to fail, so we're kinda in the same boat either way. 

Other commits look good @mstenta !

mstenta:
> I agree we could standardize this but could we simplify it?

Sure we can consider simplifying. But let's at least start with consistency with the standard adopted in #906. If we are going to change it then we should change all of the ones we added in that PR as well.

Personally I think the ugliness of the comments incentivizes us (and future devs) to both a) avoid adding more of them in the future, and b) try to fix/remove the ones that are there (as we come across them again later). Including the specific error makes that a bit easier and might encourage new approaches/ideas. I don't feel super strongly about keeping them, though, so just my 2 cents. :-)

I'll go ahead and merge this PR as-is so we can move forward with next steps. I'm eager to rebase all the 4.x branches on top.

If you want to open a follow up PR to propose reformatting the `@phpstan-ignore` comments, please do!

PR #912, opened 2025-01-28
Document how to add a field via update hook

mstenta:
This adds some documentation to our module developers guide which covers how to add a new field to an existing entity type via an update hook.

It also specifically mentions the need for running `\Drupal::service('entity_field.manager')->rebuildBundleFieldMap();` when adding bundle fields, per: https://www.drupal.org/project/drupal/issues/3045509#comment-15684438

### DIFF
--- a/docs/development/module/fields.md
+++ b/docs/development/module/fields.md
@@ -104,6 +104,73 @@ function mymodule_farm_entity_bundle_field_info(EntityTypeInterface $entity_type
 }
 ```
 
+### Update hook
+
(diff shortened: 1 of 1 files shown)


ktohalloran (approved):
I think this looks good and is clearly written. Thanks for including the code example and the notes about usage--this should be really helpful documentation!

paul121 (approved):

mstenta:
Thanks for reviewing @ktohalloran and @paul121!

PR #914, opened 2025-01-30
Build docker images with ubuntu-22.04

paul121:
This is a workaround for an upstream issue with QEMU and binfmt on 24.04 https://github.com/tonistiigi/binfmt/issues/215

This is probably the easiest way to get our tests passing again. We can revert once this fixed upstream


mstenta:
I force pushed to include a comment in the YML to make it clear why we're pinning temporarily.

mstenta (approved):

PR #916, opened 2025-01-31
Allow modules to add base fields to default CSV importers

mstenta:
This adds a new `hook_farm_import_csv_base_fields()` hook to the `farm_import_csv` module, which allows modules to return a list of base field machine IDs that should be included in default CSV importers.

This will also enable the https://github.com/farmOS/farmOS/pull/849 PR to add the  `farm` base field to asset CSV importers.

This takes the same approach as #915. It is not technically dependent on it, but I have based the branch off of it (so it includes all of that PR's commits as well), because I will be opening two more PRs that build on both.

### DIFF
--- a/modules/core/import/modules/csv/src/Plugin/Derivative/CsvImportMigrationBase.php
+++ b/modules/core/import/modules/csv/src/Plugin/Derivative/CsvImportMigrationBase.php
@@ -5,7 +5,9 @@ declare(strict_types=1);
 namespace Drupal\farm_import_csv\Plugin\Derivative;
 
 use Drupal\Component\Plugin\Derivative\DeriverBase;
+use Drupal\Core\Entity\EntityFieldManagerInterface;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\Extension\ModuleHandlerInterface;
 use Drupal\Core\Field\BaseFieldDefinition;
 use Drupal\Core\Plugin\Discovery\ContainerDeriverInterface;
 use Drupal\Core\StringTranslation\StringTranslationTrait;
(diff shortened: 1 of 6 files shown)


modules/core/import/modules/csv/src/Plugin/Derivative/CsvImportMigrationBase.php
```diff
@@ -144,14 +181,14 @@ public function getDerivativeDefinitions($base_plugin_definition): array {
...
-   * @param \Drupal\Core\Field\FieldDefinitionInterface $field_definition
+   * @param \Drupal\Core\Field\BaseFieldDefinition|\Drupal\entity\BundleFieldDefinition $field_definition
    *   The field definition.
    * @param array &$mapping
    *   The migration process mapping.
    * @param array &$columns
    *   The column descriptions from third-party settings.
    */
-  protected function addFieldMapping(FieldDefinitionInterface $field_definition, array &$mapping, array &$columns): void {
+  protected function addFieldMapping(BaseFieldDefinition|BundleFieldDefinition $field_definition, array &$mapping, array &$columns): void {
```
paul121: Both `BaseFieldDefinition` and `BundleFieldDefinition` implement `FieldDefinitionInterface`, I don't think we need to be more explicit here?

Any reason this is separate from the previous commit? https://github.com/farmOS/farmOS/pull/916/commits/208317c4c838b7fc6b56dd57d84fbb5c22555d6b

paul121 (approved):
LGTM just one small question

modules/core/import/modules/csv/src/Plugin/Derivative/CsvImportMigrationBase.php
mstenta: Oops looks like I accidentally munged those two commits in a rebase while I was working. I will clean those up and force push...

Regarding `FieldDefinitionInterface`: my original hope was that we could use that instead, but doing so causes PHPStan to complain:

```
 ------ ---------------------------------------------------------------------------------------------------- 
  Line   web/profiles/farm/modules/core/import/modules/csv/src/Plugin/Derivative/CsvImportMigrationBase.php  
 ------ ---------------------------------------------------------------------------------------------------- 
  297    Call to an undefined method Drupal\Core\Field\FieldDefinitionInterface::getCardinality().           
  297    Call to an undefined method Drupal\Core\Field\FieldDefinitionInterface::getCardinality().           
 ------ ----------------------------------------------------------------------------------------------------
```

The `getCardinality()` method is defined in `FieldStorageDefinitionInterface`, so both `BaseFieldDefinition` and `BundleFieldDefinition` have it (because they implement that interface in addition to `FieldDefinitionInterface`), but `FieldDefinitionInterface` does not.

Our original `addBundleField()` method declared (in the docblock but not in the argument definition) that it expected `BundleFieldDefinition`, so PHPStan was content with that. When I renamed the method to `addFieldMapping()`, I tried changing the declaration to expect `FieldDefinitionInterface` but that caused PHPStan to complain, so I took the simple approach of using a union `BaseFieldDefinition|BundleFieldDefinition`. Maybe there's a better way... but that's why I took that approach FWIW.

mstenta: Force pushed... the two commits should be much clearer now.

paul121: Ah yea, makes sense!

Just a quick search, I think we could potentially use [intersection types](https://php.watch/versions/8.1/intersection-types) to do something like `FieldDefinitionInterface&FieldStorageDefinitionInterface` ... basically add all the interfaces that we are expecting.

... but `BaseFieldDefinition|BundleFieldDefinition` is fine with me! If anything it might be more explicit than what we actually need, but these are the only two things we expect here anyways.

PR #917, opened 2025-01-31
Add support for boolean fields in CsvImportMigrationBase::addFieldMapping()

mstenta:
This adds support for boolean field types in `CsvImportMigrationBase::addFieldMapping()`, which is the method used to dynmically add base and bundle fields to the default farmOS CSV importers. This means that bundle fields of type `boolean` will be automatically added, and base fields can be added by implementing `hook_farm_import_csv_base_fields()` (added in #916).

This PR builds on top of #916, so it includes all of that PR's commits as well. This PR is technically only one commit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [Add support for geometry fields in CSV exports #884](https://github.com/farmOS/farmOS/pull/884)
 - [Allow modules to add base fields to farmOS Views #915](https://github.com/farmOS/farmOS/pull/915)
 - [Allow modules to add base fields to default CSV importers #916](https://github.com/farmOS/farmOS/pull/916)
+- [Add support for boolean fields in CsvImportMigrationBase::addFieldMapping() #917](https://github.com/farmOS/farmOS/pull/917)
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


paul121 (commented):
+1 to supporting boolean fields... just trying to think how this works.

What values does it accept? are `true/false`, `null`, `0/1` all valid values? I hope an empty string `''` or `' '` doesn't result in populate the field as `false` ?

mstenta:
> +1 to supporting boolean fields... just trying to think how this works.
> 
> What values does it accept? are true/false, null, 0/1 all valid values?

We added this `boolean` plugin in #815. It uses PHP's built-in `filter_var($value, FILTER_VALIDATE_BOOLEAN)`.

This comment has some more info: https://github.com/farmOS/farmOS/pull/815#discussion_r1766870455

> I hope an empty string '' or ' ' doesn't result in populate the field as false ?

I think it results in the default value being used for that field. We use that to our advantage in the `is_movement` field of log importers (also added by PR #815), so `seeding` logs default to `TRUE` and `activity` logs default to `FALSE`, but you can optionally override that in the CSV.

I'll give it a quick test to confirm this...

mstenta:
> I'll give it a quick test to confirm this...

Tested leaving `is_group_assignment` blank on an `activity` log import, and it ends up as `FALSE` on the log.

mstenta:
Just remembered how we're handling this:

https://github.com/farmOS/farmOS/blob/31c50a55a949018f8726ffd4674e3286b04b229e/modules/core/migrate/src/Plugin/migrate/process/Boolean.php#L38-L42

So a boolean field that defaults to `TRUE` will be set to `TRUE` if the field is left blank in the CSV.

paul121 (approved):

PR #921, opened 2025-02-05
Disable fail-fast strategy for automated tests

paul121:
Currently if any job in the `test` job matrix fails, other in-progress and queued jobs in the matrix are cancelled. This PR changes that behavior to disable the [`fail-fast`](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast) strategy and let all `test` matrix jobs run, even if one failed. This does not cause the `test` job itself to pass in the event any one job fails.

The current behavior is frustrating when actively "debugging" some changes and the tests against one DB randomly fail, causing you to manually restart jobs to see test results again. We should still re-run the tests to make sure they pass on all DBs, but this will save time when debugging and chasing errors.

I've also been experiencing an issue lately where PHPUnit errors thrown in Browser test cases (specifically those searching for html/text on the page, I think?) crash with a generic error when run with `paratest`. But these same errors return the standard test report w/ useful info when run with `phpunit`, which we currently only run in our `sqlite` matrix dbms case. As a result I have been turning off all `test` jobs except for `sqlite` in my draft PRs. Turning off `fail-fast` will help when debugging these situation as well.

### DIFF
--- a/.github/workflows/deliver.yml
+++ b/.github/workflows/deliver.yml
@@ -120,6 +120,7 @@ jobs:
     runs-on: ubuntu-latest
     needs: build
     strategy:
+      fail-fast: false
       matrix:
         dbms:
          - pgsql


mstenta (approved):

mstenta:
Makes sense to me!

paul121:
> I've also been experiencing an issue lately where PHPUnit errors thrown in Browser test cases (specifically those searching for html/text on the page, I think?) crash with a generic error when run with paratest. But these same errors return the standard test report w/ useful info when run with phpunit, which we currently only run in our sqlite matrix dbms case. As a result I have been turning off all test jobs except for sqlite in my draft PRs. Turning off fail-fast will help when debugging these situation as well.

This will be fixed by increasing the `shm_size` in the selenium chrome container: https://github.com/farmOS/farmOS/pull/967/commits/98395b2b00c977166e94c4e15501e1a49fe7d7b4

We could consider reverting this change re: fail-fast strategy, but IMO this still is useful for cases when there are random failures. It doesn't hurt to let all the tests run to completion.

PR #920, opened 2025-02-05
Update oauth2 static scopes for pluggable granularity

paul121:
The next release of simple_oauth will have support for pluggable granularity on oauth2 scopes: https://www.drupal.org/node/3492888

We provide some static oauth2 scopes and need to update them to use `granularity_id` and `granularity_configuration`. This old `granularity` property is deprecated until v7 but we might as well fix it now: https://git.drupalcode.org/project/simple_oauth/-/blob/3ab9ede24b45016d9059877c3a901150309588a6/modules/simple_oauth_static_scope/src/Plugin/Oauth2ScopeManager.php#L126

These changes have already been committed to simple_oauth 6.0.x so I added a temp commit to unpin the dependency for our testing in the meantime.

PR #922, opened 2025-02-07
Implement ManagedRolePermissions access policy and simple_oauth scope granularity

paul121:
For more info see https://www.drupal.org/project/farm/issues/3476250

This PR includes the commits from #920 - we need a new release of simple_oauth before we can implement the scope granularity parts of this PR. Hopefully this will happen in time for our next 3.4.0 release!

Implementing the access policy was pretty straightforward. I just changed `ManagedRolePermissionsManager::getManagedPermissionsForRole()` to be a public method so we have an easy way to get the full list of permissions. Previously we only had to check if a single permission was included on the role.

The Simple OAuth scope granularity is pretty straightforward as well, it's just changes in tests that get a little more involved. Basically, this updates `FarmApiTest` to make all of it's JSON:API requests using OAuth2 authorization instead of session auth (as mentioned [here](https://www.drupal.org/project/farm/issues/3476250#comment-15784842). I also added some tests to ensure that requests fail with `403` when users do not have sufficient permission.

Last, I added a simple test for the scope granularity plugin itself and a test case for the caching of managed role permissions (the access policy API makes this much more robust!)


modules/core/api/tests/src/Kernel/FarmApiTest.php
```diff
@@ -78,26 +82,42 @@ public function setUp(): void {
...
+
+    // Configure client with password grant and no default scopes.
+    $this->client
+      ->set('scopes', [])
+      ->set('grant_types', ['password'])
+      ->save();
+
+    // Create a second user to ensure we are not using user id 1.
+    $this->user = $this->createUser();
+    $this->assertEquals(2, $this->user->id());
```
paul121: This was needed because `simple_oauth` has special logic to grant all access for user number 1: https://git.drupalcode.org/project/simple_oauth/-/blob/2a553c1c1a6a4a269a056f493024034463331490/src/Authentication/TokenAuthUser.php#L90-93

I think simple_oauth will need to update this at some point. But for now this is an easy work around and shouldn't be impacted by changes in the future.

modules/core/role/src/Access/ManagedRolePermissionsAccessPolicy.php
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Grants permissions based on the user's role and managed role permissions.
+ */
+final class ManagedRolePermissionsAccessPolicy extends AccessPolicyBase {
+
+  public function __construct(
+    protected EntityTypeManagerInterface $entityTypeManager,
+    protected ManagedRolePermissionsManagerInterface $managed_role_permissions_manager,
+  ) {}
```
paul121, line 22: @mstenta not sure if you had heard about this, but we can start leveraging "constructor property promotion" to decrease a lot of the boilerplate required around the `__construct` method! https://www.drupal.org/project/drupal/issues/3278431

I remembered this when referencing/copying from the core [`UserRolesAccessPolicy`](https://git.drupalcode.org/project/drupal/-/blob/10.4.2/core/lib/Drupal/Core/Session/UserRolesAccessPolicy.php?ref_type=tags)

Also, comment docblocks for `__construct()` can be omitted: https://www.drupal.org/node/3365111

modules/core/role/src/Plugin/ScopeGranularity/ManagedRolePermissions.php
```diff
@@ -0,0 +1,84 @@
...
+use Drupal\simple_oauth\Plugin\ScopeGranularity\Role;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * The Managed Role Permissions scope granularity plugin.
+ */
+#[ScopeGranularity(
+  'managed_role_permissions',
+  new TranslatableMarkup('Managed Role Permissions'),
+)]
```
paul121, line 21: I'm not sure what is the best naming for this scope granularity plugin. Developers (static scopes) and Users (scope config entities via UI) just need to know that they need to use a "Managed Role Permissions" granularity instead of "Role" granularity if they want any of the _role's managed permissions_ to be considered when authorizing via oauth.

paul121:
> we need a new release of simple_oauth before we can implement the scope granularity parts of this PR. Hopefully this will happen in time for our next 3.4.0 release!

Hmm I'm just thinking... technically there is a breaking change here. If someone has configured an OAuth2 scope to use a role that has managed role permissions, they must change this to use the new `managed_role_permissions` granularity provided by this PR. Otherwise OAuth2 requests will not be granted these "managed role permissions" and could break existing API integrations.

Maybe we keep our `FarmRoleStorage` override to preserve this functionality until we can make a breaking change in v4.0? :thinking:

mstenta:
Removed this from the 3.4.0 milestone since a new release of `simple_oauth` hasn't been made yet, and we are otherwise ready to tag 3.4.0.

mstenta:
Rebased onto updated #920, which is about to be merged.

mstenta:
I changed the base branch of this PR to `4.x`, and rebased/resolved merge conflicts.

modules/core/role/src/Plugin/ScopeGranularity/ManagedRolePermissions.php
paul121: @mstenta mentioned just naming this `managed_role`. I think that makes sense to me. In documentation we do have the concept of "Managed Roles" but less so in the code here

mstenta: I've pushed a commit that changes this.

mstenta:
> changed `ManagedRolePermissionsManager::getManagedPermissionsForRole()` to be a public method

Since we're making this public I suggest we rename it to `getManagedRolePermissions()`. I've pushed a commit that does this.

modules/core/role/src/Access/ManagedRolePermissionsAccessPolicy.php
```diff
@@ -0,0 +1,51 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function calculatePermissions(AccountInterface $account, string $scope): RefinableCalculatedPermissionsInterface {
+    $calculated_permissions = parent::calculatePermissions($account, $scope);
+
+    /** @var \Drupal\user\RoleInterface[] $user_roles */
+    $user_roles = $this->entityTypeManager->getStorage('user_role')->loadMultiple($account->getRoles());
```
paul121, line 31: There might be an issue here. Wondering if we need to make sure we include "standard" role permissions in this logic :thinking: 

Reason being, our `ManagedRolePermissionManager::getManagedPermissionsForRole($role)` function only returns the "managed role" permission strings, none of the "standard" role permissions that you are still able to declare with a "managed role". This works fine when providing our own `FarmRoleStorage` class because we delegate to the `parent::isPermissionInRoles()` function to cover the standard role permission logic, and add "managed permission" logic in a second step.

However, this access policy logic doesn't cover the standard role permissions. At first I was wondering if we could simply extend the core `UserRolesAccessPolicy` but that class is `final`, so we cannot.

But thinking more, assuming that all access policies are checked (in some order of priority), I guess this wouldn't be an issue for standard browser session auth. The `UserRolesAccessPolicy` will still be run to cover the "standard" role permissions. BUT... this might be an issue for the `managed_role` OAuth Scope Granularity case....

modules/core/role/src/Access/ManagedRolePermissionsAccessPolicy.php
```diff
@@ -0,0 +1,51 @@
...
+      $calculated_permissions
+        ->addItem(new CalculatedPermissionsItem(
+          $this->managed_role_permissions_manager->getManagedPermissionsForRole($role),
+          $role->isAdmin(),
+        ))
+        ->addCacheableDependency($role);
+    }
+
+    return $calculated_permissions;
+  }
```
paul121, line 42: There might be an issue here. Wondering if we need to make sure we include "standard" role permissions in this logic :thinking: 

Reason being, our `ManagedRolePermissionManager::getManagedPermissionsForRole($role)` function only returns the "managed role" permission strings, none of the "standard" role permissions that you are still able to declare with a "managed role". This works fine when providing our own `FarmRoleStorage` class because we delegate to the `parent::isPermissionInRoles()` function to cover the standard role permission logic, and add "managed permission" logic in a second step.

However, this access policy logic doesn't cover the standard role permissions. At first I was wondering if we could simply extend the core `UserRolesAccessPolicy` but that class is `final`, so we cannot.

But thinking more, assuming that all access policies are checked (in some order of priority), I guess this wouldn't be an issue for standard browser session auth. The `UserRolesAccessPolicy` will still be run to cover the "standard" role permissions. BUT... this might be an issue for the `managed_role` OAuth Scope Granularity case....

paul121: > BUT... this might be an issue for the managed_role OAuth Scope Granularity case....

Oh! Our `class ManagedRole extends Role` and uses that parent logic - so that shouldn't be an issue. I think this is all OK. We should make sure the tests cover both of these "standard" permission cases

modules/core/role/tests/src/Kernel/ScopeGranularityTest.php
paul121, line 1: Let's make sure we include a "standard" permission in this new test, as well as for the existing `ManagedRolePermissionsTest` case. Neither of the existing `farm_test` and `farm_test_manager` roles that this PR uses include one in their `permissions: { }` list.

I think we could simply add a `standard role permission` to one or both of these roles and add a check for it in tests. @mstenta not sure how this is impacted by other refactoring of Role and API modules...

paul121: I've added a permission and tests that cover this!

paul121:
Okay, just force-pushed some changes here. This changes how the OAuth tests were working by adding a new `FarmApiOauthTestBase` class. This class contains all the changes required to get OAuth working, and `FarmApiOauthRoleTest` adds a new test case for `farm_viewer` access via API.

I also added a couple test cases to just make sure "standard role permissions" are sill compatible with our managed role access policy and scope granularity. All was working as expected!

mstenta:
LGTM! Rebased onto latest 4.x. Let's confirm that tests pass...

mstenta (changes requested):
Linting and tests are failing...

paul121:
The test failures are due to a 500 error when getting an OAuth access token. I believe it is related to what PHP Stan caught:

> Access to an undefined property                                                    
         Drupal\Tests\farm_api_oauth\Kernel\FarmApiOauthTestBase::$scope.  
         
Will see if adding the property fixes this..

paul121:
Hmm that didn't fix it. Same errors, will have to look at this later.

mstenta:
Rebased onto latest `4.x` commit, and updated for #982.

mstenta:
Fixed the failing tests...

- Renamed new `apiRequest()` calls (added in #988) to `assertApiRequest()`.
- Installed the new `farm_viewer` module (added in #982).
- Set up the `private://` filesystem in `FarmApiOauthTestBase` to fix warnings from `SimpleOauthTestTrait::setUpKeys()`.

I also merged the phpcs and phpstan fix commits into the earlier commits they were fixing, and fixed another new one.

mstenta (approved):

mstenta:
This is all ready to merge, but I'm going to wait until #1005, #1006, and #1007 go in, since those make changes to some of the files/hooks that this touches.

I've started a `4.x-access-policy` branch that is rebased on top of `4.x-hooks` (#1007) and resolved the merge conflicts, so it's all ready to go.

paul121:
> Fixed the failing tests...

Are those bullet points what fixed the tests or was it something else? It would make sense if it was the bullet points since all of those changed in recent refactors but also surprised it wouldn't have been more obvious. Just am particularly curious as this relates to OAuth and want to understand fully.

mstenta:
> Are those bullet points what fixed the tests or was it something else?

Yes! It was the first two bullet points that fixed the tests. The third one fixed warnings, but the tests passed regardless.

paul121:
Great! This looks good then.

mstenta:
Great thanks for reviewing @paul121! I'll wait for #1007 to merge then I'll merge this.

mstenta:
> This is all ready to merge, but I'm going to wait until #1005, #1006, and #1007 go in, since those make changes to some of the files/hooks that this touches.
> 
> I've started a `4.x-access-policy` branch that is rebased on top of `4.x-hooks` (#1007) and resolved the merge conflicts, so it's all ready to go.

#1007 has been merged. I've synced this PR's `3.x-access-policy` branch with my `4.x-access-policy` branch and force pushed it back to the PR. Let's confirm tests pass and then merge!

PR #923, opened 2025-02-11
Pin PHP version to 8.3 in Dockerfile

mstenta:
This PR pins the PHP version in the official farmOS Docker images to 8.3. This will allow us to be more intentional about updating the PHP version moving forward.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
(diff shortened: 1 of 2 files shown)


mstenta:
This was inspired by the fact that we accidentally updated to PHP 8.3 when we updated to the `drupal:10.4` image.

And it looks like 10.5 will default to PHP 8.4: https://github.com/docker-library/drupal/pull/281#pullrequestreview-2607443037

Maybe we'll end up following upstream anyway, but IMO it's better to be explicit about it, and mention it in our release notes when we do.

paul121 (approved):

PR #925, opened 2025-02-11
Add farmOS-map snapshot behavior

paul121:
This is dependent on this PR being included in a new release of farmOS-map: https://github.com/farmOS/farmOS-map/pull/202 but opening a PR to start reviewing the integration of that change here...

I think we have two options:

1. Enable the new Snapshot behavior on all farmOS-map instances (very simple)
2. Make it a configurable option site-wide in farmOS, doing so would make it possible to turn it on/off, but would affect all users. A little more work, but something we already do, so it's not introducing much more complexity for us to maintain.

Thoughts? @mstenta @symbioquine @Farmer-Eds-Shed @Fosten 

In this PR I have implemented option 1. I think there is minimal impact/clutter and little issue to add this to all maps.

### DIFF
--- a/modules/core/map/js/farm_map.js
+++ b/modules/core/map/js/farm_map.js
@@ -38,6 +38,9 @@
       };
       const instance = farmOS.map.create(element, mapOptions);
 
+      // Enable the snapshot behavior.
+      instance.addBehavior("snapshot");
+
       // Expose settings on the instance so other behaviors don't need to know how to look them up in drupalSettings
       instance.farmMapSettings = drupalSettings.farm_map[drupalSettingsKey] || {};
 


paul121:
Here is what the Snapshot icon looks like. Clicking the camera/snapshot expands the two options to save/download a JPEG, or print (which makes it easy to save as PDF):
Screenshot from 2025-02-11 12-51-02

mstenta:
This is great!

One thought might be: do we want to make this option configurable, like we do with the sidebar behavior?

https://github.com/farmOS/farmOS/blob/3.x/modules/core/map/js/farmOS.map.behaviors.enable_side_panel.js

https://github.com/farmOS/farmOS/blob/d999b1b2e2bbf3a92bb945ae0b5ff50510618f84/modules/core/map/src/Form/MapSettingsForm.php#L44-L50

mstenta:
Another thing to test: does the new button overlap other buttons on small screens?

paul121:
Seems OK when rendering for Pixel 7 screen size in my browser:
Screenshot from 2025-02-13 09-23-57

paul121:
Closing this in favor of #926 

On the dev call today we discussed doing more testing of this feature before enabling it by default for everyone

PR #928, opened 2025-02-17
Delete associated plan_record entities when a plan is deleted

mstenta:
This adds some logic to the `plan` module to automatically delete `plan_record` entities that are associated with a `plan` entity when the plan is deleted.

As of right now it's possible to delete a plan without cleaning up the `plan_record` entities (and there is no UI for deleting `plan_record` entities), resulting in orphaned records in the database.

More context/discussion: https://farmos.discourse.group/t/module-beta-release-guidance-and-best-practices/2219/12

### DIFF
--- a/modules/core/plan/plan.module
+++ b/modules/core/plan/plan.module
@@ -33,9 +33,14 @@ function plan_help($route_name, RouteMatchInterface $route_match) {
 function plan_plan_delete(EntityInterface $entity) {
 
   // Delete all plan_record entities associated with the plan.
-  $plan_records = \Drupal::entityTypeManager()->getStorage('plan_record')->loadByProperties(['plan' => $entity->id()]);
-  foreach ($plan_records as $plan_record) {
-    $plan_record->delete();
+  $plan_record_storage = \Drupal::entityTypeManager()->getStorage('plan_record');
+  $plan_ids = $plan_record_storage->getQuery()
+    ->condition('plan', $entity->id())
+    ->accessCheck(FALSE)
+    ->execute();
+  foreach ($plan_ids as $id) {
+    $plan_record = $plan_record_storage->load($id);
+    $plan_record?->delete();
   }
 }
 
--- a/modules/core/plan/plan.module
+++ b/modules/core/plan/plan.module
@@ -38,9 +38,12 @@ function plan_plan_delete(EntityInterface $entity) {
     ->condition('plan', $entity->id())
     ->accessCheck(FALSE)
     ->execute();
-  foreach ($plan_ids as $id) {
-    $plan_record = $plan_record_storage->load($id);
-    $plan_record?->delete();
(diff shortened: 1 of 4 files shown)


modules/core/plan/plan.module
```diff
@@ -26,6 +27,18 @@ function plan_help($route_name, RouteMatchInterface $route_match) {
...
+ */
+function plan_plan_delete(EntityInterface $entity) {
+
+  // Delete all plan_record entities associated with the plan.
+  $plan_records = \Drupal::entityTypeManager()->getStorage('plan_record')->loadByProperties(['plan' => $entity->id()]);
+  foreach ($plan_records as $plan_record) {
+    $plan_record->delete();
+  }
+}
+
```
paul121: This LGTM. The only issue I can think of is if a plan had many many plan record entities associated with it, this might crash when loading all of them? I'm sure if we realistically need to worry about this though. We could probably do better by loading just the plan_record entity IDs into an array, and then use each ID to load and delete entity objects one at a time

Reminds me of this issue https://www.drupal.org/project/drupal/issues/2577417

paul121 (approved):

modules/core/plan/plan.module
mstenta: I had a similar thought @paul121. I wondered about using Batch API but I'm not sure that's possible in `hook_ENTITY_TYPE_delete()`. I like your suggestion of loading them one at a time. I've pushed a commit that refactors the hook to do that (loading IDs via an entity query first).

Tests are still passing. Re-requesting your review...

paul121 (approved):

modules/core/plan/plan.module
```diff
@@ -26,6 +27,23 @@ function plan_help($route_name, RouteMatchInterface $route_match) {
...
+ */
+function plan_plan_delete(EntityInterface $entity) {
+
+  // Delete all plan_record entities associated with the plan.
+  $plan_record_storage = \Drupal::entityTypeManager()->getStorage('plan_record');
+  $plan_ids = $plan_record_storage->getQuery()
+    ->condition('plan', $entity->id())
+    ->accessCheck(FALSE)
+    ->execute();
+  foreach ($plan_ids as $id) {
```
wotnak: this can be optimized a bit by processing in chunks instead of every entity separately, for example something like:
```php
if (!is_array($plan_ids) || count($plan_ids) < 1) {
  return;
}
foreach (array_chunk($plan_ids, 100) as $chunk) {
  $plan_records = $plan_record_storage->loadMultiple($chunk);
  $plan_record_storage->delete($plan_records); 
}
```

mstenta: Good idea @wotnak. I've added another commit. Tests pass, but please review+approve when you have a chance. Thanks!

https://github.com/farmOS/farmOS/pull/928/commits/a06087645c29b715591a6706bd0d045ffec34d4e

wotnak (approved):

PR #931, opened 2025-02-27
Cast processed text exported to CSV as a string

mstenta:
This fixes the following error that occurs when "Export processed text" is selected from the CSV export action:

`Uncaught PHP Exception TypeError: "Drupal\farm_csv\Normalizer\TextLongFieldItemNormalizer::normalize(): Return value must be of type ArrayObject|array|string|int|float|bool|null, Drupal\filter\Render\FilteredMarkup returned" at /opt/repos/farmOS/modules/core/csv/src/Normalizer/TextLongFieldItemNormalizer.php line 32`

This started happening when strict typing was enabled in #902.

I noticed it in Farmier logs.


mstenta:
PHPStan did not catch this because it was returning `$field_item->get('processed')->getValue()`, and it has no way of knowing what type that is. So I also added a `/** @var \Drupal\filter\Render\FilteredMarkup $processed_text */` comment to declare the type before casting it to a string.

PR #932, opened 2025-03-07
Fix TypeError on invalid /asset/%asset_id/assets path

mstenta:
Fix `TypeError: "Drupal\farm_location\AssetLocation::isLocation(): Argument #1 ($asset) must be of type Drupal\asset\Entity\AssetInterface, null given`

This happens if you try to visit /asset/%asset_id/assets with a non-numeric %asset_id.

This appeared in Farmier logs. It's a pretty obscure bug because it requires a user to manually type in an invalid URL, but no harm in making our logic a bit more defensive to prevent fatal errors.

Notably, this also changes the return from `AccessResult::allowed()` to `AccessResult::forbidden()`. I'm not sure why it was originally defaulting to an "allowed" return. That doesn't make much sense. I wrote the logic in 2021, so I take the blame. :-)


mstenta:
This caused a bug. Opened a PR to fix/revert it: https://github.com/farmOS/farmOS/pull/951

PR #938, opened 2025-03-24
Fix TypeError: CircularAssetLocationConstraintValidator::getAssetsByLocationRecursively(): Argument #2 ($timestamp) must be of type int, string given

mstenta:
Saw this error in Farmier logs:

`TypeError: Drupal\farm_location\Plugin\Validation\Constraint\CircularAssetLocationConstraintValidator::getAssetsByLocationRecursively(): Argument #2 ($timestamp) must be of type int, string given, called in /opt/drupal/web/profiles/farm/modules/core/location/src/Plugin/Validation/Constraint/CircularAssetLocationConstraintValidator.php on line 77 in Drupal\farm_location\Plugin\Validation\Constraint\CircularAssetLocationConstraintValidator->getAssetsByLocationRecursively() (line 120 of /opt/drupal/web/profiles/farm/modules/core/location/src/Plugin/Validation/Constraint/CircularAssetLocationConstraintValidator.php).`

We should be casting the log timestamp to an integer before passing it to `CircularAssetLocationConstraintValidator::getAssetsByLocationRecursively()`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - [Fix Warning: Undefined array keys in hook_views_pre_render() #937](https://github.com/farmOS/farmOS/pull/937)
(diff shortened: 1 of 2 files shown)


paul121 (approved):
Interesting that PHPStan didn't catch this! All the entity field level types are challenging but it seems it might default to strings and at least give a warning? :shrug:

Issue #939, opened 2025-03-24
Issues with image orientation


mstenta:
Automatic image rotation has been broken for a while. I dug into it a bit this morning and I want to document what I've learned here.

Background: when a picture is taken vertically with a cell phone camera, it is often saved horizontally but contains an `Orientation` EXIF attribute that tells software how it should be displayed.

Source: https://www.howtogeek.com/254830/why-your-photos-dont-always-appear-correctly-rotated/

> Unfortunately, there's a small caveat. Digital camera hardware just couldn't handle saving the image directly in rotated form. So rather than performing the computationally intensive task of rotating the entire image, the camera would add a small piece of data to the file, noting which orientation the image should be in. It adds this information to the [Exif data](https://www.howtogeek.com/203592/what-is-exif-data-and-how-to-remove-it/) that all photos have (which includes the model of camera you took it with, the orientation, and possibly even [the GPS location where the photo was taken](https://www.howtogeek.com/211427/how-to-see-exactly-where-a-photo-was-taken-and-keep-your-location-private/)).
> 
> In theory, then, you could open that photo with an application, it would look at the Exif tags, and then present the photo in the correct rotation to you. The image data is saved in its original, unrotated form, but the Exif tag allows applications to correct it.

farmOS includes the Drupal [EXIF Orientation](https://www.drupal.org/project/exif_orientation) module, which has logic in it to read this `Orientation` EXIF attribute and perform the necessary rotation on the image file when it is saved.

However, this is not currently working. Attached is a test image file that you can use to replicate it. Upload this to any `image` field in farmOS (running in the official farmOS Docker image) and notice that it does not get rotated.

The reason for this is very simple (and worthy of the face-palm emoji): it relies on the `exif_read_data()` function, which is provided by the `exif` PHP extension, which is not installed by default in the `drupal` Docker image that we build our image on. 🤦

Therefore, the fix is very simple: we just need to add `docker-php-ext-install exif` to our `Dockerfile`. I've tested this locally and it fixes the issue.

HOWEVER... while I was investigating this, I learned that the act of rotating the image **strips all other EXIF data from the image file!** farmOS should always strive to maintain original data, so this is bad, and we should *not* fix this rotation issue without first preventing that data loss.

[...]

PR #940, opened 2025-03-24
Install and configure ImageMagick

mstenta:
See: #939

This installs+configures the `imagick` PHP extension and Drupal module.

The end result is that EXIF information is preserved in derivative images generated by Drupal.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,6 +31,7 @@
         "drupal/fraction": "^3.0",
         "drupal/geofield": "^1.40",
         "drupal/gin": "4.0.6",
+        "drupal/imagick": "^1.12",
         "drupal/inline_entity_form": "^1.0@RC",
         "drupal/inspire_tree": "^1.0",
         "drupal/jsonapi_extras": "^3.22",
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -21,6 +21,8 @@ RUN apt-get update && apt-get install -y \
     postgresql-client \
     # Install libgeos-c1v5 so geos php extension can use libgeos_c.so.1.
     libgeos-c1v5 \
+    # Install libmagickwand-dev for ImageMagick. \
+    libmagickwand-dev \
   && rm -rf /var/lib/apt/lists/* \
   && apt-get clean
 
@@ -36,6 +38,17 @@ RUN apt-get update && apt-get install -y libgeos-dev \
   && ( tar xzf /opt/php-geos.tar.gz -C /opt/ \
     && cd /opt/php-geos-${PHP_GEOS_VERSION} \
     && ./autogen.sh \
+    && ./configure \
+    && make \
+    && make install \
+    )
+
+# Build and install the ImageMagick PHP extension.
+ARG IMAGICK_VERSION=52ec37ff633de0e5cca159a6437b8c340afe7831
+ADD https://github.com/Imagick/imagick/archive/${IMAGICK_VERSION}.tar.gz /opt/php-imagick.tar.gz
+RUN ( tar xzf /opt/php-imagick.tar.gz -C /opt/ \
+    && cd /opt/imagick-${IMAGICK_VERSION} \
+    && phpize \
     && ./configure \
     && make \
     && make install \
@@ -59,7 +72,7 @@ RUN echo "ServerName localhost" >> /etc/apache2/apache2.conf
 
 # Enable PHP dependencies.
 COPY --from=php-dependencies /usr/local/lib/php/extensions/ /usr/local/lib/php/extensions/
-RUN docker-php-ext-enable bcmath exif geos
+RUN docker-php-ext-enable bcmath exif geos imagick
 
 # Add custom PHP configurations.
 COPY conf.d/ /usr/local/etc/php/conf.d
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -13,6 +13,17 @@ FROM drupal:10.4-php8.3 AS baseimage
 ENV BUILD_PATH=/var/farmOS
 ENV DRUPAL_PATH=/opt/drupal
 
+# Install apt dependencies.
+RUN apt-get update && apt-get install -y \
+    # Install git and unzip (needed by Composer).
+    git unzip \
+    # Install postgresql-client so Drush can connect to the database.
+    postgresql-client \
+    # Install libgeos-c1v5 so geos php extension can use libgeos_c.so.1.
+    libgeos-c1v5 \
+  && rm -rf /var/lib/apt/lists/* \
+  && apt-get clean
+
 ##
 # Build/install PHP extensions.
 FROM baseimage AS php-dependencies
@@ -52,17 +63,6 @@ RUN docker-php-ext-enable bcmath exif geos
 
 # Add custom PHP configurations.
 COPY conf.d/ /usr/local/etc/php/conf.d
-
-# Install apt dependencies.
-RUN apt-get update && apt-get install -y \
-    # Install git and unzip (needed by Composer).
-    git unzip \
-    # Install postgresql-client so Drush can connect to the database.
-    postgresql-client \
-    # Install libgeos-c1v5 so geos php extension can use libgeos_c.so.1.
-    libgeos-c1v5 \
-  && rm -rf /var/lib/apt/lists/* \
-  && apt-get clean
 
 # Set the COMPOSER_MEMORY_LIMIT environment variable to unlimited.
 ENV COMPOSER_MEMORY_LIMIT=-1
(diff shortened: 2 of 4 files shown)


mstenta:
I removed the commit that uninstalls the Drupal `exif_orientation` module. That will be part of a separate PR focused on image orientation. This PR is *only* concerned with installing and configuring ImageMagick now.

mstenta:
Note: If we merge #941, and then merge this PR, we'll need to remove/reset the `Orientation` EXIF value on the derivative image that's saved by the `farm_image_auto_orient` effect. Otherwise, the `Orientation` will still be set in the derivative image, but it will be wrong, and browsers will auto-rotate the image *again*.

I think we should handle that in this PR, because ImageMagick provides an easy way of doing this, whereas doing it in vanilla PHP is more difficult.

See: https://git.drupalcode.org/project/imagick/-/blob/373da420d5fff9f7a8a9d5dca46846c28e3797c2/src/Plugin/ImageToolkit/Operation/imagick/Autorotate.php#L65

(We can also consider dropping our custom `farm_image_auto_orient` effect at that time, and using the one that comes with the `imagick` module... but that would mean adding a hard dependency on ImageMagick.)

symbioquine:
> (We can also consider dropping our custom `farm_image_auto_orient` effect at that time, and using the one that comes with the `imagick` module... but that would mean adding a hard dependency on ImageMagick.)

I'd argue for something slightly different...

With farmOS 4 we should have a soft dependency on ImageMagick such that, if available, it is used for producing the downsized preview/thumbnail images and thus keeping the image metadata intact. This allows image rotation (and other display concerns like colors) to be handled at display time.

Other than that, I don't think we should be manipulating the images on the server side. i.e. We shouldn't be auto-rotating them there at all - except possibly as a fallback if ImageMagick is missing and we have no choice about losing the metadata while downsizing images with GD.

IMHO, the image dimension problem you [identified in #939](https://github.com/farmOS/farmOS/issues/939#issuecomment-2749211290) is a separate bug.

mstenta:
> IMHO, the image dimension problem you https://github.com/farmOS/farmOS/issues/939#issuecomment-2749211290 is a separate bug.

Perhaps. I worry that it opens the possibility for more issues. I suppose the first step is figuring out what the fix is for the bug I highlighted. I suspect digging into that will reveal the potential complexities of other contexts. I hope that I'm pleasantly surprised though! :-)

paul121:
> The end result is that EXIF information is preserved in derivative images generated by Drupal.

Hmm trying to catch up and I guess I'm a little lost now. Is this really the TLDR? Do you mean that with #941 the derivative `webp` image that is generated after applying effects loses the EXIF metadata from original image?

Seeing this comment I suppose this is the case, just surprising to me! https://github.com/farmOS/farmOS/issues/939#issuecomment-2749354895

modules/core/entity/modules/fields/farm_entity_fields.install
```diff
@@ -9,6 +9,33 @@ declare(strict_types=1);
...
+
+    // Install the imagick module.
+    if (!\Drupal::service('module_handler')->moduleExists('imagick')) {
+      \Drupal::service('module_installer')->install(['imagick']);
+    }
+
+    // Make imagick the default image toolkit.
+    $config = \Drupal::configFactory()->getEditable('system.image');
+    $config->set('toolkit', 'imagick');
+    $config->save();
```
paul121: What is the default toolkit?

Is the `imagick` toolkit provided by the drupal module or the extension?

composer.json
```diff
@@ -32,6 +32,7 @@
         "drupal/fraction": "^3.0",
         "drupal/geofield": "^1.40",
         "drupal/gin": "4.0.6",
+        "drupal/imagick": "^1.12",
```
paul121, line 34: I see there is an alpha v2 release, just curious if we should be tracking what is changing in the major release and make sure it will work for us going forward

mstenta:
> Do you mean that with #941 the derivative `webp` image that is generated after applying effects loses the EXIF metadata from original image?
> 
> Seeing this comment I suppose this is the case, just surprising to me! [#939 (comment)](https://github.com/farmOS/farmOS/issues/939#issuecomment-2749354895)

Yes. EXIF data is lost whenever the GD toolkit is used to create (or edit) an image. GD is the default toolkit that is included with PHP, and it does not handle EXIF data at all.

> When using the current GD methodologies, you are reading content from an image and manipulating it.  By then writing that content to a brand new file, you are losing the EXIF data.
> ...
> For purposes when you want to retain EXIF data, it is recommended that you compile in and use the PECL Imagick extension.

Source: https://www.php.net/manual/en/book.image.php#87566

> GD doesn't deal with Exif of IPTC data at all as its beyond the scope of what GD does.

Source: https://stackoverflow.com/a/10182983

modules/core/entity/modules/fields/farm_entity_fields.install
mstenta: Drupal uses `gd` by default.

https://git.drupalcode.org/project/drupal/-/blob/11.x/core/modules/system/config/install/system.image.yml?ref_type=heads#L1

GD is included with PHP (https://www.php.net/manual/en/book.image.php), whereas ImageMagick is a third-party extension that needs to be installed along with a Drupal module to add it as an option in Drupal.

The PHP extension adds the `Imagick` class, with methods for image manipulation: https://www.php.net/manual/en/book.imagick.php

The Drupal module provides a plugin that serves as a wrapper around the `Imagick` class methods so Drupal knows how to use them.

There are currently two Drupal contrib modules that provide ImageMagick integration, but only one of them uses the PHP extension. The other one bypasses PHP entirely and runs shell commands to achieve the same goals.

- https://www.drupal.org/project/imagick (uses the `imagick` PHP extension)
- https://www.drupal.org/project/imagemagick (uses shell commands)

This PR installs the `imagick` PHP extension and installs the `imagick` Drupal module to make use of it.

composer.json
mstenta: Agreed - and in general we should review the `imagick` module in more detail. I'm not thrilled about adding the dependency, to be honest. And with #941 we will no longer be modifying original files anymore (that I know of), so really this PR is just about maintaining EXIF data in derivative images. The two main issues this solves are orientation (and allowing browsers to auto-rotate) and color profiles (there may be others). #941 solves the first issue by rotating the derivative images directly, and the second issue has never been reported in practice, so may not be a high priority to solve.

So I think our decision should be based on whether or not the benefits (maintaining EXIF data in derivative images) are worth the tradeoffs (adding both the `imagick` PHP extension and `imagick` Drupal module as dependencies).

mstenta:
I just noticed something that this breaks...

By default, Drupal core's stock image styles (`large`, `medium`, `thumbnail`, and `wide`) include an effect to convert images to `webp` format.

Screenshot 2025-03-26 at 08-14-44 Edit Convert effect on style Large (480×480) Drush Site-Install

After installing the `imagick` module, this effect appears to be broken. Images are no longer converted to `webp`, and the configuration form is replaced with a different one, and seems to be showing default values (`3fr` is the first option in the dropdown).

Screenshot 2025-03-26 at 08-21-17 Edit Convert effect on style Large (480×480) Drush Site-Install

And it seems that the images are being converted to JPEGs (even when they are already JPEGs). They end up with an additional `.JPG` extension. For example:

- An uploaded JPG image becomes ends up with a double extension: http://localhost/system/files/styles/large/private/farm/log/2025-03/1000010267_5.jpg.JPG?itok=BoYQI3DD
- An uploaded WEBP image becomes a JPG image: http://localhost/system/files/styles/large/private/farm/log/2025-03/1000010267_5.webp.JPG?itok=G3JBQcxQ

This seems like it must be a bug in the `imagick` module. :thinking:

mstenta:
> This seems like it must be a bug in the `imagick` module. 🤔

Switching the Drupal image toolkit back to `gd` does not fix this. It seems that the `imagick` effect is still used simply because the `imagick` module is installed. Uninstalling `imagick` seems to revert it back to what it was before, and images are converted to WEBP again.

So the `imagick` module is a bit invasive, in this regard, and installing/uninstalling it probably requires some additional actions to ensure that the effects it overrides are configured properly.

modules/core/entity/modules/fields/farm_entity_fields.install
paul121: Thanks for the explanation!

mstenta:
I've rebased this on top of #941, which was merged. Now, the logic for installing and configuring the `imagick` module is in the new `farm_image` module's `hook_install()` and `hook_update_NAME()` functions (instead of the `farm_entity_fields` module).

mstenta:
If we do decide to move forward with this, then we should add a dependency on the `imagick` PHP extension like we will do in #1013.

PR #941, opened 2025-03-25
Automatically rotate derivative images based on EXIF Orientation

### LINKED ISSUE #939: Issues with image orientation

mstenta:
Fixes #939

This PR adds a new "Auto-orient" image effect to the default image styles that ship with farmOS. This effect automatically rotates derivative images based on the EXIF `Orientation` attribute.

The [EXIF Orientation](https://www.drupal.org/project/exif_orientation) module is uninstalled by this PR, and a new `farm_image` module is introduced. Notably, this does not *remove* `drupal/exif_orientation` from `composer.json`, because doing so would prevent Drupal from being able to uninstall it. So we should wait until 4.x to remove it entirely.


mstenta:
The `farm_image_auto_orient` image effect is loosely based on the following implementations found in contrib:

- [EXIF Orientation](https://www.drupal.org/project/exif_orientation) module: https://git.drupalcode.org/project/exif_orientation/-/blob/87e23ab0cd2b237f05d0aeda2bc23f441f54452e/exif_orientation.module#L62
- [Image Effects](https://www.drupal.org/project/image_effects) module:
  - `AutoOrientImageEffect`: https://git.drupalcode.org/project/image_effects/-/blob/730a1e6f5f947ef4d6eec5d359c6968ce93594f9/src/Plugin/ImageEffect/AutoOrientImageEffect.php
  - `AutoOrient` (ImageMagick): https://git.drupalcode.org/project/image_effects/-/blob/730a1e6f5f947ef4d6eec5d359c6968ce93594f9/src/Plugin/ImageToolkit/Operation/imagemagick/AutoOrient.php
  - `AutoOrient` (GD2): https://git.drupalcode.org/project/image_effects/-/blob/730a1e6f5f947ef4d6eec5d359c6968ce93594f9/src/Plugin/ImageToolkit/Operation/gd/AutoOrient.php
- [Imagick](https://www.drupal.org/project/imagick) module:
  - `AutorotateImageEffect`: https://git.drupalcode.org/project/imagick/-/blob/373da420d5fff9f7a8a9d5dca46846c28e3797c2/src/Plugin/ImageEffect/AutorotateImageEffect.php
  - `Autorotate`: https://git.drupalcode.org/project/imagick/-/blob/373da420d5fff9f7a8a9d5dca46846c28e3797c2/src/Plugin/ImageEffect/AutorotateImageEffect.php

All of them work very similarly, by reading the EXIF data and performing transformations based on the value.

Here is my reasoning for rolling our own image effect instead of using one of the contrib ones:

- I want to drop our `exif_orientation` dependency, because it alters the *original* image, and without ImageMagick that means that EXIF data is lost during the transformation. See #939 for a lot more detail.
- The [Image Effects](https://www.drupal.org/project/image_effects) module requires the [ImageMagick](https://www.drupal.org/project/imagemagick) module, which we will not be adding. If anything, we will add the [Imagick](https://www.drupal.org/project/imagick) module (PR: #940)
- This PR takes priority over PR #940. We need to fix image orientation issues first, regardless. So we need our own image effect.
- If we decide to merge #940, then we will be adding the [Imagick](https://www.drupal.org/project/imagick) module, which comes with its own `Autorotate` image effect that serves the same purpose. However, this only works with ImageMagick, not GD2, which means we would need to require ImageMagick, which would be a breaking change. We can consider doing that in farmOS 4.x, and at that time we can decide if we want to replace our custom image effect.

mstenta:
Also, IMO this can be merged as a patch release because it fixes a bug (#939). #940 will need to be merged in a minor release, at the very least, if not a major release (4.x).

modules/core/image/src/Plugin/ImageEffect/AutoOrientImageEffect.php
```diff
@@ -0,0 +1,204 @@
...
+        $result = $image->apply('mirror', ['x_axis' => TRUE]);
+        if ($result) {
+          $result = $image->apply('rotate', ['degrees' => 90]);
+        }
+        break;
+
+      case 8:
+        $result = $image->apply('rotate', ['degrees' => 270]);
+        break;
+    }
```
symbioquine: For larger chunks of code that appear directly derived from [other code](https://git.drupalcode.org/project/image_effects/-/blob/730a1e6f5f947ef4d6eec5d359c6968ce93594f9/src/Plugin/ImageToolkit/Operation/gd/AutoOrient.php#L45-86), my understanding is that there should be some sort of inline attribution indicating where the code was copied/modified from.

We can argue about whether this code is too much of boiler-plate for the GPL attribution/copyright-listing requirements to come into play (obligatory IAANAL), but at the very least I'd argue for perma-linking to the source of the copied code since that also has come in handy to me in the past when troubleshooting errors that crept into adapted code.

modules/core/image/farm_image.install
```diff
@@ -0,0 +1,32 @@
...
+  // Add auto-orient effect to Drupal core image styles.
+  $styles = [
+    'large',
+    'medium',
+    'thumbnail',
+    'wide',
+  ];
+  foreach ($styles as $name) {
+    $style = ImageStyle::load($name);
+    $style->addImageEffect([
```
symbioquine: Is there any argument for being more defensive here - i.e. if one style doesn't exist or fails to load, should it fail entirely or proceed to the others?

modules/core/entity/farm_entity.post_update.php
```diff
@@ -24,3 +24,12 @@ function farm_entity_post_update_enforce_plan_eri(&$sandbox) {
 function farm_entity_post_update_rebuild_bundle_field_maps(&$sandbox = NULL) {
   \Drupal::service('entity_field.manager')->rebuildBundleFieldMap();
 }
+
+/**
+ * Uninstall EXIF Orientation module.
+ */
+function farm_entity_post_update_uninstall_exif_orientation() {
+  if (\Drupal::service('module_handler')->moduleExists('exif_orientation')) {
+    \Drupal::service('module_installer')->uninstall(['exif_orientation']);
```
symbioquine: What happens if somebody copied the `farm_entity.info.yml` file's dependencies into a contrib module and inadvertently (or intentionally) depends on the `exif_orientation` module? Would this force their module to be uninstalled also, or would this fail?

docker/Dockerfile
```diff
@@ -30,8 +30,8 @@ RUN apt-get update && apt-get install -y libgeos-dev \
     && make install \
     )
 
-# Install the BCMath PHP extension.
-RUN docker-php-ext-install bcmath
+# Install the BCMath and Exif PHP extensions.
+RUN docker-php-ext-install bcmath exif
```
symbioquine, line 34: Do we need to update any of our documentation regarding farmOS' expectations about its environment?

modules/core/image/src/Plugin/ImageEffect/AutoOrientImageEffect.php
```diff
@@ -0,0 +1,204 @@
...
+
+/**
+ * Automatically rotates an image based on EXIF Orientation.
+ */
+#[ImageEffect(
+  id: "farm_image_auto_orient",
+  label: new TranslatableMarkup("Auto-orient"),
+  description: new TranslatableMarkup("Automatically rotates an image based on EXIF Orientation.")
+)]
+class AutoOrientImageEffect extends ConfigurableImageEffectBase implements ContainerFactoryPluginInterface {
```
symbioquine: Interesting, I hadn't seen this (new?) style of class documentation yet...

modules/core/image/src/Plugin/ImageEffect/AutoOrientImageEffect.php
```diff
@@ -0,0 +1,204 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function applyEffect(ImageInterface $image) {
+
+    // Only proceed if the exif PHP extension is loaded.
+    if (!extension_loaded('exif')) {
+      return TRUE;
+    }
```
symbioquine, line 73: Should this check also occur at the top of `transformDimensions`? The [docs](https://api.drupal.org/api/drupal/core%21modules%21image%21src%21ConfigurableImageEffectBase.php/class/ConfigurableImageEffectBase/8.9.x) imply but don't really guarantee that it won't be called if no effect was applied here...

mstenta:
Force pushed to fix `phpcs` and `phpstan` complaints. Thanks for the review @symbioquine! 'll dig into your feedback next.

modules/core/image/src/Plugin/ImageEffect/AutoOrientImageEffect.php
mstenta: Yea! It's a new feature of PHP 8: https://www.php.net/manual/en/language.attributes.overview.php

We are in the process of migrating from Symfony's "annotations" to these new attributes: https://www.drupal.org/project/farm/issues/3461548

modules/core/image/src/Plugin/ImageEffect/AutoOrientImageEffect.php
mstenta: Agreed! I also added a comment to `transformDimensions()` because that was also inspired by Image Effects code.

Commit: https://github.com/farmOS/farmOS/pull/941/commits/29cb9cac312364f35c0bef84c7b5799d289223de

modules/core/image/src/Plugin/ImageEffect/AutoOrientImageEffect.php
mstenta: Yes - good point. Fixup commit: https://github.com/farmOS/farmOS/pull/941/commits/2415dde6b049eb9e5c186876da8270bdfa00aef8

modules/core/image/farm_image.install
mstenta: Good call - fixup commit: https://github.com/farmOS/farmOS/pull/941/commits/1ec09787d03eda62b5ab60ca2976c286d6b96187

modules/core/entity/farm_entity.post_update.php
mstenta: Yes that's true. We got burned by that issue ourselves when a contrib module uninstalled a submodule we were depending on.

I'll look into checking whether or not other modules depend on `exif_orientation` before uninstalling it.

The catch is: since we're adding the `exif` PHP extension, if the `exif_orientation` module stays installed then it will start working (it wasn't because we didn't have `exif`). And that means that original images will start to get rotated, losing their EXIF data in the process. But if someone is depending on `exif_orientation`, then I suppose that's on them...

docker/Dockerfile
mstenta: It is a soft dependency, but good idea to at least mention it in the server requirements docs. I added a commit: https://github.com/farmOS/farmOS/pull/941/commits/dac2375568da587d7812957ff343169d85ba867b

modules/core/entity/farm_entity.post_update.php
mstenta: fixup: https://github.com/farmOS/farmOS/pull/941/commits/3d03f2d525bb722d0373ae249dc1d37d735ae8ad

symbioquine (approved):

mstenta:
Thanks for the reviews and approval @symbioquine! I squashed the fixup commits, resolved your review conversations, and force pushed.

modules/core/image/src/Plugin/ImageEffect/AutoOrientImageEffect.php
```diff
@@ -0,0 +1,217 @@
...
+
+    // Only work with supported image types.
+    $mime_types = [
+      'image/jpeg',
+      'image/png',
+      'image/tiff',
+    ];
```
paul121: Just noting that `tiff` isn't included in our `image` field file extensions: https://github.com/farmOS/farmOS/blob/b99e1ae519bf79c8e28ca76ab922d30a13d20427/modules/core/field/src/FarmFieldFactory.php#L631-L632

I read that `gif` doesn't include exif metadata so that's OK. Also wondering if we should add support for `webp` images, but could be a followup.

modules/core/image/src/Plugin/ImageEffect/AutoOrientImageEffect.php
```diff
@@ -0,0 +1,217 @@
...
+   * Load EXIF Orientation value from an image file.
+   *
+   * @param string $uri
+   *   The image file URI.
+   *
+   * @return int|null
+   *   Returns an integer representing the EXIF Orientation value, or NULL if no
+   *   Orientation key was found.
+   */
+  protected function loadExifOrientation($uri) {
```
paul121: Can add some types here
```suggestion
  protected function loadExifOrientation(string $uri): ?int {
```

modules/core/image/src/Plugin/ImageEffect/AutoOrientImageEffect.php
```diff
@@ -0,0 +1,217 @@
...
+      return;
+    }
+
+    // This logic is adapted from the Drupal Image Effects module.
+    // @see https://git.drupalcode.org/project/image_effects/-/blob/730a1e6f5f947ef4d6eec5d359c6968ce93594f9/src/Plugin/ImageEffect/AutoOrientImageEffect.php
+    $dimensions['width'] = $dimensions['width'] ? (int) $dimensions['width'] : NULL;
+    $dimensions['height'] = $dimensions['height'] ? (int) $dimensions['height'] : NULL;
+    if ($dimensions['width'] && $dimensions['height']) {
+      $orientation = $this->loadExifOrientation($uri);
+      if (in_array($orientation, [5, 6, 7, 8])) {
```
paul121, line 175: Check that returned orientation value is not null

```suggestion
      $orientation = $this->loadExifOrientation($uri);
      if ($orientation && in_array($orientation, [5, 6, 7, 8])) {
```

modules/core/image/src/Plugin/ImageEffect/AutoOrientImageEffect.php
```diff
@@ -0,0 +1,217 @@
...
+    // Otherwise return NULL.
+    return NULL;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildConfigurationForm(array $form, FormStateInterface $form_state) {
+    return $form;
+  }
```
paul121: Since this effect doesn't use any configuration, could we extend directly from `ImageEffectBase`?

If we did have configuration I think we would need to define the `image.effect.farm_image_auto_orient` schema, but without configuration I think we're OK. I see the core `image_module_test` only provides a schema for the configurable image effect it provides, but the non configurable effects are not included: https://git.drupalcode.org/project/drupal/-/blob/10.5.x/core/modules/image/tests/modules/image_module_test/config/schema/image_module_test.schema.yml?ref_type=heads

modules/core/image/farm_image.install
```diff
@@ -0,0 +1,34 @@
...
+ */
+function farm_image_install() {
+
+  // Add auto-orient effect to Drupal core image styles.
+  $styles = [
+    'large',
+    'medium',
+    'thumbnail',
+    'wide',
+  ];
```
paul121, line 23: I agree it makes sense to edit these BUT I'm also curious what the other option would look like, adding our own `farm_large`, etc image styles that we could more easily manage ourselves. I'm just not familiar with image styles.

If we did this, would we just update the image field formatters to use our specified `farm_*` image style(s)?

modules/core/image/src/Plugin/ImageEffect/AutoOrientImageEffect.php
```diff
@@ -0,0 +1,217 @@
...
+use Psr\Log\LoggerInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Automatically rotates an image based on EXIF Orientation.
+ */
+#[ImageEffect(
+  id: "farm_image_auto_orient",
+  label: new TranslatableMarkup("Auto-orient"),
+  description: new TranslatableMarkup("Automatically rotates an image based on EXIF Orientation.")
```
paul121: Very minor but these strings could use single quotes to be more consistent?

paul121 (changes requested):
Just some small things to change but this plan LGTM!

modules/core/image/src/Plugin/ImageEffect/AutoOrientImageEffect.php
mstenta: That's true. I included `tiff` in the allowed mime types because TIFFs support EXIF data. But you're right, `webp` also supports EXIF, so it should be included in the allowed mime types too.

And both should be added to our allowed file extensions in image fields, IMO.

mstenta: Added a commit to this PR: https://github.com/farmOS/farmOS/pull/941/commits/8dbb9a87c3690f5f2b4574dfcf8dec4c7d18bd60

> And both should be added to our allowed file extensions in image fields, IMO.

I will open a separate PR for this.

mstenta: Ah! Actually we do allow TIFFs, but in the `file` field, not the `image` field. I wonder if there was a reason for that. :thinking: I'll follow up on that in the separate PR...

https://github.com/farmOS/farmOS/blob/b99e1ae519bf79c8e28ca76ab922d30a13d20427/modules/core/field/src/FarmFieldFactory.php#L626-L627

modules/core/image/src/Plugin/ImageEffect/AutoOrientImageEffect.php
mstenta: Good call. Fixup: https://github.com/farmOS/farmOS/pull/941/commits/bb961e86611a5e003b5f0176380ecbf91a7f8256

modules/core/image/src/Plugin/ImageEffect/AutoOrientImageEffect.php
mstenta: I had that originally and PHPStorm highlighted it, saying:

> Condition is unnecessary because it is checked by 'in_array($orientation, [5, 6, 7, 8])' 

In other words: `NULL` is not in the array `[5, 6, 7, 8]`, so it will work as expected.

modules/core/image/src/Plugin/ImageEffect/AutoOrientImageEffect.php
mstenta: Ah good point! Fixup: https://github.com/farmOS/farmOS/pull/941/commits/c105ca32137c7811653842d808c69b34821c4900

modules/core/image/farm_image.install
mstenta: Yes, that's another option we could consider. The Drupal `image` module installs these styles by default - and there's no way of preventing it from adding them because they are `config/install/image.style.*.yml` files.

https://git.drupalcode.org/project/drupal/-/tree/10.4.x/core/modules/image/config/install?ref_type=heads

If we add our own, and leave the Drupal core ones untouched, then we'll have duplicates/unused styles, which feels messy to me. I would probably want to consider removing the Drupal core ones if we did that. But that has it's own considerations (more code in our `hook_install()` to delete them, is it a breaking change? etc).

Up until now we haven't needed to think about this. And in this particular instance, I think it's simpler and more elegant to just add our effect to the core styles.

I'm open to considering providing our own, but I think we should have a good reason for it.

modules/core/image/src/Plugin/ImageEffect/AutoOrientImageEffect.php
mstenta: Ah good catch! Fixup: https://github.com/farmOS/farmOS/pull/941/commits/2e9c9788261390d3e58bfece0351d2c2455f4817

mstenta:
Thanks for the review @paul121! I responded to all your feedback. Requesting re-review... :-)

modules/core/image/src/Plugin/ImageEffect/AutoOrientImageEffect.php
mstenta: Followup PR: https://github.com/farmOS/farmOS/pull/942

modules/core/image/src/Plugin/ImageEffect/AutoOrientImageEffect.php
paul121: Oh, great!

modules/core/image/farm_image.install
paul121: I agree, if we get into any more complex image styles then we could consider it. Sounds good.

paul121 (approved):

mstenta:
Thanks @paul121! Squashed fixups and resolved conversations. Force pushed for final tests to run before merging...

PR #942, opened 2025-03-26
Allow WebP files in image fields

mstenta:
This adds the ability to upload `*.webp` files to image fields in farmOS.

Spawned from this comment: https://github.com/farmOS/farmOS/pull/941#discussion_r2012940992


mstenta:
FWIW I also tested adding `tif tiff` but it seems that Drupal's image field type doesn't allow TIFFs at a lower level.

paul121 (approved):

PR #943, opened 2025-03-28
Fix PHPStan errors: Access to an undefined property Drupal\Core\Field\FieldItemInterface::$value

mstenta:
The following PHPStan errors started happening in automated tests today:

```
 ------ ------------------------------------------------------------------------------------------
  Line   web/profiles/farm/modules/core/field/src/Plugin/Field/FieldFormatter/HideableBoolean.php
 ------ ------------------------------------------------------------------------------------------
  73     Access to an undefined property Drupal\Core\Field\FieldItemInterface::$value.
         💡 Learn more: https://phpstan.org/blog/solving-phpstan-access-to-undefined-property
  73     Access to an undefined property Drupal\Core\Field\FieldItemInterface::$value.
         💡 Learn more: https://phpstan.org/blog/solving-phpstan-access-to-undefined-property
  78     Access to an undefined property Drupal\Core\Field\FieldItemInterface::$value.
         💡 Learn more: https://phpstan.org/blog/solving-phpstan-access-to-undefined-property
  81     Access to an undefined property Drupal\Core\Field\FieldItemInterface::$value.
         💡 Learn more: https://phpstan.org/blog/solving-phpstan-access-to-undefined-property
 ------ ------------------------------------------------------------------------------------------

 ------ -------------------------------------------------------------------------------------------
  Line   web/profiles/farm/modules/core/field/src/Plugin/Field/FieldFormatter/UrlLinkFormatter.php
 ------ -------------------------------------------------------------------------------------------
  33     Access to an undefined property Drupal\Core\Field\FieldItemInterface::$value.
         💡 Learn more: https://phpstan.org/blog/solving-phpstan-access-to-undefined-property
  36     Access to an undefined property Drupal\Core\Field\FieldItemInterface::$value.
         💡 Learn more: https://phpstan.org/blog/solving-phpstan-access-to-undefined-property
  37     Access to an undefined property Drupal\Core\Field\FieldItemInterface::$value.
         💡 Learn more: https://phpstan.org/blog/solving-phpstan-access-to-undefined-property
  43     Access to an undefined property Drupal\Core\Field\FieldItemInterface::$value.
         💡 Learn more: https://phpstan.org/blog/solving-phpstan-access-to-undefined-property
 ------ -------------------------------------------------------------------------------------------

 ------ -------------------------------------------------------------------------------------------------
  Line   web/profiles/farm/modules/core/inventory/src/Plugin/Field/FieldFormatter/InventoryFormatter.php
 ------ -------------------------------------------------------------------------------------------------
  29     Access to an undefined property Drupal\Core\Field\FieldItemInterface::$value.
         💡 Learn more: https://phpstan.org/blog/solving-phpstan-access-to-undefined-property
 ------ -------------------------------------------------------------------------------------------------
```

This PR fixes them by replacing `$item->value` with `$item->getValue()`.

### DIFF
--- a/modules/core/field/src/Plugin/Field/FieldFormatter/UrlLinkFormatter.php
+++ b/modules/core/field/src/Plugin/Field/FieldFormatter/UrlLinkFormatter.php
@@ -28,6 +28,7 @@ class UrlLinkFormatter extends FormatterBase {
 
     // This is a clone of Drupal core's UriLinkFormatter, but it only renders
     // a link if the URI is a valid URL.
+    /** @var \Drupal\Core\Field\Plugin\Field\FieldType\UriItem $item */
     foreach ($items as $delta => $item) {
       if (!$item->isEmpty()) {
         if (filter_var($item->value, FILTER_VALIDATE_URL)) {
--- a/modules/core/inventory/src/Plugin/Field/FieldFormatter/InventoryFormatter.php
+++ b/modules/core/inventory/src/Plugin/Field/FieldFormatter/InventoryFormatter.php
@@ -25,6 +25,7 @@ class InventoryFormatter extends FormatterBase {
   public function viewElements(FieldItemListInterface $items, $langcode) {
     $elements = [];
 
+    /** @var \Drupal\farm_inventory\Plugin\Field\FieldType\InventoryItem $item */
     foreach ($items as $delta => $item) {
       $summary = $item->value;
       if (!empty($item->units)) {
--- a/modules/core/field/src/Plugin/Field/FieldFormatter/HideableBoolean.php
+++ b/modules/core/field/src/Plugin/Field/FieldFormatter/HideableBoolean.php
@@ -63,6 +63,7 @@ class HideableBoolean extends BooleanFormatter {
 
     $formats = $this->getOutputFormats();
 
+    /** @var \Drupal\Core\Field\Plugin\Field\FieldType\BooleanItem $item */
     foreach ($items as $delta => $item) {
       $format = $this->getSetting('format');
 
--- a/modules/core/inventory/src/Plugin/Field/FieldType/InventoryItem.php
+++ b/modules/core/inventory/src/Plugin/Field/FieldType/InventoryItem.php
@@ -12,6 +12,10 @@ use Drupal\Core\TypedData\DataDefinition;
 
 /**
  * Plugin implementation of the 'Inventory' field type.
+ *
+ * @property string $measure
+ * @property string $value
+ * @property string $units
  */
 #[FieldType(
   id: 'inventory',


mstenta:
I tested locally and was unable to replicate until I ran `composer update` and these two dependencies were updated:

```
  - Upgrading phpstan/phpstan (1.12.21 => 1.12.23): Extracting archive
  - Upgrading mglaman/phpstan-drupal (1.3.3 => 1.3.4): Extracting archive
```

After that, I was able to replicate locally.

mstenta:
> `mglaman/phpstan-drupal (1.3.3 => 1.3.4)`

https://github.com/mglaman/phpstan-drupal/releases/tag/1.3.4

My guess is this: https://github.com/mglaman/phpstan-drupal/pull/822

modules/core/inventory/src/Plugin/Field/FieldFormatter/InventoryFormatter.php
```diff
@@ -26,7 +26,7 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
     $elements = [];
 
     foreach ($items as $delta => $item) {
-      $summary = $item->value;
+      $summary = $item->getValue();
```
wotnak: `$item->getValue()` returns an array while `$item->value` returns a single value, in the case of this field a string that would correspond to `$item->getValue()['value']`

with the current changes, when viewing an asset with inventory, an error is displayed and `Current inventory` value displays `Array (Count)` instead of the actual value
image
image

wotnak (changes requested):

modules/core/inventory/src/Plugin/Field/FieldFormatter/InventoryFormatter.php
paul121: This is true and should fix the issue. I think we could also do `$item->getValue()->get('value')`. I might prefer this but don't think we have a standard practice here.

modules/core/field/src/Plugin/Field/FieldFormatter/UrlLinkFormatter.php
```diff
@@ -30,17 +30,17 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
     // a link if the URI is a valid URL.
     foreach ($items as $delta => $item) {
```
paul121, line 32:
```suggestion
    /** @var \Drupal\Core\Field\Plugin\Field\FieldType\UriItem $item */
    foreach ($items as $delta => $item) {
```

This might be a better solution. I believe we can provide a better type `[Field]Item` hint for `$item`.

The change in phpstan-drupal added a stub for the `FormatterInterface::viewElements()` function which specifies that each is item is a generic `FieldItemInterface`, which has no `$value` property definition: https://github.com/mglaman/phpstan-drupal/pull/822

But phptan-drupal provides stub files for nearly all of the core field types to add correct property annotations. Nearly all of these specify the `$value` property: https://github.com/mglaman/phpstan-drupal/tree/main/stubs/Drupal/Core/Field/Plugin/Field/FieldType

paul121 (changes requested):

modules/core/inventory/src/Plugin/Field/FieldFormatter/InventoryFormatter.php
mstenta: Thanks for testing this @wotnak!

modules/core/field/src/Plugin/Field/FieldFormatter/UrlLinkFormatter.php
mstenta: Ah nice thanks for the explanation @paul121!

I tested the `/** @var ...` line locally and it works! The only one that is still affected is `InventoryFormatter`, which is for our own custom `InventoryItem` field type... so `phpstan-drupal` doesn't provide a stub for that one.

Adding ` * @property string $value` to our `InventoryItem` docblock comment fixes it.

I pushed a fixup commit demonstrating this. In undoes the changes to `->value` and adds the type comments instead. What do you think? (cc @wotnak)

https://github.com/farmOS/farmOS/pull/943/commits/7499b4cde7b446e3154ecdbda246f9cb22826e72

wotnak: Looks ok.

For consistency, might be good to also declare types for other InventoryItem's properties (`measure`, `units`).

For narrowing types, I usually prefer to use `instanceof` since it actually checks the type while running the code instead of only assuming one for type checker tools. So here instead of `@var` comment we could use `if (!$item instanceof UriItem) { continue; }` inside the foreach.

wotnak (approved):

mstenta:
> For consistency, might be good to also declare types for other InventoryItem's properties (`measure`, `units`).

Good point @wotnak!

Pushed a fixup commit: https://github.com/farmOS/farmOS/pull/943/commits/74d622ef9bfbecc98f57ac7e0d30262cd6831102

> For narrowing types, I usually prefer to use `instanceof` since it actually checks the type

I'm open to doing this in a followup! For now I just want to get tests passing. But feel free to propose a PR that adds runtime type checking if you think it's easy!

wotnak (approved):

modules/core/field/src/Plugin/Field/FieldFormatter/UrlLinkFormatter.php
paul121: Ah interesting @wotnak ! I had also added some `assert()` checks in tests, but feel like `instanceof()` is better for runtime code. Curious if you have any other insights on runtime assertions vs instanceof.

https://github.com/farmOS/farmOS/pull/906#discussion_r1923073105

modules/core/field/src/Plugin/Field/FieldFormatter/UrlLinkFormatter.php
```diff
@@ -28,6 +28,7 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
 
     // This is a clone of Drupal core's UriLinkFormatter, but it only renders
     // a link if the URI is a valid URL.
+    /** @var \Drupal\Core\Field\Plugin\Field\FieldType\UriItem[] $items */
```
paul121:
```suggestion
    /** @var \Drupal\Core\Field\Plugin\Field\FieldType\UriItem $item */
```

@mstenta minor detail but `$items` isn't actually an array of `UriItem`. [`FieldItemListInterface`](https://git.drupalcode.org/project/drupal/-/blob/10.5.x/core/lib/Drupal/Core/Field/FieldItemListInterface.php?ref_type=heads#L28) implements `ListInterface` which allows it to be iterated over in the foreach loop and yield each separate field item.

This would be important if we were using other methods from `$items` that are unique to `FieldItemListInterface`. I think we should add the correct type hint on `$item` itself.

paul121 (changes requested):

modules/core/field/src/Plugin/Field/FieldFormatter/UrlLinkFormatter.php
paul121: If for some reason that isn't working we might need to move the type hint to be inside the foreach loop. But I think I have done this before

mstenta: Ah thanks @paul121. Makes sense.

Fixup commit: https://github.com/farmOS/farmOS/pull/943/commits/3ceb4eaf6073b1e70588c34c28941621e860a5e7

modules/core/field/src/Plugin/Field/FieldFormatter/UrlLinkFormatter.php
wotnak: @paul121 Assertions are fine for tests, but for actual application code it's usually better to use other ways of checking values since assert() by default throws an error when the assertion is not upheld and, depending on [php configuration](https://www.php.net/manual/en/ini.core.php#ini.zend.assertions), they can even be completely ignored at runtime.

paul121 (approved):

PR #944, opened 2025-03-30
Replace instances of "complete" log status with "done"

mstenta:
We reference a log status called "complete" in a few places in our docs. It should be "done".

### DIFF
--- a/modules/core/log/tests/src/Kernel/LogTest.php
+++ b/modules/core/log/tests/src/Kernel/LogTest.php
@@ -91,8 +91,8 @@ class LogTest extends KernelTestBase {
     $log_ids = $this->logQueryFactory->getQuery(['timestamp' => $now])->accessCheck(FALSE)->execute();
     $this->assertNotContains($foo_log->id(), $log_ids, 'Log query results can be filtered by timestamp.');
 
-    // Set the status of one log to complete.
-    $bar_log->set('status', 'complete');
+    // Set the status of one log to "done".
+    $bar_log->set('status', 'done');
     $bar_log->save();
 
     // Test that results can be filtered by status.
--- a/docs/development/api/changes.md
+++ b/docs/development/api/changes.md
@@ -226,7 +226,7 @@ for more information.
 To illustrate, this is how to filter activity logs by their completed status:
 
 - farmOS 1.x: `/log.json?type=activity&done=1`
-- farmOS 2.x: `/api/log/activity?filter[status]=complete`
+- farmOS 2.x: `/api/log/activity?filter[status]=done`
 
 ### Text format
 
--- a/docs/guide/index.md
+++ b/docs/guide/index.md
@@ -37,7 +37,7 @@ Asset details popup.
 
 Below the map you will see "Upcoming tasks" and "Late tasks" on the left, which
 summarize Logs that are still in "pending" status. From here you can quickly
-select Logs and mark them as "complete", reschedule them, clone them, or delete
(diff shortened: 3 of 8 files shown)


mstenta:
Found one place in an automated test where we set log status to `complete`, so I tacked on a commit for that as well: https://github.com/farmOS/farmOS/pull/944/commits/29ca6b3dd4505d43f7b7fb1cfa76f57b85791f0a

wotnak:
Found two more places in code comments where logs are referred as complete instead of done:

https://github.com/farmOS/farmOS/blob/a6c6b75686a124102398435feedcdeb164ef3c1e/modules/core/location/farm_location.base_fields.inc#L166

https://github.com/farmOS/farmOS/blob/a6c6b75686a124102398435feedcdeb164ef3c1e/modules/core/location/src/AssetLocation.php#L288

modules/core/log/tests/src/Kernel/LogTest.php
```diff
@@ -91,8 +91,8 @@ public function testLogQueryFactory() {
     $log_ids = $this->logQueryFactory->getQuery(['timestamp' => $now])->accessCheck(FALSE)->execute();
     $this->assertNotContains($foo_log->id(), $log_ids, 'Log query results can be filtered by timestamp.');
 
-    // Set the status of one log to complete.
-    $bar_log->set('status', 'complete');
+    // Set the status of one log to "done".
+    $bar_log->set('status', 'done');
```
symbioquine, line 95: How was this test passing before?

symbioquine: Oh, I see... sort of. Shouldn't it fail if you set an invalid status?

mstenta: The reason this didn't cause a test failure is because it's only testing to see if changing this logs status away from `pending` removes it from lists of `pending` logs (which it does).

Technically it's possible to set any value you want as a status. It only gets checked if we run `$entity->validate()` and deal with violations, which we aren't doing in this test.

mstenta:
> Found two more places in code comments where logs are referred as complete instead of done:

Thanks @wotnak! I saw these too and was on the fence about changing them.

> https://github.com/farmOS/farmOS/blob/a6c6b75686a124102398435feedcdeb164ef3c1e/modules/core/location/farm_location.base_fields.inc#L166

This one is in a translatable string, which I want to be careful about changing, because it nullifies any translations that have been contributed. I felt that it's fine as-is, since the meaning should be understood. But I'm open to reconsidering!

> https://github.com/farmOS/farmOS/blob/a6c6b75686a124102398435feedcdeb164ef3c1e/modules/core/location/src/AssetLocation.php#L288

Similarly this felt unnecessary since it's a comment and the meaning is understood in context.

wotnak:
In my opinion it would be good to change it in both places. Consistent language in the ui and docs makes it easier to use the application. Leaving a single place in the ui where a specific concept is called differently than everywhere else feels like a bug. Translations can always be updated. For translations contributed on https://localize.drupal.org/translate/projects/farm that have this string translated, we could potentially even copy the current translation for the new version of the string when it is released.

symbioquine (approved):

mstenta:
> In my opinion it would be good to change it in both places.

Points taken! Thanks @wotnak. :-)

Added another fixup commit for review: https://github.com/farmOS/farmOS/pull/944/commits/d21126dcec3fb780455879809ea58f91d22de18d

wotnak (approved):

PR #945, opened 2025-03-30
Add abandoned log status

mstenta:
See: https://www.drupal.org/project/farm/issues/330460

And forum discussion: https://farmos.discourse.group/t/ability-to-create-abandoned-logs/1636

Opening this as a draft PR to kick off discussion. Right now all it does is add the status as an option and updates the documentation to include it.

TBD:

- Do we need to change any other code to accommodate this?
- Do we need any automated tests?
- ... ?

### DIFF
--- a/docs/guide/index.md
+++ b/docs/guide/index.md
@@ -36,9 +36,8 @@ clicking on a location and then clicking on the available links within the
 Asset details popup.
 
 Below the map you will see "Upcoming tasks" and "Late tasks" on the left, which
-summarize Logs that are still in "pending" status. From here you can quickly
-select Logs and mark them as "done", reschedule them, clone them, or delete
-them.
+summarize Logs that have a "pending" status. From here you can quickly select
+Logs and mark them as "done", reschedule them, clone them, or delete them.
 
 To the right you will see a summary of various farm metrics, including total
 Asset and Log counts. Additional metrics can be added via modules.
--- a/docs/model/type/log.md
+++ b/docs/model/type/log.md
@@ -62,10 +62,15 @@ Logs always have a timestamp which indicates when they took place.
 
 #### Status
 
-Logs can be marked as "pending" or "done", to indicate whether they are
-"planned" or "actual" events. Every change that is made to a Log is stored as
-a revision, so that it's possible to see how a plan evolves over time until it
-eventually becomes a canonical record of the event that took place.
+Logs can be marked as "pending", "done", or "abandoned". Pending Logs represent
+planned events, while done Logs represent events that actually happened. If an
+event is never completed and there is no intention of completing it, the Log
+can be marked as "abandoned" to make that clear while preserving it for future
+reference.
+
+Every change that is made to a Log is stored as a revision, so that it's
+possible to see how a plan evolves over time until it eventually becomes a
+canonical record of the event that took place.
 
 #### Flags
 
--- a/modules/asset/group/tests/src/Kernel/GroupTest.php
+++ b/modules/asset/group/tests/src/Kernel/GroupTest.php
@@ -174,7 +174,60 @@ class GroupTest extends KernelTestBase {
     // Assert that the animal's cache tags were not invalidated.
     $this->assertEntityTestCache($animal, TRUE);
 
+    // Create an "abandoned" log that assigns the animal to the second group
+    // and confirm that it still has the same group membership as before.
+    /** @var \Drupal\log\Entity\LogInterface $abandoned_log */
+    $abandoned_log = Log::create([
+      'type' => 'test',
+      'status' => 'abandoned',
+      'is_group_assignment' => TRUE,
+      'group' => ['target_id' => $second_group->id()],
+      'asset' => ['target_id' => $animal->id()],
+    ]);
+    $abandoned_log->save();
+    $this->assertEquals($first_group->id(), $this->groupMembership->getGroup($animal)[0]->id(), 'Abandoned group assignment logs do not affect membership.');
+    $this->assertCorrectAssets([], $this->groupMembership->getGroupMembers([$second_group]), TRUE, 'Groups with only abandoned membership have zero members.');
+
+    // Assert that the animal's cache tags were not invalidated.
+    $this->assertEntityTestCache($animal, TRUE);
+
+    // Change the pending log's status to "abandoned" and confirm that the group
+    // membership is still unchanged.
+    $second_log->set('status', 'abandoned');
+    $second_log->save();
+    $this->assertEquals($first_group->id(), $this->groupMembership->getGroup($animal)[0]->id(), 'Abandoned group assignment logs do not affect membership.');
+    $this->assertCorrectAssets([], $this->groupMembership->getGroupMembers([$second_group]), TRUE, 'Groups with only abandoned membership logs have zero members.');
+
+    // Assert that the animal's cache tags were not invalidated.
+    $this->assertEntityTestCache($animal, TRUE);
+
     // When the log is marked as "done", the asset's membership is updated.
+    $second_log->set('status', 'done');
+    $second_log->save();
+    $this->assertEquals($second_group->id(), $this->groupMembership->getGroup($animal)[0]->id(), 'A second group assignment log updates group membership.');
+    $this->assertCorrectAssets([$animal], $this->groupMembership->getGroupMembers([$second_group]), TRUE, 'Completed group assignment logs add group members.');
+
+    // Assert that the animal's cache tags were invalidated.
+    $this->assertEntityTestCache($animal, FALSE);
+
+    // Re-populate a cache value dependent on the animal's cache tags.
+    $this->populateEntityTestCache($animal);
+
+    // When a log is changed from "done" to "abandoned", the asset's membership
+    // is reverted to what it was previously.
+    $second_log->set('status', 'abandoned');
+    $second_log->save();
+    $this->assertEquals($first_group->id(), $this->groupMembership->getGroup($animal)[0]->id(), 'Abandoned group assignment logs do not affect membership.');
+    $this->assertCorrectAssets([], $this->groupMembership->getGroupMembers([$second_group]), TRUE, 'Groups with only abandoned membership logs have zero members.');
+
+    // Assert that the animal's cache tags were invalidated.
+    $this->assertEntityTestCache($animal, FALSE);
+
+    // Re-populate a cache value dependent on the animal's cache tags.
+    $this->populateEntityTestCache($animal);
+
+    // When a log is changed from "abandoned" to "done", the asset's membership
+    // is updated.
     $second_log->set('status', 'done');
     $second_log->save();
     $this->assertEquals($second_group->id(), $this->groupMembership->getGroup($animal)[0]->id(), 'A second group assignment log updates group membership.');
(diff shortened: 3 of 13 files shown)


mstenta:
The intention is to consider this for farmOS 4.x.

mstenta:
> We would also need an “Abandon logs” bulk action, so that you can easily abandon all of last year’s pending logs at once.

https://farmos.discourse.group/t/ability-to-create-abandoned-logs/1636/4?u=mstenta

docs/model/type/log.md
```diff
@@ -62,10 +62,14 @@ Logs always have a timestamp which indicates when they took place.
 
 #### Status
 
-Logs can be marked as "pending" or "done", to indicate whether they are
-"planned" or "actual" events. Every change that is made to a Log is stored as
-a revision, so that it's possible to see how a plan evolves over time until it
-eventually becomes a canonical record of the event that took place.
+Logs can be marked as "pending", "done", or "abandoned". Pending Logs represent
+planned events, while done Logs represent events that actually happened. If an
+event is never completed, and there is no intention of completing it, the :og
```
paul121:
```suggestion
event is never completed, and there is no intention of completing it, the log
```

mstenta: Oops good catch @paul121. Force pushed this fix. :-)

Fosten:
Here’s another doc-ish place that would need to be updated, so that instructions on the /logs page reflect this change.

https://github.com/farmOS/farmOS/blob/9c27b0770df8e43f5c23d8d58012c99ad7f7de7a/modules/core/ui/views/farm_ui_views.module#L44

mstenta:
Thanks @Fosten! I force-pushed the commit to update that help text. I also searched all occurrences of the "pending" string and updated a few other things.

symbioquine (changes requested):
It seems like there should be some additional tests added as part of this change showing that the abandoned status behaves as expected with the group/movement/inventory logic - e.g. test that a latest log in the abandoned status does not affect an asset's location.

mstenta:
Pushed a commit that adds abandoned log tests to the location, inventory, and group membership logic.

modules/core/inventory/tests/src/Kernel/InventoryTest.php
```diff
@@ -151,6 +151,22 @@ public function testSimpleAssetInventory() {
...
 
+    // Add an abandoned adjustment, and confirm that it does not affect the
+    // current inventory.
+    $log = $this->adjustInventory($asset, 'increment', '1');
+    $log->set('status', 'abandoned');
+    $log->save();
+
+    // Re-populate a cache value dependent on the asset's cache tags.
+    $this->populateEntityTestCache($asset);
+    $log->save();
```
symbioquine: Are these lines are actually needed? Naively, it seems like they weaken the test since the second `$log->save()` is basically a no-op and the cache invalidation logic could be aware of that...

Shouldn't the `populateEntityTestCache` call on line 145 be sufficient?

mstenta: Ah yes you're right. I copied and pasted it from above but it isn't necessary. I will remove it and force push.

docs/guide/index.md
```diff
@@ -36,9 +36,8 @@ clicking on a location and then clicking on the available links within the
 Asset details popup.
 
 Below the map you will see "Upcoming tasks" and "Late tasks" on the left, which
-summarize Logs that are still in "pending" status. From here you can quickly
-select Logs and mark them as "done", reschedule them, clone them, or delete
-them.
+summarize Logs that have a "pending" status. From here you can quickly select
+Logs and mark them as "done", reschedule them, clone them, or delete them.
```
symbioquine, line 40: Nit: I wonder if this can be more succinct and also mention the abandoned status:

```suggestion
Logs to rescheduled, cloned, deleted, or marked as "done"/"abandoned".
```

docs/model/type/log.md
```diff
@@ -62,10 +62,14 @@ Logs always have a timestamp which indicates when they took place.
 
 #### Status
 
-Logs can be marked as "pending" or "done", to indicate whether they are
-"planned" or "actual" events. Every change that is made to a Log is stored as
-a revision, so that it's possible to see how a plan evolves over time until it
-eventually becomes a canonical record of the event that took place.
+Logs can be marked as "pending", "done", or "abandoned". Pending Logs represent
+planned events, while done Logs represent events that actually happened. If an
+event is never completed, and there is no intention of completing it, the Log
```
symbioquine: Nit: unnecessary comma

```suggestion
event is never completed and there is no intention of completing it, the Log
```

docs/model/type/log.md
```diff
@@ -62,10 +62,14 @@ Logs always have a timestamp which indicates when they took place.
...
+can be marked as "abandoned" to preserve it for future reference.
```
symbioquine: Nit:

```suggestion
can be marked as "abandoned" to make that clear while preserving it for future reference.
```

symbioquine (approved):
Some minor nitpicks - to take or leave. Otherwise LGTM!

modules/core/inventory/tests/src/Kernel/InventoryTest.php
mstenta: Ah! That caused tests to fail! I've restored the `$this->populateEntityTestCache($asset);` line (but not the `$log->save();` line) and they are passing again. Maybe @paul121 understands why, but keeping it there follows the pattern of the rest of the tests, so I don't think we necessarily need to understand it for the purpose of this PR. :sweat_smile:

docs/guide/index.md
mstenta: That section refers to the "Upcoming tasks" and "Late tasks" on the dashboard and the bulk actions available on them. We don't provide a bulk action for "abandoning" logs in this PR. That might be a good idea for a followup, and at that point we could adjust this text IMO.

docs/model/type/log.md
mstenta: Good eye! Fixed!

docs/model/type/log.md
mstenta: :+1: Done!

symbioquine (approved):

modules/core/inventory/tests/src/Kernel/InventoryTest.php
paul121: Interesting.. that is correct, neither of these lines should be necessary. Populating the asset cache _after_ saving the log isn't useful here, it will always result in a the asset cache not being invalidated.

The issue is that the `adjustInventory()` function saves the log with a `done` status as the last step, and in that instant the asset's cache is invalidated: https://github.com/farmOS/farmOS/blob/33218a94085f359d6232f1ebb408389a9197506e/modules/core/inventory/tests/src/Kernel/InventoryTest.php#L384

When the log's status is changed to `abandoned` this would cause the asset's cache to be invalidated again... so maybe this is why there is an additional `$log->save()` in the existing previous test, to try and simulate a new log being created with a `pending` status? I'm not sure that this existing test is effective either.

I think it might make most sense to add a `status` parameter to the `adjustInventory` function. This would make testing these edge cases with the log status a little easier to read. Does that make sense?

modules/asset/group/tests/src/Kernel/GroupTest.php
```diff
@@ -174,6 +174,16 @@ public function testGroupMembership() {
...
+    // Change the log status to "abandoned" and confirm that the group
+    // membership is unchanged.
+    $second_log->set('status', 'abandoned');
+    $second_log->save();
+    $this->assertEquals($first_group->id(), $this->groupMembership->getGroup($animal)[0]->id(), 'Abandoned group assignment logs do not affect membership.');
+    $this->assertCorrectAssets([], $this->groupMembership->getGroupMembers([$second_group]), TRUE, 'Groups with only abandoned membership logs have zero members.');
+
+    // Assert that the animal's cache tags were not invalidated.
+    $this->assertEntityTestCache($animal, TRUE);
+
```
paul121: This test covers changing a group membership log from `pending` -> `abandoned`.

It might be worth adding another test to cover creating a new `abandoned` log (expect no cache invalidation) or even changing from `done` -> `abandoned` (expect cache invalidation)

modules/core/location/tests/src/Kernel/LocationTest.php
```diff
@@ -318,6 +318,16 @@ public function testAssetLocation() {
...
+    // Change the log status to "abandoned" and confirm that the asset location
+    // is unchanged.
+    $second_log->set('status', 'abandoned');
+    $second_log->save();
+    $this->assertEquals($this->logLocation->getLocation($first_log), $this->assetLocation->getLocation($asset), 'Asset with abandoned movement log has original location');
+    $this->assertEquals($this->logLocation->getGeometry($first_log), $this->assetLocation->getGeometry($asset), 'Asset with abandoned movement log has original geometry.');
+
+    // Assert that the asset's cache tags were not invalidated.
+    $this->assertEntityTestCache($asset, TRUE);
+
```
paul121: Ditto. Just depends how involved we want to get with tests...

modules/core/inventory/tests/src/Kernel/InventoryTest.php
mstenta: Gotcha - yea it looks like there some existing funkiness in the test, and I ended up just replicating that with my new test.

> I think it might make most sense to add a `status` parameter to the `adjustInventory` function. This would make testing these edge cases with the log status a little easier to read. Does that make sense?

Yes! I'll do that and refactor the existing tests.

I'll also remove the extraneous calls to populate the cache... that should only happen after we confirm that a cache miss occurs, if I'm understanding correctly. That would be consistent with how the `LocationTest` and `GroupTest` logic is structured too.

I'll push these as pre-commits to the abandoned log commits.

modules/asset/group/tests/src/Kernel/GroupTest.php
mstenta: Good idea. I'll add these and amend the relevant commit.

modules/core/location/tests/src/Kernel/LocationTest.php
mstenta: Agreed. Will do.

modules/core/inventory/tests/src/Kernel/InventoryTest.php
mstenta: These are the new commits I pushed:

- https://github.com/farmOS/farmOS/pull/945/commits/e385912ed6f8463a2400ce420fa8e71dd4ef4750
- https://github.com/farmOS/farmOS/pull/945/commits/e8074e8e51942d558253341117110fe7d58356aa
- https://github.com/farmOS/farmOS/pull/945/commits/705a9ed3e7efa6adb8d8e4c16c9bcd7977e71c9b
- https://github.com/farmOS/farmOS/pull/945/commits/81548551d698caec58e316bb1e30e0a4c64f5d2e

mstenta:
Rebased onto latest `4.x` and addressed @paul121's feedback above, including some cleanup/refactoring of `InventoryTest` (before the "abandoned" tests are added).

I've added additional tests to `LocationTest`, `GroupTest`, and `InventoryTest`. Probably easiest to just review them from scratch now: https://github.com/farmOS/farmOS/pull/945/commits/ae4f25814fa674f9a5f0ef75c58f482c8392124e

mstenta:
Rebased onto latest 4.x.

paul121 (approved):

symbioquine (approved):

PR #946, opened 2025-04-01
Add Google Maps base layers

paul121:
Dependent on a new release of farmOS-map. See:
- https://www.drupal.org/project/farm/issues/3515907
- https://github.com/farmOS/farmOS-map/pull/208#pullrequestreview-2730927397

This is basically a copy/paste of the `farm_map_mapbox` module. The JS behavior has been modified to add Google Map Tile layers with slightly different logic


modules/core/map/modules/google/farm_map_google.info.yml
```diff
@@ -0,0 +1,7 @@
+name: farmOS Google Map
+description: Adds Google Map layers to farmOS maps.
+type: module
+package: farmOS Maps
+core_version_requirement: ^10
```
mstenta: We'll need to change this to `^11`, but that will be caught automatically when we try to merge it into `4.x`.

modules/core/map/modules/google/farm_map_google.links.task.yml
```diff
@@ -0,0 +1,4 @@
+farm_map_google.settings:
+  title: 'Google Maps'
```
mstenta, line 2: I wonder if we should change this to simply `title: 'Google'`, so the three tabs that are visible under `/farm/settings/map/` are: Map / Google / Mapbox

mstenta (changes requested):
This all LGTM! I haven't tested but skimmed and compared against our current `farm_map_mapbox` module.

One minor suggestion: I think we should change "Google Map" to "Google Maps" everywhere we mention it, to make sure we're using the "official" name (with the "s" on the end). I don't know if it matters from a legal or trademark perspective, but we might as well be consistent with what Google calls it itself.

paul121:
Yeah, re: naming of Google Maps I wasn't sure. There is a distinction when talking about it at the API level, the Google Map Tiles API, but I suppose it is correct to say the API is used to display "Google Maps": https://developers.google.com/maps/documentation/tile/policies#displaying-the-google-logo-with-third-party-renderers

> When you use the Map Tiles API to display Google Maps using a third-party renderer, ...

paul121:
Chatting today on the dev call, we're curious if we should add a `conflicts` section to our `farmos/farmos` composer.json to say that the 4.x release conflicts with `drupal/farm_map_google`

mstenta:
Curious if Drupal.org's composer dependency resolution magic would see `core_version_requirement: ^9 || ^10` and prevent it from being included in a farmOS 4.x codebase (Drupal 11). Then we wouldn't need to add a `conflict` to our `composer.json` (which @symbioquine pointed out would technically need to stay there forever...)

symbioquine:
> Curious if Drupal.org's composer dependency resolution magic would see `core_version_requirement: ^9 || ^10` and prevent it from being included in a farmOS 4.x codebase (Drupal 11). Then we wouldn't need to add a `conflict` to our `composer.json` (which @symbioquine pointed out would technically need to stay there forever...)

So the action item here is taking a 4.x install and attempting to do `composer require drupal/farm_map_google` and seeing if it fails with that version compatibility error right?

mstenta:
> So the action item here is taking a 4.x install and attempting to do `composer require drupal/farm_map_google` and seeing if it fails with that version compatibility error right?

Yes! I just tried that...

```bash
$ ./composer require drupal/farm_map_google
./composer.json has been updated
Running composer update drupal/farm_map_google
Loading composer repositories with package information
Updating dependencies
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Root composer.json requires drupal/farm_map_google * -> satisfiable by drupal/farm_map_google[1.0.0, ..., 1.0.4].
    - drupal/farm_map_google[1.0.0, ..., 1.0.3] require drupal/core ^9 -> found drupal/core[9.0.0-alpha1, ..., 9.5.x-dev] but the package is fixed to 11.2.2 (lock file version) by a partial update and that version does not match. Make sure you list it as an argument for the update command.
    - drupal/farm_map_google 1.0.4 requires drupal/core ^9 || ^10 -> found drupal/core[9.0.0-alpha1, ..., 9.5.x-dev, 10.0.0-alpha1, ..., 10.6.x-dev] but the package is fixed to 11.2.2 (lock file version) by a partial update and that version does not match. Make sure you list it as an argument for the update command.

Use the option --with-all-dependencies (-W) to allow upgrades, downgrades and removals for packages currently locked to specific versions.
You can also try re-running composer require with an explicit version constraint, e.g. "composer require drupal/farm_map_google:*" to figure out if any version is installable, or "composer require drupal/farm_map_google:^2.1" if you know which you need.

Installation failed, reverting ./composer.json and ./composer.lock to their original content.
```

So it looks like this is all set! There isn't a risk of this module being used with farmOS v4+, assuming that Composer is being used to manage dependencies. If someone has manually downloaded/installed it (outside of Composer), then it might cause issues, but I think that's outside of our control. And we no longer recommend that method of installing modules in v4+.

mstenta:
Rebased onto latest `4.x`, along with the commit from #1038 (which is required for this to work), and updated this PR to be based on `4.x`.

I cleaned up a few small things, including changing "Google Map" to "Google Maps" in a few places (as discussed above).

Tested it and it's working great! Nice to have another option for satellite base layers included in farmOS core!

Thanks again @paul121 for all the work on this! I will go ahead and merge this soon!

mstenta (approved):

mstenta:
Pushed a small addition to add this to the modules form, so folks can install it. :-)

PR #864, opened 2024-08-02
Build multi-platform Docker images/manifests

symbioquine:
The goal of this change is to make it trivial to run farmOS on the most common ARM architectures without the need for a local `docker build` step.

Unfortunately, the change is fairly involved in order to keep the caching and testing working the same way.

Ideally, we'd be able to do a single `docker buildx build --push --platform COMMA_SEPARATED_PLATFORMS_HERE --tag` command, but that wouldn't work for how we test pull requests and handle creating the various tags conditionally - in addition to making sure our test workflows run against the same images prior to actually pushing them to Docker hub.

Instead, we have to do the more manual strategy of using `docker buildx` to build/cache each platform's image individually, then only at the end do we create the final multi-architecture manifest.

The other thing that makes this change slightly larger is that the existing workflow hard-codes the Docker image name `farmos/farmos` so in order to test that has to be changed in many places. This change has moved that prefix into a variable so it can be changed in a single place to make verifying the changes easier.

References:

* https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/
* https://github.com/docker/setup-qemu-action
* https://github.com/docker/setup-buildx-action

PR #947, opened 2025-04-03
Minor GitHub Action variable improvements

mstenta:
This PR makes two small changes to our `deliver.yml` GitHub Action:

1. Wrap variables in curly brackets for consistency.
2. Add a FARMOS_DEV_BRANCH env var to reduce duplication.

The first change is primarily for consistency. Most of our shell variables are already wrapped in curly brackets (eg: `${MYVAR}` instead of `$MYVAR`), so this just wraps the ones that aren't. From a functional standpoint there isn't a big difference, but I think it's good to pick one and use it consistently. It's easier to tell future contributors to match the existing pattern than it is to decide each case individually and explain why.

See:

- https://stackoverflow.com/questions/8748831/when-do-we-need-curly-braces-around-shell-variables
- https://nickjanetakis.com/blog/why-you-should-put-braces-around-your-variables-when-shell-scripting

The second change is in preparation for the forthcoming farmOS `4.x` branch (and future ones).

### DIFF
--- a/.github/workflows/deliver.yml
+++ b/.github/workflows/deliver.yml
@@ -65,12 +65,12 @@ jobs:
       - name: Build and save the farmOS Docker image
         run: |
           COMMON_BUILD_ARGS="--load --build-arg FARMOS_REPO=https://github.com/${FARMOS_REPO} --build-arg FARMOS_VERSION=${FARMOS_VERSION} --platform ${{ matrix.DOCKER_PLATFORM_NAME }}"
-          PROD_IMG_NAME=$DOCKER_IMG_PREFIX:3.x-${{ matrix.platform }}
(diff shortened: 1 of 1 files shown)


paul121 (approved):

symbioquine:
Changes look fine to me, but it's a lot of small details... it might be worth pushing a test release to your fork and see if it all still runs as expected. I did that in this PR: https://github.com/farmOS/farmOS/pull/864 (It's probably similar to this commit: https://github.com/farmOS/farmOS/commit/0fbb14829cf6b02228cded52cc222f941eb9929e)

mstenta:
Took a few tries to get it to use my fork properly, but the PR changes worked! :sweat_smile:

mstenta:
Here is the release it created: https://github.com/mstenta/farmOS/releases/tag/3.4.5

symbioquine (approved):

PR #948, opened 2025-04-10
Maintain separate CHANGELOG.md files in each branch

mstenta:
Coming out of discussion on the dev call and in the forum (see https://farmos.discourse.group/t/managing-changelog-md-with-3-x-and-4-x-branches/2268), this PR adopts a new strategy for maintaining our `CHANGELOG.md` file(s) moving forward.

Instead of including ALL releases across all branches of farmOS, the `CHANGELOG.md` file in each branch will only include releases from that branch. So the `2.x` changelog will only include `2.x` releases, `3.x` will only include `3.x` releases, and so on.

Links are added to the bottom of `CHANGELOG.md` to give quick access to release notes on previous branches. We only started using `CHANGELOG.md` in `2.x`, so for `1.x` it links to the Drupal.org releases, filtered for "7.x-1" (because the `1.x` branch of farmOS was actually `7.x-1.x`, following the old standard of including the Drupal major version, before we adopted semantic versioning).

This makes it easier to maintain two branches in parallel, which we will most likely be doing with `3.x` and `4.x` in the near future.


symbioquine (approved):
LGTM!

wotnak (approved):

paul121 (approved):

PR #949, opened 2025-04-10
Start the official farmOS 4.x branch

mstenta:
I propose that we start an official `4.x` branch!

(Actually... I already created a `4.x` branch in the farmOS GitHub repo https://github.com/farmOS/farmOS, but it's currently at the same commit as `3.x`, so that I could open this against it. I have not pushed `4.x` to the Drupal.org repo https://git.drupalcode.org/project/farm yet.)

Once we push `4.x` to the Drupal.org repo and tag a "development snapshot" release we will have the ability to assign d.o issues to a `4.x-dev` version (like this one, which is currently assigned to the `3.x-dev` version: https://www.drupal.org/project/farm/issues/3488918), BUT it means we won't be able to force-push `4.x` anymore. So an important consideration here is that starting an official `4.x` branch means that we are going to be maintaining `3.x` and `4.x` in parallel, and will no longer be able to rebase `4.x` as changes are merged into `3.x`. We've avoided taking this step to maintain flexibility up until now, but I think we're ready.

It also means that we'll be able to start opening more pull requests against `4.x`. I have a bunch of `4.x-*` branches already that I'm eager to open PRs for.

The commits in this PR are just the first steps. It doesn't even upgrade to Drupal 11 yet (that's in my `4.x-d11` branch). All this does is:

- Change all references of `3.x` to `4.x` in code, docs, and workflows.
- Remove `3.x` release notes from `CHANGELOG.md` (note that this PR is on top of #948, so it follows the pattern established there).
- Remove stuff that we deprecated in `3.x`:
    - d7 Asset migration source plugin (see https://www.drupal.org/project/farm/issues/3410701)
    - d7 Plan migration source plugin (see https://www.drupal.org/project/farm/issues/3498064)
    - Data stream migration plugins (see https://www.drupal.org/project/farm/issues/3498067)
    - `drupal/exif_orientation` module (see https://github.com/farmOS/farmOS/pull/941)
- Update minimum PHP version to 8.3.


mstenta:
I pushed a  `4.x` branch to https://github.com/farmOS/composer-project as well so that GitHub Actions can run.

https://github.com/farmOS/composer-project/compare/3.x...4.x

CHANGELOG.md
```diff
@@ -386,26 +22,4 @@ farmOS 2.x release notes are available in the 2.x branch's
...
-[3.2.0]: https://github.com/farmOS/farmOS/releases/tag/3.2.0
-[3.1.2]: https://github.com/farmOS/farmOS/releases/tag/3.1.2
-[3.1.1]: https://github.com/farmOS/farmOS/releases/tag/3.1.1
-[3.1.0]: https://github.com/farmOS/farmOS/releases/tag/3.1.0
-[3.0.1]: https://github.com/farmOS/farmOS/releases/tag/3.0.1
-[3.0.0]: https://github.com/farmOS/farmOS/releases/tag/3.0.0
-[3.0.0-beta3]: https://github.com/farmOS/farmOS/releases/tag/3.0.0-beta3
-[3.0.0-beta2]: https://github.com/farmOS/farmOS/releases/tag/3.0.0-beta2
-[3.0.0-beta1]: https://github.com/farmOS/farmOS/releases/tag/3.0.0-beta1
+[Unreleased]: https://github.com/farmOS/farmOS/compare/3.x..HEAD
```
symbioquine: I'm wondering if this should be a "three dot" diff link and should be between 3.x and 4.x for now...

```suggestion
[Unreleased]: https://github.com/farmOS/farmOS/compare/3.x...4.x
```

Ref: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-comparing-branches-in-pull-requests#three-dot-and-two-dot-git-diff-comparisons

CHANGELOG.md
```diff
@@ -7,374 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
```
symbioquine, line 9: Shouldn't this PR add a `### Removed` section - with at least the removal of the deprecated stuff?

Also maybe a `### Changed` section with a little note about updating docs/workflow references to 4.x and PHP version to 8.3?

symbioquine (changes requested):

CHANGELOG.md
mstenta: Ah I think the original intention was to replace the `3.x` with `4.x` - so it would be `4.x..HEAD` (following the same pattern as what we do now in `3.x`). But this is a good question... it sounds like you're suggesting we *temporarily* change it to `3.x...4.x`, but then change it to `4.x..HEAD` at some point in the future? Would that be when the first release is made, perhaps?

I'm inclined to keep the number of things we need to remember small, so I'd probably vote for just changing it to `4.x..HEAD` now and then we're good to go.

Although maybe I'm misunderstanding your suggestion.

CHANGELOG.md
mstenta: Yes good point! I'll push a fixup commit with some of these details under the `[Unreleased]` heading.

mstenta: > note about updating docs/workflow references to 4.x

I don't think we need to include this in the release notes, but I will add a point about PHP 8.3.

mstenta: Done. I ended up adding the changes to each relevant commit, rather than a single fixup commit.

> I will add a point about PHP 8.3.

One other change I made is to say "8.3+" instead of just "8.3". This might require some follow-up though, to confirm that it works with PHP 8.4.

CHANGELOG.md
symbioquine: > `git diff A..B`

For clarity I'll refer to `A` as the "base" branch/tag and `B` as the "topic" branch/tag. (Which is my understanding of how the documentation I linked above refers to them.)

> it sounds like you're suggesting we _temporarily_ change it to `3.x...4.x`, but then change it to `4.x..HEAD` at some point in the future? Would that be when the first release is made, perhaps?

Yeah, I guess there's a couple things going on:

1. Currently the "unreleased" diff link doesn't work at all (because `HEAD` *is* the `3.x` branch).
2. The base branch for the "unreleased" diff link will have to change again/anyway when the first release is made. IMHO, it should be `3.x...4.x` right now and `4.0.0..4.x` in the future.
3. The link for the first release will need to the be a bit special going forward also. IMHO, it should be `3.x...4.0.0` once the first release is made.

**Now:**

```md
[unreleased]: https://github.com/farmOS/farmOS/compare/3.x...4.x
```

**After first release:**

```md
[unreleased]: https://github.com/farmOS/farmOS/compare/4.0.0..4.x
[4.0.0]: https://github.com/farmOS/farmOS/compare/3.x...4.0.0
```

I believe we actually want the "three-dot" diff in the specific case where we're comparing between 3.x and 4.x/4.0.0 because we want to show the commits that are new to 4.x since the last common/shared ancestor commit. This stack overflow answer has helpful diagrams: https://stackoverflow.com/a/7256391

It is debatable but in my opinion, we don't really want the "two-dot" comparison for that case because it would change over time in the event that we make new commits to the 3.x branch. Those commits are irrelevant to the "story of farmOS version 4" - again IMHO 🤓.

mstenta: Ah so it looks like we somehow ended up diverging from https://keepachangelog.com/en/1.1.0/ on this! In their example, they use 3 dots:

`[unreleased]: https://github.com/olivierlacan/keep-a-changelog/compare/v1.1.1...HEAD`

But in our current 3.x branch, we use 2:

`[Unreleased]: https://github.com/farmOS/farmOS/compare/3.4.4..HEAD`

Maybe the third dot got deleted somewhere along the way. :sweat_smile: 

OK well I think what you said makes sense. And maybe we should actually consider changing it in 3.x to:

`[Unreleased]: https://github.com/farmOS/farmOS/compare/3.4.4...3.x`

... Because if we change the default branch to `4.x`, but still have unreleased changes on `3.x` at that point, then it would behave incorrectly. So maybe we should just standardize on using the branch name itself instead of `HEAD` all the time?

So to summarize, we should:

1. Change to `...3.x` in the `3.x` branch.
2. Change to `3.x...4.x` in this PR.
3. After the first release, change to `4.0.0...4.x`.

Note that this is slightly different from your suggestion, because it maintains 3 dots throughout, in line with https://keepachangelog.com/en/1.1.0/.

Does that make sense?

symbioquine: > OK well I think what you said makes sense. And maybe we should actually consider changing it in 3.x to:
> 
> `[Unreleased]: https://github.com/farmOS/farmOS/compare/3.4.4...3.x`

Agreed!



> So to summarize, we should:
> 
>     1. Change to `...3.x` in the `3.x` branch.
> 
>     2. Change to `3.x...4.x` in this PR.
> 
>     3. After the first release, change to `4.0.0...4.x`.
> 
> 
> Note that this is slightly different from your suggestion, because it maintains 3 dots throughout, in line with https://keepachangelog.com/en/1.1.0/.
> 
> Does that make sense?

I think so...

Two dots seems slightly more logical to me for the in-branch comparisons, but I'm not sure I can make a strong enough argument in that direction to justify diverging from the keep-a-changelog guidance. (I looked back through their history and it looks like they've always used three-dots.)

CHANGELOG.md
mstenta: > Change to `...3.x` in the `3.x` branch.

Done: https://github.com/farmOS/farmOS/commit/98145e2e5d9492802f342ac97f7cbac0d0542714

> Change to `3.x...4.x` in this PR.

Done: https://github.com/farmOS/farmOS/pull/949/commits/1bb19302d00c5cf073a36b681e8c3f4192579a5a#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR25

> After the first release, change to `4.0.0...4.x`.

Will do when the time comes! :smile:

CHANGELOG.md
mstenta: > Change to `...3.x` in the `3.x` branch.

Done: https://github.com/farmOS/farmOS/commit/98145e2e5d9492802f342ac97f7cbac0d0542714

> Change to `3.x...4.x` in this PR.

Done: https://github.com/farmOS/farmOS/pull/949/commits/1bb19302d00c5cf073a36b681e8c3f4192579a5a#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR25

> After the first release, change to `4.0.0...4.x`.

Will do when the time comes! :smile:

paul121: Looks like we had 3 dots for 3.2.2: https://github.com/farmOS/farmOS/commit/b54f3532a94e7e3b2fe0c52bff21b2aced0dd1b6

And lost one with 3.2.3 : https://github.com/farmOS/farmOS/commit/889737c2a3627856a9523dbb7887a6251d98e103

I didn't even know there was difference! Fun to learn :smile:

paul121 (approved):

CHANGELOG.md
mstenta: Ah ha! There it is. :-)

Maybe we should suggest that keepachangelog.com provide a GitHub Action for validating our `CHANGELOG.md` - to avoid similar mistakes in the future. :laughing:

mstenta:

Great! Assuming @symbioquine approves, I think this is ready!

Since we are close to releasing 3.4.5, I will probably wait until that is done, then do one last rebase, so we're starting `4.x` at the latest commit of `3.x`. At that point we will be maintaining both branches in parallel.

symbioquine (approved):
LGTM!

mstenta:
`4.x` has begun! :tada:

PR #950, opened 2025-04-14
Fix hidden entity fields

mstenta:
I noticed recently that the "Authored by" and "Authored on" fields were being displayed on records:

Screenshot from 2025-04-14 09-23-18
Screenshot from 2025-04-14 09-25-14

We intentionally hide those via code in our `farm_entity_fields.module`:

https://github.com/farmOS/farmOS/blob/8bcc1c09a266d668d59d3e4ca9d83ba752ae37b0/modules/core/entity/modules/fields/farm_entity_fields.module#L68-L73

I traced this to a recent change I made in #911:

https://github.com/farmOS/farmOS/commit/9950c97c12a00736f6ad38b8ec3a33ffeb3ae23f

While debugging with XDebug, I found that `switch ($options['hidden'])` is ending up in `case 'form':` when `$options['hidden']` is `TRUE`.

Screenshot from 2025-04-14 09-35-58

This SO question (and the first answer) explains what's happening: https://stackoverflow.com/questions/8829229/why-switchtrue-in-php-with-a-strange-logic

To fix this, I changed the code from a `switch` statement to `if` + `elseif` + `else`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- [Fix hidden entity fields #950](https://github.com/farmOS/farmOS/pull/950)
+
 ## [3.4.4] 2025-04-03
 
 ### Changed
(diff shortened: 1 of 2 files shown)


paul121 (approved):

mstenta:
PHPStan wasn't happy with my solution, so I took a slightly different approach. I reverted https://github.com/farmOS/farmOS/commit/9950c97c12a00736f6ad38b8ec3a33ffeb3ae23f and then refactored the original code to declare the `$options['hidden']` variable type as `bool|string`. This seems to please the PHPStan gods. :-)

paul121 (approved):

PR #951, opened 2025-04-23
Fix missing Assets tab on locations

mstenta:
In farmOS, assets that are designated as "locations" have an "Assets" tab on them that shows a list of assets currently in that location. A bug was introduced recently that prevents this tab from showing.

Originally reported in: https://farmos.discourse.group/t/how-to-list-all-the-current-plant-quantities-in-my-vegetable-garden-field/2273/16

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - [Fix hidden entity fields #950](https://github.com/farmOS/farmOS/pull/950)
+- [Fix missing Assets tab on locations #951](https://github.com/farmOS/farmOS/pull/951)
 
 ## [3.4.4] 2025-04-03
 
--- a/modules/core/ui/views/src/Access/FarmLocationAssetViewsAccessCheck.php
+++ b/modules/core/ui/views/src/Access/FarmLocationAssetViewsAccessCheck.php
@@ -50,10 +50,10 @@ class FarmLocationAssetViewsAccessCheck implements AccessInterface {
    */
   public function access(RouteMatchInterface $route_match) {
 
-    // If there is no integer "asset" parameter, deny access.
+    // If there is no "asset" parameter, bail.
     $asset_id = $route_match->getParameter('asset');
-    if (empty($asset_id) || !is_int($asset_id)) {
-      return AccessResult::forbidden();
+    if (empty($asset_id)) {
+      return AccessResult::allowed();
     }
 
     // Allow access if the asset is a location.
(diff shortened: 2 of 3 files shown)


mstenta:
Opening this as a draft with a first commit that adds an automated test to demonstrate the issue. I will push a fix commit and open this for review next...

mstenta:
This was introduced by #932, which adds a check to see if the asset ID that is passed in via URL (`/asset/[id]/assets`) is an integer. This does not return `TRUE` because the ID is a string.

I would like to believe that I tested this, but maybe I just assumed we had an automated test for it. I don't remember, but it was my bad. I added an automated test so it won't happen again.

I may propose simply reverting the changes in #932 because now that I'm looking at it, similar changes should have probably been considered for other places as well. For example, these other classes that have similar logic:

- https://github.com/mstenta/farmOS/blob/a192967c8d847a9b171892bd2d93c3c13565e029/modules/core/ui/views/src/Access/FarmInventoryAssetViewsAccessCheck.php
- https://github.com/mstenta/farmOS/blob/a192967c8d847a9b171892bd2d93c3c13565e029/modules/core/ui/views/src/Access/FarmAssetLogViewsAccessCheck.php
- https://github.com/mstenta/farmOS/blob/a192967c8d847a9b171892bd2d93c3c13565e029/modules/core/ui/views/src/Access/FarmAssetChildrenViewsAccessCheck.php

mstenta:
The new test failed, which demonstrates the issue.

https://github.com/mstenta/farmOS/actions/runs/14621813931/job/41025271945

I pushed a commit that reverts the change made by #932, to demonstrate that it was the cause.

mstenta:
Relevant from #932:

> This appeared in Farmier logs. It's a pretty obscure bug because it requires a user to manually type in an invalid URL, but no harm in making our logic a bit more defensive to prevent fatal errors.
>
> Notably, this also changes the return from `AccessResult::allowed()` to `AccessResult::forbidden()`. I'm not sure why it was originally defaulting to an "allowed" return. That doesn't make much sense. I wrote the logic in 2021, so I take the blame. :-)

I will look into:

a) How to best fix the obscure bug that #932 aimed to fix.
b) Whether we should keep that second change (`AccessResult::forbidden()` instead of `AccessResult::allowed()`)
c) Doing this in all the relevant places, not just the "Assets" tab.

Maybe we want to address those in a separate PR, and merge this one first to fix the bug, which is a lot worse than the one that #932 aimed to fix.

mstenta:
> Maybe we want to address those in a separate PR, and merge this one first to fix the bug, which is a lot worse than the one that https://github.com/farmOS/farmOS/pull/932 aimed to fix.

I think I would like to approach it in that order.

I added a `CHANGELOG.md` line for this PR. Ready for review...

symbioquine (approved):
LGTM!

paul121 (approved):

PR #952, opened 2025-04-25
Gracefully handle invalid entity IDs in Views access checks

mstenta:
This is a followup to #951 (and includes it's commits), which fixes a bug introduced by #932.

This PR fixes the bug that #932 aimed to fix again, but in a better and more complete way, and with a lot more tests. :-)

For background, read comments from #951 starting here:

https://github.com/farmOS/farmOS/pull/951#issuecomment-2824808928


mstenta:
Force pushed to fix `phpstan` and `phpcs` issues.

modules/core/ui/views/tests/src/Functional/FarmUiViewsTest.php
```diff
@@ -21,20 +22,49 @@ class FarmUiViewsTest extends FarmBrowserTestBase {
...
+   * Run all tests.
+   */
+  public function testAll() {
+    $this->doTestAssetViews();
+    $this->doTestLogViews();
+    $this->doTestAssetsByLocationView();
+    $this->doTestAssetChildrenView();
+    $this->doTestAssetInventoryView();
+    $this->doTestAssetLogsView();
+  }
```
paul121: Is there a reason for combining here instead of running in separate test methods? Is this order important or are there dependencies between test methods?

It seems this introduces the need for `deleteAllEntities()` method and other complexity?

modules/core/ui/views/src/Access/FarmTaxonomyTermEntityViewsAccessCheck.php
```diff
@@ -71,15 +71,22 @@ public function __construct(string $base_entity_type, EntityTypeManagerInterface
...
+    // filter validation returns a 404.
     $term_id = $route_match->getParameter('taxonomy_term');
-    $entity_bundle = $route_match->getParameter('entity_bundle');
-
-    if (empty($term_id) || empty($entity_bundle)) {
-      return AccessResult::forbidden();
+    /** @var \Drupal\taxonomy\TermInterface|null $term */
+    $term = $this->taxonomyTermStorage->load($term_id);
+    if (is_null($term)) {
+      return AccessResult::allowed();
```
paul121, line 81: Trying to remember what the rationale is for either of these access results... I saw you ponder it various times across the issues @mstenta but curious if you figured anything out?

I'm not sure if this is correct - but the thought I had is that in theory someone might want to provide a route that matches this pattern but uses an invalid value, maybe like `/asset/property/assets`. If we use `AccessResult::forbidden()` then it might hard-code a 403? But `AccessResult::allowed()` is more graceful?

Or maybe we should use `AccessResult::neutral()` ?? I think the route isn't allowed unless something responds `allowed`, so `neutral` might be a fine default.

In fact I see that we use `AccessResult::allowedIf()` in some of these access checks, which will return `netural` in the false condition: https://git.drupalcode.org/project/drupal/-/blob/11.x/core/lib/Drupal/Core/Access/AccessResult.php?ref_type=heads#L69-81

paul121 (commented):
I think we just punted on doing this extra validation when we first added these features... but nice to get these edge cases covered!

modules/core/ui/views/tests/src/Functional/FarmUiViewsTest.php
mstenta: This is the reason: https://github.com/farmOS/farmOS/pull/797

Combining functional tests reduces test run time.

> Is this order important or are there dependencies between test methods?

No the order doesn't matter, and yes I added the `deleteAllEntities()` logic to keep everything clean and separate between them. I think that the tests would still pass if we removed the `deleteAllEntities()` - but it felt better to have it explicit, to prevent potential issues in the future.

Splitting them up into separate tests would achieve the same goal, at the expense of longer test run time. I felt that these were all similar enough that it didn't warrant separate bootstraps, but didn't want to have them all in the same test method that we started with (https://github.com/farmOS/farmOS/pull/952/commits/c81ed75f67b70fb2a436e9bc30082a07c19e0d96), so I split them into separate methods run by `testAll()`, and added the `deleteAllEntities()` logic as a simple way to clean up in between. The same `deleteAllEntities()` code could be used if we kept them all in the same method too, so it feels natural to keep it IMO.

modules/core/ui/views/src/Access/FarmTaxonomyTermEntityViewsAccessCheck.php
mstenta: So... there are some layers here...

First, all of these are simply Views that define their own routes, access checks (simple permission based), and contextual filter validation. For example, the "Assets by Location" display (`farm_asset.page_location`) adds a route at `/asset/%asset/assets`, requires the "View any assets" permission, and adds a Contextual filter for "Asset: Current location" (to filter the list of assets to ones that are currently located in `%asset`. The Contextual filter is configured to "validate" that `%asset` is indeed an asset, and if it isn't then it shows "Page not found" (404).

We don't want the "Assets" (`/asset/%asset/assets`) tab to show on all assets, though... only on "location" assets. There isn't any way to do this within the Views configuration, so we added an access check with the logic we needed (check `is_location == TRUE`). These access checks run *before* Views gets the request and processes the Contextual filter validation.

Within the context of the asset check, we only have the ability to return access results (allowed, forbidden, or neutral). We can't return a 404, like Views does, even though that might be more desirable. (Side note: maybe there is a way to force that from within an access check... but we aren't doing that now.)

So... within our access checks we essentially want to either: a) step in and deny access (so the tab doesn't show), or b) pass it off to Views to decide (which will then check the user permission and validate the contextual filter).

> I'm not sure if this is correct - but the thought I had is that in theory someone might want to provide a route that matches this pattern but uses an invalid value, maybe like `/asset/property/assets`. If we use `AccessResult::forbidden()` then it might hard-code a 403? But `AccessResult::allowed()` is more graceful?

We could test this, but I think that a route of `/asset/property/assets` would work fine, because it is a different route than `/asset/%/assets`, and so our access check wouldn't get applied to it in the first place. Here is where we add the access check, you can see that it's applied to the `view.farm_log.page_asset` route by name, not by path: https://github.com/farmOS/farmOS/blob/9d6d78e4f9d8b8aa791b8a65a11018a19a366b5e/modules/core/ui/views/src/Routing/RouteSubscriber.php#L24

Maybe the broader question is: how does Drupal handle wildcard routes vs a similar route with a fixed value where the wildcard is. My assumption is that when a request is made to `/asset/property/assets`, Drupal will give precedence to the `/asset/property/assets` route over the `/asset/%/assets` route. I haven't tested this, though. But it's sort of a separate question from all this.

> Or maybe we should use `AccessResult::neutral()` ?? I think the route isn't allowed unless something responds `allowed`, so `neutral` might be a fine default.

We could consider switching from `AccessResult::allowed()` to `AccessResult::neutral()`, and I think it would have the same effect. I don't feel strongly either way. `AccessResult::neutral()` might make more sense, since our goal is to pass it off to Views in those cases, and if `AccessResult::allowedIf()` returns `neutral` then it would actually be more consistent for us to do the same.

I'll add a commit to change to `AccessResult::neutral()` and we can see if tests pass.

mstenta: https://github.com/farmOS/farmOS/pull/952/commits/43e1849bf7e6505e5c74d41f273b94c321c8b48d

mstenta: Interesting... returning `AccessResult::neutral()` causes the tests to fail, because an invalid asset ID ends up as an access denied, rather than passing it to Views where it would result in 404.

I traced the logic in core and this is where it ends up: https://git.drupalcode.org/project/drupal/-/blob/1026286778ac8cf0d64ebaa9288768098e583eaa/core/lib/Drupal/Core/Routing/AccessAwareRouter.php#L120

Basically, Drupal core will return access denied `if (!$access_result->isAllowed())`, and a "neutral" access result does not return true for that.

It seems that the Views access check isn't being taken into account during this step. Maybe Views access checking happens after Drupal core's initial `AccessAwareRouter` checks, but because "neutral" ends up as "denied" in that pass it never makes it through to Views. Not really sure!

What do you think? Worth digging deeper? Or OK to leave it as `AccessResult::allowed()` for now? This is sort of a separate decision IMO, because the current `3.x` code uses `AccessResult::allowed()`, so we're keeping that the same.

paul121: > because an invalid asset ID ends up as an access denied, rather than passing it to Views where it would result in 404.

Oh interesting. Yea this makes sense. The 404 logic happens once Views gets a hold of it.

> What do you think? Worth digging deeper? Or OK to leave it as AccessResult::allowed() for now? 

Yea, leaving `AccessResult::allowed()` makes sense to me!

mstenta: Cool! I removed that last commit and force-pushed. Ready for final review!

modules/core/ui/views/tests/src/Functional/FarmUiViewsTest.php
paul121: Makes sense. If the tests were all very different I'd argue to keep them separate but agree, all the logic is quite similar here.

Maybe moving `deleteAllEntities()` into the `testAll()` function would make this dynamic easier to understand? Normally test methods don't need to do any cleanup.

```suggestion
  /**
   * Run all tests.
   */
  public function testAll() {
  
    // Run each farm UI views test in a single method to decrease test time.
    // Delete all entities between running each test.
    $this->doTestAssetViews();
    $this->deleteAllEntities();
    
    $this->doTestLogViews();
    $this->deleteAllEntities();
    
    $this->doTestAssetsByLocationView();
    $this->deleteAllEntities();
    
    $this->doTestAssetChildrenView();
    $this->deleteAllEntities();
    
    $this->doTestAssetInventoryView();
    $this->deleteAllEntities();
    
    $this->doTestAssetLogsView();
    $this->deleteAllEntities();
  }
```

mstenta: Ah yea I like that idea! Standby for a fixup...

mstenta: https://github.com/mstenta/farmOS/commit/b44dc9d4e5940152780bafab81d9d829b50a69df

paul121 (approved):

PR #953, opened 2025-04-25
Validate entities generated by quick trait methods

mstenta:
This PR adds `$entity->validate()` calls to the `farm_quick` module's trait methods for creating assets, logs, quantities, and terms, to prevent creation of invalid entities. If violations are found, it throws an `EntityMalformedException`.

This is not great UX for the end-user, but it prevents potentially saving invalid entities, which is more important.

See this chat for more context and potential next steps for 4.x: https://irc.farmos.org/bot/log/farmOS/2025-04-25


mstenta:
Converted back to draft because there are test failures.

mstenta:
OK let's see if all tests pass now.

In debugging the tests, I ended up making another small change: I moved the `farm_asset_reference` View from the `farm_entity_views` module to the `farm_log_asset` module. That View probably should have been moved back when we created the `farm_log_asset` module. It is the only View in `farm_entity_views`, which otherwise is largely concerned with lower level workaround for things like https://www.drupal.org/node/2489476 and https://www.drupal.org/project/entity/issues/3206703. Moving the View required a few small adjustments to dependencies in other tests, but all-in-all it's a nice small improvement and cleanup.

The reason I needed to move it in order to fix the tests for this PR is because that `farm_asset_reference` View is used for validating the `asset` field on logs, and since we're calling `$log->validate()` in this PR now, it wasn't able to find the View and therefore the `ValidReferenceConstraint` failed, which meant that tests for any quick forms that used the `asset` field failed because the logs didn't validate.

paul121 (commented):
This looks good to me.

But what happens in the UI if this exception is thrown and never caught? A somewhat nasty error or crash page?

If so, I think we could catch this exception in the parent quick form submission logic and provide a somewhat nicer message "Error submitting quick form" ? Doesn't really help the user know what is wrong, but at least they wouldn't end up with a blank screen

https://github.com/farmOS/farmOS/blob/9d6d78e4f9d8b8aa791b8a65a11018a19a366b5e/modules/core/quick/src/Form/QuickForm.php#L143-L145

mstenta:
> I think we could catch this exception in the parent quick form submission logic and provide a somewhat nicer message "Error submitting quick form" ? Doesn't really help the user know what is wrong, but at least they wouldn't end up with a blank screen

That's a great idea @paul121 - let me give that a try...

mstenta:
I added a commit that catches `EntityMalformedException`s during quick form submit and shows two messages to the user: a generic message, and the message `EntityMalformedException`'s message.

Thoughts on both? One or the other? IMO the generic one is a bit better UX, because it explains the situation in simple terms. But the second one tells you specifically what entity type failed, and includes the quick form ID, which makes reporting the bug easier.

paul121 (approved):
LGTM! I think there is value in both messages. Could maybe combine them together into one message as "Generic message: detail" but I don't have a strong opinion.

PR #954, opened 2025-04-26
Link to upcoming tasks with appropriate sort direction

mstenta:
This is a very minor UX improvement - I would even consider it a small "bugfix"...

On the farmOS dashboard, under "Upcoming tasks", there is a link to "View all upcoming logs" which links to `/logs` with preset filters (pending logs in the future).

https://github.com/farmOS/farmOS/blob/9d6d78e4f9d8b8aa791b8a65a11018a19a366b5e/modules/core/ui/views/farm_ui_views.views_execution.inc#L75

However, the "Upcoming tasks" block sorts logs by timestamp ascending, whereas `/logs` defaults to sorting by timestamp *descending*.

This simply adds `&order=timestamp&sort=asc` to the "View all upcoming logs" link, so that the user experience is more consistent as they jump from the truncated list on the dashboard to the full list on `/logs`.

Notably, we don't need to do the same thing for "View all late logs", because the "Late tasks" dashboard block sorts by timestamp descending, which is already the same as `/logs`.


paul121 (approved):
Makes sense to me

PR #955, opened 2025-04-27
Set asset/log flags via CSV importers

mstenta:
This adds support for setting asset and log flags via the default CSV importers.


symbioquine (approved):
LGTM!

paul121 (approved):

PR #958, opened 2025-04-28
Add a hook for excluding fields from CSV importers

mstenta:
This adds a `hook_farm_import_csv_exclude_fields()` hook so that modules can exclude fields from the default CSV importers.

This replaces the simple logic we have now that simply excludes "hidden" fields:

https://github.com/farmOS/farmOS/commit/c125581241c850ce51987718233911e71235e547#diff-0e77ffc0e1e4eca5cac2f4f650d5696f0bfd1a8d5fbe22d209958a02d2d6c844L205-L210

This logic is too simple. There are cases where it's desirable to import data into a "hidden" field. I am about to open a pull request that adds CSV import support for config fields, which will require this.

This is a breaking change, because it means that any fields that are "hidden" (and therefore excluded from the default CSV importers currently) will need to decide if they want to explicitly implement the new hook in order to continue excluding them. Otherwise, they will become an available column in CSV importers automatically.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ will ensure that all necessary changes have been applied before updating to 4.x.
 ### Added
 
 - [Add support for attributes in farmOS plugin types #963](https://github.com/farmOS/farmOS/pull/963)
+- [Add a hook for excluding fields from CSV importers #958](https://github.com/farmOS/farmOS/pull/958)
 
 ### Changed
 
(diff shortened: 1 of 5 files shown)


mstenta:
Note: I intentionally based this branch on top of #957, which converts the `quick` field into a base field. The `quick` field is currently excluded from CSV importers because it is hidden. When it gets converted to a base field, then it will be excluded because base fields need to be explicitly included. So, as long as this PR is merged after #957 we don't need to implement `hook_farm_import_csv_exclude_fields()` in the `farm_quick` module to explicitly exclude the `quick` field.

paul121 (approved):

PR #959, opened 2025-04-28
Include config fields in CSV importers

mstenta:
This automatically adds config fields (fields that are defined via configuration, rather than base/bundle fields) to default CSV importers.

It builds on the work that was done in #916. It requires #958 because the logic for excluding "hidden" fields will exclude config fields as well.

It also adds support for number fields (`integer` and `decimal`). This allows the `plant_type` config fields to become available for import (`crop_family`, `companions`, `maturity_days`, `harvest_days`, `transplant_days`).


paul121 (approved):
I did not test these changes myself, but the additions to existing tests look good :+1:

PR #960, opened 2025-04-29
Change Animal asset is_castrated attribute to is_sterile

mstenta:
See https://www.drupal.org/project/farm/issues/3494784

### DIFF
--- a/docs/development/api/changes.md
+++ b/docs/development/api/changes.md
@@ -16,6 +16,7 @@
   to a status of `active`.
 - Plans now have additional options for their `status` attribute: `planning`,
   `done` and `abandoned`, in addition to the existing `active` option.
+- The Animal asset `is_castrated` attribute has been renamed to `is_sterile`.
 
 ## 3.x vs 2.x
 
--- a/modules/asset/animal/src/Plugin/Asset/AssetType/Animal.php
+++ b/modules/asset/animal/src/Plugin/Asset/AssetType/Animal.php
@@ -44,10 +44,10 @@ class Animal extends FarmAssetType {
           'view' => -35,
         ],
       ],
-      'is_castrated' => [
+      'is_sterile' => [
         'type' => 'boolean',
-        'label' => $this->t('Castrated'),
-        'description' => $this->t('Has this animal been castrated?'),
+        'label' => $this->t('Sterile'),
+        'description' => $this->t('Is this animal sterile?'),
         'weight' => [
           'form' => 26,
         ],
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,7 @@ requirements (inherited from Drupal 11):
 - [Update Drupal to v11.2, PHPUnit to v11, PHPStan to v2 #980](https://github.com/farmOS/farmOS/pull/980)
 - [Update Docker base image to Debian Trixie #992](https://github.com/farmOS/farmOS/pull/992)
 - [Change how assets and plans are archived #986](https://github.com/farmOS/farmOS/pull/986)
+- [Change Animal asset is_castrated attribute to is_sterile #960](https://github.com/farmOS/farmOS/pull/960)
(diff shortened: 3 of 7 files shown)


modules/asset/animal/src/Plugin/Asset/AssetType/Animal.php
```diff
@@ -43,10 +43,10 @@ public function buildFieldDefinitions() {
           'view' => -35,
         ],
       ],
-      'is_castrated' => [
+      'is_sterile' => [
         'type' => 'boolean',
-        'label' => $this->t('Castrated'),
-        'description' => $this->t('Has this animal been castrated?'),
+        'label' => $this->t('Sterile'),
+        'description' => $this->t('Is this animal sterile?'),
```
symbioquine, line 50: Does/Should this support an `undefined` state where the sterility of the animal is unknown?

docs/development/api/changes.md
```diff
@@ -1,5 +1,9 @@
 # API Changes
 
+## 4.x vs 3.x
+
+- The Animal asset `is_castrated` attribute was renamed to `is_sterile`.
```
symbioquine: Not sure it affects this PR, but it occurs to me that `is_sterile` could also be a useful attribute for plant/seed/compost assets - i.e. triploid, non-pollen-producing, infertile hybrid, hot composted, etc 🤓.

modules/asset/animal/src/Plugin/Asset/AssetType/Animal.php
mstenta: Hmm perhaps - this brings back memories of a similar discussion around another boolean field recently. :thinking: 

We might be constrained by what Drupal's boolean field supports. I think it's possible for boolean fields to have `null` value, but I think the way the form widget works means that it's not possible to deliberately set it to null. Once it's loaded into an asset edit form and saved, the value gets set as true or false in the database.

I think to support 3 distinct options we would need to use a different field type (eg: select list with options of `yes`, `no`, `unknown`). That would expand the scope of this PR, which is simply renaming the existing field/behavior. This also means that the API would change. This is already a breaking change by renaming, but that would mean any applications using it would also need to change the way they handle values.

I don't have strong feelings either way, but I don't know if I have the capacity to take this much further myself. We've also talked about "log based attributes" as a next step, which would replace this field entirely. Maybe we defer to that future work?

docs/development/api/changes.md
mstenta: True! It wouldn't be too hard to add it to other asset types. Just a matter of deciding which. Plant and Seed make sense... you also mentioned Compost but I think the "sterile" meaning is slightly different there (aka `is_sterilized` or `is_kept_sterile`). I remember talking about this on a call together a while back...

We could also defer to the future "log based attributes" work on this...

modules/asset/animal/src/Plugin/Asset/AssetType/Animal.php
symbioquine: I think you're right about deferring this. Though I do think it's a can of worms we'll need to get into...

Representing an explicit unknown state seems like it would be helpful for intake/processing scenarios. e.g. A animal shelter that receives a truckload of abandoned dogs - one probably wouldn't want to assume they're either already sterilized or not before going through and confirming on an animal by animal basis.

The sex of animals/plants (rabbits, ducks, hops, persimmon, ginko, etc) is another case where it could be important to be able to explicitly model the attribute as unknown and then also model *when* it becomes known.

mstenta: Yea those are good examples.

docs/development/api/changes.md
symbioquine: Agreed on deferring this.

I guess you're also right about compost maybe being a different case. Modeling levels of heterogeneity, contamination, fecundity over time all seem related though.

symbioquine (approved):

PR #961, opened 2025-04-29
Add Term Merge module

mstenta:
This adds the Drupal [Term Merge](https://www.drupal.org/project/term_merge) module to farmOS, along with a simple `farm_ui_term` module that depends on it and adds the necessary permission.

For more information, see this forum topic: https://farmos.discourse.group/t/merge-module-testing/1635

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@ requirements (inherited from Drupal 11):
 - [Allow API access to organizations #1010](https://github.com/farmOS/farmOS/pull/1010)
 - [Issue #3304608: Add an "abandoned" log status](https://www.drupal.org/project/farm/issues/3304608)
 - [Add default plan status options: "planning", "done", "abandoned" #986](https://github.com/farmOS/farmOS/pull/986)
+- [Add Term Merge module #961](https://github.com/farmOS/farmOS/pull/961)
 - [Add support for attributes in farmOS plugin types #963](https://github.com/farmOS/farmOS/pull/963)
 - [Add a hook for excluding fields from CSV importers #958](https://github.com/farmOS/farmOS/pull/958)
 - [Include config fields in CSV importers #959](https://github.com/farmOS/farmOS/pull/959)
--- a/composer.json
+++ b/composer.json
@@ -43,6 +43,7 @@
         "drupal/simple_oauth_password_grant": "^2.1",
         "drupal/state_machine": "^1.12",
         "drupal/subrequests": "3.0.12",
+        "drupal/term_merge": "^2.0",
         "drupal/token": "^1.15",
         "drupal/views_geojson": "^1.3",
         "drush/drush": "^13.0",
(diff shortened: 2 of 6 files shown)


paul121:
I don't have a strong opinion on this.  It does sound useful, but only once your farmOS instance gets "messy" :-) As such I do think it would be nice if this was an optional module and not always enabled with `farm_ui`. Otherwise it is just more "things" clutter/etc for new users to learn. :shrug:

mstenta:
> it would be nice if this was an optional module and not always enabled with `farm_ui`.

I thought about this too.

We need to wrap it in a "ui" module either way, so that we can add the necessary permission. Since the permission is the only bit that's needed (on top of enabling the module), I originally thought about just sticking it into an existing UI module, but decided it made sense to keep as it's own "term UI" module. This gives site admins the option to remove it if they want. So the flexibility is there, it's just a question of whether or not it should be installed by default.

If we don't, then we need to add it to `setup/modules` so that non-admins can find/enable it.

> Otherwise it is just more "things" clutter/etc for new users to learn.

My feeling is that this is a small enough "thing" that it isn't a burden, and it's purpose is intuitively clear. All it adds is a "Merge" tab to the term list page (where you would intuitively look for it).

Screenshot from 2025-05-28 06-58-16

Screenshot from 2025-05-28 06-58-30

This is the sort of thing that you don't know you need until you need it, and is so small that you might not think to look for it as an optional module. IMO someone who finds themselves with a cluttered taxonomy would look for some kind of option in the Setup > Taxonomy section.

I think the intuitive feature benefit outweighs the "clutter" of adding one tab. Honestly, this feels to me like a feature that should be part of Drupal core taxonomy UI.

> It does sound useful, but only once your farmOS instance gets "messy" :-)

I think this is a very common problem. Whether due to user error or bugs in code, I've seen exact duplicate terms in almost every site I've looked at - especially in the Units vocabulary (eg: multiple terms named "lbs"). It could be that there were bugs in the past that created them, or bugs right now that we're unaware of (or *future* bugs in core/contrib modules!). Those would be separate issues in themselves, of course, but IMO there's a need to provide some more control to help users manage their vocabularies, even if it's not about exact duplicates. The other thing I see all the time are terms that *should* be merged (like "lb", "lbs", "pound", "pounds").

Without this module there is no way to deal with these, other than manually cleaning them up (and ALL the entities that reference them). I've personally experienced what it feels like to realize that - it's very frustrating and disheartening! And because the burden of cleaning them up manually is so high, most users will simply live with the mess (myself included)! Over time that mess wears on you. :sweat_smile:

So in my mind this is a very small addition (a single "Merge" tab, low/no maintenance burden) with a very big UX benefit.

symbioquine (approved):
I'm in favor of including this by default

mstenta:
Rebased onto `4.x`, updated for Drupal 11, and changed the `drupal/term_merge` version constraint from `^2.0@beta` to `^2.0`. I also tested the update hook and term merge module manually again. All is working as expected.

paul121 (approved):

PR #963, opened 2025-05-08
Add support for attributes in farmOS plugin types

paul121:
See https://www.drupal.org/project/farm/issues/3461548

This adds support for attributes to all of our core plugin types. Contrib is still able to use annotations, but we will drop support for this in a future release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- [Add support for attributes in farmOS plugin types #963](https://github.com/farmOS/farmOS/pull/963)
+
 ### Changed
 
 - farmOS 4.x requires PHP 8.3+.
+
+### Deprecated
+
+- [farmOS core plugin type annotations are deprecated](https://www.drupal.org/node/3523485)
 
 ### Removed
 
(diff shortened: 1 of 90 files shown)


mstenta (changes requested):
This looks great @paul121! I skimmed through all the changes and they make sense. Tests are all green, so I'm confident that it's all working.

Question: I see that we extend Drupal core's `DefaultPluginManager` in most (all?) cases... and it looks like that provides support for both Annotations and Attributes out-of-the-box for backwards compatibility. That's great! I assume we can deprecate and remove farmOS Annotations on our own schedule, as long as it happens before that class removes the BC logic. What will it look like to remove Annotations? Will we be making changes to all the `__construct()` methods on our plugin manager classes again?

I think the last thing to do is add `@deprecated` comments to all the relevant Annotation classes to indicate when they will be removed, and create a Drupal.org change record to link to. I think one CR can cover all of the plugin types, rather than creating one for each.

paul121:
> I assume we can deprecate and remove farmOS Annotations on our own schedule, as long as it happens before that class removes the BC logic. What will it look like to remove Annotations? Will we be making changes to all the __construct() methods on our plugin manager classes again?

Yep, we can remove whenever we like. Core made this work by changing the parameters in the `__construct` function. They changed the annotation parameter to expect an *attribute* class, but for backwards compatibility continue to support providing an annotation class here. Going forward you need to provide the attribute class first, and if you want to support annotations, just pass an annotation class as the last parameter.

For us, we'll just drop the last annotation class parameter once we do not want to support annotations. And via your suggestion I've already prepared a commit that does this!

https://github.com/paul121/farmOS/commits/4.x-attributes-remove-farm-core-annotations/
https://github.com/paul121/farmOS/commit/99b67e7227c681e4e9f3e79240fe8880c0911623

Once we remove annotations there is a minor cleanup we can do as well: https://github.com/paul121/farmOS/commit/0a974489a4621d7fb0422195ed8d4ce0e272f0f7

mstenta (approved):
Looks good! Thanks @paul121!

PR #964, opened 2025-05-08
Remove dependency on JSON:API Extras module

mstenta:
We currently use the [JSON:API Extras](https://www.drupal.org/project/jsonapi_extras) module for two purposes:

1. Rename the `/jsonapi` endpoint(s) to `/api`.
2. Disable "internal" resources (mostly a bunch of Drupal core config entity types) to keep `/api` cleaner and closer to https://farmos.org/model/.

I propose that we drop this dependency and implement these two changes ourselves in the `farm_api` module.

There are few benefits to doing so:

- One less contrib dependency to hold us back from major upgrades! :tada: 
- This module is currently causing tests to fail in the `4.x-d11` branch (this is what kicked off this effort).
- We currently have a bunch of YML files in https://github.com/farmOS/farmOS/tree/9c27b0770df8e43f5c23d8d58012c99ad7f7de7a/modules/core/api/config/optional for disabling "internal" Drupal core resource types, and we haven't been keeping up with maintaining those.
- Simplifies how these changes are applied, because we don't need all the extra options that `jsonapi_extras` provides.

This PR changes the way that resources are disabled. Rather than explicitly *disabling* them by name, we add a hook that allows modules to declare that certain *entity types* are *allowed*. Everything else gets disabled. The `farm_api` module implements the hook for most core farmOS entity types, and there are a few other implementations sprinkled throughout for other random ones that should be available via API.


paul121 (approved):
This LGTM once tests are passing!

Only comment I have is if we could consider doing a simple yaml plugin discovery instead of a hook, similar to how we collect our `managed_role_permissions` definitions. Unless we need `hook_alter` functionality (which isn't implemented here) I think that would be my preference.

I mostly suggest this because hooks are going out (?) and it avoids needing `.module` files :-)

mstenta:
Tests were failing because `EntryPointTest` was looking for `node` module resources, which are not explicitly allowed now, and were not explicitly disabled before.

@paul121 I removed that from the test. Is that OK? This test was written a long time ago, and I can't tell if that check was in there for a specific reason. Do you have any recollection?

Here is the diff to fix the test failure: https://github.com/farmOS/farmOS/compare/c1c20fd655c285d59308fea25d793d919e8ab6dc..fef01e1e38fe7ed9beb26d77e503102847b23f7f

mstenta:
> Only comment I have is if we could consider doing a simple yaml plugin discovery instead of a hook, similar to how we collect our `managed_role_permissions` definitions. Unless we need `hook_alter` functionality (which isn't implemented here) I think that would be my preference.

Actually, I thought about adding the `hook_alter` functionality, but figured it could be done in a follow-up. It does make sense that modules might want to explicitly *disallow* certain resources, and that's not currently possible. So I'm inclined to just go ahead and add that now, and keep it as hooks.

> I mostly suggest this because hooks are going out (?) and it avoids needing `.module` files :-)

Hooks aren't going away (for better or worse), and we will be moving them to OOP classes in 4.x (I'm eager to do this as soon as #849 is merged). This will remove the need for `.module` files! See: https://www.drupal.org/node/3442349

We also have this long-standing issue, which is postponed on that work: https://www.drupal.org/project/farm/issues/3415475

paul121 (approved):
>  This test was written a long time ago, and I can't tell if that check was in there for a specific reason. Do you have any recollection?

That looks fine to me. I think this test was largely copied from Drupal core since we are altering that entrypoint page in a custom way. 

Sounds good, onwards with hooks :+1:

modules/core/api/farm_api.api.php
```diff
@@ -28,6 +28,32 @@ function hook_farm_api_meta_alter(array &$data) {
...
+ * @param string[] $entity_types
+ *   An array of entity type IDs allowed by other modules.
+ */
+function hook_farm_api_allow_resource_types_alter(&$entity_types) {
+
+  // Disable view entities.
+  if (in_array('view', $entity_types)) {
+    unset($entity_types[array_search('view', $entity_types)]);
+  }
+}
```
symbioquine, line 55: * Why allow the 'view' entity type only to later remove it again?
* Maybe the function docs can say the "why" not the "what" - roughly what the function does is trivial (and in the function's name), but why we want to do that (and sometimes how it works) is less clear.

modules/core/api/farm_api.module
```diff
@@ -53,3 +53,30 @@ function farm_api_entity_base_field_info(EntityTypeInterface $entity_type) {
...
+    'plan_type',
+    'plan_record',
+    'plan_record_type',
+    'quantity',
+    'quantity_type',
+    'taxonomy_term',
+    'taxonomy_vocabulary',
+    'user',
+    'user_role',
+  ];
```
symbioquine, line 81: I'm struggling a bit with the rational for some modules declaring their own allowed resource types and others being declared here in the `farm_api` module.

I thought at first it was something about modules which are commonly disabled (declare theirs separately), but then `plan`/`data_stream`/etc would be out in their respective modules...

modules/core/api/farm_api.api.php
mstenta: This particular file is `farm_api.api.php`, which is for *documenting* hooks provided by the `farm_api` module. These two are just examples, and are not actually used (although I did add similar ones to the module that's used in automated testing).

symbioquine: ah, right! Makes sense.

modules/core/api/farm_api.module
mstenta: Great question - I'll explain my rationale, but open to feedback on it.

All of the modules listed in this hook (implemented in the `farm_api` module) are ones that are not in the `farm_*` namespace. Similar to the `log` module (which is a separate repo), `asset`, `plan`, `quantity`, etc aren't *technically* specific to farmOS. In theory, they could be split out to contrib modules if there was ever a desire to do so. I made the decision very early on to keep them similar to `log` in that way.

Then, we have `farm_*` modules which augment those modules for farmOS's needs. For example, the `farm_entity` module is responsible for enabling bundle plugins for all our entity types.

I'm fairly confident that we've been consistent with that design/decision process, although I wouldn't be surprised if a few things have slipped in here and there. But that's the reason it's split up this way. Hope that makes sense!

symbioquine: I think that makes sense.

To paraphrase; the things listed here are entity types which farmOS is "built of" - but which aren't themselves supposed to be "aware" that they're part of farmOS. i.e. They shouldn't have farmOS-specific hooks in them.

That brings up two points that probably aren't strictly part of this PR:

* I think that relationship (the directionality of the module dependencies) could be enforced by a test
* Perhaps that design/decision should be documented somewhere that folks digging into the codebase will tend to find/read?

mstenta: Yes that sums up the design. And yea agreed, we could document that somewhere to make it more clear. Stepping back, it might be nice to provide some kind of "deep dive" overview of the codebase in a form of documentation somewhere. We don't really provide something like that currently.

> I think that relationship (the directionality of the module dependencies) could be enforced by a test

Interesting idea. Certainly worth exploring! Although I wonder how that would be implemented. Either way I agree it is a separate issue from this PR.

symbioquine (approved):
LGTM!

P.S. I love seeing this: image

mstenta:
> P.S. I love seeing this: image

Yesss! Haha same. And that's just stuff removed from *our* codebase. It doesn't include removing all of the code in `jsonapi_extras`! ;-)

Thanks for the review!

PR #965, opened 2025-05-08
Allow more granular access to views

paul121:
For more background see this issue: [Allow more granular access on default farmOS Views](https://www.drupal.org/project/farm/issues/3415653)

This PR also includes a few simple commits adding entity `query_access` handlers to our main farmOS core entity types. Borrowed (and rebased) from the old MR here: https://www.drupal.org/project/farm/issues/3413263

Overview of these (proposed!) changes:

1. Add a collection permission for all core entity types eg: `access {entity_type} collection`.

This pattern was recently adopted/supported by Drupal core: https://www.drupal.org/node/2955178.

Notably, Drupal core does not have a way to "provide" this permission, unlike what we are accustomed to seeing via the `entity` module's `permission_provider` system. In some ways the `entity` module was already trying to do this via the `access {entity_type} overview` permission that they provide (and is currently present in farmOS). However, we have never used this permission, and IMO would be better to follow the Drupal core pattern going forward.

---

2. Provide a custom routing access check that ensures users have "entity collection access"

The logic for this access check is as follows:
- Require that the user has the `access {entity_type} collection` permission
- Require that the user has at least ONE of the following view permissions: `view any {entity_type}` or `view own {entity_type}` or (for any 1 bundle) `view any {bundle} {entity_type}` or `view own {bundle} {entity_type}`
- OR if a the route is only designed to display a single bundle - then require that the user has a view permission for that specific bundle, if not either of the higher level `view any {entity_type}` or `view own {entity_type}` permissions

The rationale for this logic is that if we have a page listing a collection of *all* or *many* asset types, we don't want to require that the user has the `view any asset` permission. As long as the user has access to at least one `view * * asset` permission, this listing of all assets should still be visible to them. Thanks to entity `query_access` handlers, all the assets on this page will be filtered to only show those that the user has proper view permissions for (`any/own` and `bundle any/own`).

We also cannot depend on only the collection permission `access asset collection` by itself. Even though we can rely on `query_access` handlers for fundamental security, there is still a UX issue where if a page/view is only going to display `equipment` assets, but the user has no access to see any `equipment` assets, this page would still be accessible in menus and routes. Denying access solves this issue.

Implementing this as a custom access check is nice because it allows this logic to be re-used in many contexts, even outside of views. All you need to do is set the access check like so:

```yaml
module.route:
  path: '/path'
  #... standard routing definition
  requirements:
    _asset_collection_access: 'TRUE'
```
---

3. Provide a `farm_entity_collection` views access plugin 

This is the last step required to connect the entity collection access check to all of our default views routes. While it would be possible to manually and tediously alter each and every view route via a `RoutingSubscriber`.... I think implementing a custom views access plugin will be much easier. This also has the added benefit that this "entity collection access" logic can more easily be used by contrib views, as well as views that don't create pages (blocks, attachments, etc).

### DIFF
--- a/modules/core/entity/modules/access/src/Plugin/views/access/EntityCollection.php
+++ b/modules/core/entity/modules/access/src/Plugin/views/access/EntityCollection.php
@@ -2,7 +2,7 @@
 
 declare(strict_types=1);
 
-namespace Drupal\farm_entity_views\Plugin\views\access;
+namespace Drupal\farm_entity_access\Plugin\views\access;
 
 use Drupal\Core\Cache\Cache;
 use Drupal\Core\Cache\CacheableDependencyInterface;
@@ -10,7 +10,7 @@ use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Session\AccountInterface;
 use Drupal\Core\StringTranslation\TranslatableMarkup;
-use Drupal\farm_entity\Access\EntityCollectionAccessCheck;
+use Drupal\farm_entity_access\Access\EntityCollectionAccessCheck;
 use Drupal\views\Attribute\ViewsAccess;
 use Drupal\views\Plugin\views\access\AccessPluginBase;
 use Symfony\Component\DependencyInjection\ContainerInterface;
--- a/modules/core/entity/src/Access/EntityCollectionAccessCheck.php
+++ b/modules/core/entity/src/Access/EntityCollectionAccessCheck.php
@@ -1,0 +1,109 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Drupal\farm_entity\Access;
+
+use Drupal\Core\Access\AccessResult;
+use Drupal\Core\Entity\EntityTypeBundleInfo;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\Routing\Access\AccessInterface;
+use Drupal\Core\Routing\RouteMatchInterface;
+use Drupal\Core\Session\AccountProxyInterface;
+
+/**
+ * Checks access for the specified entity type collection.
+ */
+class EntityCollectionAccessCheck implements AccessInterface {
+
+  /**
+   * The current user.
+   *
+   * @var \Drupal\Core\Session\AccountProxyInterface
+   */
+  protected $currentUser;
+
+  /**
+   * The entity type manager.
+   *
+   * @var \Drupal\Core\Entity\EntityTypeManagerInterface
+   */
+  protected $entityTypeManager;
+
+  /**
+   * The entity type bundle info service.
+   *
+   * @var \Drupal\Core\Entity\EntityTypeBundleInfoInterface
+   */
+  protected $entityTypeBundleInfo;
+
+  /**
+   * The entity type ID.
+   *
+   * @var string
+   */
+  protected $entityTypeId;
+
+  /**
+   * FarmAssetLogViewsAccessCheck constructor.
+   *
+   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_type_manager
+   *   The entity type manager service.
+   */
+  public function __construct(AccountProxyInterface $current_user, EntityTypeManagerInterface $entity_type_manager, EntityTypeBundleInfo $entity_type_bundle_info, string $entity_type_id) {
+    $this->currentUser = $current_user;
+    $this->entityTypeManager = $entity_type_manager;
+    $this->entityTypeBundleInfo = $entity_type_bundle_info;
+    $this->entityTypeId = $entity_type_id;
+  }
+
+  /**
+   * A custom access check.
+   *
+   * @param \Drupal\Core\Routing\RouteMatchInterface $route_match
+   *   The route match.
+   */
+  public function access(RouteMatchInterface $route_match) {
+
+    // Bail if the entity type does not exist.
+    if (!$this->entityTypeManager->hasDefinition($this->entityTypeId)) {
+      // TODO: Cache tag for when entity type is later installed?
+      return AccessResult::forbidden();
+    }
+
+    // Load entity type.
+    $entity_type = $this->entityTypeManager->getDefinition($this->entityTypeId);
+
+    // Check that user has the access {entity_type} collection permission.
+    $collection_permission = AccessResult::allowedIfHasPermission($this->currentUser, $entity_type->getCollectionPermission());
+
+    // View any/own {entity_type} permissions grant access to all bundles.
+    $view_permissions = [
+      "view any $this->entityTypeId",
+      "view own $this->entityTypeId",
+    ];
+
+    // Load available bundles for the entity type.
+    $entity_bundles = $this->entityTypeBundleInfo->getBundleInfo($this->entityTypeId);
+    $all_bundles = array_keys($entity_bundles);
+
+    // Check if the route is expecting a specific bundle.
+    $bundle = $route_match->getParameter($entity_type->getBundleEntityType()) ?: $route_match->getParameter('arg_0');
+    if ($bundle && in_array($bundle, $all_bundles)) {
+      $view_permissions[] = "view any $bundle $this->entityTypeId";
+      $view_permissions[] = "view own $bundle $this->entityTypeId";
+    }
+    // Else add view permissions for all bundles.
+    else {
+      foreach ($all_bundles as $bundle) {
+        $view_permissions[] = "view any $bundle $this->entityTypeId";
+        $view_permissions[] = "view own $bundle $this->entityTypeId";
+      }
+    }
+
+    // Grant permission if the user has at least one of the view permissions.
+    $view_permission = AccessResult::allowedIfHasPermissions($this->currentUser, $view_permissions, 'OR');
+    return $collection_permission->andIf($view_permission);
+  }
+
+}
--- a/modules/core/entity/src/Access/EntityCollectionAccessCheck.php
+++ b/modules/core/entity/src/Access/EntityCollectionAccessCheck.php
@@ -10,45 +10,17 @@ use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Routing\Access\AccessInterface;
 use Drupal\Core\Routing\RouteMatchInterface;
 use Drupal\Core\Session\AccountInterface;
-use Drupal\Core\Session\AccountProxyInterface;
 
 /**
  * Checks access for the specified entity type collection.
  */
 class EntityCollectionAccessCheck implements AccessInterface {
 
-  /**
-   * The current user.
-   *
-   * @var \Drupal\Core\Session\AccountProxyInterface
-   */
-  protected $currentUser;
-
-  /**
-   * The entity type manager.
-   *
-   * @var \Drupal\Core\Entity\EntityTypeManagerInterface
-   */
-  protected $entityTypeManager;
-
-  /**
-   * The entity type ID.
-   *
-   * @var string
-   */
-  protected $entityTypeId;
-
-  /**
-   * FarmAssetLogViewsAccessCheck constructor.
-   *
-   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_type_manager
-   *   The entity type manager service.
-   */
-  public function __construct(AccountProxyInterface $current_user, EntityTypeManagerInterface $entity_type_manager, string $entity_type_id) {
-    $this->currentUser = $current_user;
-    $this->entityTypeManager = $entity_type_manager;
-    $this->entityTypeId = $entity_type_id;
-  }
+  public function __construct(
+    protected AccountInterface $currentUser,
+    protected EntityTypeManagerInterface $entityTypeManager,
+    protected string $entityTypeId,
+  ) {}
 
   /**
    * A custom access check.
--- a/modules/core/entity/src/Access/EntityCollectionAccessCheck.php
+++ b/modules/core/entity/src/Access/EntityCollectionAccessCheck.php
@@ -5,10 +5,11 @@ declare(strict_types=1);
 namespace Drupal\farm_entity\Access;
 
 use Drupal\Core\Access\AccessResult;
-use Drupal\Core\Entity\EntityTypeBundleInfo;
+use Drupal\Core\Access\AccessResultInterface;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Routing\Access\AccessInterface;
 use Drupal\Core\Routing\RouteMatchInterface;
+use Drupal\Core\Session\AccountInterface;
 use Drupal\Core\Session\AccountProxyInterface;
 
 /**
@@ -31,13 +32,6 @@ class EntityCollectionAccessCheck implements AccessInterface {
   protected $entityTypeManager;
 
   /**
-   * The entity type bundle info service.
-   *
-   * @var \Drupal\Core\Entity\EntityTypeBundleInfoInterface
-   */
-  protected $entityTypeBundleInfo;
-
-  /**
    * The entity type ID.
    *
    * @var string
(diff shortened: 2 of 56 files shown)


modules/core/role/modules/roles/config/install/user.role.farm_tester.yml
mstenta, line 1: Seems like this could be added to a test module.

modules/core/ui/location/farm_ui_location.routing.yml
```diff
@@ -5,6 +5,7 @@ farm.locations:
     _form: '\Drupal\farm_ui_location\Form\LocationHierarchyForm'
   requirements:
     _permission: 'access locations overview'
+    _asset_collection_access: 'TRUE'
```
mstenta, line 8: Should this also be added to the `farm.asset.locations` route below this one?

(Also makes me wonder if there are other places we should use this? :thinking:)

modules/core/ui/views/config/install/views.view.farm_asset.yml
```diff
@@ -7,12 +7,12 @@ dependencies:
   module:
     - asset
     - entity_browser
+    - farm_entity_views
```
mstenta: I think this means we need to add an explicit dependency on `farm_entity_views` to the `farm_ui_views` module. It already depends on `farm_entity`, but I like to add explicit dependencies whenever we add logic that directly depends on a sub-module. That way if we ever refactor in the future (eg: remove `farm_entity_views` from `farm_entity.info.yml`), we don't lose that by accident.

modules/core/ui/views/config/install/views.view.farm_log_quantity.yml
```diff
@@ -1592,13 +1593,12 @@ display:
           content: 'Displaying @start - @end of @total'
       display_extenders: {  }
     cache_metadata:
-      max-age: -1
+      max-age: 0
```
mstenta, line 1591: What is this change for? There are two more similar changes further down in this file. Is it intentional, or just a result of re-exporting the View?

modules/core/entity/src/Access/EntityCollectionAccessCheck.php
```diff
@@ -0,0 +1,133 @@
...
+   * A custom access check.
+   *
+   * @param \Drupal\Core\Routing\RouteMatchInterface $route_match
+   *   The route match.
+   */
+  public function access(RouteMatchInterface $route_match) {
+
+    // Bail if the entity type does not exist.
+    if (!$this->entityTypeManager->hasDefinition($this->entityTypeId)) {
+      // TODO: Cache tag for when entity type is later installed?
```
mstenta: Are these `TODO`s going to be completed in this PR? (There is a second one in `checkEntityCollectionAccess()` below as well).

Nit: `TODO` should be `@todo`.

modules/core/entity/farm_entity.module
mstenta, line 1: FYI: I will release a new 3.0.0 version of `drupal/log` soon so we can remove this. I'll push a commit to this PR that updates `composer.json` and reverts the changes in this file when I do.

modules/core/entity/modules/access/src/Plugin/views/access/EntityCollection.php
```diff
@@ -0,0 +1,157 @@
...
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Session\AccountInterface;
+use Drupal\Core\StringTranslation\TranslatableMarkup;
+use Drupal\farm_entity\Access\EntityCollectionAccessCheck;
+use Drupal\views\Attribute\ViewsAccess;
+use Drupal\views\Plugin\views\access\AccessPluginBase;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\Routing\Route;
+
+#[ViewsAccess(
```
mstenta, line 22: We should add an `AnnotationToAttributeConfiguration` configuration to `rector.php` for this type of attribute.

modules/core/entity/src/Access/EntityCollectionAccessCheck.php
```diff
@@ -0,0 +1,133 @@
...
+      return AccessResult::forbidden();
+    }
+
+    // Check if the route is expecting a specific bundle.
+    $entity_type = $this->entityTypeManager->getDefinition($this->entityTypeId);
+    $bundle = $route_match->getParameter($entity_type->getBundleEntityType()) ?: $route_match->getParameter('arg_0');
```
mstenta: Is `arg_0` always going to be the parameter name, or is that specific to Views?

mstenta (changes requested):
All makes sense, thanks for the thorough explanations @paul121!

I think this approach make sense. I left some comments on little things here and there.

It would be worth adding some automated tests for this, since it's pretty mission-critical. Might be able to get away with just adding some tests to `FarmUiViewsTest.php` that check the Views output. That would cover the low level access class logic too, but maybe it makes sense to have some dedicated tests just for that as well.

mstenta:
I rebased this onto the latest `4.x` commit and added a commit that updates `drupal/log` to `3.0.0`, which provides the `query_access` handler the `access log collection` permission.

modules/core/ui/location/farm_ui_location.routing.yml
paul121: I don't think it is needed below since the access check also checks that asset is a location. This will also change a little with #849 and refactoring of location hierarchy form

modules/core/entity/src/Access/EntityCollectionAccessCheck.php
paul121: Yes they should be completed in this PR. I didn't remember the syntax at the time but later I did see somewhere we do use the entity type list cache tag for this use-case... need to find that again.

modules/core/entity/src/Access/EntityCollectionAccessCheck.php
paul121: That is specific to views. And it's not consistent in our views IIRC. Sometimes views do have the parameter named as `asset_type`. I think there is a Drupal core issue to try and standardize this.

The more robust solution for this would be to implement configuration in the views access check plugin so that the argument name can be configured when using the access plugin on the view. That would support use cases when the asset type might not be the first parameter `arg_0` or not be named `asset_type`. At least I think something like this would be possible.

modules/core/ui/views/config/install/views.view.farm_log_quantity.yml
paul121: Unsure, partially why this commit is WIP. I think this means that the previous access check using the user permission allowed these views to be cached. With this proposed views entity collection access check it currently is not cacheable. So I'm not sure... do we just need to add the right cache context? The user's permissions?

Also I think there are additional views we should add this to, like the view that provides list of assets for referencing in entity reference widgets?

mstenta:
Rebased onto `4.x` again, and fixed merge conflicts now that #967 is in. This was mainly adapting this PR's changes to the new `ContentEntityType` attributes (which were annotations previously).

modules/core/ui/views/config/install/views.view.farm_log_quantity.yml
mstenta: Interesting - yea it would be worth understanding the caching considerations, and exploring whether or not we can still support caching of these Views.

mstenta: And yes - a thorough review of all the Views we provide is needed! Happy to help with that. Maybe on today's dev call?

mstenta:
We will also need to make the relevant changes to #849...

paul121:
The test failures are due to a 500 error when getting an OAuth access token. I believe it is related to what PHP Stan caught:

> Access to an undefined property                                                    
         Drupal\Tests\farm_api_oauth\Kernel\FarmApiOauthTestBase::$scope.  
         
Will see if adding the property fixes this.

mstenta:
Rebased onto latest `4.x` commit and fixed merge conflicts.

mstenta:
I started working on the remaining bits for this.

I took the current branch and rebased it onto #1007, and pushed it to my fork, with the intention of force-pushing it to this PR after #1007 is merged.

You can see the commits in their current state here: https://github.com/mstenta/farmOS/compare/4.x-hooks...4.x-views-perms

Here is a quick overview of what I've done:

- Moved the "Update drupal/log to ^3.0." commit to the front, and dropped the temporary "Add query_access handler and log collection permission" commit.
- (Left all of @paul121's commits untouched.)
- Added support for Organization entity collection permission checking (now that #849 is merged).
- Add `ViewsAccess` `AnnotationToAttributeConfiguration` check to Rector config, since we are adding our first `ViewsAccess` plugin.
- Added a dependency on `farm_entity_views` to the `farm_ui_views` module (for the `farm_entity_collection` access plugin).
- Updated `EntityCollection` and `EntityCollectionAccessCheck` to use constructor property promotion (#1005) and autowiring (#1006).

Remaining todos:

- Make the `arg_0` parameter name configurable per View in `EntityCollectionAccessCheck`.
- Finish @paul121's "WIP update core views to use farm_entity_collection views access" commit.
- Add cache tags in `EntityCollectionAccessCheck` so that the cache is invalidated when new entity types are installed (see `TODO`s in code.

>>
>> Should this also be added to the farm.asset.locations route below this one?
>
> I don't think it is needed below since the access check also checks that asset is a location. This will also change a little with https://github.com/farmOS/farmOS/pull/849 and refactoring of location hierarchy form

@paul121 I wanted to double check on this one last time, as well. There are now three routes in `farm_ui_location.routing.yml`. Do we need to add `_asset_collection_access: 'TRUE'` to either of the other two? If not, let's briefly document why in these comments for future reference.

mstenta:
Tests are also failing.

I investigated the first one (`Drupal\Tests\farm_quick\Kernel\QuickFormTest::testQuickFormSubmission()`), which tests a basic quick form submission that creates a log with a quantity attached. It fails because the log entity does not validate, because `ValidReferenceConstraintValidator` checks to see if the `quantity` entity is allowed to be referenced, but because the test is run as user 0, the `QueryAccessHandlerBase::buildEntityOwnerConditions()` returns `NULL`, which causes `QueryAccessHandlerBase::buildConditions()` to add `$conditions->alwaysFalse();`. Kernel tests always run as the anonymous user, and notably this is only an issue in this one because it installs `farm_entity`.

mstenta:
Update: I pushed a commit to my branch that fixes failing kernel tests by adding `createUser()` calls with the necessary permissions.

There are still some functional tests that are failing, but some/all of those are probably related to Views configuration, so let's focus on finishing those updates and see what's left after that.

mstenta:
For the sake of testing, here is the config YML for the `farm_tester` role that @paul121 originally included in a temporary commit in this PR:

```yml
langcode: en
status: true
dependencies:
  enforced:
    module:
      - farm_role
id: farm_tester
label: 'Tester'
weight: 1
is_admin: false
permissions:
  - access asset collection
  - access log collection
  - access quantity collection
third_party_settings:
  farm_role:
    access:
      config: false
      entity:
        view all: false
        create all: false
        update all: false
        delete all: false
        type:
          quantity:
            view own:
              - standard
          asset:
            create:
              - structure
              - equipment
            view any:
              - equipment
              - land
            view own:
              - structure
            update any:
              - equipment
            update own:
              - structure
            delete own:
              - structure
          log:
            create:
              - activity
              - observation
            view any:
              - activity
            view own:
              - observation
            update any:
              - activity
            update own:
              - observation
            delete own:
              - observation
```

This was originally added to the `farm_role` module, which has since been removed (#982). I have removed this commit from my branch (not yet force pushed to this PR), so I want to preserve it here for ongoing testing purposes. This can be copied and pasted into the config import form: /admin/config/development/configuration/single/import

mstenta:
Update: I have updated all core Views to use the new entity collection Views access plugin...

> There are still some functional tests that are failing, but some/all of those are probably related to Views configuration, so let's focus on finishing those updates and see what's left after that.

... and functional tests are passing! :tada:

While working on this, I realized that there was an implicit circular dependency between `farm_entity` and `farm_entity_views`. I fixed this by moving both the `EntityCollectionAccessCheck` and `EntityCollection` Views access plugin class to a new `farm_entity_access` module.

Previously, `EntityCollectionAccessCheck` was in `farm_entity`, which depends on `farm_entity_views`, and `EntityCollection` was in `farm_entity_views`, which uses  `EntityCollectionAccessCheck`, thus creating an implicit circular dependency.

This is all still in my `4.x-views-perms` fork... not yet pushed to this PR.

https://github.com/farmOS/farmOS/compare/4.x...mstenta:farmOS:4.x-views-perms

mstenta:
I pushed a commit that removes the two `TODO`s that were left in the `EntityCollectionAccessCheck` class: https://github.com/farmOS/farmOS/commit/234e7d854a8525956d9031ce0546433eea9f9b9d

My thought is: this access check only runs on paths that we add it to, and will therefore only be cached on those paths. So presumably, if a new entity type is installed, there will be new paths with their own access checks added for that entity type, and the access results of those paths won't be cached until they are accessed. So I don't *think* we need to worry about these `TODO`s...

@paul121 Do you think that makes sense? Maybe I'm overlooking a scenario that you thought about earlier...

mstenta:
> Make the `arg_0` parameter name configurable per View in `EntityCollectionAccessCheck`.

> That is specific to views. And it's not consistent in our views IIRC. Sometimes views do have the parameter named as `asset_type`. I think there is a Drupal core issue to try and standardize this.
> 
> The more robust solution for this would be to implement configuration in the views access check plugin so that the argument name can be configured when using the access plugin on the view. That would support use cases when the asset type might not be the first parameter `arg_0` or not be named `asset_type`. At least I think something like this would be possible.

I'm digging into this to understand it better...

These are all of the core Views and displays that use bundle arguments (the bundle is the first argument unless noted below):

- `farm_asset_geojson` (both displays: `full` and `centroid`)
- `farm_asset`
  - `page_type`
  - `page_term` (asset bundle is second argument)
- `farm_log`
  - `page_type`
  - `page_term` (log bundle is second argument)
  - `page_asset` (asset bundle is second argument)
- `farm_log_quantity.page_type`
- `farm_organization.page_type`
- `farm_organization_asset.page_type` (asset bundle is second argument)
- `farm_organization_log.page_type` (log bundle is second argument)

I put a breakpoint in `EntityCollectionAccessCheck` and it seems that the `$bundle` is correctly determined in all of these *except* the `page_term` displays of the `farm_asset` and `farm_log` Views.

mstenta:
> I put a breakpoint in `EntityCollectionAccessCheck` and it seems that the `$bundle` is correctly determined in all of these *except* the `page_term` displays of the `farm_asset` and `farm_log` Views.

In these cases the argument is called `entity_bundle` (not `asset_type`, `log_type`, or `arg_0` as it is in other contexts).

mstenta:
> > I put a breakpoint in `EntityCollectionAccessCheck` and it seems that the `$bundle` is correctly determined in all of these _except_ the `page_term` displays of the `farm_asset` and `farm_log` Views.
> 
> In these cases the argument is called `entity_bundle` (not `asset_type`, `log_type`, or `arg_0` as it is in other contexts).

Ah... this "`entity_bundle`" argument name is defined in the `farm_asset` and `farm_log` Views config:

https://github.com/farmOS/farmOS/blob/de35e390b76fa5da79f77e3489c26aac64eb37ab/modules/core/ui/views/config/install/views.view.farm_asset.yml#L1906

https://github.com/farmOS/farmOS/blob/de35e390b76fa5da79f77e3489c26aac64eb37ab/modules/core/ui/views/config/install/views.view.farm_log.yml#L2608

Changing that to `%asset_type` and `%log_type`, respectively, sets `$bundle` to the correct value. I will push a commit that makes this change.

mstenta:
> Changing that to `%asset_type` and `%log_type`, respectively, sets `$bundle` to the correct value. I will push a commit that makes this change.

This still feels brittle to me, but maybe it's enough to merge this PR and figure out how to improve it as a next step? :thinking:

mstenta:
> Changing that to `%asset_type` and `%log_type`, respectively, sets $bundle to the correct value. I will push a commit that makes this change.

Oops this causes tests to fail. I will need to investigate that later...

mstenta:
> Oops this causes tests to fail. I will need to investigate that later...

I fixed the failing tests by taking a different approach (suggested by @paul121): rather than trying to change the name of the `entity_bundle` parameter (which is referenced in other places in logic), I reworked the conditional `$bundle` assignment logic in `EntityCollectionAccessCheck` so that it look for an `entity_bundle` parameter (in addition to looking for `arg_0`, `asset_type`, `log_type`, etc).

Tests are all green now. I think this is ready to merge! I want to get a final review of my branch from @paul121 before I force-push it to this PR and then merge into `4.x`.

paul121:
Just reviewed this and LGTM! https://github.com/farmOS/farmOS/compare/4.x...mstenta:farmOS:4.x-views-perms

mstenta:
Great - thanks again for reviewing @paul121!

I'll go ahead and rebase my branch, force-push, and merge into 4.x...

mstenta (approved):

PR #966, opened 2025-05-10
Add map layer for "Other Location" assets

paul121:
Basically, any assets that are not an asset type that is a location by default (land, structure vs equipment, plant) will not be shown on the dashboard nor locations page. We can quite easily display these additional location assets by including them in a new "Other locations" layer on the map.

I think this could be especially useful for the map on location hierarchy form for organizations https://github.com/farmOS/farmOS/pull/849#issuecomment-2868168802


mstenta:
Makes sense to me! Good idea @paul121. I tested this and it works as expected. Rebased, added `CHANGELOG.md` line, and running tests/linting before merging...

symbioquine (approved):
LGTM

PR #967, opened 2025-05-20
Update Drupal core to 11.x

mstenta:
See [Issue #3488916: Update Drupal core to 11.x in farmOS](https://www.drupal.org/project/farm/issues/3488916)

This PR updates Drupal from 10 to 11.

I am starting it in "draft" status because we are waiting for one more contrib dependency (`jsonapi_schema`), but it is ready to review otherwise. This PR temporarily removes `jsonapi_schema` so that tests can run without it. I expect a D11 compatible release of `jsonapi_schema` very soon - so we can either tack that onto the end of this, or restore it in a followup PR, depending on when it's available.

We will also want to update to the latest 11.x version right before merging this. Currently it's pinning 11.1.5, but the latest D11 release is 11.1.7. I'll wait until this is ready to merge before updating it again.


mstenta:
This still needs `CHANGELOG.md` updates... I'll push a commit for that soon...

mstenta:
> I expect a D11 compatible release of `jsonapi_schema` very soon - so we can either tack that onto the end of this, or restore it in a followup PR, depending on when it's available.


JSON:API Schema 8.x-1.0 is now available with Drupal 11 support! :tada: https://www.drupal.org/project/jsonapi_schema/releases/8.x-1.0

AND all the patches we were maintaining for it have been merged! :tada: :tada:

I have removed the commit that removed it, and replaced it with a commit that updates to the new version and removes the patches from our `composer.json`.

> We will also want to update to the latest 11.x version right before merging this. Currently it's pinning 11.1.5, but the latest D11 release is 11.1.7. I'll wait until this is ready to merge before updating it again.

I updated to the latest 11.1.7, so we are fully ready to review! Removing this from draft status...

pcambra:
@mstenta wondering if we could include this core patch for postgresql https://www.drupal.org/project/drupal/issues/2920527 that is kind of required for the multienant use case, as it is a core patch, there's no way for contribs to alter the project composer.json

mstenta:
> @mstenta wondering if we could include this core patch for postgresql https://www.drupal.org/project/drupal/issues/2920527

@pcambra I saw that and haven't had a chance to look closer yet. I would prefer to not block this PR on that, personally. Especially since it isn't an issue for farmOS core itself (yet). We can always add it later as a separate PR if there's enough need/interest in doing so. For now I think it's fine to require anyone experimenting with the multitenant module to apply the patch themselves in their own project's `composer.json`.

modules/asset/land/src/Entity/FarmLandType.php
```diff
@@ -5,30 +5,31 @@
...
+  label: new TranslatableMarkup('Land type'),
+  label_collection: new TranslatableMarkup('Land types'),
+  entity_keys: [
+    'id' => 'id',
+    'label' => 'label',
+  ],
+  handlers: [
+    'access' => '\Drupal\entity\EntityAccessControlHandler',
+    'permission_provider' => '\Drupal\entity\EntityPermissionProvider',
+  ],
```
paul121: Whenever we reference classes within properties of new plugin class attributes I think we should switch to using the `ClassName::class` convention (which requires a `use Drupal\module\ClassName;` statement) rather than typing the full class string. There is better support in editors for this type of class reference and would be useful whenever searching for usages or refactoring a given class.

Took a quick look for `ConfigEntityType` plugins and looks like Drupal core has switched to doing this for most core config entity type plugins. I assume this will work for all plugins that use attributes.

This change could also be in a follow up PR, happy to work on that. Also curious if there is a PHPStan rule for this.

```suggestion
  handlers: [
    'access' => EntityAccessControlHandler::class,
    'permission_provider' => EntityPermissionProvider::class,
  ],
```

modules/core/entity/farm_entity.module
```diff
@@ -68,7 +68,7 @@ function farm_entity_entity_type_build(array &$entity_types) {
   }
 
   // Enable the use of bundle plugins on specific entity types.
-  foreach (['asset', 'log', 'plan', 'plan_record', 'quantity'] as $entity_type) {
+  foreach (['asset', 'data_stream', 'log', 'plan', 'plan_record', 'quantity'] as $entity_type) {
```
paul121: If we make this change I think we also need to include `data_stream` in the logic of `farm_entity_entity_field_storage_info_alter()`. Otherwise contributed bundle fields will not have the correct module provider and can cause issues when installing/uninstalling the modules that provide these bundle fields. It seems that `plan_record` might already be missing from this logic.

I think we could also remove the `data_stream` module dependency on `entity`, and add a dependency on `farm_entity` instead UNLESS we move the `DataStreamTypeManager` and etc `DataStreamType` plugin scaffolding to the `farm_entity` module like we do for our other entity types.

paul121 (changes requested):

modules/asset/land/src/Entity/FarmLandType.php
mstenta: Good idea. I added a commit that changes all of these in the `handlers: [...]` sections of the attributes. There may be other places in code we could change too, but this is most relevant to this PR IMO. It would be great if there was a CodeSniffer rule we could use for this moving forward, but at least having them in our core entity types will be a great start, because that's often where I copy and paste from. :smile:

This was a bit of a tedious process, and while PHPStorm makes it easy to "import" the class, in many cases we are using the `entity` module's version, which extends a core class of the same name. I was careful to double check each one to be sure I used the right class, but a second set of eyes wouldn't hurt.

mstenta: (Force-pushed that commit to fix CodeSniffer's "Use statements should be sorted alphabetically" complaints.

paul121: awesome! I'll review

paul121: LGTM!

modules/core/entity/farm_entity.module
mstenta: Spoke with @paul121 about this and we decided to take a different/simpler approach. Instead of setting `bundle_plugin_type` in `farm_entity`, we set it in `data_stream`, using the same `hook_entity_type_build()`.

It's more complicated to allow data stream bundle fields to be added via `hook_farm_entity_bundle_field_info()`, but we might pursue that in a followup.

Here is the new commit: https://github.com/farmOS/farmOS/pull/967/commits/ed7bd01b5e36e7150de39669e083e0972ab15503

paul121 (approved):

mstenta:
Merged!! :tada:

PR #970, opened 2025-05-29
Remove .tugboat/config.yml

mstenta:
On the dev call today @paul121 said he didn't think this was being used, and that we could remove it.


paul121 (approved):
It seems like Drupal.org is still using Tugboat. We just aren't using MRs often enough to test. Ah well. Maybe one day the dream of deploy previews for every change can become a reality...

PR #971, opened 2025-06-02
Update hosting documentation for 4.x

mstenta:
This makes a few key changes to farmOS hosting documentation, per recent conversations among the maintainers and community.

In summary, this PR:

- Adds a PostgreSQL container to the `docker-compose.production.yml` example configuration, along with corresponding updates to `docs/hosting/install.md` to describe it.
- Splits the "Server requirements" documentation out of `docs/hosting/install.md` into its own `docs/hosting/requirement.md`. This helps to slim down `docs/hosting/install.md` and make it less overwhelming.
- Splits the "packaged release" install and update docs out to a separate `docs/hosting/prebuilt.md`, with emphasis that packaged release deployments are no longer recommended. Also adds a suggestion for how to build your own packaged releases with Composer if your hosting environment doesn't have Docker/Composer.
- Splits the Docker install and update docs out to their own `docs/hosting/doocker.md` page, which further simplifies/generalizes `docs/hosting/install.md` and `docs/hosting/update.md`.


botlfarm (commented):
These look good to me as a non-technical review. I think it give a much clearer starting point for someone new to this type of stuff.

symbioquine (commented):
Just skimmed this for now - happy to take a more detailed proof-reading pass later (maybe tomorrow morning).

Overall it looks good.

I wonder if it might make sense to also point folks at more generic Drupal hosting resources for some of the general stuff - troubleshooting, logging, file upload sizes/locations, backups, etc? It feels like we're often alluding to that as a resource for folks to unblock themselves in the forums/issues, but maybe we should highlight it more strongly in these docs?

docs/hosting/docker.md
```diff
@@ -0,0 +1,119 @@
...
+This page covers how to run farmOS in Docker. Once the containers are running,
+continue with the [farmOS installation steps](/hosting/install).
+
+## Docker Run
+
+The following command will run farmOS in a Docker container (replace `x.y.z`
+with the latest stable release version from the
+[GitHub releases page](https://github.com/farmOS/farmOS/releases)):
+
+    docker run --rm -p 80:80 -v "${PWD}/sites:/opt/drupal/web/sites" farmos/farmos:4.x.y
```
paul121: `x.y.z` or `4.x.y` ?

paul121 (commented):
Seems pretty good to me. Just one little inconsistency here

docs/hosting/docker.md
mstenta: Good eye @paul121! I started to change that but didn't finish.

My thought was to replace `4.x.y` with `x.y.z` everywhere so that it's one less `4.*` that needs to be replaced when the major version increases to `5` later. :smile:

I've rebased and amended the docker commit to fix this (and also change it in the `docker-compose.production.yml` to match).

mstenta:
> I wonder if it might make sense to also point folks at more generic Drupal hosting resources for some of the general stuff - troubleshooting, logging, file upload sizes/locations, backups, etc? It feels like we're often alluding to that as a resource for folks to unblock themselves in the forums/issues, but maybe we should highlight it more strongly in these docs?

Yea great idea @symbioquine. Can we merge this and use one of the next calls to brainstorm which sections would benefit from more links? Easier to look at that together I think, and it probably doesn't need to block this. Happy to draft a follow-up PR during the dev call!

docs/hosting/requirements.md
```diff
@@ -0,0 +1,81 @@
...
+a PostgreSQL database Docker container alongside a farmOS container.
+
+## SSL
+
+Although not strictly a requirement, some features (like the "Geolocate" button
+on maps) will only work over a secure connection. SSL is also recommended if you
+are streaming sensor data into farmOS, to keep your sensor's private key a
+secret.
+
+A common strategy is to use [Nginx](https://nginx.org) as a reverse proxy with
```
symbioquine, line 59:
```suggestion
A common strategy is to use [NGINX](https://nginx.org) as a reverse proxy with
```

I didn't find an NGINX trademark page, but it appears that in places where "NGINX" is capitalized on their site, they capitalize the whole term. e.g. see https://nginx.org/en/community.html

docs/hosting/requirements.md
```diff
@@ -0,0 +1,81 @@
...
+certificate issuance, and renewal can be automated via cron.
+
+These resources may be helpful:
+
+- [Drupal HTTPS Information](https://www.drupal.org/https-information)
+- [Reverse Proxy Forum Post](https://farmos.discourse.group/t/running-behind-reverse-proxy/108) -
+  Includes links to related GitHub issues and examples of how others have
+  configured reverse proxies serving HTTPS.
+- [Local HTTPS](/development/environment/https) - Documentation for running an
+  Nginx reverse proxy with self-signed certificates for local farmOS
```
symbioquine, line 71:
```suggestion
  NGINX reverse proxy with self-signed certificates for local farmOS
```

docs/hosting/requirements.md
```diff
@@ -0,0 +1,81 @@
...
+
+- PostgreSQL 16+
+- MariaDB 10.6+
+- MySQL/Percona 8.0+
+- SQLite 3.45+
+
+See [farmOS in Docker][/hosting/docker/] for an example configuration that runs
+a PostgreSQL database Docker container alongside a farmOS container.
+
+## SSL
```
symbioquine, line 52:
```suggestion
## HTTPS
```

docs/hosting/requirements.md
```diff
@@ -0,0 +1,81 @@
...
+
+Although not strictly a requirement, some features (like the "Geolocate" button
+on maps) will only work over a secure connection. SSL is also recommended if you
```
symbioquine, line 55:
```suggestion
on maps) will only work over a secure connection. HTTPS is also recommended if you
```

docs/hosting/requirements.md
```diff
@@ -0,0 +1,81 @@
...
+are streaming sensor data into farmOS, to keep your sensor's private key a
+secret.
+
+A common strategy is to use [Nginx](https://nginx.org) as a reverse proxy with
+SSL termination, which listens on port 443 and forwards to farmOS on port 80.
+[Let's Encrypt](https://letsencrypt.org) is a good option for free SSL
```
symbioquine, line 61:
```suggestion
HTTPS termination, which listens on port 443 and forwards to farmOS on port 80.
[Let's Encrypt](https://letsencrypt.org) is a good option for free HTTPS
```

docs/hosting/docker.md
```diff
@@ -0,0 +1,119 @@
...
+
+**Important**: Using the `latest` Docker tag is not recommended, because updates
+require manual steps. See [Updating](/hosting/docker/#updating) below.
+
+This will pull the farmOS Docker image, provision a farmOS web server container
+listening on port 80, and bind-mount a `sites` directory into the container for
+persistence of settings and uploaded files (see [Persistence](#persistence)
+below).
+
+This does not include a database or reverse proxy for SSL termination.
```
symbioquine, line 34:
```suggestion
This does not include a database or reverse proxy for HTTPS termination.
```

symbioquine (approved):
LGTM w/ minor nitpicks that could probably be left for later since this is mostly moving the affected text around.

docs/hosting/requirements.md
symbioquine: In case it's not obvious, I'm proposing changing all our references from "SSL" to "HTTPS" - to avoid the whole "SSL vs TLS (SSL is deprecated)" issue.

mstenta:
Thanks @symbioquine! Yea let's make those changes in a separate PR. In my mind this one is focused on shifting things around and demoting the "packaged" release deployment method.

docs/hosting/requirements.md
mstenta: Ah interesting. I suppose we could update this to match what the community calls it. I checked Wikipedia and it's referred to as "Nginx" there: https://en.wikipedia.org/wiki/Nginx

docs/hosting/requirements.md
mstenta: Cool! Yea I'm OK with doing this. We've been using the "colloquial" terminology (referring to everything as "SSL") for a while. :sweat_smile: 

We should make sure we're consistent everywhere. For example, I see that we also recommend making a directory called `ssl` to store the certificates: https://github.com/farmOS/farmOS/blob/53c1d579d7900226925577887f3c149da4f73885/docs/development/environment/https.md?plain=1#L8

Let's do this in a separate PR.

PR #975, opened 2025-06-03
Generate OAuth2 scopes for each managed role

mstenta:
Currently we provide static OAuth2 scopes for the `farm_manager`, `farm_viewer`, and `farm_worker` roles in the module that provides them (`farm_role_roles`):

https://github.com/farmOS/farmOS/blob/53c1d579d7900226925577887f3c149da4f73885/modules/core/role/modules/roles/farm_role_roles.oauth2_scopes.yml

And we detect whether or not the `simple_oauth_password_grant` module is installed and add the `password` grant to them automatically:

https://github.com/farmOS/farmOS/blob/53c1d579d7900226925577887f3c149da4f73885/modules/core/role/modules/roles/farm_role_roles.module

This means that any module providing a managed role must do the same in order to make it available as a scope. That's a lot of boilerplate that we could provide on behalf of managed roles automatically.

This PR moves the logic to the new `farm_api_oaut` module (introduced in #973) and automatically generates scopes for all managed roles.

This is a follow-up to #973, and includes its commit.


paul121 (commented):
We talked about this on a dev call some.. and after a little more investigation I don't think we should default to making _all_ managed roles available as oauth scopes. The issue is that there is no way to limit which scopes are available to different consumers/oauth clients. Once a scope exists, any user with sufficient access to that scope (permission or role) can authorize 1st or 3rd party clients access with that scope.

Normally this is the exact feature we are seeking :smile:  But I also think this is problematic because there are scenarios when users might have additional roles/permissions to access or modify particularly sensitive data (maybe personally identifiable information, organizational data, financial data) that simply never needs to be interacted with via API. If it's not necessary to make these data available via API, doing so only increases the potential risk of leaking sensitive data.

That said, I don't think this scenario is something we can protect against with the default viewer, worker and manager roles and scopes. When a site uses these roles they are assuming pretty standard & generic all-or-nothing access logic.

For example, if only `animal` assets or `purchase` and `sale` logs are considered to be particularly sensitive and should have access limited to only certain users, the default roles should not be used. Instead, new (managed) roles should be created that provide only the access each logical group of users might need. Static scopes can then be created for only the roles that need access via API. This would support a scenario where an admin user who can access sensitive data might want to use an external integration to manage more general `land` assets and `observation` logs, but at the same time never allow access to sensitive data via API.

But now thinking about this further... I wonder if FarmOS core should really just be recommending that `dynamic` scopes be configured on each site, instead of relying on `static` scopes that are either always on, or always off. This additional configuration would make using simple_oauth for custom API integrations a bit "more complex", but I think there is an argument that this is ultimately the most secure, requiring users themselves to enable each scope (role/permission) that they want to be available via API. With dynamic scopes users can still create  `managed_role` scopes for "managed roles", but importantly, they would choose which scopes to create.

For API integrations that do expect/require specific scopes, or for use-cases that need to streamline an oauth configuration across multiple sites, they can provide such static scopes within a custom module. At the end of the day, however, users can still choose to throw out static scopes and use dynamic scopes instead. In which case the users would need to replicate any static scopes in their dynamic scope configuration for these integrations to work. But even having this capability is important in the context of "owning your own data".

paul121:
> I wonder if FarmOS core should really just be recommending that dynamic scopes be configured on each site, instead of relying on static scopes that are either always on, or always off.

I know we want FarmOS to have some kind of an "out of the box" API experience. But currently this only works with the `farm_api_default_consumer` module to create a default `farm` client and enable the `password` grant. Perhaps this module should also be the one to change the simple_oauth config to use `static` scopes *and* provide static scopes for all managed roles. This would be a convenience feature for easy setup, but not recommended for more custom and advanced API integrations.

mstenta:
We decided on the monthly call today to scrap this PR for now and just keep things how they are for the time being.

Changing from `static` to `dynamic` is a much bigger conversation (and presumably a much larger breaking change). This probably warrants a forum topic to discuss all the considerations. I doubt we'll be able to do that in 4.x.

PR #979, opened 2025-07-02
Require PHP 8.4

mstenta:
This updates farmOS to run on PHP 8.4 in the Docker image, and to require it in `composer.json`.

PR #980, opened 2025-07-02
Update Drupal to v11.2, PHPUnit to v11, PHPStan to v2

mstenta:
This PR updates Drupal core to 11.2.

It also updates PHPUnit to v11 (which requires #979) and PHPStan v2.

There are still some PHPStan failures and tests are not passing, so this needs some work, but I wanted to push what I have so far up and open a draft PR to get things rolling.


mstenta:
I rebased this onto latest #979 changes, which should eliminate some (all?) of the failing tests. Let's see what happens...

mstenta:
Hmm there are 4 test failures like this related to the selenium version... :thinking: 

```
1) Drupal\Tests\farm_location\Functional\LocationTest::testComputedAssetLocation
Behat\Mink\Exception\DriverException: This driver requires Selenium version 3 or lower, but version 4.1.2 (revision 9a5a329c5a) was found.

Please use the "mink/webdriver-classic-driver" Mink driver or switch to Selenium Server 2.x/3.x.
```

mstenta:
It appears that Gin has released a new 5.0.0 that only works with Drupal ^11.2, and the 4.1.0 only works with <11.2, so we will need to update to Gin 5.x in order to update to Drupal 11.2.

It seems that this is related to Gin being absorbed into Drupal core in 11.3 (see https://www.drupal.org/about/core/blog/drupal-core-will-adopt-gin-admin-theme-to-replace-claro).

I'm skimming through some of the release notes (currently the latest release is 5.0.3), and it sounds like we may need to assess how this affects our dependency on the `toolbar` / `gin_toolbar` modules...

- https://www.drupal.org/project/gin/releases/5.0.0
- https://www.drupal.org/project/gin/releases/5.0.1
- https://www.drupal.org/project/gin/releases/5.0.2
- https://www.drupal.org/project/gin/releases/5.0.3

Specifically these issues:

- https://www.drupal.org/project/gin/issues/3503575
- https://www.drupal.org/project/gin/issues/3531241
- https://www.drupal.org/project/gin_toolbar/issues/3531236

Which all seem to stem from this: https://www.drupal.org/project/drupal_cms/issues/3530945

mstenta:
Oh excellent! I figured out the issue that was causing tests to fail. I hadn't noticed earlier but the `drupal/core-dev` `composer.json` was changed in Drupal 10.3 to require `lullabot/mink-selenium2-driver` instead of `behat/mink-selenium2-driver`. I didn't dig in deep enough to understand exactly why this broke our tests, but changing it in our `composer-project.json` and adding the new `goog:` prefix to `chromeOptions` (as recommended here: https://www.drupal.org/node/3240792) did the trick!

So now we just need to fix the PHPStan failures and figure out if any other changes are necessary for Gin 5.x...

mstenta:
I did some manual testing of Gin 5.x locally and everything seems to work as expected.

> I'm skimming through some of the release notes (currently the latest release is 5.0.3), and it sounds like we may need to assess how this affects our dependency on the `toolbar` / `gin_toolbar` modules...

I don't think we need to make any changes right now. Gin just made `toolbar` optional, but it still works with it enabled. We currently depend on `admin_toolbar`, which depends on `toolbar`, so teasing that apart would be a bigger task, and doesn't need to block this PR.

Notably, I noticed that we are NOT actually depending on `gin_toolbar`. Perhaps this is also worth a separate issue... :thinking:

mstenta:
So I *think* the only remaining work here is the PHPStan test failures...

mstenta:
Fixed the PHPStan issues.

Added a `CHANGELOG.md` line that mentions Drupal minor version bump and PHPUnit and PHPStan major version bumps.

Also added a commit that changes our use of `$entity->original` to `$entity->getOriginal()` (per https://www.drupal.org/node/3295826) and manually tested them to make sure they worked the same.

Tests should all be green now! This is now ready for review.

modules/core/field/src/FarmFieldFactory.php
```diff
@@ -48,12 +48,12 @@ public function bundleFieldDefinition(array $options = []): BundleFieldDefinitio
...
    *
-   * @param \Drupal\Core\Field\BaseFieldDefinition &$field
-   *   A base field definition object.
+   * @param \Drupal\Core\Field\BaseFieldDefinition|\Drupal\entity\BundleFieldDefinition $field
+   *   A base or bundled field definition object.
    * @param array $options
    *   An array of options.
    */
-  protected function buildFieldDefinition(BaseFieldDefinition &$field, array $options = []) {
+  protected function buildFieldDefinition(BaseFieldDefinition|BundleFieldDefinition $field, array $options = []) {
```
paul121: Instead of adding multiple types I wonder if we should add an assertion inside `FarmFieldFactory::bundleFieldDefinition` to ensure that a `BundleFieldDefinition` is indeed returned from the public function (and do the same for `::baseFieldDefinition`/`BaseFieldDefinition`).

I think this would fix the PHPStan error by itself and be an improvement, too. Given that the base vs bundle field type has important implications it wouldn't hurt to have some runtime assertions that stop things if this were ever wrong and got changed by `::buildFieldDefinition`.

Something like:
```php
public function bundleFieldDefinition(array $options = []): BundleFieldDefinition {
    $field = BundleFieldDefinition::create($options['type']);
    $this->buildFieldDefinition($field, $options);
    assert($field instanceof BundleFieldDefinition);
    return $field;
  }

paul121 (approved):

modules/core/field/src/FarmFieldFactory.php
mstenta: I don't think that `assert()`s will help in this case, because ultimately we need to be able to pass either base or bundle field definition objects to `buildFieldDefinition()`, so the method parameter type declaration should reflect that. The fact that it doesn't right now is a bug.

In the process, I also realized that all of the `::modify*Field()` methods ALSO needed their parameter type declarations to be updated in the same way. So I added an additional commit that does this. PHPStan doesn't pick up on this, but we know it's true.

mstenta:
Rebased on top of #986 and #992, updated the new `$entity->getOriginal()` instance from #986 (see https://github.com/farmOS/farmOS/pull/986#discussion_r2277805357), and force pushed.

modules/core/field/src/FarmFieldFactory.php
mstenta: Discussed a bit more with @paul121 in chat and realized that it wasn't the type change that fixed the PHPStan failure. It was simply removing the `&` before `$field` in the `buildFieldDefinition()` method arguments. For whatever reason that was throwing off PHPStan. I also removed the same `&`s from all the `::modify*Field()` methods. They are unnecessary because objects are always passed by reference.

mstenta:
My rebase accidentally restored `PlanStorage` (which was removed in #986), which caused PHPStan failures. Fixed and force-pushed.

Also changed the way one of the earlier PHPStan failures is fixed, per the conversation with @paul121 above.

This should be all green and good to go now! I'll wait for tests to pass and then merge...

PR #982, opened 2025-07-03
Move/split default role modules

mstenta:
This PR makes two changes:

1. Moves the `farm_role_account_admin` module to a new `farm_account_admin` module.
2. Splits the `farm_role_roles` module into three new modules: `farm_manager`, `farm_viewer`, `farm_worker`.

These new modules are now in `modules/roles` instead of buried in `modules/core/role/modules/account_admin` and `modules/core/role/modules/roles`. This improves their visibility within the codebase and standardizes how we provide default roles, in the same way that we provide default asset types, log types, quantity types, etc.

It also allows site admins to decide which roles they need without enabling all of them. So if you only need a "Manager" role but not a "Viewer" or "Worker", you can do that now.

The old modules are still present, but are marked as `obsolete` so they can't be installed. They each provide a new update hook that handles installing the new module(s) and ensuring that users keep the roles they had. The `farm_role_roles` update hook is smart about installing only the roles that are currently used, so eg if "Viewer" is not granted to any users then `farm_viewer` will not be installed.

I created two draft change records and linked to them from the `CHANGELOG.md` (to be published when this is merged):

- https://www.drupal.org/node/3527786
- https://www.drupal.org/node/3527787


symbioquine (approved):
LGTM

modules/role/worker/farm_worker.module
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Implements hook_oauth2_scope_info_alter().
+ */
+function farm_worker_oauth2_scope_info_alter(array &$scopes) {
+
+  // Enable the password grant for static role scopes.
+  if (\Drupal::moduleHandler()->moduleExists('simple_oauth_password_grant')) {
+    $target_scopes = [
+      'farm_manager',
```
paul121:
```suggestion
      'farm_worker',
```

paul121 (changes requested):
One typo but then this LGTM!

modules/role/worker/farm_worker.module
mstenta: Good catch thanks @paul121!!

mstenta:
Merged and published change records.

PR #984, opened 2025-07-04
Document farm_ext_ module namespacing guidance

symbioquine:
The existing module development namespacing guidance actually makes it more likely that contrib modules will eventually conflict with core farmOS modules.

This change describes a new prefix `farm_ext_` for external contrib modules and promises that farmOS itself will never ship a module thus named as part of core.

### DIFF
--- a/docs/development/module/index.md
+++ b/docs/development/module/index.md
@@ -29,6 +29,13 @@ For example, if you were to build a module that adds a new log type called
 this module is made to work with farmOS, and is not designed to be installed in
 other Drupal sites more generally.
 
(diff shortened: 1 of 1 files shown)


symbioquine:
We talked about this in the dev call and I thought I'd open a PR for it to capture the idea. If we want to proceed with it, I'll add the CHANGELOG entry to get it ready for merging...

mstenta:
I think this is a good addition - thanks @symbioquine! LGTM!

mstenta (approved):

paul121 (approved):

PR #985, opened 2025-07-31
Rename primary tab in edit forms to General

mstenta:
## Context

#770 implemented the Gin content form for our entity types (assets, logs, etc) and moved fields to vertical tabs within the form and sidebar. More discussion here for context: https://farmos.discourse.group/t/ui-improvements-editing-forms/1743

One of the decisions made during this work was to set the title of the primary vertical tab (the one at the top that is visible by default) to the label of the bundle being edited.

For example, if you are editing an Activity log, the label of the primary tab is "Activity". See the green text in the following screenshot:

Screenshot from 2025-07-31 14-34-28

Or, if you're editing an Animal asset, the label of the primary tab is "Animal":

Screenshot from 2025-07-31 14-35-26

## Proposal

I would like to propose that we rename this tab to simply "General".

Here is a screenshot of this change:

Screenshot from 2025-07-31 14-42-05

### Reasoning

My reasoning is two-fold...

First, I worry that the current titles (which vary based on the bundle) can potentially cause confusion/ambiguity in the form. If the tab is "Plant", that implies that it contains information about the Plant asset. But in reality, it's just "general" information about the plant asset. And all the other tabs ALSO contain information about the plant asset. So it feels a bit odd.

Second, while this strangeness is probably overcome quickly by new users, after they've had a chance to wrap their mind around the form and what is included, specifically in the context of core assets and logs, I think the ambiguity becomes greater with *plan* entities.

For example, this is what it looks like where you're adding a Crop plan (from https://www.drupal.org/project/farm_crop_plan):

Screenshot from 2025-07-31 15-02-25

... and adding a Grazing plan (from https://www.drupal.org/project/farm_grazing_plan):

Screenshot from 2025-07-31 15-02-29

The fact that "plan" is not included in the name makes "Crop" even more confusing. This is sort of a separate issue with the way plans types are named generally (which I may also propose a change for in a separate PR soon), but it serves to emphasize the ambiguity of using the bundle name as the title of the primary tab (and was my primary motivation for suggesting this change).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,7 @@ requirements (inherited from Drupal 11):
 - [The farm_account_admin role has moved to a new farm_account_admin module](https://www.drupal.org/node/3527786)
 - [The farm_manager, farm_worker, and farm_viewer roles have moved to their own modules](https://www.drupal.org/node/3527787)
 - [Allow plans without active status in farm_plan View #978](https://github.com/farmOS/farmOS/pull/978)
+- [Rename primary tab in edit forms to General #985](https://github.com/farmOS/farmOS/pull/985)
 
 ### Deprecated
 
--- a/modules/core/ui/theme/src/Form/GinContentFormBase.php
+++ b/modules/core/ui/theme/src/Form/GinContentFormBase.php
@@ -89,7 +89,7 @@ class GinContentFormBase extends ContentEntityForm implements RenderCallbackInte
     $field_groups = [
       'default' => [
         'location' => 'main',
-        'title' => 'Default',
+        'title' => $this->t('General'),
         'weight' => -50,
       ],
       'meta' => [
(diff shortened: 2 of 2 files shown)


mstenta:
@Fosten made a good point on the dev call today:

There is no indication what bundle an entity is when you are *editing* it.

One potential solution to this would be to add the bundle name to the "Edit [entity-label]" page title, so it is "Edit [bundle-label]: [entity-label]"

@symbioquine also suggested adding the entity type label to both the "Add" and "Edit" page titles. For example: "Add activity log" and "Edit activity log: apply manure".

paul121 (approved):

mstenta:
I'm going to merge this.

> One potential solution to this would be to add the bundle name to the "Edit [entity-label]" page title, so it is "Edit [bundle-label]: [entity-label]"
> 
> @symbioquine also suggested adding the entity type label to both the "Add" and "Edit" page titles. For example: "Add activity log" and "Edit activity log: apply manure".

Let's tackle these in follow-ups.

PR #988, opened 2025-08-05
Fix logic for forcing entity revision creation

mstenta:
Back in 2021, during the 2.x upgrade, we added logic to the `farm_entity` module to force the creation of entity revisions. This was done to ensure that revisions are created via API requests.

Original issue for more context: https://www.drupal.org/node/3216766

A recent commit broke this logic. This PR fixes it and adds an automated test to prevent future regressions.

It also moves the two hooks responsible for this logic from `farm_entity.module` to `farm_api.module`. My reasoning is: this is only necessary for API requests (but unfortunately we can't tell where the update is happening, so we take a heavy-handed approach), and now that the API is optional (#974), we can allow normal revision logic when the API is not in use. It also makes sense for the automated test for this to live in the `farm_api` module's kernel tests IMO. 

Here is the commit that broke the logic: https://github.com/farmOS/farmOS/commit/a512c3042e7eff93c6b628882f20df5db08c6bed

There are two things wrong with the current logic which causes the hook to bail early:

1. `$entity_type = $entity->get('type')->getEntity();` loads the content entity, not the bundle config entity. This causes `$entity_type instanceof RevisionableEntityBundleInterface` to return `FALSE`
2. The `$entity->getEntityType()->isRevisionable()` check changed to `$entity_type->getEntityType()->isRevisionable()`, which does not work if `$entity_type` is a bundle config entity (as it should be).

(Note that this PR builds on top of #980, because it touches code that changed in that PR, so it includes all of those commits as well. This PR only adds 2 commits by itself.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,6 +83,7 @@ requirements (inherited from Drupal 11):
 ### Fixed
 
 - [Fix plan_record bundle field providers #969](https://github.com/farmOS/farmOS/pull/969)
+- [Fix logic for forcing entity revision creation #988](https://github.com/farmOS/farmOS/pull/969)
 
 ## farmOS 3.x
 
--- a/modules/core/api/tests/src/Kernel/FarmApiTest.php
+++ b/modules/core/api/tests/src/Kernel/FarmApiTest.php
@@ -5,8 +5,10 @@ declare(strict_types=1);
 namespace Drupal\Tests\farm_api\Kernel;
 
 use Drupal\Component\Serialization\Json;
+use Drupal\Core\Entity\EntityInterface;
 use Drupal\KernelTests\KernelTestBase;
 use Drupal\Tests\user\Traits\UserCreationTrait;
+use Drupal\asset\Entity\AssetInterface;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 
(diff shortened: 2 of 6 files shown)


modules/core/api/farm_api.module
```diff
@@ -7,6 +7,97 @@
...
+    || !$entity instanceof FieldableEntityInterface
+  ) {
+    return;
+  }
+
+  // Force create new revision as json api doesn't do that by default.
+  // @see https://www.drupal.org/project/drupal/issues/2993557
+  // @see https://www.drupal.org/project/drupal/issues/2795279
+  // @see https://github.com/json-api/json-api/pull/824
+  $entity_type = $entity->get($entity->getEntityType()->getKey('bundle'))->entity;
```
paul121: Is there a reason this line can't be `$entity->getEntityType()` ?

```suggestion
  $entity_type = $entity->getEntityType();
```

paul121:
And should we commit this fix to v3?

modules/core/api/farm_api.module
mstenta: Ah yes if that works then it will be much cleaner. I'll make that change and see if tests still pass.

mstenta:
> And should we commit this fix to v3?

Sure I can cherry-pick this to 3.x after it's merged.

modules/core/api/farm_api.module
paul121: I could be missing something... but think this might work

mstenta: I tested this and `$entity->getEntityType()` does not return the same object as `$entity->get($entity->getEntityType()->getKey('bundle'))->entity`.

If I send an API request to update an `asset` entity:

`$entity->getEntityType()` returns a `\Drupal\Core\Entity\ContentEntityType` object, which represents the "entity type" (`asset`).

`$entity->get($entity->getEntityType()->getKey('bundle'))->entity` returns a `\Drupal\asset\Entity\AssetType` object, which represents the "entity bundle type" (`asset_type`).

In this case we actually want the `asset_type` entity, so that we can check `$entity_type instanceof RevisionableEntityBundleInterface` and `$entity_type->shouldCreateNewRevision()`.

Notably, we do use `$entity->getEntityType()` to check `$entity->getEntityType()->isRevisionable()`.

So I think this is working as expected.

mstenta:
OK so... I just discovered that the test I added passes even without this "fix" :face_with_spiral_eyes:.

Our code is still broken, as described in the issue description above, but it seems like revisions are being created via JSON:API `PATCH` requests regardless. :thinking: 

I'm trying to remember how I discovered this... but I can't :laughing: I was not testing the API directly, and I could have sworn that I made sure the test was failing before I fixed it. So I'm a bit confused now...

Maybe something else is setting `$entity->setNewRevision(TRUE)`??

I'll see if I can figure out what... but either way this PR is still valid. The test is valuable to have to ensure it works as desired, and our current code (even though it's not making a difference) is broken.

mstenta:
> our current code (even though it's not making a difference) is broken

Just to be clear, I know that it is broken, and I know that this PR fixes it, because I have been testing it with XDebug.  The automated tests just check to see if our desired outcome happens (a revision is created after an API `PATCH` request), but it doesn't test that the `farm_api` code itself is responsible for that. So if revisions are being created by something else, it would still pass.

mstenta:
Update: I added the new test (without the fix) on `3.x` and it still passes. So it seems that JSON:API `PATCH` requests have been making new revisions regardless of our code for a while.

Is it possible that our code was *never* necessary? Hard to say because we didn't write a test for it originally. Or did core make a change in the past few years that caused them to start generating revisions?

It's a bit hard to determine where `setNewRevision(TRUE)` is being called in the process of an API request, but I'll do a little more digging to see if I can isolate it...

Either way, it's still a good idea to add a test to our `farm_api` module to make sure this continues to work as expected.

And it may still be worth keeping our code, but perhaps for a more general reason. Originally it was added to handle JSON:API requests, but ultimately the way it was implemented does not limit it to API requests. It runs whenever an entity is being saved. This means that it ensures revisions are created by default even when PHP code is used to save an entity. Without it, developers must remember to call `setNewRevision(TRUE)` themselves, which is easy to forget. This results in loss of granular change records, because the new save essentially overwrites the existing current revision, so you lose that distinction.

With that in mind, I'm rethinking this PR a bit... perhaps instead of moving the logic to `farm_api.module`, it should be kept in `farm_entity.module`, as a general-purpose guarantee that new revisions are created.

mstenta:
> Is it possible that our code was _never_ necessary? Hard to say because we didn't write a test for it originally. Or did core make a change in the past few years that caused them to start generating revisions?

I found this issue from 2021 which provides the historical context: https://www.drupal.org/project/farm/issues/3216766

However, I also found this change record from the JSON:API module from before it was merged into core, which suggests that it should work as expected: https://www.drupal.org/node/3042037

mstenta:
> instead of moving the logic to `farm_api.module`, it should be kept in `farm_entity.module`, as a general-purpose guarantee that new revisions are created.

I've made these changes, added an additional general purpose automated test to the `farm_entity` module, and kept the new automated test for JSON:API coverage, just for good measure.

paul121 (approved):

modules/core/api/farm_api.module
paul121: Cool - yes - that makes sense.

mstenta:
Backported to 3.x: https://github.com/farmOS/farmOS/commit/7e04a3554e7c07d13b1d28a30d7f321230c2551d

PR #989, opened 2025-08-07
Inject entity_type.manager and current_user service dependencies into QuickFormBase class

mstenta:
The `QuickFormBase` class currently includes the `messenger` service via dependency injection, so that quick forms can set messages, but it doesn't include any other services. This means that downstream quick form developers often need to inject commonly needed services in every class.

This PR proposes that we inject two additional services into `QuickFormBase`:

- `entity_type.manager`
- `current_user`

Both of these are currently being injected separately into every quick form that farmOS core provides, so by moving them to `QuickFormBase` it reduces the amount of boilerplate needed in classes that extend from that.

- https://github.com/farmOS/farmOS/blob/286f13b607d20a0545e4fa0fa6b662f946a60cc6/modules/quick/birth/src/Plugin/QuickForm/Birth.php#L130-L134
- https://github.com/farmOS/farmOS/blob/286f13b607d20a0545e4fa0fa6b662f946a60cc6/modules/quick/group/src/Plugin/QuickForm/Group.php#L100-L102
- https://github.com/farmOS/farmOS/blob/286f13b607d20a0545e4fa0fa6b662f946a60cc6/modules/quick/inventory/src/Plugin/QuickForm/Inventory.php#L101-L103
- https://github.com/farmOS/farmOS/blob/286f13b607d20a0545e4fa0fa6b662f946a60cc6/modules/quick/movement/src/Plugin/QuickForm/Movement.php#L102-L104
- https://github.com/farmOS/farmOS/blob/286f13b607d20a0545e4fa0fa6b662f946a60cc6/modules/quick/planting/src/Plugin/QuickForm/Planting.php#L99-L102

... and in contrib quick forms:

- https://git.drupalcode.org/project/farm_eggs/-/blob/2.x/src/Plugin/QuickForm/Eggs.php?ref_type=heads#L86
- https://git.drupalcode.org/project/farm_weather/-/blob/2.x/src/Plugin/QuickForm/Precipitation.php?ref_type=heads#L67
- (probably others)

The `entity_type.manager` service is often used to load entity storage, and the `current_user` service is used to get the current user's timezone for date fields.

This would be a breaking change, so I'm proposing it for 4.0.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,7 @@ requirements (inherited from Drupal 11):
 - [The farm_manager, farm_worker, and farm_viewer roles have moved to their own modules](https://www.drupal.org/node/3527787)
 - [Allow plans without active status in farm_plan View #978](https://github.com/farmOS/farmOS/pull/978)
 - [Rename primary tab in edit forms to General #985](https://github.com/farmOS/farmOS/pull/985)
+- [Inject entity_type.manager and current_user service dependencies into QuickFormBase class #989](https://github.com/farmOS/farmOS/pull/989)
 
 ### Deprecated
 
--- a/docs/development/module/quick.md
+++ b/docs/development/module/quick.md
@@ -77,7 +77,7 @@ class Harvest extends QuickFormBase {
     $form['timestamp'] = [
       '#type' => 'datetime',
       '#title' => $this->t('Date'),
-      '#default_value' => new DrupalDateTime('now', \Drupal::currentUser()->getTimeZone()),
+      '#default_value' => new DrupalDateTime('now', $this->currentUser->getTimeZone()),
       '#required' => TRUE,
     ];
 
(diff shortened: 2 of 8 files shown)


mstenta:
> the `current_user` service is used to get the current user's timezone for date fields.

I can imagine a future where this is no longer necessary, if we provide a custom form element that handles user timezone automatically. I think that would be a nice improvement.

So of the two, with that potential future in mind, perhaps `current_user` is less important to include here (and if we were to remove it in the future that would necessitate another breaking change). Although having access to the current user can be useful for other things, so perhaps it would be fine to leave it in anyway.

mstenta:
> perhaps `current_user` is less important to include here

Also worth noting: `$this->currentUser()` is available as a method on all classes that extend `FormBase`.

I'm not sure what best practice is, however, because that method calls `\Drupal::currentUser()`, which means dependency injection is not used. I wonder if that method will go away in the future, and if it's better practice to use our own... or if we should just follow core's pattern until then... ?

mstenta:
Discussed with @paul121 and I'm going to refactor these changes so that we just use `$this->currentUser()` in core quick forms and don't inject `current_user` as a dependency ourselves.

Otherwise this is ready to merge.

mstenta:
> Also worth noting: `$this->currentUser()` is available as a method on all classes that extend `FormBase`.

Ah I was misguided! While it is true that `$this->currentUser()` is available as a method on all classes that extend `FormBase`, `QuickFormBase` does *not* extend from `FormBase`, because it is technically a plugin class, and we load those plugins into a separate `QuickForm extends FormBase` class.

So there *is* in fact value in injecting `current_user` into `QuickFormBase`. Therefore I am keeping this PR as originally suggested.

I made one small additional tweak to our quick form development docs to suggest using `$this->currentUser` instead of `\Drupal::currentUser()`.

PR #993, opened 2025-08-15
Fix CSV importer compatibility with migrate_source_ui v1.3+

mstenta:
The `migrate_source_ui` module introduced changes to their `MigrateSourceUiForm` class that broke backwards compatibility with our `CsvImportForm` class (which extends it).

@see https://www.drupal.org/project/migrate_source_ui/issues/3537749

This PR also unpins, unpatches, and updates `migrate_tools` and `migrate_plus` because the patches we were applying were merged upstream and included in new releases. The `migrate_source_ui` module then made its own changes in response to those updates, which broke our CSV importers.


mstenta:
Backported to `3.x`: https://github.com/farmOS/farmOS/commit/e184c10571020b86c41e4121dc14b40ae59ddbec

Issue #924, opened 2025-02-11
CSV import validation messages not showing


mstenta:
I just noticed that when you import an activity log via CSV, and it is not valid (eg: missing required fields) there are no errors shown to indicate what the problem is.

I am 99% positive they were showing when we originally built this feature. 🤔

Fosten:
I believe it only shows you the number of rows that fail, but not what the issue is.

This may be piggybacking on required fields, but… is there a way to expand the error message to cover invalid reference field values?

For example, I would often compare the failed row number in the “Imported Records” view to find typos or renamed values in my CSV.  

Would be nice if it said something like: “Row 17 could not be imported because Location: Gadren Bedd 1 does not exist."

mstenta:
> Would be nice if it said something like: “Row 17 could not be imported because Location: Gadren Bedd 1 does not exist."

It should be doing something like this already IIRC. So I think there's a bug that's preventing it...

mstenta:
Here is the commit that originally added the functionality: https://github.com/farmOS/farmOS/commit/4b180a5c98b29bdf2bfa03e4fe8d6947f95f5fb6

That is currently using `$event->logMessage($this->t('Row @rownum: @message', ['@rownum' => $record_number, '@message' => $message->message]), 'warning');`... and I can see that in Drupal's watchdog logs in `/admin/reports/dblog`, but it isn't showing in the UI. I nearly positive that it used to...

mstenta:
> I am 99% positive they were showing when we originally built this feature. 🤔

I looked back through https://www.drupal.org/project/farm/issues/3382539 and https://github.com/farmOS/farmOS/pull/722 but couldn't find any evidence of this, so it's just my (imperfect) memory that it worked before.

What I did find is that the `migrate_source_ui` module provides a `StubMigrationMessage` class, which *does* use Drupal's messenger service to add a message to the screen. The `MigrateSourceUiForm::submitForm()` method (which we inherit) creates a `new MigrateBatchExecutable` with a `new StubMigrationMessage()` passed in as a parameter.

- https://git.drupalcode.org/project/migrate_source_ui/-/blob/c0c0ef2c6878e8799d4da574e697dc3fcf8caea1/src/StubMigrationMessage.php#L25
- https://git.drupalcode.org/project/migrate_source_ui/-/blob/c0c0ef2c6878e8799d4da574e697dc3fcf8caea1/src/Form/MigrateSourceUiForm.php#L182

[...]

PR #994, opened 2025-08-16
Remove dependency on Migrate Source UI module

mstenta:
This pull request refactors our `CsvImportForm` class to remove our dependency on the [Migrate Source UI](https://www.drupal.org/project/migrate_source_ui) module.

(This is similar to the way we removed our dependency on JSON:API Extras #964.)

We leveraged the Migrate Source UI module as a dependency early on when we added CSV Importers to farmOS v2, because it provided a form for uploading CSVs to use as a migration source file. The module is quite simple, and there are some benefits to dropping the dependency and replicating+maintaining the logic we need ourselves.

For one, it provides things we don't need, like support for other file types. We also have to jump through some hoops with our own code in order to override/disable certain things, move uploaded files to the `private://csv` directory, and register file usage. By absorbing only the logic we need into our module, we can simplify the code quite a bit.

Also, we had to fix our importers recently because the upstream module introduced breaking changes (#993). The maintainer stated that they do not consider forms covered by the BC promise (even though the form is the only thing the module provides): https://www.drupal.org/project/migrate_source_ui/issues/3537749#comment-16229671

So with all of those considerations, and since we are working on `4.x` and have the opportunity to remove dependencies right now, I thought it made sense.

This PR refactors our code, removes the dependency, and uninstalls the module, but it does not remove `drupal/migrate_source_ui` from our `composer.json`. This will happen in `4.x`.

I have a separate `4.x-remove-migrate-source-ui` that is identical to this one, except that it does not provide an update hook, adds this branch's update hook to `hook_removed_post_updates()`, and removes `drupal/migrate_source_ui` from our `composer.json`. I will open a PR for that next.


mstenta:
Also worth noting: we had to workaround an issue introduced previously by `migrate_source_ui` which broke our validation messages. See https://github.com/farmOS/farmOS/issues/924

It sounds like they are going to fix it soon, which will cause a new issue in farmOS because validation messages will be duplicated.

Here is my comment from February 2025:

> We will probably work around this issue in farmOS in the short term, but adding our own explicit call to `\Drupal::messenger()->addMessage()`. The only concern I have is that `migrate_source_ui` will restore `MigrateBatchExecutable` in a subsequent release and then we'll have *double* messages. It would be good to coordinate to avoid that.

https://www.drupal.org/project/migrate_source_ui/issues/3507162#comment-15991297

This PR avoids the need to coordinate with a potential new release of `migrate_source_ui`.

paul121 (approved):
Untested, but LGTM!

PR #996, opened 2025-08-29
Title improvements to entity forms, buttons, and Views

mstenta:
This PR makes a few improvements to the titles of entity forms, action buttons, and Views throughout the farmOS UI. I decided to bundle all of these proposed changes into a single PR because they are similar. They are motivated by a few different conversations we've had recently in development calls.

Summary of changes:

### Entity edit form page title

When editing an existing entity (eg: asset, log, plan), the entity edit form page title will now include the bundle label. For example, if you are editing an Activity log named "Spread compost", the edit page title will be "Edit Activity: *Spread compost*". Previously it was "Edit *Spread compost*". This is a follow-up to #985, which removed the bundle label from the primary tab in the edit form. See this comment for more context: https://github.com/farmOS/farmOS/pull/985#issuecomment-3164938642

**Before**:

Screenshot from 2025-08-29 13-42-24

**After**:

Screenshot from 2025-08-29 13-50-29

### Entity add form page title

When adding a new entity, the entity add form page title will now always be "Add [bundle]". This fixes a minor annoyance in Drupal core's default logic, which only uses the bundle label if there is more than one bundle available for a given entity type, otherwise it uses the entity type label. For example, before this change, if you only have the Activity log type enabled, and you go to /add/log/activity, the title would be "Add log", but if you had another log type enabled it would be "Add Activity". Now it will always be "Add activity".

**Before**:

Screenshot from 2025-08-29 13-41-51

**After**:

Screenshot from 2025-08-29 13-43-10

### Entity "By type" Views titles

In all entity Views with a "By type" display (eg: /logs/activity), the title of the page will now be "[view-title]: [bundle]" (eg: "Logs: Activity"). Previously it was "[bundle] [entity-type-plural]" (eg: "Activity logs").

(see before/after pics below, because they include both this change and the next one)

### Add entity action link button titles

The action link buttons for adding entities of a specific bundle are now titled "Add [entity-type]: [bundle]" (eg: "Add Log: Activity"). Previously they were "Add [bundle] [entity-type]" (eg: "Add Activity Log").

(the before/after pics below include both this change and the previous one)

**Before**:

Screenshot from 2025-08-29 13-56-36

Screenshot from 2025-08-29 13-56-32

**After**:

Screenshot from 2025-08-29 13-57-10

Screenshot from 2025-08-29 13-57-07

Worth noting: the last two ("Entity "By type" Views titles" and "Add entity action link button titles") address an issue that has become apparent with plan bundle names, specifically. Oftentimes, the name of a plan bundle will include the name "plan" in it (eg: "Crop plan"). This resulted in Views titled "Crop plan plans" and action buttons titled "Add Crop plan plan", so the workaround was to omit "plan" from the bundle name, which felt weird from a plan DX perspective. Now that is no longer necessary - this proposed change simply rearranges the title a bit to lessen the redundancy of it.

### DIFF
--- a/modules/core/ui/views/farm_ui_views.views_execution.inc
+++ b/modules/core/ui/views/farm_ui_views.views_execution.inc
@@ -125,12 +125,12 @@ function farm_ui_views_views_pre_render(ViewExecutable $view) {
   }
 
   // If this is a "By type" display and a bundle argument is specified, load
-  // the bundle label and set the title.
+  // the bundle label and add it to the title.
   $bundle = farm_ui_views_get_bundle_argument($view, $view->current_display, $view->args);
   if (!empty($bundle)) {
     $bundles = \Drupal::service('entity_type.bundle.info')->getBundleInfo($view->getBaseEntityType()->id());
     if (!empty($bundles[$bundle])) {
-      $title = $bundles[$bundle]['label'] . ' ' . $view->getBaseEntityType()->getPluralLabel();
+      $title = $view->getTitle() . ': ' . $bundles[$bundle]['label'];
     }
   }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,7 @@ requirements (inherited from Drupal 11):
 - [Rename primary tab in edit forms to General #985](https://github.com/farmOS/farmOS/pull/985)
 - [Inject entity_type.manager and current_user service dependencies into QuickFormBase class #989](https://github.com/farmOS/farmOS/pull/989)
 - [Allow the "type" base field view display to be configured #990](https://github.com/farmOS/farmOS/pull/990)
+- [Title improvements to entity forms, buttons, and Views #996](https://github.com/farmOS/farmOS/pull/996)
 
 ### Deprecated
 
--- a/modules/core/entity/farm_entity.module
+++ b/modules/core/entity/farm_entity.module
@@ -17,7 +17,7 @@ use Drupal\Core\Form\FormStateInterface;
 use Drupal\entity\EntityAccessControlHandler;
 use Drupal\entity\EntityPermissionProvider;
 use Drupal\farm_entity\BundlePlugin\FarmEntityBundlePluginHandler;
-use Drupal\farm_entity\Routing\DefaultHtmlRouteProvider;
+use Drupal\farm_entity\Routing\BundleEntityTypeRouteProvider;
 
 /**
  * Implements hook_modules_installed().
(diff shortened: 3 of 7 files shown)


symbioquine (approved):
LGTM!

modules/core/ui/views/farm_ui_views.views_execution.inc
```diff
@@ -125,12 +125,12 @@ function farm_ui_views_views_pre_render(ViewExecutable $view) {
...
 
   // If this is a "By type" display and a bundle argument is specified, load
-  // the bundle label and set the title.
+  // the bundle label and add it to the title.
   $bundle = farm_ui_views_get_bundle_argument($view, $view->current_display, $view->args);
   if (!empty($bundle)) {
     $bundles = \Drupal::service('entity_type.bundle.info')->getBundleInfo($view->getBaseEntityType()->id());
     if (!empty($bundles[$bundle])) {
-      $title = $bundles[$bundle]['label'] . ' ' . $view->getBaseEntityType()->getPluralLabel();
+      $title = $view->getTitle() . ': ' . $bundles[$bundle]['label'];
```
paul121, line 133: Will this title still be translated?

paul121 (approved):
Just one question, but LGTM

modules/core/ui/views/farm_ui_views.views_execution.inc
mstenta: Good question... but maybe it's a broader one... this PR isn't making any change to the type of the `$title` variable that we are generating. It's a string before and after (confirmed with debugger, and confirmed that `$view->getTitle()` method returns a string).

However, I see that elsewhere in our hook (which sets the title in the context of taxonomy term Views, as opposed to this change which affects the "By type" assets/logs/etc Views) we set `$title` to a translatable string. So we might want to test translations of these titles more thoroughly in general.

That would be a separate/existing issue, though, so I don't think it needs to block this change.

mstenta: I just tested with German translations and it looks like these titles were never being properly translated. :-(

I'll open up a follow-up bug report...

mstenta: https://github.com/farmOS/farmOS/issues/997

paul121: > this PR isn't making any change to the type of the $title variable that we are generating. It's a string before and after 

I wasn't as concerned with the type, more so just that I'm not sure if the views titles are translated themselves? 
Going forward I guess we should make sure they are, might be covered w/ config translations?

Bundle labels (singular and plural) are translated, I think? So I'm surprised this wasn't already working. Hmm.

PR #1000, opened 2025-09-04
Check that real path exists before loading exif data

paul121:
When the image file does not exist an error is thrown for an empty first argument to exif_read_data() and causes a page crash.

This can happen when working from a database dump and you do not have the same images/files from the production environment, as reported here: https://github.com/Rothamsted-Ecoinformatics/farm_rothamsted/issues/843


mstenta:
Rebased onto 3.x, fixed `phpcs`, and merged. Thanks @paul121!

mstenta:
Cherry-picked onto 4.x: https://github.com/farmOS/farmOS/commit/0c9d64a5020e01f5f6b37314ac494e480cf49288

PR #1004, opened 2025-10-15
Do not allow entity revisions to be reverted

mstenta:
As discovered in https://github.com/farmOS/farmOS/pull/849#issuecomment-3408126282, Drupal does not run entity constraint validation when entities are reverted to previous revisions. farmOS modules provide entity constraint validation logic that may depend on other entities in the database (not just the data in the entity being reverted). This can lead to data that is in an invalid state.

This PR takes a very simple approach to solving this by preventing entities from being reverted to previous revisions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,6 +92,7 @@ requirements (inherited from Drupal 11):
 - [Fix logic for forcing entity revision creation #988](https://github.com/farmOS/farmOS/pull/969)
 - [Fix CSV importer compatibility with migrate_source_ui v1.3+ #993](https://github.com/farmOS/farmOS/pull/993)
 - [Check that real path exists before loading exif data #1000](https://github.com/farmOS/farmOS/pull/1000)
+- [Do not allow entity revisions to be reverted #1004](https://github.com/farmOS/farmOS/pull/1004)
 
 ## farmOS 3.x
 
--- a/modules/core/entity/farm_entity.module
+++ b/modules/core/entity/farm_entity.module
@@ -91,6 +91,27 @@ function farm_entity_entity_type_build(array &$entity_types) {
 }
 
 /**
+ * Implements hook_entity_type_alter().
+ */
+function farm_entity_entity_type_alter(array &$entity_types) {
+  /** @var \Drupal\Core\Entity\EntityTypeInterface[] $entity_types */
+
+  // Remove the form for reverting entity revisions.
+  // We do this because farmOS modules provide entity constraint validation
+  // logic that may depend on other entities in the database (not just the data
+  // in the entity being reverted), and Drupal does not validate entities when
+  // they are reverted. This could result in entities that no longer validate,
+  // leaving the database in a state that would normally be considered invalid.
+  foreach ($entity_types as $entity_type) {
+    if ($entity_type->hasLinkTemplate('revision-revert-form')) {
+      $links = $entity_type->getLinkTemplates();
+      unset($links['revision-revert-form']);
+      $entity_type->set('links', $links);
+    }
+  }
+}
+
+/**
  * Implements hook_entity_field_storage_info_alter().
  *
  * @todo https://www.drupal.org/project/farm/issues/3194206
--- a/modules/core/entity/tests/src/Functional/FarmEntityRevisionsTest.php
+++ b/modules/core/entity/tests/src/Functional/FarmEntityRevisionsTest.php
@@ -1,0 +1,75 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Drupal\Tests\farm_entity\Functional;
+
+use Drupal\Tests\farm_test\Functional\FarmBrowserTestBase;
+
+/**
+ * Tests that entity revisions cannot be reverted.
+ *
+ * @group farm
+ */
+class FarmEntityRevisionsTest extends FarmBrowserTestBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  protected static $modules = [
+    'farm_entity',
+    'farm_entity_test',
+  ];
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function setUp(): void {
+    parent::setUp();
+
+    // Create and login a user with access to entity revisions.
+    $user = $this->createuser([
+      'view any asset',
+      'view any log',
+      'view any plan',
+      'view all asset revisions',
+      'view all log revisions',
+      'view all plan revisions',
+      'revert all asset revisions',
+      'revert all log revisions',
+      'revert all plan revisions',
+    ]);
+    $this->drupalLogin($user);
+  }
+
+  /**
+   * Test that entity revisions cannot be reverted.
+   */
+  public function testFarmEntityRevisions() {
+    $entity_types = [
+      'asset',
+      'log',
+      'plan',
+    ];
+    foreach ($entity_types as $entity_type) {
+      $storage = \Drupal::entityTypeManager()->getStorage($entity_type);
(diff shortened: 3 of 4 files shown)


mstenta:
To be clear: I don't know if this is the best way to solve the issue - but I wanted to get something simple together to start the thinking...

This only disables access to the form at `/[entity-type]/[id]/revisions/[revision-id]/revert`. Notably it does not disable/hide the "Revert" buttons from `/[entity-type]/[id]/revisions`, and gives no indication to the user *why* access is denied when they click them.

There may be other ways to revert revisions, as well. Needs more investigation...

There are also permissions for reverting revisions, which we grant to "managed" roles. We should probably stop granting those. Although we can't prevent those permission from being granted to other roles.

mstenta:
An important question we need to consider with this is: do we need to allow revisions to be reverted? It is, of course, a very nice feature to have. But how often is it necessary? And is requiring that changes be reverted manually "good enough"?

mstenta:
I think I found a better way to do this. The form for reverting revisions is provided by the `entity` module, but only if `revision-revert-form` is declared in the entity type annotation:

https://git.drupalcode.org/project/entity/-/blob/276e51fc92ba8159b5e0783d07f671dfd75956b6/src/Routing/RevisionRouteProvider.php#L77

This is also used to add the "Revert" buttons to `/[entity-type]/[id]/revisions`:

https://git.drupalcode.org/project/entity/-/blob/276e51fc92ba8159b5e0783d07f671dfd75956b6/src/Controller/RevisionOverviewController.php#L69

Here is where we (and the `log` module) declare `revision-revert-form`:

- https://github.com/farmOS/farmOS/blob/29e892d601fe9a831433c9f67737cf15b7f125a3/modules/core/asset/src/Entity/Asset.php#L78
- https://github.com/farmOS/farmOS/blob/29e892d601fe9a831433c9f67737cf15b7f125a3/modules/core/plan/src/Entity/Plan.php#L76
- https://git.drupalcode.org/project/log/-/blob/2bbb45e9c97dca0af0d07efd219733392327d3ad/src/Entity/Log.php#L80

(Also, the `organization` entity does the same, although #849 has not been merged yet).

Notably, `quantity` and `plan_record` entities do not declare `revision-revert-form`.

So perhaps the best approach is to simply remove that `revision-revert-form` declaration from entity definitions. This will remove the form and button for reverting revisions from the UI.

mstenta:
> So perhaps the best approach is to simply remove that `revision-revert-form` declaration from entity definitions.

It appears this is not currently possible. :-(

See core issue: https://www.drupal.org/project/drupal/issues/3124206

Edit: I found a way around that...

```php
/**
 * Implements hook_entity_type_alter().
 */
function farm_entity_entity_type_alter(array &$entity_types) {
  /** @var \Drupal\Core\Entity\EntityTypeInterface[] $entity_types */

  // Remove the form for reverting entity revisions.
  // We do this because farmOS modules provide entity constraint validation
  // logic that may depend on other entities in the database (not just the data
  // in the entity being reverted), and Drupal does not validate entities when
  // they are reverted. This could result in entities that no longer validate,
  // leaving the database in a state that would normally be considered invalid.
  foreach ($entity_types as $entity_type) {
    if ($entity_type->hasLinkTemplate('revision-revert-form')) {
      $links = $entity_type->getLinkTemplates();
      unset($links['revision-revert-form']);
      $entity_type->set('links', $links);
    }
  }
}
```

mstenta:
> So perhaps the best approach is to simply remove that `revision-revert-form` declaration from entity definitions.

Refactored and force pushed to use this approach instead.

> There are also permissions for reverting revisions, which we grant to "managed" roles. We should probably stop granting those. Although we can't prevent those permission from being granted to other roles.

I also removed the line for adding the `revert all [entity-type] revisions` permission to managed roles.

This includes a functional test to confirm that it's possible to view `/[entity-type]/[id]/revisions`, that the "Revert" text does not appear anywhere in that page, and that `/[entity-type]/[id]/revisions/[revision-id]/revert` returns a 404.

This is now ready for review.

symbioquine (approved):
LGTM!

Of course an even better fix would be finding a way to hook into the revision reversion and trigger validation...

symbioquine:
I also checked and it looks like the JSON:API support for revisions is read-only: https://www.drupal.org/docs/core-modules-and-themes/core-modules/jsonapi-module/revisions

mstenta:
Thanks for the review @symbioquine!

> Of course an even better fix would be finding a way to hook into the revision reversion and trigger validation...

Yea agreed. Maybe we can dig into that in the future and restore this feature.

I'm also very keen to see this issue finished up in the Diff module: https://www.drupal.org/project/diff/issues/2452523

That would allow users to see a comparison of changes between revisions. In lieu of a "Revert" button, that would at least make it easier to figure out what changed and manually change it back (rather than reverting the entire revision and potentially undoing changes that you want to keep).

paul121 (approved):
Bummer that the "Revert" button is still shown even if you don't have the `revert` permission (I didn't test but sounds like & looks like that from reading the code you linked to).

If that wasn't the case I would suggest just removing the permission, rather than forcibly removing the functionality so others could still use it if necessary. As is it's just another little thing for us to track & maintain 😅

mstenta:

Yea, it would be nice if we could keep the revert functionality. Maybe we can restore it in the future, if we have time to dig deeper into this. Perhaps there's a way to run constraint validation on revert. Let's consider that a next step.

paul121:
So I was doing some research in to D11 core support for revisions recently... it seems some of the features from Entity have been "merged" to core while also being somewhat upgraded (*@paul121 expresses some frustration about the ubiquitous nature of Entity module while also having subpar documentation/maintenance... :frowning_face: *)

Notably, it appears that now access is checked to only show the `Revert` button when the user has permission: https://git.drupalcode.org/project/drupal/-/blob/11.2.7/core/lib/Drupal/Core/Entity/Controller/VersionHistoryController.php?ref_type=tags#L102-118

It is possible some of the expected link template patterns for revisions changed (didn't compare them too closely): https://git.drupalcode.org/project/drupal/-/blob/11.2.7/core/lib/Drupal/Core/Entity/Routing/RevisionHtmlRouteProvider.php?ref_type=tags#L23-37

Also looked at the core Revision Revert Form to see if validation had been added (I'm not sure if we are currently using this or a form from Entity module?) - but the core form has not changed since 2023 and does not appear to perform any validation. https://git.drupalcode.org/project/drupal/-/blob/11.2.7/core/lib/Drupal/Core/Entity/Form/RevisionRevertForm.php?ref_type=tags

mstenta:
> (*@paul121 expresses some frustration about the ubiquitous nature of Entity module while also having subpar documentation/maintenance... ☹️ *)

:laughing: I know what you mean... I hope that one day the `entity` module is no longer necessary, because the parts we need get absorbed into core. That was partly the purpose of the `farm_entity` module: to encapsulate all of those dependencies on `entity` - although that focus may have blurred since we created in in 2.x... I would still like to think of it that way as much as possible...

> Also looked at the core Revision Revert Form to see if validation had been added (I'm not sure if we are currently using this or a form from Entity module?) - but the core form has not changed since 2023 and does not appear to perform any validation. https://git.drupalcode.org/project/drupal/-/blob/11.2.7/core/lib/Drupal/Core/Entity/Form/RevisionRevertForm.php?ref_type=tags

I searched the core issue queue for existing issues about this, but couldn't find any, so I created one: https://www.drupal.org/project/drupal/issues/3556840

Thanks for looking into all of this! Perhaps we should create an issue in https://www.drupal.org/project/issues/farm to lay out a roadmap for a future selves...

PR #1005, opened 2025-10-16
Use PHP 8 constructor property promotion

mstenta:
This refactors all of the PHP class constructors in farmOS to make use of PHP 8 constructor property promotion.

https://php.watch/versions/8.0/constructor-property-promotion

Drupal core is in the process of doing the same: https://www.drupal.org/project/drupal/issues/3278431

This also removes docblock comments from all constructors, because they are no longer required by Drupal coding standards. See: https://www.drupal.org/project/coder/issues/3400560

There is one big commit in the middle of this branch (called "Use PHP 8 constructor property promotion.") which makes the actual changes that are the namesake of this PR. The purpose of all of the commits leading up to that are to refactor our constructors to make them as simple as possible, in preparation for using property promotion. All of the commits after are minor cleanups or linting fixes. A Rector rule is also added to ensure future classes use property promotion when possible.

### DIFF
--- a/modules/core/map/src/Element/FarmMap.php
+++ b/modules/core/map/src/Element/FarmMap.php
@@ -40,8 +40,8 @@ class FarmMap extends RenderElementBase {
    *   appended to 'farm-map-' as the map element ID if one has not already
    *   been provided.
    *
-   * @return \Drupal\farm_map\Element\FarmMap
-   *   The farm map render element.
+   * @return array
+   *   The farm map render element build array.
    *
    * @see \Drupal\farm_map\Event\MapRenderEvent
    */
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,6 +68,7 @@ requirements (inherited from Drupal 11):
 - [Inject entity_type.manager and current_user service dependencies into QuickFormBase class #989](https://github.com/farmOS/farmOS/pull/989)
 - [Allow the "type" base field view display to be configured #990](https://github.com/farmOS/farmOS/pull/990)
 - [Title improvements to entity forms, buttons, and Views #996](https://github.com/farmOS/farmOS/pull/996)
+- [Use PHP 8 constructor property promotion #1005](https://github.com/farmOS/farmOS/pull/1005)
 
 ### Deprecated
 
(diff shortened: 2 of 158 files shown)


symbioquine (approved):

modules/core/map/src/Element/FarmMap.php
```diff
@@ -40,8 +40,8 @@ public function getInfo() {
    *   appended to 'farm-map-' as the map element ID if one has not already
    *   been provided.
    *
-   * @return \Drupal\farm_map\Element\FarmMap
-   *   The farm map render element.
+   * @return array
+   *   The farm map render element build array.
```
paul121, line 44: I think we should maybe change this to `FarmMap|array`. We return the value of `$event->element` which is typed as either: https://github.com/farmOS/farmOS/blob/0157bf49b1b6d56f76861ceed5a7f56929beec2a/modules/core/map/src/Event/MapRenderEvent.php#L20-L25

I think there is something weird behind the scenes here where it ends up as either an array or the actual render element. I never fully figured it out which is why I kept both types... Probably not a biggies because I don't think these type doc changes would break strict type rules, but maybe worth looking into again.

paul121 (approved):

mstenta:
Rebased onto latest `4.x` commit, added a line to `CHANGELOG.md`, and merged.

Thanks for the reviews @paul121 and @symbioquine!

@paul121 I'll take a deeper look at the `FarmMap` docblock and see if I can make sense of it...

modules/core/map/src/Element/FarmMap.php
mstenta: I double checked this, and notably this PR also changes the typehint of that class attribute. Now it is promoted from the constructor property as an `array`: 

https://github.com/farmOS/farmOS/blob/b8565eafe65e765e242492dea331ab60fb268478/modules/core/map/src/Event/MapRenderEvent.php#L20-L24

This `MapRenderEvent` is instantiated in `FarmMap::preRenderMap()` and the `$element` variable that is passed into it is an array:

https://github.com/farmOS/farmOS/blob/b8565eafe65e765e242492dea331ab60fb268478/modules/core/map/src/Element/FarmMap.php#L113

PR #1006, opened 2025-10-16
Autowire injected service dependencies

mstenta:
~**Note:** this PR builds on top of #1005 and therefore includes all of its commits. When that is merged I will rebase this so it only includes the relevant commits for these changes. In the meantime, to see this PR's changes by themselves, use this link: https://github.com/mstenta/farmOS/compare/4.x-property-promotion...4.x-autowire~ Edit: #1005 has been merged.

This leverage's [Symfony's autowiring](https://symfony.com/doc/current/service_container/autowiring.html) to reduce a lot of boilerplate and cross-file maintenance. Combined with #1005, most of the classes in farmOS are much easier to jump into and understand for a newcomer.

This PR includes 4 commits, which can be summarized as follows:

- ~`Patch drupal/geofield for Issue #3545961: Provide geofield.geophp and geofield.wkt_generator service aliases for autowiring` - This is just a small patch to the [Geofield](https://drupal.org/project/geofield) to support autowiring in farmOS. We use two of that module's services in some of our classes, but the latest release of that module does not currently provide an autowiring alias for them, which means we would need to explicitly add `#[Autowire(service: 'geofield.geophp')]` in each of their constructors to use autowiring. The Geofield maintainers have already committed aliases to the dev branch, so we will be able to remove this patch as soon as they release a new stable version that includes those changes. At that point, our automated builds will start to fail (because this patch will no longer apply), which will tell us it's time to remove the patch.~ **Edit:** A new version of Geofield was released which includes this. We no longer need to patch. Now this first commit just bumps our required Geofield version to ^1.66.
- `Provide aliases for farmOS core services.` - This provides the same kind of aliases as the Geofield module to all the core services that farmOS provides. This allows Symfony to infer the service based on the interface that is passed into the constructor. Drupal core recently did the same thing: https://www.drupal.org/node/3323122
- `Autowire farmOS core services.` - This adds `autowire: true` and removes `arguments: [...]` from all module `*.services.yml` files. We will no longer need to keep these arguments (lists of service aliases) in sync with the classes that they point to. It also adds `#[Autowire(service: '...')]` in a few constructors for services that do not have an alias.
- `Use AutowireTrait to autowire ContainerInjectionInterface classes.` - A lot of our classes implement `ContainerInjectionInterface`, which requires that we add a `create()` method that declares the names of the services our class requires. These must correspond with the constructor parameters so that Symfony can initialize the class and pass in the correct dependencies to the constructor. Drupal core now provides an `AutowireTrait` that adds the `create()` method for us. This method uses reflection to figure out which services need to be passed, and allows us to remove (and stop maintaining) a lot of our `create()` methods.

Worth noting, this comment in Drupal core's `AutowireTrait`:

```
 * This trait uses reflection and may cause performance issues with classes
 * that will be instantiated multiple times.
```

I haven't experienced any performance issues in testing, so IMO we can address those individually if/when they are encountered.

Also worth noting, this PR only adds `AutowireTrait` to classes that implement `ContainerInjectionInterface`. There are other classes that still require `create()` methods to be maintained (with different method arguments), for now...

- `ContainerFactoryPluginInterface` - Plugin classes (see upstream issue: https://www.drupal.org/project/drupal/issues/3452852)
  - `public static function create(ContainerInterface $container, array $configuration, $plugin_id, $plugin_definition);`
- `ContainerDeriverInterface` - Deriver classes (I couldn't find an upstream issue for this)
  - `public static function create(ContainerInterface $container, $base_plugin_id)`

... so hopefully we can remove those in the future when Drupal core figures out how they are going to handle them.

### DIFF
--- a/modules/core/api/src/Controller/FarmEntryPoint.php
+++ b/modules/core/api/src/Controller/FarmEntryPoint.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Drupal\farm_api\Controller;
 
+use Drupal\Core\DependencyInjection\AutowireTrait;
 use Drupal\Core\Extension\ProfileExtensionList;
 use Drupal\Core\Session\AccountInterface;
 use Drupal\jsonapi\CacheableResourceResponse;
@@ -12,7 +13,6 @@ use Drupal\jsonapi\JsonApiResource\JsonApiDocumentTopLevel;
 use Drupal\jsonapi\JsonApiResource\NullIncludedData;
 use Drupal\jsonapi\JsonApiResource\ResourceObjectData;
 use Drupal\jsonapi\ResourceType\ResourceTypeRepositoryInterface;
-use Symfony\Component\DependencyInjection\ContainerInterface;
 
 /**
  * Extend the core jsonapi EntryPoint controller.
@@ -31,23 +31,14 @@ use Symfony\Component\DependencyInjection\ContainerInterface;
  */
 class FarmEntryPoint extends EntryPoint {
 
+  use AutowireTrait;
+
   public function __construct(
     ResourceTypeRepositoryInterface $resourceTypeRepository,
     AccountInterface $user,
     protected ProfileExtensionList $profileExtensionList,
   ) {
     parent::__construct($resourceTypeRepository, $user);
-  }
-
-  /**
-   * {@inheritdoc}
-   */
-  public static function create(ContainerInterface $container) {
-    return new static(
-      $container->get('jsonapi.resource_type.repository'),
-      $container->get('current_user'),
-      $container->get('extension.list.profile'),
-    );
   }
 
   /**
--- a/modules/core/data_stream/modules/notification/src/Form/DataStreamNotificationForm.php
+++ b/modules/core/data_stream/modules/notification/src/Form/DataStreamNotificationForm.php
@@ -4,18 +4,21 @@ declare(strict_types=1);
 
 namespace Drupal\data_stream_notification\Form;
 
+use Drupal\Core\DependencyInjection\AutowireTrait;
 use Drupal\Core\Entity\EntityForm;
 use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Form\SubformState;
 use Drupal\data_stream_notification\NotificationConditionManagerInterface;
 use Drupal\data_stream_notification\NotificationDeliveryManagerInterface;
-use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\DependencyInjection\Attribute\Autowire;
 
 /**
  * Data stream notification entity form.
  */
 class DataStreamNotificationForm extends EntityForm {
+
+  use AutowireTrait;
 
   /**
    * The data stream notification entity.
@@ -25,19 +28,11 @@ class DataStreamNotificationForm extends EntityForm {
   protected $entity;
 
   public function __construct(
+    #[Autowire(service: 'plugin.manager.data_stream_notification_condition')]
     protected NotificationConditionManagerInterface $conditionManager,
+    #[Autowire(service: 'plugin.manager.data_stream_notification_delivery')]
     protected NotificationDeliveryManagerInterface $deliveryManager,
   ) {}
-
-  /**
-   * {@inheritdoc}
-   */
-  public static function create(ContainerInterface $container) {
-    return new static(
-      $container->get('plugin.manager.data_stream_notification_condition'),
-      $container->get('plugin.manager.data_stream_notification_delivery'),
-    );
-  }
 
   /**
    * {@inheritdoc}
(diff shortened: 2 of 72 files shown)


mstenta:
Some relevant Drupal change records for reference:

- https://www.drupal.org/node/3218156
- https://www.drupal.org/node/3396179
- https://www.drupal.org/node/3323122

mstenta:
> `ContainerFactoryPluginInterface` - Plugin classes (see upstream issue: https://www.drupal.org/project/drupal/issues/3452852) 

Oh snap! This just merged into Drupal core 11.x 2 days ago! So we'll be able to remove most of our remaining `create()` methods soon! I'll start a follow-up PR for that...

mstenta:
> I haven't experienced any performance issues in testing, so IMO we can address those individually if/when they are encountered.

Also just found this...

https://www.drupal.org/project/drupal/issues/3294266

So it sounds like performance won't be a concern once that gets in...

mstenta:
> Oh snap! This just merged into Drupal core 11.x 2 days ago! So we'll be able to remove most of our remaining `create()` methods soon! I'll start a follow-up PR for that...

https://github.com/farmOS/farmOS/pull/1008

mstenta:
> Patch drupal/geofield for Issue #3545961: Provide geofield.geophp and geofield.wkt_generator service aliases for autowiring

Ah looks like the 1.66 release of Geofield includes this now. I've pushed a commit that reverts the first commit in this PR (which patches Geofield). Assuming tests pass, we can drop both commits before merging.

modules/core/api/src/Controller/FarmEntryPoint.php
```diff
@@ -94,15 +59,18 @@ public function index() {
...
     $meta['farm'] = [
       'name' => $this->config('system.site')->get('name'),
       'url' => $base_url,
-      'version' => $this->farmProfileInfo['version'],
+      'version' => $farm_profile_info['version'],
     ];
 
     // Allow modules to add additional meta information.
-    $this->moduleHandler->alter('farm_api_meta', $meta['farm']);
+    $this->moduleHandler()->alter('farm_api_meta', $meta['farm']);
```
pcambra, line 73: I don't see this one clear, is this part of autowiring?

modules/core/data_stream/modules/notification/src/Form/DataStreamNotificationForm.php
```diff
@@ -4,62 +4,35 @@
...
-   * {@inheritdoc}
-   */
-  public static function create(ContainerInterface $container) {
-    return new static(
-      $container->get('plugin.manager.data_stream_notification_condition'),
-      $container->get('plugin.manager.data_stream_notification_delivery'),
-    );
-  }
+  public function __construct(
+    #[Autowire(service: 'plugin.manager.data_stream_notification_condition')]
```
pcambra, line 31: What's the difference between doing it here and on the services.yml file?

mstenta: IMO the benefit of doing it here is we don't need to touch the `*.services.yml` file in the future if we change this class. All dependency info is contained in the same file.

modules/core/id_tag/src/Plugin/Field/FieldFormatter/IdTagFormatter.php
```diff
@@ -86,7 +67,8 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
       }
 
       // Render the type if it exists. Use the tag_type label.
-      if (!empty($item->type) && $tag_type = $this->tagTypeStorage->load($item->type)) {
+      $tag_type_storage = $this->entityTypeManager->getStorage('tag_type');
+      if (!empty($item->type) && $tag_type = $tag_type_storage->load($item->type)) {
```
pcambra, line 71: I've found a better experience to do the assigns outside of comparisons and then check the interface of the object (same with inline vars)

pcambra: An alternative option is assertions

modules/core/api/src/Controller/FarmEntryPoint.php
mstenta: Yes, now we're using the parent `EntryPoint` class's `moduleHandler()` method instead of injecting the `module_handler` ourselves.

modules/core/id_tag/src/Plugin/Field/FieldFormatter/IdTagFormatter.php
mstenta: Ah interesting. I'm not sure I understand, but perhaps we can open a follow-up PR? This PR isn't making any change to the assignment/comparison, it is just moving the `tag_type` storage loading out of the constructor.

pcambra: Yeah something like:

```
$tag_type = $tag_type_storage->load($item->type));
if ($tag_type interfaceof ImNotSureWhichInterface) {
```

mstenta: Ah yea - I'm not usually a fan of setting variables inside a condition either. :laughing:

mstenta:
#1005 has been merged. I rebased this branch on top of the latest `4.x`, removed the Geofield patch commit and the commit that reverted it, added a new commit to require Geofield ^1.66, and added a line to `CHANGELOG.md`.

mstenta:
FYI: the PHP Codesniffer failure is unrelated to this PR. It was introduced by yesterday's release of Coder (https://www.drupal.org/project/coder/releases/8.3.31). I just pushed a commit to fix it in `4.x`: https://github.com/farmOS/farmOS/commit/25266ae1cbe6e60c2f3e591b4349854a1cb636e9

paul121 (approved):

mstenta:
Merged! Thanks for the reviews @pcambra and @paul121!

PR #1009, opened 2025-10-21
Run Docker container as www-data user

mstenta:
Opening up this draft PR in the hopes that we can include it as a breaking change for 4.x.

The official farmOS Docker images currently run as `root` by default. It would be better security practice to run as `www-data` by default.

We looked into this previously (in #734), and found some considerations that warranted more review and discussion, so we decided to tackle it in its own PR.

### DIFF
--- a/.github/workflows/deliver.yml
+++ b/.github/workflows/deliver.yml
@@ -183,12 +183,14 @@ jobs:
           docker compose -f docker-compose.testing.common.yml -f docker-compose.testing.${{ matrix.dbms }}.yml config > docker-compose.yml
           # Replace the generic 'farmos/farmos:${FARMOS_DEV_BRANCH}-dev' image name/tag with the reference to our local platform-specific tagged image
           sed -i "s%farmos/farmos:${FARMOS_DEV_BRANCH}-dev%${DOCKER_IMG_PREFIX}:${FARMOS_DEV_BRANCH}-dev-${{ matrix.platform }}%g" docker-compose.yml
+      - name: Create www directory owned by www-data for bind-mount volume.
+        run: mkdir www && sudo chown 1000:1000 www
       - name: Start containers
         run: docker compose up -d
       - name: Wait until www container is ready
       # The www-container-fs-ready file is only created once we expect the containers to be online
       # so waiting for that lets us know it is safe to start the tests
-        run: until [ -f ./www/www-container-fs-ready ]; do sleep 0.1; done
+        run: sudo bash -c 'until [ -f ./www/www-container-fs-ready ]; do sleep 0.1; done'
       - name: Install pg_trgm PostgreSQL extension
       # This avoids race conditions when trying to automatically install it in concurrently run tests.
         if: matrix.dbms == 'pgsql'
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,7 @@ requirements (inherited from Drupal 11):
 
 - [farmOS 4.x requires PHP 8.4 #979](https://github.com/farmOS/farmOS/pull/979)
 - [Issue #3488916: Update Drupal core to 11.x](https://www.drupal.org/project/farm/issues/3488916)
+- [Run Docker container as www-data user #1009](https://github.com/farmOS/farmOS/pull/1009)
 - [Update Drupal to v11.2, PHPUnit to v11, PHPStan to v2 #980](https://github.com/farmOS/farmOS/pull/980)
 - [Update Docker base image to Debian Trixie #992](https://github.com/farmOS/farmOS/pull/992)
 - [Change how assets and plans are archived #986](https://github.com/farmOS/farmOS/pull/986)
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -146,6 +146,12 @@ RUN rm -r ${DRUPAL_PATH} && cp -rp ${BUILD_PATH} ${DRUPAL_PATH}
 # Create a Composer config directory for the www-data user.
 RUN mkdir /var/www/.composer && chown www-data:www-data /var/www/.composer
 
+# Run as www-data.
+USER www-data
+
 ##
 # Build the final production image.
 FROM production
+
+# Run as www-data.
+USER www-data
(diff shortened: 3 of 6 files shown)


mstenta:
This PR currently only has one commit, which simply adds `USER www-data` to `Dockerfile` before the final step where we build the production image:

```
# Run as www-data.
USER www-data

##
# Build the final production image.
FROM production
```

mstenta:
> This PR currently only has one commit, which simply adds `USER www-data` to `Dockerfile` before the final step where we build the production image:

NOTE! Tests are currently hanging on this step, so they need to be canceled manually otherwise they will not stop on their own:

```
      - name: Wait until www container is ready
      # The www-container-fs-ready file is only created once we expect the containers to be online
      # so waiting for that lets us know it is safe to start the tests
        run: until [ -f ./www/www-container-fs-ready ]; do sleep 0.1; done
```

mstenta:
Another issue: running as `www-data` doesn't currently work with our current `docker-entrypoint.sh` logic for copying files when a volume is mounted into `/opt/drupal` or `/opt/drupal/web/sites`.

This is what happens:

```
$ sudo docker run --rm -u www-data -v './www:/opt/drupal' farmos/farmos:4.x-dev
farmOS codebase not detected. Copying from pre-built files in the Docker image.
cp: cannot create regular file '/opt/drupal/./.editorconfig': Permission denied
cp: cannot create directory '/opt/drupal/./.git': Permission denied
cp: cannot create regular file '/opt/drupal/./.gitattributes': Permission denied
cp: cannot create regular file '/opt/drupal/./composer.json': Permission denied
cp: cannot create regular file '/opt/drupal/./composer.lock': Permission denied
cp: cannot create directory '/opt/drupal/./recipes': Permission denied
cp: cannot create directory '/opt/drupal/./vendor': Permission denied
cp: cannot create regular file '/opt/drupal/./phpcs.xml': Permission denied
cp: cannot create regular file '/opt/drupal/./phpstan-baseline.neon': Permission denied
cp: cannot create regular file '/opt/drupal/./phpstan.neon': Permission denied
cp: cannot create regular file '/opt/drupal/./rector.php': Permission denied
cp: cannot create regular file '/opt/drupal/./phpunit.xml': Permission denied
cp: cannot create directory '/opt/drupal/./web': Permission denied
cp: preserving times for '/opt/drupal/.': Operation not permitted
```

mstenta:
... this was previously described in these comments:

https://github.com/farmOS/farmOS/pull/734#issuecomment-1775914327

https://github.com/farmOS/farmOS/pull/734#issuecomment-1775921337

> > Docker compose files are mounting `./www:/opt/drupal`. If the path `./www` doesn't exist, it will be created being root the owner.
> 
> This is the core issue. If the `./www` directory does not exist before `docker compose up` (or `docker run`) runs, then it will be owned by root, and `docker-entrypoint.sh` (running as `www-data`) will not be able to copy files into it.
> 
> We could potentially work around this by modifying our `deliver.yml` to create the `www` directory and `chown www-data:www-data` it before we run `docker compose up`.
> 
> But we would also have to update our docs to tell end-users who are setting up their development/production environments to do that manually as well. I think that would work, but it's a bummer because it adds manual steps to the setup (always prefer removing steps to adding them). And we need to consider whether or not this will affect existing installs.
> 
> All-in-all, this is feeling like a more complicated change than I originally anticipated. Which means it will need a lot more community review and discussion before we can merge it regardless.
> 
> All the more reason to split it out to its own PR, IMO.

symbioquine:
Maybe these are obvious, but there are a few important things to note:

We're mostly talking about the case where a host directory or volume is mapped to `/opt/drupal` or `/opt/drupal/www` since those are the cases where our current `docker-entrypoint.sh` logic tries to copy files into a location which may have just been created (by the docker volume mount mechanism) or may have unknown permissions.

We should be able to update `docker-entrypoint.sh` to explicitly detect permission problems and exit with a useful error message explaining how to proceed.

The mapping between the `www-data` user inside the container and the UID of that user on the host system may not be one-to-one or necessarily obvious - so the command that needs to be run on the host system **can't** just be `sudo bash -c 'mkdir ./www && chown -R www-data:www-data ./www'`.

Instead, we can probably provide a `docker run` one-liner that avoids the UID mapping problem.

I think something like this might work:

```shell
docker run -it --rm -v $(pwd)/www:/opt/drupal farmos/farmos:3.4.5 bash -c 'chown -R www-data:www-data /opt/drupal'
```

mstenta:
> We're mostly talking about the case where a host directory or volume is mapped to `/opt/drupal` or `/opt/drupal/www` since those are the cases where our current `docker-entrypoint.sh` logic tries to copy files into a location which may have just been created (by the docker volume mount mechanism) or may have unknown permissions.

Yes - although minor correction: the two cases are when `/opt/drupal` or `/opt/drupal/web/sites` are bind-mounted (not `/opt/drupal/www`). Here is the relevant logic in `docker-entrypoint.sh`:

https://github.com/farmOS/farmOS/blob/89016826f85ba448a0c8b8da77a7cb7aebead935/docker/docker-entrypoint.sh#L10-L20

> We should be able to update `docker-entrypoint.sh` to explicitly detect permission problems and exit with a useful error message explaining how to proceed.

Great idea!

> The mapping between the `www-data` user inside the container and the UID of that user on the host system may not be one-to-one or necessarily obvious

True - although we do know that in the production image the `www-data` user ID is `33`, and in the development image it is `1000`, unless the user overrides that with `--build-arg WWW_DATA_ID=...` - see: https://github.com/farmOS/farmOS/blob/89016826f85ba448a0c8b8da77a7cb7aebead935/docker/Dockerfile#L103-L111

So I think the simplest thing to suggest, in the production environment context, is to `chown 33:33` the bind-mounted `sites` directory. This is a pretty minor change to our current documentation.

In the development context, assuming that the host system's user ID is `1000`, all they need to do is create the local `www` directory as their user (no `chown` required). This is also a minor change to current docs.

I've started working on these changes in a branch `wip` branch off of this PR, so I'll push those up for consideration... and also work on adding a useful error message to `docker-entrypoint.sh`...

mstenta:
Worked on this with @symbioquine on today's dev call. I think this is ready for review now.

symbioquine (approved):
LGTM!

paul121 (approved):
LGTM!

PR #1010, opened 2025-10-30
Allow API access to organizations

mstenta:
This is a follow-up to #849 (which added the `organization` entity type) and #964 (which replaced the `jsonapi_extras` module with our own logic for hiding internal resource types in `/api` endpoints).

This simply adds `organization` and `organization_type` to the list of "allowed" resource types that farmOS makes accessible via the `/api` endpoints.

(Thanks to @ktohalloran for reporting this omission.)


symbioquine (approved):
LGTM!

PR #1011, opened 2025-10-30
Change farmOS Update config revert logic from opt-out to opt-in

mstenta:
This PR introduces a breaking change to the `farm_update` module for the 4.x branch.

Currently, when this module is enabled, it will automatically revert configuration entities that it detects are overridden (if the active configuration is different from the YML defined in the `config/install/*.yml` file of the module that provides it). This was originally introduced (in the `2.x` branch) here: https://www.drupal.org/project/farm/issues/3220627

The rationale was that it makes it easier to ship changes to complex config entities (like Views) without needing to write update hooks for them, and without requiring site administrators to manually review and revert changes. This is still useful for some config (Views especially), but for simpler config changes it would be better practice to write intentional update hooks.

We currently provide a `hook_farm_update_excluded_config()` hook for excluding certain config from being automatically reverted. This PR changes this to `hook_farm_update_managed_config()` - thus making it an opt-in process instead of an opt-out process. It removes all of the "exclusions" we had previously, and explicitly declares all of the farmOS core Views configurations as "managed" so they will still be reverted automatically.

The benefit of doing this is it would allow end-users to override some of the core farmOS configurations without losing those changes upon cache rebuild (this would still be considered "at your own risk", and only recommended for advanced users).

It can also be very confusing to new developers when they are working locally, changing config, and then losing those changes when they rebuild the cache. And in cases where someone is managing their site's active configuration in source control, they wouldn't necessarily need to turn off the `farm_update` module (because it wouldn't be as heavy-handed about reverting changes when the cache is rebuilt).

Moving forward, we can decide if there are other core config entities that should be automatically managed. Or, we can write update hooks for changes to them as necessary.


mstenta:
Worth mentioning: on the dev call today @symbioquine was wondering if there could be some middle ground between fully-managing config (always reverting it), and manually writing update hooks, perhaps with a YML diff/patch based approach. We found this module as something to potentially consider in the future: https://www.drupal.org/project/config_patch

mstenta:
Also worth mentioning (and @symbioquine echoed on the dev call): making this change means that any configuration changes that have already been merged into `4.x` since `3.x` would no longer be automatically reverted (with the exception of the Views config we are now declaring as "managed").

I ran `git diff`s to find all changes between `3.x` and `4.x` (in `*/config/install` and `*/config/optional` directories, and this is what comes up (as of today):

```bash
$ git diff --name-only 4.x origin/3.x -- '*/config/install/*.yml'
modules/asset/animal/config/install/asset.type.animal.yml
modules/asset/compost/config/install/asset.type.compost.yml
modules/asset/equipment/config/install/asset.type.equipment.yml
modules/asset/equipment/tests/modules/farm_entity_test/config/install/log.type.test.yml
modules/asset/group/config/install/asset.type.group.yml
modules/asset/group/tests/modules/farm_group_test/config/install/asset.type.animal.yml
modules/asset/group/tests/modules/farm_group_test/config/install/asset.type.pasture.yml
modules/asset/land/config/install/asset.type.land.yml
modules/asset/material/config/install/asset.type.material.yml
modules/asset/plant/config/install/asset.type.plant.yml
modules/asset/product/config/install/asset.type.product.yml
modules/asset/seed/config/install/asset.type.seed.yml
modules/asset/sensor/config/install/asset.type.sensor.yml
modules/asset/structure/config/install/asset.type.structure.yml
modules/asset/water/config/install/asset.type.water.yml
modules/core/api/tests/modules/farm_api_test/config/install/asset.type.test.yml
modules/core/asset/tests/modules/asset_test/config/install/asset.type.default.yml
modules/core/entity/tests/modules/farm_entity_test/config/install/asset.type.test.yml
modules/core/entity/tests/modules/farm_entity_test/config/install/quantity.type.test.yml
modules/core/id_tag/tests/modules/farm_id_tag_test/config/install/asset.type.test.yml
modules/core/import/modules/csv/tests/modules/farm_import_csv_test/config/install/field.field.taxonomy_term.animal_type.test_config_field.yml
modules/core/import/modules/csv/tests/modules/farm_import_csv_test/config/install/field.storage.taxonomy_term.test_config_field.yml
modules/core/inventory/tests/modules/farm_inventory_test/config/install/asset.type.container.yml
modules/core/location/tests/modules/farm_location_test/config/install/asset.type.location.yml
modules/core/location/tests/modules/farm_location_test/config/install/asset.type.object.yml
modules/core/log/tests/modules/farm_log_test/config/install/asset.type.test.yml
modules/core/organization/tests/modules/organization_test/config/install/organization.type.default.yml
modules/core/quick/tests/modules/farm_quick_test/config/install/asset.type.test.yml
modules/core/role/modules/account_admin/config/install/user.role.farm_account_admin.yml
modules/core/role/modules/roles/config/install/user.role.farm_manager.yml
modules/core/role/modules/roles/config/install/user.role.farm_viewer.yml
modules/core/role/modules/roles/config/install/user.role.farm_worker.yml
modules/core/role/tests/modules/farm_role_test/config/install/user.role.farm_test.yml
modules/core/ui/views/config/install/views.view.farm_asset.yml
modules/core/ui/views/config/install/views.view.farm_log.yml
modules/core/ui/views/config/install/views.view.farm_log_quantity.yml
modules/organization/farm/config/install/organization.type.farm.yml
modules/organization/farm/tests/modules/farm_farm_test/config/install/asset.type.test.yml
modules/organization/farm/tests/modules/farm_farm_test/config/install/log.type.test.yml
modules/role/worker/config/install/user.role.farm_worker.yml
```

```bash
$ git diff --name-only 4.x origin/3.x -- '*/config/optional/*.yml'
modules/asset/animal/config/optional/core.entity_view_display.asset.animal.map_popup.yml
modules/core/asset/config/optional/system.action.asset_activate_action.yml
modules/core/asset/config/optional/views.view.asset_admin.yml
modules/core/data_stream/config/optional/views.view.data_stream_basic_data.yml
modules/core/organization/config/optional/system.action.organization_archive_action.yml
modules/core/organization/config/optional/system.action.organization_clone_action.yml
modules/core/organization/config/optional/system.action.organization_delete_action.yml
modules/core/organization/config/optional/system.action.organization_unarchive_action.yml
modules/core/plan/config/optional/system.action.plan_activate_action.yml
modules/core/plan/config/optional/views.view.plan_admin.yml
modules/core/ui/views/config/optional/views.view.farm_inventory.yml
modules/core/ui/views/config/optional/views.view.farm_organization.yml
modules/core/ui/views/config/optional/views.view.farm_organization_asset.yml
modules/core/ui/views/config/optional/views.view.farm_organization_log.yml
modules/core/ui/views/config/optional/views.view.farm_plan.yml
modules/organization/farm/config/optional/system.action.asset_farm_action.yml
```

These changes can be broken down into a few types:

Two that require attention:

- `asset.type.*.yml`
- `core.entity_view_display.asset.animal.map_popup.yml`

And the rest that don't require attention:

- `log.type.*.yml`
- `organization.type.*.yml`
- `quantity.type.*.yml`
- `field.field.taxonomy_term.*.*.yml`
- `field.storage.taxonomy_term.*.yml`
- `system.action.*.yml`
- `user.role.*.yml`
- `views.view.*.yml`

**Summary of the two that require attention:**

The `asset.type.*.yml` changes all simply remove `workflow: asset_default` (from #986). I have included a commit in this PR which adds some logic to #986's update hook to make these changes.

The `core.entity_view_display.asset.animal.map_popup.yml` change is related to #960, and updates the `map_popup` config accordingly. I have included a commit in this PR that adds the necessary change to #960's update hook.

**Summary of the ones that don't require attention:**

All of the `log.type.*.yml`, `organization.type.*.yml`, `quantity.type.*.yml`, `field.field.taxonomy_term.*.*.yml`, and `field.storage.taxonomy_term.*.yml` changes are either in tests, or are simply renamed/deleted files (which are not handled by `farm_update` anyway).

The `system.action.*.yml` changes are from #986 (where they are already covered by an update hook that renames the actions), and #849 (which adds new actions via new modules that are not installed on any existing sites yet).

The `user.role.*.yml` changes are from #982, which has the necessary update hooks.

And all of the `views.view.*.yml` files are declared as "managed" config now, so they will be reverted automatically.

mstenta:
I've rebased and force-pushed onto the latest `4.x`... no changes to this PR.

modules/core/update/src/FarmUpdate.php
```diff
@@ -76,32 +76,25 @@ public function rebuild(): void {
...
    * @return array
    *   An array of config item names.
    */
-  protected function getExcludedItems() {
+  protected function getManagedConfig() {
 
-    // Ask modules for config exclusions.
-    $exclude_config = $this->moduleHandler->invokeAll('farm_update_exclude_config');
+    // Ask modules for managed configuration items.
+    $managed_config = $this->moduleHandler->invokeAll('farm_update_managed_config');
```
paul121, line 89: Just trying to think if there is still a use-case for `farm_update_exclude_config`... Currently you can keep the `farm_update` module on and exclude a select few config items (including from other modules) that you want to manage yourself. With this change to opt-in that wouldn't be possible.

But instead of keeping `farm_update_exclude_config` a much simpler solution might be to simply add an alter hook `farm_update_managed_config_alter` so you could _remove_ some config from otherwise being included.

paul121 (approved):
Otherwise this looks good! Thanks for laying out the various config changes and how this would impact.

modules/core/update/src/FarmUpdate.php
mstenta: I almost did this when I was drafting this PR - it's an obvious next step - and very simple. :smile: 

I'll add a commit with tests for altering "managed" config...

mstenta:
I added the ability to alter the list of managed configuration items, along with an automated test.

https://github.com/farmOS/farmOS/pull/1011/commits/3312679e34a4ff2a5f105c945afa33ffe7ceb418

I'll wait for green on all tests and then merge this PR. Thanks for the review and input @paul121!

PR #1012, opened 2025-10-30
Remove asset_admin and plan_admin Views

mstenta:
This PR removes the `asset_admin` and `plan_admin` Views that we currently ship in our `asset` and `plan` modules. These create pages at `/admin/content/asset` and `/asset/content/plan`, respectively. No farmOS roles currently have access to these pages.

Notably, these are different from the Views provided by the `farm_ui_views` module (`farm_asset` and `farm_plan`), which provide the `/assets` and `/plans` paths (and sub-paths) for the default farmOS UI.

The only reason these Views exist in the first place is because the `asset` module was originally coped from the separate contrib `log` module (https://www.drupal.org/project/log), and then `plan` was copied from `asset`. The `log` module provides a `log_admin` View because it can be installed in other Drupal sites outside of farmOS, so it makes sense for it to ship with a "log admin" view that follows the same pattern as Drupal core's `node` module.

IMO there is no point in maintaining these `asset_admin` and `plan_admin` Views in farmOS.


mstenta:
Notably, when I created the `organization` module in #849, I did *not* create an `organization_admin` View. However, I just realized that the module still included some references to the `entity.organization.collection` route. So I just added a commit to this PR that cleans those up in the same way.

mstenta:
Oops I left `@todo`s in there to add update hooks that delete the Views on existing sites.

I added those and force pushed.

paul121 (commented):
I support removing the admin views of entities!

But I think this PR is also removing the Structure -> Asset Type menu links. I don't think these are connected to admin views, it is either a form/controller provided by Drupal core or maybe the Entity module. Removing the menu link shouldn't remove the pages themselves, but the menu link is very useful for getting to the relevant pages.

I know we don't often use Field UI & Config Fields, but I have found these pages useful for quickly creating new fields with different field types & configuring entity view/form displays.

modules/core/asset/asset.links.menu.yml
paul121, line 1: This is the menu link I think is unrelated to admin views and IMO would prefer to keep

mstenta:
Oops! That was a mistake. Good catch @paul121. I've reverted those deletions, rebased, and force pushed.

Ready for re-review! :-)

paul121 (approved):

PR #1014, opened 2025-11-19
Remove unused ManagedRolePermissionsManager service from AssignActionForm

ktohalloran:
The `AssignActionForm` of `farm_owner` was injecting a service provided by `farm_role` but then not using it, adding an unnecessary dependency on the `farm_role` module.


paul121 (approved):
Nice catch!

mstenta:
Thanks @ktohalloran! I amended the commit message (and this PR name) to match, and be more specific. Merging now!

PR #1016, opened 2025-11-23
Add update hook to update entity revisions with revision_user set to anonymous user

paul121:
_As reported in https://github.com/Rothamsted-Ecoinformatics/farm_rothamsted/issues/855 and tracking in https://github.com/Rothamsted-Ecoinformatics/farm_rothamsted/issues/861_

This is related to an issue we have since fixed concerning out logic to "force entity revision creation": https://github.com/farmOS/farmOS/pull/988

Throughout the time this bug existed it appears there was an additional side-effect: any entities entities that were _not created_ via standard entity forms had an issue where the `revision_user` was set to the Anonymous user. A common example of this would be logs/assets created via our Quick Forms. If the entity was later modified using the standard Edit form, the revision user would then be populated correctly on the new revision, but not old revisions.

Thankfully, because the user that originally created these entities is saved in the `uid` data field, we have the opportunity to update the original `revision_user` to this value.

Looking back at #988 I suppose we need to fix this for all of asset, log, plan and quantity entities.

### DIFF
--- a/modules/core/entity/farm_entity.post_update.php
+++ b/modules/core/entity/farm_entity.post_update.php
@@ -89,4 +89,16 @@ function farm_entity_post_update_update_anonymous_log_revisions(&$sandbox = NULL
       ->execute();
     \Drupal::logger('farm_entity')->notice('Updated @count plan revisions with correct revision_user.', ['@count' => $updated]);
   }
+
+  // Update quantities.
+  if (\Drupal::service('module_handler')->moduleExists('quantity')) {
+
+    // Note that quantities have a single quantity_revision table.
+    // Update anonymous quantity revision_user to the original creator uid.
+    $updated = $connection->update('quantity_revision')
+      ->expression('revision_user', 'quantity_revision.uid')
+      ->condition('revision_user', 0)
+      ->execute();
+    \Drupal::logger('farm_entity')->notice('Updated @count quantity revisions with correct revision_user.', ['@count' => $updated]);
+  }
 }
--- a/modules/core/entity/farm_entity.post_update.php
+++ b/modules/core/entity/farm_entity.post_update.php
@@ -7,6 +7,8 @@
 
 declare(strict_types=1);
 
+use Drupal\Core\Database\Database;
+
 /**
  * Enforce entity reference integrity on plan reference fields.
  */
@@ -35,4 +37,25 @@ function farm_entity_post_update_uninstall_exif_orientation() {
       \Drupal::service('module_installer')->uninstall(['exif_orientation']);
     }
   }
+}
+
+/**
+ * Update log revisions with revision_user set to anonymous user.
+ */
+function farm_entity_post_update_update_anonymous_log_revisions(&$sandbox = NULL) {
+
+  // Create a subquery to get the original creator uid.
+  $connection = Database::getConnection();
+  $subquery = $connection->select('log_field_revision', 'lfr')
+    ->fields('lfr', ['uid'])
+    ->where('lfr.id = log_revision.id')
+    ->where('lfr.revision_id = log_revision.revision_id');
+
+  // Update anonymous log revision_user to the original creator uid.
+  $updated = $connection->update('log_revision')
+    ->expression('revision_user', "($subquery)")
+    ->condition('revision_user', 0)
+    ->execute();
+
+  \Drupal::logger('farm_entity')->notice('Updated @count log revisions with correct revision_user.', ['@count' => $updated]);
 }
--- a/modules/core/entity/farm_entity.post_update.php
+++ b/modules/core/entity/farm_entity.post_update.php
@@ -43,9 +43,10 @@ function farm_entity_post_update_uninstall_exif_orientation() {
  * Update log revisions with revision_user set to anonymous user.
  */
 function farm_entity_post_update_update_anonymous_log_revisions(&$sandbox = NULL) {
-
-  // Create a subquery to get the original creator uid.
   $connection = Database::getConnection();
+
+  // Update logs.
+  // Create a subquery to get the original creator uid.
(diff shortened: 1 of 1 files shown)


paul121:
Oh I should also mention - it seems that some of our original entities that were migrated from FarmOS 1.x ended up with this same anonymous `revision_user`, too. I've seen numerous records with a revision log message like: `Migrated from farmOS 1.x on 2022-03-21`.

I've done some initial testing with this update hook and seems to be working well! My main concern is that this will update _any_ revision where the `revision_user` is 0, not only the "first" revisions. I'm not sure if there is a scenario where later revisions might have been set to the 0 anonymous user (I think this would be a bug, too?) but in that event, we can't assume the original creator `uid` would be the same revision user :shrug:

modules/core/entity/farm_entity.post_update.php
```diff
@@ -36,3 +38,67 @@ function farm_entity_post_update_uninstall_exif_orientation() {
...
+ * Update log revisions with revision_user set to anonymous user.
+ */
+function farm_entity_post_update_update_anonymous_log_revisions(&$sandbox = NULL) {
+  $connection = Database::getConnection();
+
+  // Update logs.
+  // Create a subquery to get the original creator uid.
+  $subquery = $connection->select('log_field_revision', 'lfr')
+    ->fields('lfr', ['uid'])
+    ->where('lfr.id = log_revision.id')
```
mstenta, line 52: I think this will fail if the database table names are prefixed (eg: `myprefix_log_revision` instead of `log_revision`). I assume a table alias can be set via `$connection->update('log_revision', 'lr')`... ?

But maybe it would be simpler to just use a static query for these (one `query()` instead of `select()` and `update()`).

eg:

```php
$connection->query('UPDATE {log_revision} lr SET lr.revision_user = (SELECT lfr.uid FROM {log_field_revision} lfr WHERE lfr.id = lr.id) WHERE lr.revision_user = 0');
```

Untested, but I think that's equivalent... and uses curly braces around all table names to ensure prefixing works.

mstenta (changes requested):
> My main concern is that this will update _any_ revision where the `revision_user` is 0, not only the "first" revisions.

It may be possible to limit this to the first revision with another subquery condition.

Given the static query I suggested above, it might look something like this:

```
UPDATE {log_revision} lr
SET lr.revision_user = (
  SELECT lfr.uid
  FROM {log_field_revision} lfr
  WHERE lfr.id = lr.id
)
WHERE lr.revision_user = 0
AND lr.id IN (
  SELECT MIN(lr2.id)
  FROM log_revision lr2
  GROUP BY lr2.id
);
```

(Haven't tested this...)

mstenta:
> Oh I should also mention - it seems that some of our original entities that were migrated from FarmOS 1.x ended up with this same anonymous `revision_user`, too.

This is actually OK with me, to be honest, because the migration was not performed by a particular user, so anonymous makes sense.

The way this PR works now, it will end up looking like is: the user that originally created the entity in farmOS 1.x is attributed as the one who made the "Migrated from 1.x" revision. That's weird IMO.

> A common example of this would be logs/assets created via our Quick Forms.

I wonder if we should make this PR more targeted to entities created via quick forms. We have the `asset__quick` and `log__quick` tables to reference, which would allow us to filter down to only those entities.

mstenta:
> the user that originally created the entity in farmOS 1.x is attributed as the one who made the "Migrated from 1.x" revision. That's weird IMO.

Although... maybe this is already the case for all assets/logs that were NOT created by quick forms? In which case, the weirdness is an existing issue, and doesn't need to be considered here. :laughing:

modules/core/entity/farm_entity.post_update.php
paul121: Ah not sure about table prefixing. At first I was going to use a static query like `UPDATE .. SET .. FROM .. WHERE` but saw that this syntax doesn't work on mysql which wants `UPDATE .. JOIN .. SET .. WHERE` 

The approach I ended up using with a SELECT subquery might not have those limitations but I wasn't sure. So... I thought using the Drupal API would be safer :-)

> I assume a table alias can be set via $connection->update('log_revision', 'lr')... ?

Unfortunately it doesn't support this.

mstenta: Ah good point re: MySQL... :-(

Well I think the table prefixing issue can be addressed by setting an alias in `$connection->update('log_revision', 'lr')` and then using that in `->where('lfr.id = lr.id')` (instead of `->where('lfr.id = log_revision.id')`).

The problem with `->where('lfr.id = log_revision.id')` arises if table prefixing is used, because then the table will not be called `log_revision`. It might be something else like `myprefix_log_revision`.

Drupal automatically figures this out for you if you're using standard static DB query methods, or if you wrap the table names in curly brackets in static queries.

paul121:
> It may be possible to limit this to the first revision with another subquery condition.

This looks good - except I think we need to adjust to update multiple, not only the 1st revision (`SELECT MIN(lr2.id)`). I need to review the DB data again but I think I saw some cases where there were _multiple_ initial, anonymous user revisions before a "good" one. Maybe related to #1021?

> This is actually OK with me, to be honest, because the migration was not performed by a particular user, so anonymous makes sense.

Yea, I suppose that makes sense. We just didn't have revision data to include in the migration. It would be nice if the revision history page included the `created` user to make things more clear!

mstenta:
Oops sorry @paul121 I didn't mean to close this. I just noticed that the branch was pushed to `origin` and I deleted it, didn't realize this PR was opened from that. I'm sure you meant to push it to your fork instead. If you want to update it for 4.x and reopen please do! I'll look to see if there's a way we can prevent accidentally pushing feature branches to `origin`, because I've made that mistake too...

PR #1007, opened 2025-10-16
Convert all procedural hook implementations to Hook classes

mstenta:
~Note: this PR builds on top of #1005 and #1006 , and therefore includes all of their commits. When those are merged I will rebase this so it only includes the relevant commits for these changes. In the meantime, to see this PR's changes by themselves, use this link: https://github.com/mstenta/farmOS/compare/4.x-autowire...4.x-hooks~ Edit: #1005 and #1006 have been merged.

This refactors all of farmOS's hook implementations from procedural functions to class methods that use the new `#[Hook('...')` attribute. See relevant Drupal core change records:

- https://www.drupal.org/node/3442349
- https://www.drupal.org/node/3486534

This removes all `*.module`, `*.views.inc`, and `*.views_execution.inc` files from our modules, which are where procedural hook implementations lived. All hook functions have been moved to `src/Hook/*.php` files which are autoloaded as needed, rather than being loaded in every request, so there is a performance benefit from this as well.

Another benefit, because these are now in classes, is we can use dependency injection (and autowiring) to avoid the use of `\Drupal::*` in our hook logic.

This also adds a Rector rule to ensure that all future hook implementations are added in classes, and not as procedural functions.

PR #1017, opened 2025-11-23
Fix page for adding quick form instances

mstenta:
During manual testing of #1005, #1006, and #1007 I discovered that the `/setup/quick/add` page was broken, with the following error in logs:

`Uncaught PHP Exception TypeError: "Drupal\\Component\\Utility\\Html::escape(): Argument #1 ($text) must be of type string, Drupal\\Core\\StringTranslation\\TranslatableMarkup given, called in /opt/repos/farmOS/modules/core/quick/src/Controller/QuickFormAddPage.php on line 63"`

This PR fixes it by calling `render()` on the quick form plugin label and description in that page before passing it to `Html::escape()`.

I also added some general tests of `/setup/quick/add` and `/setup/quick/add/[id]` to prevent future regressions.


mstenta:
> During manual testing of #1005, #1006, and #1007 I discovered

Just to clarify, this issue existed before those changes, so it was unrelated. I waited to open this as a follow-up.

symbioquine (approved):
LGTM

PR #1018, opened 2025-11-23
Show violation messages when validation fails in bulk action forms

mstenta:
farmOS provides a number of bulk action forms (eg: for flagging assets/logs, assigning owners, etc). All of these forms perform entity validation (`$entity->validate()`) and print a standard message if validation fails (eg: "Could not flag <a href=":entity_link">%entity_label</a>: validation failed."). But they do not provide any details on what failed, specifically.

This PR adds warning messages for each violation, so the user can see exactly what was invalid.

### DIFF
--- a/modules/core/owner/src/Form/AssignActionForm.php
+++ b/modules/core/owner/src/Form/AssignActionForm.php
@@ -188,6 +188,9 @@ class AssignActionForm extends ConfirmFormBase {
             ],
           ),
         );
+        foreach ($violations as $violation) {
+          $this->messenger()->addWarning($violation->getMessage());
+        }
         continue;
       }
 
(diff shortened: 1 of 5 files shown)


modules/core/owner/src/Form/AssignActionForm.php
```diff
@@ -188,6 +188,9 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
             ],
           ),
         );
+        foreach ($violations as $violation) {
+          $this->messenger()->addWarning($violation->getMessage());
```
symbioquine, line 192: Should these communicate which violation message goes with which entity - in case the user wants to track down which failed or why?

mstenta: I think we're covered by the existing code. Right above these new lines it already adds a generic warning with the entity label (and link). So these new lines are just adding additional warnings for each violation, which will be shown directly under the generic message.

mstenta:
(Just rebased this onto latest `4.x`. No changes.)

modules/core/owner/src/Form/AssignActionForm.php
symbioquine: Oh cool, I see that now :)

symbioquine (approved):
LGTM

paul121 (approved):

PR #1019, opened 2025-11-23
Do not use birth log label in child revision log message

mstenta:
I discovered this issue while doing manual testing for #1005, #1006, and #1007. This was unrelated to those changes, so I'm opening this as a follow-up fix.

This is related to logic we have for automatically maintaining a link between birth logs and the birthdate field on animal assets. If you create a birth log that references an animal, it will automatically update the birthday of the animal to match the timestamp of the log.

The issue occurs if you do not give the birth log a name (taking advantage of the `log` module's logic for automatically assigning log names). The following error occurs:

Uncaught PHP Exception TypeError:
"Drupal\\Component\\Utility\\Html::escape(): Argument #1 ($text) must be of type string, null given, called in
/opt/drupal/web/core/lib/Drupal/Component/Render/FormattableMarkup.php on line 238" at /opt/drupal/web/core/lib/Drupal/Component/Utility/Html.php line 433

This is because we were adding a link to the birth log in the animal asset's revision message, and passing in the blank log name (because it wasn't set yet).

This changes the logic so that we don't try to use the name of the log anymore, thereby sidestepping the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,6 +83,7 @@ requirements (inherited from Drupal 11):
 - [Do not set blank revision log messages #1029](https://github.com/farmOS/farmOS/pull/1029)
 - [Sort user selection alphabetically by name #1026](https://github.com/farmOS/farmOS/pull/1026)
 - [Sort "Active plans" by last updated #1027](https://github.com/farmOS/farmOS/pull/1027)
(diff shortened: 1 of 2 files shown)


symbioquine (approved):
LGTM

PR #1020, opened 2025-11-23
Soften farm_field module dependencies

mstenta:
I realized recently that the `farm_field` module's `FarmFieldFactory` class has an expectation that the `farm_log_asset` module is installed, because it references the `farm_asset_reference` View provided by that module:

https://github.com/farmOS/farmOS/blob/ae67d20d8c26d067d4223a1db2ac51090bf6ee31/modules/core/field/src/FarmFieldFactory.php#L356

However, `farm_log_asset` depends on `farm_field` already, because it uses `farm_field.factory` to add the `asset` field to `log` entities. So, there is an implicit circular dependency currently.

https://github.com/farmOS/farmOS/blob/ae67d20d8c26d067d4223a1db2ac51090bf6ee31/modules/core/log/modules/asset/farm_log_asset.info.yml#L9

A simple way to fix this, without major refactoring, is to accept that `farm_field` will have some knowledge of other modules (which it does), but instead of explicitly depending on them we can add checks to see if those modules are installed before making use of their features. Aka "soft dependencies" instead of "hard dependencies"...

As I was addressing this soft dependency on `farm_log_asset`, I realized we could also soften dependencies on some other modules as well. So this PR also removes the hard dependencies on `farm_map` and `fraction` from `farm_field`. It adds checks for these modules and other core modules that `farm_field` makes use of (and should otherwise have hard dependencies on, but didn't, like `file`, `image`, etc).

This means that modules that use `farm_field.factory` are responsible for ensuring that they depend on the modules that provide the field types they use, because `farm_field` will not ensure that.

One small thing this revealed is that we were depending on `farm_field` to install `farm_map`. So I moved that dependency to the `farm_location` module.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -84,6 +84,7 @@ requirements (inherited from Drupal 11):
 - [Sort user selection alphabetically by name #1026](https://github.com/farmOS/farmOS/pull/1026)
 - [Sort "Active plans" by last updated #1027](https://github.com/farmOS/farmOS/pull/1027)
 - [Do not use birth log label in child revision log message #1019](https://github.com/farmOS/farmOS/pull/1019)
+- [Soften farm_field module dependencies #1020](https://github.com/farmOS/farmOS/pull/1020)
 
 ### Deprecated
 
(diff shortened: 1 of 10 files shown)


symbioquine (approved):
LGTM

PR #1022, opened 2025-11-24
Add a Config Admin role for granting access to farmOS configuration

mstenta:
Originally proposed here: https://farmos.discourse.group/t/adding-a-settings-admin-role/2295

This adds a new "Config Admin" (`farm_config_admin`) role specifically for granting access to manage certain farmOS configuration settings.

The intention is similar to the existing "Account Admin" (`farm_account_admin`) - these roles are not enabled by default in farmOS, but a system administrator (with full Drupal user 1 access) can grant them to users in order to provide some higher-level permissions, without needing to grant full admin access.

Previously, we were starting to overload the "Account Admin" role with permissions to things unrelated to user accounts (like the `administer farm settings` permission).

This PR also shifts some of the "configuration" settings from the "Manager" role to the new "Config Admin" role. Therefore this is a breaking change because it removes some access from existing users.

It does this by splitting the `config_permissions` setting into `config_permissions` and `manager_permissions`. I took my best shot at splitting these in a way that makes sense, but curious to hear what others think.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ requirements (inherited from Drupal 11):
 - [Issue #3304608: Add an "abandoned" log status](https://www.drupal.org/project/farm/issues/3304608)
 - [Add default plan status options: "planning", "done", "abandoned" #986](https://github.com/farmOS/farmOS/pull/986)
 - [Add Term Merge module #961](https://github.com/farmOS/farmOS/pull/961)
+- [Add a Config Admin role for granting access to farmOS configuration #1022](https://github.com/farmOS/farmOS/pull/1022)
 - [Add support for attributes in farmOS plugin types #963](https://github.com/farmOS/farmOS/pull/963)
 - [Set asset/log flags via CSV importers #955](https://github.com/farmOS/farmOS/pull/955)
 - [Add a hook for excluding fields from CSV importers #958](https://github.com/farmOS/farmOS/pull/958)
--- a/docs/development/module/roles.md
+++ b/docs/development/module/roles.md
@@ -55,7 +55,7 @@ configuration):
 third_party_settings:
   farm_role:
     access:
-      config: true
+      manager: true
       entity:
         view all: true
         create all: true
--- a/docs/development/module/roles.md
+++ b/docs/development/module/roles.md
@@ -106,6 +106,8 @@ are provided by creating a `ManagedRolePermissions` plugin in the
 
 - `default_permissions`: A list of permissions that will be added to *all*
   managed roles.
+- `manager_permissions`: A list of permissions that will be added to managed
+  roles that have access to additional management features (`manager: true`).
 - `config_permissions`: A list of permissions that will be added to managed
   roles that have access to configuration (`config: true`).
 - `permission_callbacks`: A list of callbacks in controller notation that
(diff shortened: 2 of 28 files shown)


mstenta:
I pushed a small fix to grant the `access farm setup` permission to `manager` roles. It is *also* added to `config` roles.

mstenta:
Chatted with @paul121 about this this morning and we decided to add an update hook that will automatically grant the `farm_config_admin` role to all users with the `farm_account_admin` role. I didn't do this originally, because I thought that would be *adding* permissions to existing roles, but that's not actually the case.

The `farm_account_admin` role in farmOS 3.x already has `config: true`, which means it has all permissions granted by `config_permissions` already. These are removed from `farm_account_admin` in 4.x and granted to the `farm_config_admin` role. So granting the `farm_config_admin` role to existing `farm_account_admin` users is not granting more access than they already had. In fact, it is reducing them, because some `config_permissions` are moving to `manager_permissions`. And it provides administrators with a new option: they can remove the `farm_config_admin` role from users who don't need them, while still allowing them to manage users (via the `administer users` permissions that is kept).

mstenta:
> we decided to add an update hook that will automatically grant the `farm_config_admin` role to all users with the `farm_account_admin` role

This is a bit tricky because it needs to work alongside this existing change we made in 4.x: #982

#982 moved the `farm_account_admin` role from the old `farm_role_account_admin` module to a new `farm_account_admin` module. We added an update hook in `farm_role_account_admin` that uninstalls the old module, installs the new module, and ensures that the role is maintained on the users that had it.

We can't add a new update hook (for this PR) to the new `farm_account_admin` module, because it will never run on sites that are upgrading from 3.x. I'm not sure if we can add a *new* update hook to `farm_role_account_admin`, either... will it run both even if the first one uninstalls itself?

Perhaps we could just add this logic to the existing update hook in `farm_role_account_admin`, so it does *both* updates. Of course, this means that if anyone has already deployed farmOS 4.x, they won't get this new update logic, because the update hook will have already run. But that isn't the end of the world. It just means they would need to manually apply the `farm_config_admin` role to users that need it. Maybe that's the only option...

mstenta:
> Perhaps we could just add this logic to the existing update hook in `farm_role_account_admin`, so it does _both_ updates.

I have this drafted, but I currently have this branch built on top of #1024 in my fork, and they have some overlapping changes, so I'm going to wait until that is merged before updating this PR.

mstenta:
Here is the change to the update hook if you'd like to review it before it gets pushed here: https://github.com/farmOS/farmOS/commit/35f91d4770c146a873044f14f4da951e3c5bc58f

mstenta:
> I have this drafted, but I currently have this branch built on top of https://github.com/farmOS/farmOS/pull/1024 in my fork, and they have some overlapping changes, so I'm going to wait until that is merged before updating this PR.

Still waiting on #1024 but I decided to go ahead and push the rebased branch so that this PR reflects the current state, even though it also includes commits for #1024.

Use this link to see *only* the commits for this PR in the meantime: https://github.com/mstenta/farmOS/compare/4.x-merge-settings-into-setup...mstenta:farmOS:4.x-config-admin-wip

mstenta:
Rebased onto latest `4.x` commit (along with the `4.x` version of #1024). No changes.

paul121 (approved):

PR #1024, opened 2025-12-01
Merge farm_settings module into farm_setup

mstenta:
This merges the `farm_settings` module into `farm_setup`, uninstalls `farm_settings` via an update hook, and marks it as `obsolete`.

A companion branch against farmOS `4.x` does the same thing, but also removes the `farm_settings` module and the update hook (adding its name to `hook_removed_post_updates()` to ensure hosts upgrade to the latest 3.x release before 4.x - see #962): https://github.com/farmOS/farmOS/compare/4.x...mstenta:farmOS:4.x-merge-settings-into-setup

The motivation for doing this is twofold:

1. Right now the modules form is in `farm_settings`, but we are considering adding a "Setup Wizard", which would be in `farm_setup`, so it feels logical for them to be in the same module. We can't make `farm_setup` depend on `farm_settings` because that would create a circular dependency.
2. It feels more elegant to have one module. The distinction between the two is hard to describe, and sometimes hard to remember :laughing:. And the "settings form" feel like a subset of the "setup process/options".

This was briefly considered when the `farm_setup` module was first added in #706. See: https://github.com/farmOS/farmOS/pull/706#issuecomment-1690331639

Right now they are both pretty simple. `farm_setup` basically just provides the "Setup" main menu item, and `farm_settings` depends on `farm_setup` (for that menu item) and adds two forms: the settings form+menu item and the module installation form. So as it is, it's possible to have `farm_setup` installed without `farm_settings`, but it isn't possible to have `farm_settings` without `farm_setup`. They have separate permissions (`access farm setup` vs `administer farm settings`) - so it's already possible to set up users with the ability to access Setup without accessing Setup > Settings.

With all of that in mind, I don't see a strong reason to keep them separate, and merging them simplifies next steps for the setup wizard work.

PS: There is a second commit which simply moves the Gin-specific logic of adding the `form-checkboxes` class to the modules form over to the `farm_ui_theme` module, which is responsible for encapsulating all Gin-specific stuff.


mstenta:
I drafted a change record for this, which is referenced in the `lifecycle_link` of the obsolete `farm_settings` module: https://www.drupal.org/node/3559903

mstenta:
Chatted w/ @paul121 - we need to maintain the route names otherwise another module depending on that would break. We can maintain them in 3.x and change them in 4.x inside the `farm_setup` module.

mstenta:
I pushed a commit that reverts back to the original route names and links, to ensure that this is not a breaking change.

https://github.com/farmOS/farmOS/pull/1024/commits/8308123c706e9425a711e94569603d2c0dc2c824

The 4.x version of this will change them so that they are more consistent with the `farm_setup` namespacing.

paul121 (approved):
Maintaining these route/tasks names helps, but if existing modules depend on `farm_settings` and these are moving to `farm_setup`, then modules should also be updating their dependencies? I wish we could just wait to make any changes until 4.x?😅 

But if we need to now then I think this is fine, we've considered the breaking changes I can think of

mstenta:

> but if existing modules depend on `farm_settings` and these are moving to `farm_setup`, then modules should also be updating their dependencies?

Yes, they will need to update their dependencies when they declare support for v4. They don't need to update for v3. The `farm_settings` module is marked as `obsolete` by this PR, but it is not removed, and it still depends on `farm_settings`, so there is no risk to downstream modules that depend on it.

I think this is a pretty low-impact deprecation, so I think it's worth doing now. It allows us to completely remove `farm_settings` in 4.x.

mstenta:
Merged into 3.x, and published change record: https://www.drupal.org/node/3559903

mstenta:
I've merge the `4.x-merge-settings-into-setup` branch into `4.x`. This includes all of the changes from this PR, and also removes the `farm_settings` module and `3.x` update hook.

PR #1025, opened 2025-12-01
Display warning when user attempts to navigate away from form with unsaved changes

ktohalloran:
Adds a `farm_form` module that includes a JS library for displaying an alert whenever a user tries to leave a form with unsaved changes.

This approach is based on some of the elements used by the [node edit protection module](https://git.drupalcode.org/project/node_edit_protection) and combined with the addition of `farm_form` that @mstenta had already started. The discovery process involved a lot of ways of figuring out how _not_ to implement this. Ultimately, the trick here has been trying to create something that works only for a specific subset of forms of varying types (such as quick forms and entity forms, and ideally also forms within forms like the entity browser; this solution works for the first two, but not the latter), regardless of the impact of other JS libraries (such as the one added by `farmOS-map`) and Drupal's own form handling. In short, there were a lot of gotchas here that Mike and I discovered during implementation.

Anyway, there are some definite improvements that can be made, but it's in a good place to be put up as a PR for discussion.

### DIFF
--- a/modules/core/form/js/form_protection.js
+++ b/modules/core/form/js/form_protection.js
@@ -1,0 +1,56 @@
+/**
+ * @file
+ * Show warning when a user is about to navigate away from an unsaved form.
+ */
+(function ($, Drupal) {
+
+  'use strict'
+
+  // Track the initial state of forms.
+  const initialFormState = {};
+
+  // Track whether a submit was clicked "globally" over all attached DOM, so we
+  // can ignore those "onbeforeunload" events and allow form submission.
+  let submitWasClicked = false;
+
+  // Check to see if a protected form input has changed.
+  const protectedFormInputHasChanged = function() {
+    return $('form.form-protected :input').is(function(index, element) {
+      if (!element.id || !Object.hasOwn(initialFormState, element.id)) {
+        return false;
+      }
+      return $(this).serialize() !== initialFormState[element.id];
+    });
+  };
+
+  // Define the Drupal form protection behavior.
+  Drupal.behaviors.form_protection = {
+    attach: function (context) {
+
+      // Save the initial form state of all input elements.
+      $('form.form-protected :input', context).each(function () {
+        if (this.id) {
+          initialFormState[this.id] = $(this).serialize();
+        }
+      });
+
+      // Tell onbeforeunload to allow the "submit" event through.
+      $('form.form-protected').on('submit', () => {
+        submitWasClicked = true;
+      });
+
+      // Handle navigation, back button, exit etc.
+      window.onbeforeunload = function () {
+
+        // If the form has been changed, and a form submission has not occurred,
+        // warn the user that they will lose unsaved changes.
+        if (protectedFormInputHasChanged() && !submitWasClicked) {
+
+          // For very old browsers show custom text (modern browsers will ignore
+          // this text).
+          return Drupal.t('You have unsaved changes.');
+        }
+      }
+    }
+  };
+})(jQuery, Drupal);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,7 @@ requirements (inherited from Drupal 11):
 - [Add map layer for "Other Location" assets #966](https://github.com/farmOS/farmOS/pull/966)
 - [Add ability to assign plan ownership #1015](https://github.com/farmOS/farmOS/pull/1015)
 - [Show violation messages when validation fails in bulk action forms #1018](https://github.com/farmOS/farmOS/pull/1018)
+- [Add warning when navigating away from forms with unsaved changes #1025](https://github.com/farmOS/farmOS/pull/1025)
 - [Document how to use symfony/mailer for email SMTP relay #844](https://github.com/farmOS/farmOS/pull/844)
 - [Document setup of local Git repositories #1032](https://github.com/farmOS/farmOS/pull/1032)
 
--- a/modules/core/form/config/install/farm_form.settings.yml
+++ b/modules/core/form/config/install/farm_form.settings.yml
@@ -1,0 +1,1 @@
+enable_form_protection: true
--- a/modules/core/form/config/schema/farm_form.schema.yml
+++ b/modules/core/form/config/schema/farm_form.schema.yml
@@ -1,0 +1,7 @@
+farm_form.settings:
+  type: config_object
+  label: 'farmOS Form Settings'
+  mapping:
+    enable_form_protection:
+      type: boolean
+      label: 'Display warning when users attempt to navigate away from forms with unsaved changes.'
--- a/modules/core/form/farm_form.info.yml
+++ b/modules/core/form/farm_form.info.yml
@@ -1,0 +1,5 @@
+name: farmOS Forms
+description: Provides features for farmOS forms.
+type: module
+package: farmOS
+core_version_requirement: ^11
--- a/modules/core/form/farm_form.libraries.yml
+++ b/modules/core/form/farm_form.libraries.yml
@@ -1,0 +1,5 @@
+form_protection:
+  js:
+    js/form_protection.js: { }
+  dependencies:
+    - core/jquery
--- a/modules/core/form/farm_form.links.task.yml
+++ b/modules/core/form/farm_form.links.task.yml
@@ -1,0 +1,4 @@
+farm_form.settings:
+  base_route: farm.setup.settings
+  route_name: farm_form.settings
+  title: 'Forms'
--- a/modules/core/form/farm_form.managed_role_permissions.yml
+++ b/modules/core/form/farm_form.managed_role_permissions.yml
@@ -1,0 +1,3 @@
+farm_form:
+  config_permissions:
+    - administer farm forms
--- a/modules/core/form/farm_form.permissions.yml
+++ b/modules/core/form/farm_form.permissions.yml
@@ -1,0 +1,2 @@
+administer farm forms:
+  title: 'Administer farmOS form settings'
--- a/modules/core/form/farm_form.routing.yml
+++ b/modules/core/form/farm_form.routing.yml
@@ -1,0 +1,7 @@
+farm_form.settings:
+  path: 'farm/settings/forms'
+  defaults:
+    _form: '\Drupal\farm_form\Form\FormSettingsForm'
+    _title: 'Form settings'
+  requirements:
+    _permission: 'administer farm forms'
--- a/modules/core/form/src/Form/FormSettingsForm.php
+++ b/modules/core/form/src/Form/FormSettingsForm.php
@@ -1,0 +1,65 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Drupal\farm_form\Form;
+
+use Drupal\Core\DependencyInjection\AutowireTrait;
+use Drupal\Core\Form\ConfigFormBase;
+use Drupal\Core\Form\FormStateInterface;
+
+/**
+ * Provides settings form for the farm_form module.
+ */
+class FormSettingsForm extends ConfigFormbase {
+
+  use AutowireTrait;
+
+  /**
+   * Config settings.
+   *
+   * @var string
+   */
+  const SETTINGS = 'farm_form.settings';
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getFormId() {
+    return 'farm_form_settings';
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function getEditableConfigNames() {
+    return [
+      static::SETTINGS,
+    ];
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildForm(array $form, FormStateinterface $form_state) {
+    $config = $this->config(static::SETTINGS);
+    $form['enable_form_protection'] = [
+      '#type' => 'checkbox',
+      '#title' => $this->t('Enable form protection'),
+      '#description' => $this->t('Display a warning when users attempt to navigate away from forms with unsaved changes.'),
+      '#default_value' => $config->get('enable_form_protection'),
+    ];
+    return parent::buildForm($form, $form_state);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function submitForm(array &$form, FormStateInterface $form_state) {
+    $this->configFactory->getEditable(static::SETTINGS)
+      ->set('enable_form_protection', $form_state->getValue('enable_form_protection'))
+      ->save();
+    parent::submitForm($form, $form_state);
+  }
+
+}
--- a/modules/core/form/src/Traits/FarmFormProtectionTrait.php
+++ b/modules/core/form/src/Traits/FarmFormProtectionTrait.php
@@ -1,0 +1,39 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Drupal\farm_form\Traits;
+
+/**
+ * Provides a standard method for enabling form protection.
+ */
+trait FarmFormProtectionTrait {
+
+  /**
+   * This method is provided by \Drupal\Core\Form\FormBase.
+   *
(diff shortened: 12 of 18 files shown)


mstenta:
Thanks @ktohalloran!!

FYI: the failing PHPStan tests are due to an upstream change - unrelated to this PR. I've fixed them in `4.x` so this should pass once it is rebased.

> combined with the addition of `farm_form` that @mstenta had already started

For what it's worth, I started this a while back in a branch where I am exploring adding some additional custom form elements for farmOS (eg: an improved entity reference selection widget and a combined date+done widget for logs). Those aren't ready, but adding this `farm_form` module now will give us a place for those kinds of things in the future.

I'm curious to see what @symbioquine @paul121 and others think of the approach that this PR takes to detecting "dirty" forms. It was more complicated than expected, as @ktohalloran mentioned. :-)

mstenta:
Echoing a comment from @paul121 in chat:

> enabling by default in most places seems like it would be OK - but maybe make it configurable ;-)

I'm imagining there could be situations where this feature causes more harm than good, so making it configurable is a good idea.

Perhaps a simple form at `/farm/settings/form` with a single checkbox to turn on/off the behavior (in entity forms and quick forms) would be a sensible first step. In the future, perhaps this gets expanded to provide more nuanced configuration, but that can come later...

modules/core/form/js/form_protection.js
```diff
@@ -0,0 +1,76 @@
...
+
+        checkForUnsavedChanges: function () {
+            let behavior = Drupal.behaviors.form_protection;
+            let formIsDirty = false
+
+            $("form.protected :input").each(function () {
+                const elId = $(this).attr("id")
+
+                // Check new state against initial state if available; if it's not, move on.
+                if (elId && Object.keys(behavior.initialFormState).includes(elId)) {
```
symbioquine: Maybe this falls into the category of premature optimization, but I shy away from creating a new copy of the array of all the initial form state keys for every input element to be checked...

```suggestion
                if (elId && Object.hasOwn(behavior.initialFormState, elId)) {

```

ktohalloran:
After reviewing with @symbioquine in the dev call yesterday, I've pushed up commits to clean up the Javascript and use a more unique class name.

modules/core/form/js/form_protection.js
```diff
@@ -0,0 +1,76 @@
...
+                // means return false).
+                if (formIsDirty) {
+                    return false;
+                }
+            })
+            return formIsDirty;
+        },
+
+    }
+})(jQuery, Drupal)
```
symbioquine: I kind of went down a rabbit hole trying to figure out all the quirks of how browsers handle this stuff...

I ended up with a much shorter version that I think will accomplish the same thing - perhaps with fewer gotchas. (which I've tried to capture in my other comments)

(No criticism is intended in the rewrite, just that by the time I'd worked out what I think is going on I also had mostly arrived at this shorter version.)

```suggestion
/**
 * @file
 * Show warning when a user is about to navigate away from an unsaved, dirty form.
 */
(function ($, Drupal) {

    "use strict"

    const initialFormState = {};

    // Track whether submit was clicked "globally" over all attached DOM so we can ignore those
    // causes of window.onbeforeunload firing and not prevent form submission
    let submitWasClicked = false;

    const doesAnyProtectedFormContainChanges = function() {
        const anyFormElementChanged = $("form.protected :input").is(function(i, el) {
            if (!el.id or !Object.hasOwn(initialFormState, el.id)) {
                return false;
            }

            const formElementChanged = $(this).serialize() !== initialFormState[elId];

            return formElementChanged;
        });

        return anyFormElementChanged;
    };

    Drupal.behaviors.form_protection = {

        attach: function (context) {
            // Save the initial form state of any input elements in the attached DOM
            $("form.protected :input", context).each(function () {
                if (this.id) {
                  initialFormState[this.id] = $(this).serialize();
                }
            });

            // Let all form submit buttons through.
            $("input[type='submit'], button[type='submit']", context).click(function() {
                submitWasClicked = true;
            });

            // Handle navigation, backbutton, exit etc.
            window.onbeforeunload = function () {
                // If the form contains changes and this unload event was not caused by form submission
                // cause the browser to prompt the user whether they want to leave the page.
                if (doesAnyProtectedFormContainChanges() && !submitWasClicked) {
                    // For very old browsers show custom text (modern browsers do not allow customizing the text).
                    return Drupal.t("You have unsaved changes.");
                }

                // No return here since that is the preferred way to not prompt
            }
        }

    };
})(jQuery, Drupal);

```

modules/core/form/js/form_protection.js
```diff
@@ -0,0 +1,76 @@
...
+                .each(function() {
+                    $(this).click(function() {
+                        formIsDirty = behavior.checkForUnsavedChanges();
+                        // If the user clicked on a # link or the form is not dirty, let click through; otherwise,
+                        // return something other than undefined to trigger onbeforeunload.
+                        if (formIsDirty && $(this).attr("href") !== "#") {
+                            return 0;
+                        }
+                    });
+                });
```
symbioquine: I'm wondering if this is actually needed. Why do we need to maintain the `formIsDirty` state in this manner when it seems like we can just check for changes in the `window.onbeforeunload` handler?

modules/core/form/js/form_protection.js
```diff
@@ -0,0 +1,76 @@
...
+                behavior.initialFormState[$(this).attr("id")] = $(this).serialize();
+            });
+
+            // Let all form submit buttons through.
+            $("input[type='submit'], button[type='submit']").each(function() {
+                $(this).addClass("allow-submit");
+                $(this).click(function() {
+                    allowClick = true;
+                });
+            });
```
symbioquine: * I suspect adding the `allow-submit` class is not needed
* The tracking of the `allowClick` state may be needed since it appears to be hard to determine if a submit button was clicked from within the `window.onbeforeunload` handler
* `$('#some-selector').each(function() { $(this).click(someHandler) })` can be written as `$('#some-selector').click(someHandler)` - see https://api.jquery.com/each/
  > image

modules/core/form/js/form_protection.js
```diff
@@ -0,0 +1,76 @@
...
+
+    "use strict";
+
+    Drupal.behaviors.form_protection = {
+
+        initialFormState: {},
+
+        attach: function (context) {
+            const behavior = Drupal.behaviors.form_protection;
+            let allowClick = false;
```
symbioquine: * I think we can improve the naming of `allowClick` to better communicate what state is actually being tracked - i.e. something like `submitWasClicked`
* I think this state should be scoped to the top-level IIFE - otherwise I think you'd get weird behavior with multiple forms on the same page where only the most recent attached form (or form elements) would receive the protection

modules/core/form/js/form_protection.js
```diff
@@ -0,0 +1,76 @@
...
+ * @file
+ * Show warning when a user is about to navigate away from an unsaved, dirty form.
+ */
+(function ($, Drupal) {
+
+    "use strict";
+
+    Drupal.behaviors.form_protection = {
+
+        initialFormState: {},
```
symbioquine: We can move `initialFormState` up to the scope of the [IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE) and avoid the need for the `behavior.initialFormState` references.

```suggestion

    const initialFormState = {};

    Drupal.behaviors.form_protection = {

```

modules/core/form/js/form_protection.js
```diff
@@ -0,0 +1,76 @@
...
+                        if (formIsDirty && $(this).attr("href") !== "#") {
+                            return 0;
+                        }
+                    });
+                });
+
+            // Handle backbutton, exit etc.
+            window.onbeforeunload = function () {
+                if (formIsDirty && !allowClick) {
+                    allowClick = false;
```
symbioquine: There's no point in setting `allowClick` to `false` here since we can't get into this block unless it is already `false` right?

modules/core/form/js/form_protection.js
```diff
@@ -0,0 +1,76 @@
...
+                    return (Drupal.t("You have unsaved changes."));
```
symbioquine: Probably worth noting that modern browsers don't seem to allow custom messages so this is usually equivalent to returning `true` (though I haven't tested doing so). https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event#browser_compatibility

modules/core/form/js/form_protection.js
```diff
@@ -0,0 +1,76 @@
...
+                }
+            }
+        },
+
+        checkForUnsavedChanges: function () {
+            const behavior = Drupal.behaviors.form_protection;
+            let formIsDirty = false;
```
symbioquine: * I think we can avoid the `formIsDirty` variable being mutable and the need to break out of the loop by using jQuery's `is` instead of `each`: https://api.jquery.com/is/ (basically akin to [Array::some](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some)).
* No need for the `const behavior = Drupal.behaviors.form_protection;` pattern as noted above

modules/core/form/js/form_protection.js
```diff
@@ -0,0 +1,76 @@
...
+
+            $("form.entity-or-quick-form-protected :input").each(function () {
+                const elId = $(this).attr("id");
```
symbioquine: `$(this).attr("id");` can be written `this.id`

modules/core/form/js/form_protection.js
symbioquine: Worth also noting, I also haven't tested this.

ktohalloran:
Thanks for reviewing @symbioquine!I tested out your changes locally; there were a couple of changes to make but otherwise it works almost as well and is a lot more readable so I appreciate the refactor!

That said, there's one new behavior this rewrite adds that I'd be curious to get your (or whoever's) thoughts on. My version didn't work with the entity browser in that if you went into a form, only made a change via an entity browser and then tried to exit, it wouldn't stop you, but if you had made another change to another field as well as used the entity browser, it would (in other words, changing something via the entity browser didn't look like a change from initial form state but any other change did). This version allows you to exit the form without triggering an alert _any time_ you use the entity browser, so the alert doesn't trigger regardless of what else you did before using the entity browser. @mstenta suggested fixing this in the initial version via the addition of `context` to the logic for getting initial form state because Drupal attaching a form via Ajax was triggering the "initial load" logic to run again. That worked in that version, but doesn't seem to be doing the same thing here. Do you have any thoughts on how we might fix this?

symbioquine:
> Thanks for reviewing @symbioquine!I tested out your changes locally; there were a couple of changes to make but otherwise it works almost as well and is a lot more readable so I appreciate the refactor!

Glad to help!

> This version allows you to exit the form without triggering an alert _any time_ you use the entity browser, so the alert doesn't trigger regardless of what else you did before using the entity browser. @mstenta suggested fixing this in the initial version via the addition of `context` to the logic for getting initial form state because Drupal attaching a form via Ajax was triggering the "initial load" logic to run again. That worked in that version, but doesn't seem to be doing the same thing here. Do you have any thoughts on how we might fix this?

I don't have a great hypothesis for why that would be but it sounds like a deal-breaker.

Maybe one test to conduct would be adding a `console.log` statement to the top-level IIFE (e.g. after the `"use strict"`) to see if that's running more than once and another inside the `attach` method to see when that's happening. Maybe information about when those are running and what the `context` is each time will suggest further lines of investigation...

ktohalloran:
@symbioquine I figured out why the refactor was allowing changes through any time the entity browser was involved. It was because the "Select assets" button that brings up the entity browser modal is a submit button and therefore was triggering the change to `submitWasClicked`. Then, regardless of what changes had been made or were made moving forward, that flag was already set to true, so the alert was never triggered. I've updated it to only trigger that update when the submit button is not entity browser specific. So this should be back to matching the behavior of the old version.

mstenta:
> It was because the "Select assets" button that brings up the entity browser modal is a submit button and therefore was triggering the change to `submitWasClicked`.
> ...
> I've updated it to only trigger that update when the submit button is not entity browser specific.

Hmm. It looks like the logic is basically saying "if a submit button was clicked anywhere on the page, ignore form protection logic" - is that right?

This will still cause issues... there are definitely going to be more cases where a submit button is used for Ajax functionality (not submitting the form)... for example anywhere we have an "Add another item" button in a form.

Examples:

- Asset form Parent field.
- Asset form ID Tag field.
- Animal Asset Nicknames field.
- Log Quantity field.
- ... probably more

But also if any module adds their own, or provides a quick form that uses buttons like this (there are a few in the RCD module, which is the original motivation for this PR)...

I'm not sure how to work around that... but it won't make sense to add/maintain a list of exceptions for every button. :thinking:

symbioquine:
> Hmm. It looks like the logic is basically saying "if a submit button was clicked anywhere on the page, ignore form protection logic" - is that right?
> 
> This will still cause issues...

I was also thinking the logic may be flawed in terms of a `click` event not being the only thing that causes forms to get submitted. They can also be submitted via the enter key in certain fields, via javascript, and maybe other ways.

Maybe we should try changing the logic to bind to the [`submit`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/submit_event) event on the form(s) and see if that resolves both issues?

ktohalloran:
@mstenta @symbioquine good points, thanks for the suggested fixes. I'll give that a try today.

ktohalloran:
@symbioquine that did seem to fix the issue! I tested this with quick forms, entity forms, and the entity browser's modal trigger submit button and it all worked as it should. Great suggestion!

symbioquine (dismissed):
Needs a changelog entry, but otherwise looks great!

ktohalloran:
Changelog has been updated 👍

mstenta:
Thanks again @ktohalloran and @symbioquine for working on this!

I tested myself locally and it seems to be working well! There are a few minor cases where it misses changes and allows navigation, but I think that's fine for now. This is a progressive improvement, and it's OK if it isn't perfect. We just want to catch most common changes to help prevent lost work. These are the ones that I noticed did not register changes to the form state:

- Adding/removing entities via `entity_browser` (as @ktohalloran noted above - it sounded like this was fixed, but seems to still be broken).
- Removing an entity autocomplete item (eg: edit a plant asset, click "Remove" next to the plant type, then navigate away).
- Uploading a file to an image/file attachment field.

So maybe those are things we can improve on in the future, but I don't think they need to block this initial implementation.

I rebased @ktohalloran's branch onto `4.x` along with the `4.x-make-form-protection-optional` (for adding a settings form to allow disabling protection), squashed everything into a simpler set of logical commits, and force pushed to this PR.

Here are the changes I made in the process:

I took a quick pass through the JS code and made a few cleanup changes for general coding/comment standardization (we don't really have linting for JS right now, but that would be worth adding to farmOS CI/CD in the future!). No functional changes.

I changed the CSS class for protected forms from `entity-or-quick-form-protected` to simply `form-protected`. Originally it was just `protected`, and @symbioquine recommended making it more specific, but I think `entity-or-quick-form-protected` was too specific. It is possible for it to be used by other types of forms as well. I think `form-protected` will work, and there's a low risk of collision with other code/class usage.

I also realized there's a better way to apply the protection to entity forms. Rather than using multiple `hook_form_alter()`s (in `src/Hook/FormHooks.php`) we can do it once in `src/Form/GinContentFormBase.php`, which we use as the content form for all of our main entity types. This covers all of the entity types originally handled with `hook_form_alter()`s (`asset`, `log`, `organization`, `plan`), but also covers `taxonomy_term`s.

I encapsulated the code for enabling form protection into a `trait FarmFormProtectionTrait` provided by the `farm_form` module. This makes it easier for downstream code to apply the form protection consistently, and it gives us the flexibility to change the class name or library attachment behavior in the future, without requiring all downstream code to change as well.

Lastly, I granted the new `administer farm forms` permission to `config` roles (see #1022).

PR #1028, opened 2025-12-09
Update drupal/simple_oauth to ^6.1

mstenta:
Opening this draft PR as a follow-up to https://github.com/farmOS/farmOS/commit/735479980db72dd03bcc43bf81da48541197919b.

Todo: Investigate why `drupal/simple_oauth` `6.1.0` causes tests to fail.

See https://www.drupal.org/project/simple_oauth/issues/3507450

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,6 +100,7 @@ requirements (inherited from Drupal 11):
 - [Workers cannot update or delete taxonomy terms #1033](https://github.com/farmOS/farmOS/pull/1033)
 - [Allow birth quick form to create up to 20 children #1030](https://github.com/farmOS/farmOS/pull/1030)
 - [Always consider revision translations affected #1041](https://github.com/farmOS/farmOS/pull/1041)
+- [Update drupal/simple_oauth to ^6.1 #1028](https://github.com/farmOS/farmOS/pull/1028)
 
 ### Deprecated
 
--- a/composer.json
+++ b/composer.json
@@ -39,7 +39,7 @@
         "drupal/migrate_source_csv": "^3.6",
         "drupal/migrate_tools": "^6.1.2",
         "drupal/role_delegation": "^1.3",
-        "drupal/simple_oauth": "6.0.9",
+        "drupal/simple_oauth": "^6.1",
         "drupal/simple_oauth_password_grant": "^2.1",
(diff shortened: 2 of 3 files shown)


mstenta:
I think we actually need to consider this a blocker for 4.0.0 release, because we did not pin `drupal/simple_oauth` to `6.0.9` in farmOS `3.x` (only in `4.x`), which means it's possible that some deployments may have already updated to `6.1.0`. I'm not sure if there would be any issues with downgrading back to `6.0.9` during an upgrade to farmOS v4, but it would be best to avoid that IMO.

FWIW, `6.1.0` does *not* cause our tests to fail on the `3.x` branch, so I suspect that the failures are related to https://github.com/farmOS/farmOS/pull/922, which was merged into `4.x`.

mstenta:
I investigated this and found (and fixed) two things:

1. The new version of `drupal/simple_oauth` introduced a breaking change by adding a `getPermissions()` method to the `ScopeGranularityInterface`. Our custom scope granularity in the `farm_role` module did not implement this method, which broke our managed role permissions scope granularity. I added a commit to this PR which implements that method, and that fixes the failing tests.

More info here: https://www.drupal.org/project/simple_oauth/issues/3507450#comment-16454893

2. This one is a bit more mysterious... one of our tests was failing with the following infinite recursion error:

```
There was 1 error:

1) Drupal\Tests\farm_role\Kernel\FarmApiOauthRoleTest::testAllowedApiResources
PHPUnit\Framework\Exception: Uncaught PHP Exception Error: "Maximum call stack size of 8339456 bytes (zend.max_allowed_stack_size - zend.reserved_stack_size) reached. Infinite recursion?" at /opt/drupal/vendor/symfony/dependency-injection/Container.php line 101
```

I ended up finding that a change to the `Oauth2AccessPolicy` class in the `simple_oauth` module fixes it. So I opened an upstream issue to propose making that change, and included a patch in this PR in the meantime.

Here is the upstream issue, with more info: https://www.drupal.org/project/simple_oauth/issues/3572695

mstenta:
I went ahead and merged this. Our tests are passing and I don't thing we need to block `4.0.0` while we wait to hear back from the maintainers of `simple_oauth` in https://www.drupal.org/project/simple_oauth/issues/3572695. If we need to make adjustments at that point we can.

PR #1030, opened 2025-12-14
Allow birth quick form to create up to 20 children

botlfarm:
Allow more than 15 children in quick brith form

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -98,6 +98,7 @@ requirements (inherited from Drupal 11):
 - [Soften farm_field module dependencies #1020](https://github.com/farmOS/farmOS/pull/1020)
 - [Workers can only delete their own records #1033](https://github.com/farmOS/farmOS/pull/1033)
 - [Workers cannot update or delete taxonomy terms #1033](https://github.com/farmOS/farmOS/pull/1033)
+- [Allow birth quick form to create up to 20 children #1030](https://github.com/farmOS/farmOS/pull/1030)
 
 ### Deprecated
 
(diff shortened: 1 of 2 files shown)


symbioquine (approved):
LGTM

symbioquine:
> LGTM

With a Changelog entry, obviously.

mstenta:
Thanks @botlfarm!

Got me thinking... should we increase this even more? I was curious and found this:

> The largest recorded litter of piglets was 37, born to a sow on September 21, 1993, with 33 of those piglets surviving. This record was set at Eastfield House Farm in Melbourne, York, UK.

I only looked into piglets, and there may be other cases that need to support even more, but for now what do you think about increasing it to 40?

I think the only tradeoff is that the dropdown gets longer. We could consider changing it to a number field in the future, but we'd have to test if that affects how the Ajax works. So for now just changing the number is easier. :-)

> With a Changelog entry, obviously.

I can handle this when I merge...

botlfarm:
> I think the only tradeoff is that the dropdown gets longer. We could consider changing it to a number field in the future, but we'd have to test if that affects how the Ajax works. So for now just changing the number is easier. :-)

I wonder if long term it should be a "add new child" button at the bottom of the form. Similar to how the grazing plan model is being developed by the students. this way you can and as many as you want without cluttering it.

mstenta:
> I wonder if long term it should be a "add new child" button at the bottom of the form.

Agreed that would be nice.

mstenta:
Added a `CHANGELOG.md` line. I'm not going to worry about increasing this any more right now. I think @botlfarm's idea of changing it to an "Add another" button makes more sense as the next step.

mstenta:
Merged! Thanks again @botlfarm!

PR #1033, opened 2025-12-20
Limit permissions of worker role

mstenta:
I propose that remove some of the permissions currently assigned to the `farm_worker` role.

Specifically:

- Remove the ability to delete records that were created by others.
- Remove the ability to update or delete taxonomy terms.

The motivations for doing this include:

- Limit potential damage by malicious users. Generally speaking, a user with the `farm_worker` role doesn't need to be able to delete data that is not relevant to them. An exception is made here for records that the user has created themselves. So not all risk is removed, but it's a step in the right direction (perhaps the idea of "soft deletion" will be implemented, removing the need for any `delete` permissions).
- Only roles with `config: true` (eg: `farm_manager`) have `access taxonomy overview`, under the assumption that they are in charge of taxonomies. Users with the `farm_worker` role will still have the ability to create new terms (eg: via autocomplete textfields) but this removes the ability to update or delete terms. Terms don't have the same granularity of permissions as other records, so we can't differentiate between `any` and `own`, so this takes a conservative approach and further solidifies the responsibility of taxonomy maintenance to managers (and limiting workers).
- Further differentiate the `farm_worker` and `farm_manager` roles.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -87,6 +87,7 @@ requirements (inherited from Drupal 11):
 - [Do not use birth log label in child revision log message #1019](https://github.com/farmOS/farmOS/pull/1019)
 - [Soften farm_field module dependencies #1020](https://github.com/farmOS/farmOS/pull/1020)
 - [Workers can only delete their own records #1033](https://github.com/farmOS/farmOS/pull/1033)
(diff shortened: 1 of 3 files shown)


symbioquine (approved):
LGTM

mstenta:
Rebased onto latest `4.x` commit. No changes.

paul121 (approved):

PR #1034, opened 2025-12-23
Remove farm_install_modules() installation task

mstenta:
**Note: This PR is built on top of both #1024 and #1022, and therefore includes all of their commits as well. When those are merged, this will be rebased and updated to only include the commits from this branch. In the meantime, this link can be used to review *only* this branch: https://github.com/mstenta/farmOS/compare/4.x-config-admin...mstenta:farmOS:4.x-remove-modules-install-task**

See original drupal.org thread for motivation and details: https://www.drupal.org/project/farm/issues/3494872

This is a breaking change because it removes the ability to install `base`, `default`, or `all` modules via `drush site-install`. This will be replaced by a new setup wizard in a followup PR.

After these changes, only the following "base" modules will be installed during farmOS installation:

- `farm_login`
- `farm_setup`
- `farm_ui`
- `farm_update`

This means that after installing farmOS, you are dropped into a very barebones instance, without any asset types, log types, or other features. The dashboard and primary UI elements all work. So this PR is a big step backwards in terms of initial user experience. But the plan is to immediately follow this with a new "setup wizard" feature that walks the new user through installing modules that are relevant to them. This is ready for review as well, but I will open it as a separate pull request.

For testing purposes, this adds a new Drush command `drush farm_test:modules` for installing all farmOS modules. This works better than the previous `farm.modules=all` approach, which only installed modules included in `farm_modules()` (a more limited and manually maintained list that was used in the UI form). Now it automatically loads all modules that are included in the following packages:

- farmOS
- farmOS Taxonomies
- farmOS Assets
- farmOS Logs
- farmOS Quantities
- farmOS Organizations
- farmOS Defaults
- farmOS Roles
- farmOS Maps
- farmOS UI
- farmOS Quick Forms
- farmOS (Experimental)
- farmOS (Legacy)

A few related cleanup commits are also included. Happy to answer questions if any of them are unclear.


mstenta:
Rebased onto latest `4.x` commit (along with #1022 and the `4.x` version of #1024). No changes.

paul121 (approved):

PR #1035, opened 2025-12-23
Add a farmOS setup wizard

mstenta:
This adds a new "farmOS Setup Wizard" framework, which provides the ability for farmOS modules to add forms to a multi-step setup process. This is shown in a block on the dashboard after a fresh install, visible to admin (user 1) and users with the `farm_config_admin` role (from #1022). The forms are also available directly via routes in the `/setup/wizard/*` path.

This is intended to replace the module installation step that was removed in #1034, as well as to provide a new place to put setup wizard forms moving forward.

Setup wizard forms are plugins provided by modules, so it is possible for modules to provide their own steps. In the future, we can add more steps to core for things like initial farm mapping, setting up taxonomies, importing records, etc (these are all ideas that have come up during discussion).

This branch starts with three steps:

- Welcome - simply welcomes the new user to farmOS and introduces them to the setup process.
- Modules - guides the user through installing modules based on their need, by asking questions and installing relevant modules
- Resources - shows a bunch of links to the user guide, forum, etc as the last step

The commit history should describe each step of the development clearly, but happy to answer any questions!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ requirements (inherited from Drupal 11):
 
 ### Added
 
+- [Add a farmOS setup wizard #1035](https://github.com/farmOS/farmOS/pull/1035)
 - [Add an Organization entity type with a Farm bundle #849](https://github.com/farmOS/farmOS/pull/849)
 - [Allow API access to organizations #1010](https://github.com/farmOS/farmOS/pull/1010)
 - [Issue #3304608: Add an "abandoned" log status](https://www.drupal.org/project/farm/issues/3304608)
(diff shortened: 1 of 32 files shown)


mstenta:
Rebased onto latest `4.x` commit (along with #1034 , #1022, and the `4.x` version of #1024). No changes.

mstenta:
Added a `CHANGELOG.md` line for this. I put it at the top of the `Added` items, since this feels like a nice one to highlight for 4.0.0.

mstenta:
Pushed small a fix for the failing test ... just a typo (`farm_standard_quantity` should have been `farm_quantity_standard`).

PR #1048, opened 2026-02-17
Validate entity bundle in asset/log/etc page_type display logic

mstenta:
Fixes https://www.drupal.org/project/farm/issues/3249201

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - [Fix missing action links on entity collection pages #1050](https://github.com/farmOS/farmOS/pull/1050)
 - [Fix asset and plan update hooks for prefixed database tables #1051](https://github.com/farmOS/farmOS/pull/1051)
+- [Validate entity bundle in asset/log/etc page_type display logic #1048](https://github.com/farmOS/farmOS/pull/1048)
 
 ## [4.0.0-beta2] 2026-02-12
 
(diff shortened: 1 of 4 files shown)


symbioquine (approved):
No test for the GeoJSON part?

Otherwise, LGTM!

mstenta:
> No test for the GeoJSON part?

Yeaaaa... not as trivial. Although maybe there's an easy approach...

The GeoJSON logic is in a different module (`farm_ui_map`), which unfortunately doesn't have any tests currently. The tests I added in this PR were easy to tack on to the existing `farm_ui_views` functional tests, which already test the asset/log Views.

Maybe it's worth taking the time to add tests to `farm_ui_map` right now, too. And adding coverage for other logic in that module at the same time... give a mouse a cookie...

Not sure I can put in that kind of time right now, but happy to leave this PR open and wait for that. This is a very minor bug, and has gone unfixed (and unreported by users other than you, me, and @paul121) for many years, so there's no urgency to fix it.

Another (small) consideration is that functional tests take longer to run, so adding additional functional tests adds to our delivery pipeline testing time. Definitely worthwhile for better coverage of `farm_ui_map` in general. Worthwhile for this bug fix alone? Debatable...

Another approach might be to just tack on an additional bit of logic to the `farm_ui_views` test (which this PR added to) that tests the GeoJSON View, too. That would be "easy", but technically it's not the responsibility of `farm_ui_views` to know about `farm_ui_map` (and actually `farm_ui_map` depends on `farm_ui_views`), so given the choice I would rather wait. Or merge this without GeoJSON tests and add that later along with more general `farm_ui_map` tests.

symbioquine (approved):

PR #1049, opened 2026-02-18
Clear access_policy cache when modules are installed

mstenta:
I noticed a bug with the farmOS module installation forms (both the original `/setup/modules` form and the new `/setup/wizard/modules` form). After installing a quick form module, the "Quick forms" menu item does not show, and going to `/quick` results in a 403 access denied.

Rebuilding the cache fixes the problem.

Steps to reproduce:

1. Install farmOS.
2. In the setup wizard modules installation step (or in `/setup/modules`) install the Planting quick form module.
3. Notice that "Quick forms" does not appear in the main menu.
4. Run `drush cr` and notice that "Quick forms" is added to the menu.

I'm not sure if this has been an issue all along in 3.x, and why it went unnoticed. Maybe it is related to the fact that we no longer install "default" modules (#1034) during initial farmOS installation, and now leave most modules to be installed by the user afterwards (#1035).

I'm also not sure if there is a more targeted fix (eg: only rebuilding the router information), or if rebuilding the entire cache is necessary (for this an other things).

I added a test that demonstrates the problem, and a simple `FarmModulesForm::finishInstallModulesBatch()` callback that runs `drupal_flush_all_caches()` after the module installation batch operations finish.

Worth noting: the only other place we run `drupal_flush_all_caches()` (outside of tests) is when the Mapbox / Google Maps settings forms are submitted (for saving API keys). We found that was necessary in order for the API keys to take effect. Maybe there is a more targeted fix for that as well, and digging deeper into both cases would be worthwhile as a follow-up step as a refinement.

Also worth noting: we might also want to follow-up on the relationship between `FarmSetupBlockForm::finishInstallModulesBatch()` and `FarmModulesForm::finishInstallModulesBatch()`. Before this PR, the former was only used to reset the setup wizard block plugin ID to ensure that newly installed modules have a chance to insert their setup wizard steps into the process. After this PR, the former also calls the latter, to ensure caches are rebuilt. Maybe there's a better way to organize these things, but I didn't want to get too bogged down in that, so this PR is focused on fixing the bug first.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [Fix missing action links on entity collection pages #1050](https://github.com/farmOS/farmOS/pull/1050)
 - [Fix asset and plan update hooks for prefixed database tables #1051](https://github.com/farmOS/farmOS/pull/1051)
 - [Validate entity bundle in asset/log/etc page_type display logic #1048](https://github.com/farmOS/farmOS/pull/1048)
+- [Clear access_policy cache when modules are installed #1049](https://github.com/farmOS/farmOS/pull/1049)
 
 ## [4.0.0-beta2] 2026-02-12
 
--- a/modules/core/role/src/Hook/ModuleHooks.php
+++ b/modules/core/role/src/Hook/ModuleHooks.php
@@ -1,0 +1,32 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Drupal\farm_role\Hook;
+
+use Drupal\Core\Cache\CacheBackendInterface;
+use Drupal\Core\Hook\Attribute\Hook;
+use Symfony\Component\DependencyInjection\Attribute\Autowire;
+
+/**
+ * Module hook implementations for farm_role.
+ */
+class ModuleHooks {
+
+  public function __construct(
+    #[Autowire(service: 'cache.access_policy')]
+    protected CacheBackendInterface $accessPolicyCache,
+  ) {}
+
+  /**
+   * Implements hook_modules_installed().
+   */
+  #[Hook('modules_installed')]
+  public function modulesInstalled($modules, $is_syncing) {
+
+    // Clear the access_policy cache when modules are installed, so that our
+    // managed role access policy rules get refreshed.
+    $this->accessPolicyCache->deleteAll();
+  }
+
+}


symbioquine:
This only works for the module installation forms right? What about `drush en <module_name>`? Could this be the argument for finding a more general (to module installation) and targeted (to just the routing cache) fix?

mstenta:
> What about `drush en <module_name>`?

~I just tested this and it did not suffer from the same problem. So maybe `drush en` does some cache rebuilding itself internally.~

Update: see https://github.com/farmOS/farmOS/pull/1049#issuecomment-3938895166

mstenta:
I started digging into this and found that it is specifically the `access_policy` cache that needs to be cleared, so I think we can take a more targeted approach to this.

However, I also discovered that the test I added is passing regardless of the "fix" in this PR. So that needs work too...

Converted back to draft...

mstenta:
> After installing a quick form module, the "Quick forms" menu item does not show, and going to `/quick` results in a 403 access denied.

The fact that we see an "access denied" and not a "page not found" suggests that this isn't related to the routing cache. And I can confirm that `drush cc router` has no effect.

> I started digging into this and found that it is specifically the `access_policy` cache that needs to be cleared, so I think we can take a more targeted approach to this.

Running `drush cc bin access_policy` _does_ fix it.

Now I'm wondering if this is actually an issue with the caching of our new managed role access policy: #922

If that's the case, then maybe we can test for it in the `farm_role` module, instead of testing these `farm_setup` module installation forms.

I'm experimenting with that now, but so far I haven't been able to replicate the issue in a kernel test. More investigation required...

mstenta:
> After installing a quick form module, the "Quick forms" menu item does not show, and going to `/quick` results in a 403 access denied.

Also worth noting (and makes sense): this does not affect user 1, which has the super user access policy (all permissions). If I follow the steps to reproduce (above) as user 1, the `/quick` path is immediately visible in the menu, and does not 403.

So I'm doing some further testing to isolate exactly when the issue manifests... I'll document the commands I'm running and their results for reproducibility below...

mstenta:
**Test 1: Admin user only, via UI**

1. `drush site-install -y --db-url=pgsql://farm:farm@db/farm --account-pass=admin`
2. Log in as admin via UI.
3. Install "Planting Quick Form" module via `/setup/modules`.

Results:

- "Quick forms" menu item is immediately visible and accessible to admin user.

**Test 2: Admin user installs and then creates Manager user**

1. `drush site-install -y --db-url=pgsql://farm:farm@db/farm --account-pass=admin`
2. Log in as admin via UI.
3. Install "Planting Quick Form" module via `/setup/modules`.
4. Create test user with Manager role:
    1. `drush en -y farm_manager`
    2. `drush user:create test --password=test`
    3. `drush user:role:add farm_manager test`
5. Log in as test user via UI.

Results:

- "Quick forms" menu item is immediately visible and accessible to test user.

**Test 3: Manager user is create first, then admin installs module**

1. `drush site-install -y --db-url=pgsql://farm:farm@db/farm --account-pass=admin`
2. Create Manager user:
    1. `drush en -y farm_manager`
    2. `drush user:create test --password=test`
    4. `drush user:role:add farm_manager test`
5. Log in as admin via UI.
6. Install "Planting Quick Form" module via `/setup/modules`.
7. Log in as test user via UI.

Results:

- "Quick forms" menu item is immediately visible and accessible to test user.

**Test 4: Manager+Config Admin user installs module**

1. `drush site-install -y --db-url=pgsql://farm:farm@db/farm --account-pass=admin`
2. Create user with Manager and Config Admin roles:
    1. `drush en -y farm_manager farm_config_admin`
    2. `drush user:create test --password=test`
    3. `drush user:role:add farm_config_admin test`
    4. `drush user:role:add farm_manager test`
5. Log in as test user via UI.
6. Install "Planting Quick Form" module via `/setup/modules`.

Results:

- "Quick forms" menu item is _NOT_ visible to test user. Visiting `/quick` directly shows "Access denied".
- Logging out and logging back in as test user does _NOT_ fix the issue.
- Running `drush cc bin access_policy` _DOES_ fix the issue. "Quick forms" menu item is immediately visible and accessible to test user upon refresh afterwards.

mstenta:
> * Running `drush cc bin access_policy` _DOES_ fix the issue. "Quick forms" menu item is immediately visible and accessible to test user upon refresh afterwards.

Ah! Just noticed something else during this test... which raises the urgency of this issue...

After step 4 ("Install "Planting Quick Form" module via `/setup/modules`."), the "Add Asset" action link is _also_ missing from the dashboard and going to `/asset/add` results in an "Access denied".

So this is affecting much more than just quick forms...

mstenta:
OK here is another interesting data point...

I tested installing the "Planting Quick Form" via Drupal's build-in module installation form (`/admin/modules`) as the test user (I granted the `farm_config_admin` role the `administer modules` permission first), and the same issue occurs.

This further confirms that this is _not_ specifically an issue with our module installation form.

> > What about `drush en <module_name>`?
> 
> I just tested this and it did not suffer from the same problem. So maybe `drush en` does some cache rebuilding itself internally.

I also tested this again... this time following the steps from "Test 4" above, but instead of installing via `/setup/modules` I installed via `drush en farm_quick_planting`.

The problem _DOES_ occur! So I was wrong before... probably because I tested as the admin user.

mstenta:
> Now I'm wondering if this is actually an issue with the caching of our new managed role access policy: https://github.com/farmOS/farmOS/pull/922

Assuming that this is the case (which seems to be confirmed by my testing), then I think there are two ways to fix it:

1. We implement `hook_modules_installed()` in the `farm_role` module and clear the `access_policy` cache bin whenever modules are installed.
2. We add additional cache contexts to our managed role access policy, so the cache is invalidated only when necessary.

I have a working implementation of the first approach, because it's very easy. I will push that as a commit to show what it looks like.

The second approach would be the ideal one, but I still need to confirm that it would actually work. And the challenge might be figuring out which cache context(s) we need to include. :thinking: 

This is where we can declare the contexts to vary by: https://github.com/farmOS/farmOS/blob/6012587ab9221c6ff101db13f29c693f84dcc0dc/modules/core/role/src/Access/ManagedRolePermissionsAccessPolicy.php#L44-L49

mstenta:
> 1. We implement `hook_modules_installed()` in the `farm_role` module and clear the `access_policy` cache bin whenever modules are installed.
> ...
> 
> I have a working implementation of the first approach, because it's very easy. I will push that as a commit to show what it looks like.

I reverted my original commit, and added a new commit that implements `hook_modules_installed()`. I have tested this manually and it fixes the issue.

Once we figure out the approach we want to take I will discard the reverted code, but for now I'll leave it in the branch for reference.

> I'm experimenting with that now, but so far I haven't been able to replicate the issue in a kernel test. More investigation required...

I still haven't figured out how to replicate the issue in a kernel test... so that still needs to be figured out, regardless of the approach we take to solve this.

symbioquine:
> I have a working implementation of the first approach, because it's very easy. I will push that as a commit to show what it looks like.
> 
> The second approach would be the ideal one, but I still need to confirm that it would actually work. And the challenge might be figuring out which cache context(s) we need to include. 🤔

Even that first approach seems a lot more targeted than the original "clear all caches" approach so I'm good with that if the second approach feels too complicated.

mstenta:
> Even that first approach seems a lot more targeted than the original "clear all caches" approach so I'm good with that if the second approach feels too complicated.

Agreed... and as I'm thinking more about it, I don't know if the second approach is viable at all. As far as I can tell, there are no relevant cache contexts that would do what we need.

This page describes the available contexts in Drupal core (assuming this is up to date and comprehensive): https://www.drupal.org/docs/develop/drupal-apis/cache-api/cache-contexts

I think what we would need is a context that represents "a list of all the installed modules" - so that when that list changes, the context is invalidated. But I don't think Drupal core provides a context like that.

mstenta:
> I still haven't figured out how to replicate the issue in a kernel test... so that still needs to be figured out, regardless of the approach we take to solve this.

I'm having a hard time adding an automated test for this.

I'm trying to add to our existing `ManagedRolePermissionsTest::testManagedRolePermissionsCaching()` test method, but I can't get the test to fail without the fix. It seems that caching isn't being used at all in the kernel test, so the permissions are being recalculated on every call to `hasPermission()` anyway. Without the permissions being cached, we can't replicate this issue.

It also makes me wonder if our existing tests are working as intended. I tried to make them fail too, but couldn't.

@paul121 I wonder if you have any thoughts on this?

mstenta:
> I'm having a hard time adding an automated test for this.

I'm going to go ahead and merge the fix without an automated test, so that we can release 4.0.0-beta3.

I've pushed my work-in-progress to a `4.x-test-access-policy-caching` branch on my fork: https://github.com/farmOS/farmOS/compare/4.x...mstenta:farmOS:4.x-test-access-policy-caching

I think it is "code complete"... we just need to figure out why the access policies aren't being cached in the kernel test more generally.

PR #1050, opened 2026-02-20
Fix missing action links on entity collection pages

mstenta:
I noticed that the "Add Asset" action link was missing from the /assets collection page in 4.x (but works in 3.x). As far as I can recall, nothing changed in the way we generate these links between 3.x and 4.x. This led me down a rabbit hole...

Tl;dr: I'm still not sure exactly what changed, but I've refactored the way we generate these action links, and added automated tests so we'll know if they break in the future.

### Observations:

I installed a fresh 4.x instance with the `asset`, `log`, `organization`, and `plan` modules turned on:

```bash
$ drush site-install -y --db-url=pgsql://farm:farm@db/farm --account-pass=admin
$ drush en asset plan log organization
```

The dashboard shows all expected action links:

Screenshot from 2026-02-20 07-26-02

However, the same action link should also be appearing on entity collection pages (`/assets`, `/logs`, `/organizations`, `/plans`), but in practice they are inconsistent. "Add Log" and "Add Organization" are present, but "Add Asset" and "Add Plan" are missing.

Screenshot from 2026-02-20 07-28-19

Screenshot from 2026-02-20 07-28-32

Screenshot from 2026-02-20 07-28-42

Screenshot from 2026-02-20 07-28-55

_Notably, all the other actions links derived by our `FarmActions` class (see below) appear to be working as expected._

### Investigation:

This is where those action links are currently built in the `farm_ui_action` module's `FarmActions` deriver class:

https://github.com/farmOS/farmOS/blob/6012587ab9221c6ff101db13f29c693f84dcc0dc/modules/core/ui/action/src/Plugin/Derivative/FarmActions.php#L65-L85

And specifically, this is where they are added to the dashboard:

https://github.com/farmOS/farmOS/blob/6012587ab9221c6ff101db13f29c693f84dcc0dc/modules/core/ui/action/src/Plugin/Derivative/FarmActions.php#L84

And this is where they are added to the collection pages (built by Views):

https://github.com/farmOS/farmOS/blob/6012587ab9221c6ff101db13f29c693f84dcc0dc/modules/core/ui/action/src/Plugin/Derivative/FarmActions.php#L74

So, what's going on? And why are they working for `log` and `organization`, but not for `asset` and `plan`? :thinking: 

Well... I discovered that I could fix `/assets` and `/plans` by changing line 74 in `FarmActions` (linked above) from:

`$this->derivatives[$name]['appears_on'][] = 'view.farm_' . $type . '.page';`

to:

`$this->derivatives[$name]['appears_on'][] = 'entity.' . $type . '.collection';`

This restores the "Add Asset" and "Add Plan" action links! But it removes the "Add Organization" link, and causes TWO "Add Log" links to appear. :face_with_spiral_eyes: 

Screenshot from 2026-02-20 08-01-07

Screenshot from 2026-02-20 08-01-24

Screenshot from 2026-02-20 08-01-35

Screenshot from 2026-02-20 08-01-47

After much head scratching, I figured out why...

The Log module implements its own "Add Log" action link in `log.links.action.yml`, which appears on the `entity.log.collection`: https://git.drupalcode.org/project/log/-/blob/2bbb45e9c97dca0af0d07efd219733392327d3ad/log.links.action.yml#L1-5 

So, by changing line 74 in `FarmActions`, we are essentially adding our own action link to the `entity.log.collection` route alongside the `log` module's own action link.

As for the missing "Add Organization" action link, I realized that was because we didn't include `organization` in the `farm_ui_view` module's implementation of `hook_entity_type_build()`:

https://github.com/farmOS/farmOS/blob/6012587ab9221c6ff101db13f29c693f84dcc0dc/modules/core/ui/views/src/Hook/EntityHooks.php#L14-L33

That hook alters the entity type definitions to override the "collection" links, so that they point to the pages we generate via Views provided by `farm_ui_views`. We should have added `organization` to that hook during #849, but that was overlooked.

The part that I don't understand, which is at the crux of all this, is: why does Drupal interpret the route name of `/assets` (and others) as `entity.asset.collection` in farmOS 4.x, while it is interpreted as `view.farm_asset.page` in farmOS 3.x? :thinking: 

Ultimately, this is better! Our desire is to add action links to the "collection" pages, and in 3.x we had to reference the Views route names explicitly, but that logic would break if anyone wanted to override those paths with their own routes. So it's better to add action links to the generic `entity.[entity-type].collection` routes, if possible.

It's possible that this was a bug in Drupal 10, which was fixed in Drupal 11... and this "fix" broke our code, which was technically a workaround.

I wish I could pinpoint exactly what changed, but I wasn't able to find a related Drupal core issue or change record. Although to be fair, this is a bit of a tricky one to search for...

### Proposed Solution:

I've reworked the way that these action links are generated. Instead of deriving our own generic "Add [entity-type]" action links in the `FarmActions` deriver class, I added them via `[entity-type].links.action.yml` definitions in the `asset`, `organization`, and `plan` modules, following the same pattern that the `log` module already employs. This fixes the duplicate "Add Log" action link, and restores the "Add Asset" and "Add Plan" links.

This only adds them to the entity collection pages. So then I also implemented `hook_menu_local_actions_alter()` to alter these default action links to add additional routes that they should appear on, including the dashboard and `/user/%/logs`, and removed the related logic from the `FarmAction` deriver class.

To fix the "Add Organization" link, I added `organization` to the `farm_ui_view` module's implementation of `hook_entity_type_build()`.

Lastly, I added automated test coverage for all these links, so we'll know if this breaks in the future.

### DIFF
--- a/modules/core/ui/action/tests/src/Functional/ActionsTest.php
+++ b/modules/core/ui/action/tests/src/Functional/ActionsTest.php
@@ -68,43 +68,43 @@ class ActionsTest extends FarmBrowserTestBase {
     // Test dashboard buttons.
     $this->drupalGet('/dashboard');
     $this->assertSession()->statusCodeEquals(200);
-    $this->assertSession()->linkExists('Add Asset');
-    $this->assertSession()->linkExists('Add Log');
-    $this->assertSession()->linkExists('Add Organization');
-    $this->assertSession()->linkExists('Add Plan');
+    $this->assertActionLinkExists('Add Asset', '/asset/add');
+    $this->assertActionLinkExists('Add Log', '/log/add');
+    $this->assertActionLinkExists('Add Organization', '/organization/add');
+    $this->assertActionLinkExists('Add Plan', '/plan/add');
 
     // Test entity lists.
     $this->drupalGet('/assets');
     $this->assertSession()->statusCodeEquals(200);
-    $this->assertSession()->linkExists('Add Asset');
+    $this->assertActionLinkExists('Add Asset', '/asset/add');
     $this->drupalGet('/logs');
     $this->assertSession()->statusCodeEquals(200);
-    $this->assertSession()->linkExists('Add Log');
+    $this->assertActionLinkExists('Add Log', '/log/add');
     $this->drupalGet('/organizations');
     $this->assertSession()->statusCodeEquals(200);
-    $this->assertSession()->linkExists('Add Organization');
+    $this->assertActionLinkExists('Add Organization', '/organization/add');
     $this->drupalGet('/plans');
     $this->assertSession()->statusCodeEquals(200);
-    $this->assertSession()->linkExists('Add Plan');
+    $this->assertActionLinkExists('Add Plan', '/plan/add');
 
     // Test per-bundle entity lists.
     $this->drupalGet('/assets/test');
     $this->assertSession()->statusCodeEquals(200);
-    $this->assertSession()->linkExists('Add Asset: Test');
+    $this->assertActionLinkExists('Add Asset: Test', '/asset/add/test');
     $this->drupalGet('/logs/test');
     $this->assertSession()->statusCodeEquals(200);
-    $this->assertSession()->linkExists('Add Log: Test');
+    $this->assertActionLinkExists('Add Log: Test', '/log/add/test');
     $this->drupalGet('/organizations/test');
     $this->assertSession()->statusCodeEquals(200);
-    $this->assertSession()->linkExists('Add Organization: Test');
+    $this->assertActionLinkExists('Add Organization: Test', '/organization/add/test');
     $this->drupalGet('/plans/test');
     $this->assertSession()->statusCodeEquals(200);
-    $this->assertSession()->linkExists('Add Plan: Test');
+    $this->assertActionLinkExists('Add Plan: Test', '/plan/add/test');
 
     // Test /user/%uid/logs.
     $this->drupalGet('/user/' . $this->user->id() . '/logs');
     $this->assertSession()->statusCodeEquals(200);
-    $this->assertSession()->linkExists('Add Log');
+    $this->assertActionLinkExists('Add Log', '/log/add');
 
     // Test links to /log/add/[bundle]?asset=[id] on asset pages.
     /** @var \Drupal\asset\Entity\AssetInterface $asset */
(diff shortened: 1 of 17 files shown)


modules/core/ui/action/tests/src/Functional/ActionsTest.php
```diff
@@ -0,0 +1,133 @@
...
+    $log->save();
+    $this->drupalGet('/asset/' . $asset->id());
+    $this->assertSession()->statusCodeEquals(200);
+    $this->assertSession()->linkExists('Add Log: Test');
+    $this->drupalGet('/asset/' . $asset->id() . '/logs');
+    $this->assertSession()->statusCodeEquals(200);
+    $this->assertSession()->linkExists('Add Log: Test');
+    $this->drupalGet('/asset/' . $asset->id() . '/logs/test');
+    $this->assertSession()->statusCodeEquals(200);
+    $this->assertSession()->linkExists('Add Log: Test');
```
symbioquine: Should we be checking the link urls here too - like line 109 implies?

symbioquine (approved):
Weird issue!

One minor point about the tests, but otherwise LGTM!

mstenta:
> The part that I don't understand, which is at the crux of all this, is: why does Drupal interpret the route name of `/assets` (and others) as `entity.asset.collection` in farmOS 4.x, while it is interpreted as `view.farm_asset.page` in farmOS 3.x? 🤔
> 
> Ultimately, this is better! Our desire is to add action links to the "collection" pages, and in 3.x we had to reference the Views route names explicitly, but that logic would break if anyone wanted to override those paths with their own routes. So it's better to add action links to the generic `entity.[entity-type].collection` routes, if possible.
> 
> It's possible that this was a bug in Drupal 10, which was fixed in Drupal 11... and this "fix" broke our code, which was technically a workaround.
> 
> I wish I could pinpoint exactly what changed, but I wasn't able to find a related Drupal core issue or change record. Although to be fair, this is a bit of a tricky one to search for...

I dug a little deeper into this...

Local action links are rendered via Drupal core's `LocalActionsBlock` (a block plugin that the Gin theme places at the top of its `content` region). Here is where it loads actions for a given route:

https://git.drupalcode.org/project/drupal/-/blob/b33878d18ac83debd721b1d4af5741505a56dc01/core/lib/Drupal/Core/Menu/Plugin/Block/LocalActionsBlock.php#L67

I put a breakpoint on that line in both farmOS 4.x and 3.x, and visited `/assets` in my browser. In both cases, `$route_name` was `entity.asset.collection`. I was expecting 3.x to be `view.farm_asset.page`, but it was not... so something else is going on. :thinking:

Next, I stepped into `$this->localActionManager->getActionsForRoute($route_name)` to figure out why the "Add Asset" action link _is_ being added in 3.x...

This method is identical in Drupal 11.3.x and 10.6.x:

11.3.x: https://git.drupalcode.org/project/drupal/-/blob/7b2531c3d796e0878cf039af4b3940bff3f9ac8c/core/lib/Drupal/Core/Menu/LocalActionManager.php#L165-209

10.6.x: https://git.drupalcode.org/project/drupal/-/blob/0e353af9a7db0dc9725aa0c4bb5acfed8aa911f7/core/lib/Drupal/Core/Menu/LocalActionManager.php#L165-209

... so there must be something downstream of this that is causing the "Add Asset" action link to match in 3.x.

modules/core/ui/action/tests/src/Functional/ActionsTest.php
mstenta: +1 yea that's a good idea. I'll look into that...

mstenta:
> ... so there must be something downstream of this that is causing the "Add Asset" action link to match in 3.x.

Ah ha! I figured it out! And I was wrong... the `FarmAction` derived action links were broken in 3.x too!

The reason they worked is because we _also_ defined them in `[entity-type].links.action.yml`! AKA exactly what this PR is adding!

We had removed those `[entity-type].links.action.yml` items in #1012, because we assumed that they were not being used. But apparently our deriver wasn't working, and those were filling in the gap for us, so we never noticed that the deriver was broken. It may have always been broken! :shrug:

So, this PR is taking the correct approach IMO, and is a nice cleanup of that logic more generally. Plus the tests will make sure this doesn't happen again. :-)

modules/core/ui/action/tests/src/Functional/ActionsTest.php
mstenta: I made a helper method for testing both the link text and URL: https://github.com/farmOS/farmOS/pull/1050/changes/87aa0e007bf253cd5cb10248686f3be221a701a7

PR #1053, opened 2026-02-24
Move "Add log" action and asset field population logic to farm_log_asset module

mstenta:
This is a simple cleanup PR that moves two bits of code from the `farm_log` module to the `farm_log_asset` module:

- The "Add log" action (`asset_log_add_action`)
- The `hook_entity_prepare_form()` implementation that is responsible for populating the asset field in the log edit form when `?asset=[id]` is passed in as a query string

These should live in `farm_log_asset`, because that module is responsible for adding the `asset` base field to `log` entities, so it is a better home for these than `farm_log`.


mstenta:
I have tested these features before and after this update and they work as expected.

PR #1056, opened 2026-03-05
Deprecate asset/organization/quantity entity events

mstenta:
See [Issue #3576633: Deprecate and remove asset/organization/quantity entity events](https://www.drupal.org/project/farm/issues/3576633) for full description.

### DIFF
--- a/modules/core/asset/src/Event/AssetEvent.php
+++ b/modules/core/asset/src/Event/AssetEvent.php
@@ -8,7 +8,11 @@ use Drupal\Component\EventDispatcher\Event;
 use Drupal\asset\Entity\AssetInterface;
 
 /**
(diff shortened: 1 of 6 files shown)


mstenta:
This is ready for review, but still needs a CHANGELOG.md update. I'm targeting this for inclusion in 4.1.0, so I will wait until it's ready to merge before adding that.

mstenta:
I'm going to merge the first two commits of this before 4.1.0, which refactor our event listeners to use hooks instead. This is necessary because the Log 3.1.0 release deprecates `LogEvent`, which causes PHPStan to fail. Might as well do this refactor now, and then the deprecation of asset/organization/quantity events is a simple change for 4.1.0.

PR #1060, opened 2026-03-07
Fix exposed taxonomy term reference filters in Views

mstenta:
The plugin settings must have changed at some point, causing this to revert to an autocomplete that does not filter by taxonomy. This changes it back to a multi-select field that only shows log category terms.

**Before**:

Screenshot from 2026-03-07 08-22-57

Screenshot from 2026-03-07 08-22-43

**After**:

Screenshot from 2026-03-07 08-24-38

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- [Fix exposed log category filter in Views #1060](https://github.com/farmOS/farmOS/pull/1060)
+
 ## [4.0.0-beta3] 2026-02-21
 
 ### Added
--- a/modules/core/ui/views/config/install/views.view.farm_log.yml
+++ b/modules/core/ui/views/config/install/views.view.farm_log.yml
@@ -937,7 +937,7 @@ display:
           entity_field: category
           plugin_id: taxonomy_index_tid
           operator: or
-          value: {  }
+          value: null
           group: 1
           exposed: true
           expose:
@@ -973,6 +973,16 @@ display:
           hierarchy: false
           limit: true
           error_message: true
+          sub_handler: 'default:taxonomy_term'
+          sub_handler_settings:
+            target_bundles:
+              log_category: log_category
+            sort:
+              field: name
+              direction: asc
+            auto_create: false
+            auto_create_bundle: ''
+          widget: select
         flag_value:
           id: flag_value
           table: log__flag
--- a/modules/core/ui/views/config/optional/views.view.farm_organization_log.yml
+++ b/modules/core/ui/views/config/optional/views.view.farm_organization_log.yml
@@ -975,7 +975,7 @@ display:
           entity_field: category
           plugin_id: taxonomy_index_tid
           operator: or
-          value: {  }
+          value: null
           group: 1
           exposed: true
           expose:
@@ -1011,6 +1011,16 @@ display:
           hierarchy: false
           limit: true
           error_message: true
+          sub_handler: 'default:taxonomy_term'
+          sub_handler_settings:
+            target_bundles:
+              log_category: log_category
+            sort:
+              field: name
+              direction: asc
+            auto_create: false
+            auto_create_bundle: ''
+          widget: select
         flag_value:
           id: flag_value
           table: log__flag


mstenta:
Hmm this fixed the issue, but causes tests to fail due to a "missing schema" issue:

> `Drupal\Core\Config\Schema\SchemaIncompleteException: Schema errors for views.view.farm_log with the following errors: views.view.farm_log:display.default.display_options.filters.category_target_id.value variable type is NULL but applied schema class is Drupal\Core\Config\Schema\Sequence, views.view.farm_log:display.default.display_options.filters.category_target_id.sub_handler missing schema, views.view.farm_log:display.default.display_options.filters.category_target_id.sub_handler_settings missing schema, views.view.farm_log:display.default.display_options.filters.category_target_id.widget missing schema`

I'm investigating...

mstenta:
Something funky is going on... I fixed the issue earlier by simply editing the category filter settings in the Views, specifying the "Vocabulary" to use (which was not set), and changing the "Selection type" from "Autocomplete" to "Select list". This results in the `sub_handler`, `sub_handler_settings`, and `widget` data being added to the configuration.

But those are NOT valid attributes of the `taxonomy_index_tid` filter plugin. They ARE attributes of the `entity_reference` plugin, and if I remove the filter and add it again through the UI, it gets saved as the `entity_reference` instead of `taxonomy_index_tid`.

Some things to note (still investigating):

- We have some logic in the `farm_entity_views` module that forces the `taxonomy_index_tid` plugin to be used for taxonomy reference field filters:
    - https://github.com/farmOS/farmOS/blob/b6da0b7126932ef1cf6df8c9c6fe7ab726835273/modules/core/entity/modules/views/src/Hook/EntityHooks.php#L23
    - https://github.com/farmOS/farmOS/blob/b6da0b7126932ef1cf6df8c9c6fe7ab726835273/modules/core/entity/modules/views/src/FarmEntityViewsData.php#L14
    - https://github.com/farmOS/farmOS/blob/b6da0b7126932ef1cf6df8c9c6fe7ab726835273/modules/core/entity/modules/views/src/EntityViewsDataTaxonomyFilterTrait.php#L53
- The exposed "Units" field in our quantity Views (`farm_log_quantity` and `farm_inventory`) uses the `taxonomy_index_tid`. And removing it and recreating it does NOT result in the `entity_reference` plugin being used instead, which is different behavior than I'm seeing with the log Views. :thinking:
    - https://github.com/farmOS/farmOS/blob/b6da0b7126932ef1cf6df8c9c6fe7ab726835273/modules/core/ui/views/config/install/views.view.farm_log_quantity.yml#L1101
    - https://github.com/farmOS/farmOS/blob/b6da0b7126932ef1cf6df8c9c6fe7ab726835273/modules/core/ui/views/config/optional/views.view.farm_inventory.yml#L1149
 - The log Views (`farm_log` and `farm_organization_log`) and quantity Views (`farm_log_quantity` and `farm_inventory`) are the only core Views provided by farmOS that have exposed taxonomy term reference filters explicitly declared in their YML configuration.
 - For all other term reference fields (which are all bundle-specific) we add them to Views dynamically at runtime via `hook_views_pre_view()`. I stepped through with a debugger and found that they end up using the `entity_reference` plugin.
     - https://github.com/farmOS/farmOS/blob/b6da0b7126932ef1cf6df8c9c6fe7ab726835273/modules/core/ui/views/src/Hook/ViewsExecutionHooks.php#L73
     - https://github.com/farmOS/farmOS/blob/b6da0b7126932ef1cf6df8c9c6fe7ab726835273/modules/core/ui/views/src/FarmUiViewsHelper.php#L223
     - https://git.drupalcode.org/project/drupal/-/blob/378e65203ea8be5dcc029b5d684fea187b76319f/core/modules/views/src/ViewExecutable.php#L2249

The fact that the dynamically-added bundle fields end up as `entity_reference` might be a separate issue that we want to follow-up on. They work fine as-is, but the advantage of the `taxonomy_index_tid` plugin (and why we chose to override `entity_reference` to use that instead) is because it allows showing the term hierarchy in the multi-select filter widget. So we do *not* currently have that for bundle-specific filters, and it would be better if we did.

mstenta:
> * We have some logic in the `farm_entity_views` module that forces the `taxonomy_index_tid` plugin to be used for taxonomy reference field filters:

This seems like it would be the most likely culprit, but I need to dig in to understand why (if it is).

Open questions:

- Why would adding `sub_handler` (etc) configuration fix the issue, if the `plugin_id` is still set to `taxonomy_index_tid`? I would think those settings would be ignored because they are specific to the `entity_reference` plugin.
    - Is it possible that the `entity_reference` plugin is being used, despite the `taxonomy_index_tid` plugin declared in YML?
- Why does adding term filter through the UI result in the `entity_reference` plugin being used? I would think that it would respect our override and use `taxonomy_index_tid`.
    - This means a higher likelihood of future regressions, since any edits through the UI could potentially break the filter again.
- What changed (and when) that broke the category filter originally? It is broken in both `3.x` and `4.x`, which means it predates the Drupal 11 update.

mstenta:
> * Is it possible that the `entity_reference` plugin is being used, despite the `taxonomy_index_tid` plugin declared in YML?

It appears that this is indeed happening.

To test, I started by reverting this PR's change, which removed the extra `sub_handler` (etc) settings.

If I put a breakpoint in the `valueForm()` method (for building the exposed filter form widget) of both the `EntityReference` and `TaxonomyIndexTid` filter plugin classes, I can see that `EntityReference` is being used for the `category_target_id` filter in the `farm_log` View. But `TaxonomyIndexTid` is being used for the `units` filter in the `farm_log_quantity` View.

This would explain why the filter was broken, and why this PR initially appeared to fix it. The declared `plugin_id: taxonomy_index_tid` was being ignored, and the `entity_reference` plugin was being used instead, but it was missing the necessary `sub_handler`, `sub_handler_settings`, and `widget` configuration. This PR added that, but kept `plugin_id: taxonomy_index_tid` - which resulted in the `missing schema` test failures.

mstenta:
> * We have some logic in the `farm_entity_views` module that forces the `taxonomy_index_tid` plugin to be used for taxonomy reference field filters:

With a breakpoint on the line linked above (https://github.com/farmOS/farmOS/blob/b6da0b7126932ef1cf6df8c9c6fe7ab726835273/modules/core/entity/modules/views/src/EntityViewsDataTaxonomyFilterTrait.php#L53), when I run `drush --xdebug cr` (to rebuild Views table data) I can see that the `category_target_id` filter plugin ID is being set to `taxonomy_index_tid`, so that's working.

Maybe it's being overridden/altered later in the Views table data rebuild process?

mstenta:
> I can see that the `category_target_id` filter plugin ID is being set to `taxonomy_index_tid`, so that's working.
> 
> Maybe it's being overridden/altered later in the Views table data rebuild process?

It is being overridden...

... by this: https://github.com/farmOS/farmOS/blob/b6da0b7126932ef1cf6df8c9c6fe7ab726835273/modules/core/ui/views/src/Hook/ViewsHooks.php#L38

That was added here: https://github.com/farmOS/farmOS/commit/8beda88c1c75812e58686ca7dde6c4acaa07a500#diff-d2c72dcde6d75447b11247148391e6104d73de42f0e97118fcfe09c336dc6209

... which was done to make use of the patch in this Drupal core issue (which was later merged in #900): https://www.drupal.org/node/3347343

So this has probably been an issue since farmOS v3.3.0 (the first release after that commit).

mstenta:
> I can see that `EntityReference` is being used for the `category_target_id` filter in the `farm_log` View. But `TaxonomyIndexTid` is being used for the `units` filter in the `farm_log_quantity` View.

Ah... the reason it's happening to `category_target_id` and not `units` is because this condition doesn't pass for `units`: https://github.com/farmOS/farmOS/blob/b6da0b7126932ef1cf6df8c9c6fe7ab726835273/modules/core/ui/views/src/Hook/ViewsHooks.php#L37

The `units` field is a single-value base-field on `quantity` entities, whereas the `category` field is a multi-value field on `log` entities. So, in the Views `$data` array, the `units` filter is in `$data['quantity']['units']['filter']`, whereas the `category` filter is in `$data['log__category']['category_target_id']['filter']`.

So, the `Drupal\farm_ui_views\Hook\ViewsHooks::viewsDataAlter()` method is overriding the filter plugin ID for entity reference fields, but only for ones that have a dedicated database table (bundle fields and multi-value base field).

This reveals two separate issues with `Drupal\farm_ui_views\Hook\ViewsHooks::viewsDataAlter()`:

1. It is overriding the `taxonomy_index_tid` plugin ID, which is breaking the log category exposed filter (the original issue of this PR).
2. It is not overriding plugin IDs for single-value base fields like `units`. We don't want it to override `units` specifically, but we would want it to override other single-value entity reference base fields that are not taxonomy term references. I don't think farmOS core has any other fields like this, but it would be an issue if anyone added one.

mstenta:
I have reverted the first commit from this PR. It was a broken fix.

mstenta:
The first thing I'm going to do is move the `entity_reference` plugin alter logic from the `farm_ui_views` module to the `farm_entity_views` module, since they are doing similar things and conflicting with each other. Better to manage the relationship between the related logic in one module. It probably should have been added there originally instead of `farm_ui_views`.

mstenta:
Ah another consideration: `EntityViewsData::addReverseRelationships()` (which we are overriding with our trait via https://github.com/farmOS/farmOS/blob/b6da0b7126932ef1cf6df8c9c6fe7ab726835273/modules/core/entity/modules/views/src/FarmEntityViewsData.php#L14) appears to only be intended for defining base field relationships... not bundle field relationships.

So this is potentially a third issue currently, in addition to...

> 1. It is overriding the `taxonomy_index_tid` plugin ID, which is breaking the log category exposed filter (the original issue of this PR).
> 2. It is not overriding plugin IDs for single-value base fields like `units`. We don't want it to override `units` specifically, but we would want it to override other single-value entity reference base fields that are not taxonomy term references. I don't think farmOS core has any other fields like this, but it would be an issue if anyone added one.

3. The filter plugin of taxonomy term reference bundle fields is not being set to `taxonomy_index_tid`.

mstenta:
> 3. The filter plugin of taxonomy term reference bundle fields is not being set to `taxonomy_index_tid`.

Oh wait, that is this issue, already described above:

> The fact that the dynamically-added bundle fields end up as `entity_reference` might be a separate issue that we want to follow-up on. They work fine as-is, but the advantage of the `taxonomy_index_tid` plugin (and why we chose to override `entity_reference` to use that instead) is because it allows showing the term hierarchy in the multi-select filter widget. So we do _not_ currently have that for bundle-specific filters, and it would be better if we did.

Maybe we can fix all three together!

mstenta:
> Ah another consideration: `EntityViewsData::addReverseRelationships()` ... appears to only be intended for defining base field relationships... not bundle field relationships.

Wait. I'm wrong about this. I see that bundle fields ARE being altered in this method as well. But they are probably being overridden just like the category filter is.

So maybe we should move the `hook_views_data_alter()` logic that sets the plugin to `entity_reference` into this method, so it's all together. This would ensure that it works in cooperation with the `taxonomy_index_tid` override, and we could also ensure it works with both base and bundle fields.

> 1. It is overriding the `taxonomy_index_tid` plugin ID, which is breaking the log category exposed filter (the original issue of this PR).

This would no longer be an issue because we could ensure taxonomy reference fields use `taxonomy_index_tid`.

> 2. It is not overriding plugin IDs for single-value base fields...

The logic in `EntityViewsData::addReverseRelationships()` is smarter about which fields are in their own dedicated tables, so I think this would be fixed too.

> 3. The filter plugin of taxonomy term reference bundle fields is not being set to `taxonomy_index_tid`.

This might just start working once we merge/refactor the code. But we should check that these dynamically-added exposed filters still work afterwards...

mstenta:
> So maybe we should move the `hook_views_data_alter()` logic that sets the plugin to `entity_reference` into this method, so it's all together.

Alternatively, we could merge all the logic into `hook_views_data_alter()` (the other direction).

Or, merge both into a new `hook_field_views_data_alter()`, which is more tailored towards altering fields. This is what Drupal core uses to alter taxonomy term reference filters to use `taxonomy_index_tid`:

https://git.drupalcode.org/project/drupal/-/blob/378e65203ea8be5dcc029b5d684fea187b76319f/core/modules/taxonomy/src/Hook/TaxonomyViewsHooks.php#L67

Actually, now I wonder why that isn't covering our needs for `taxonomy_index_tid` already, and why we need `EntityViewsData::addReverseRelationships()`... :thinking:

mstenta:
> Actually, now I wonder why that isn't covering our needs for `taxonomy_index_tid` already, and why we need `EntityViewsData::addReverseRelationships()`... 🤔

Oh... `hook_field_views_data_alter()` only works for config fields. Not base/bundle fields...

mstenta:
I've restored my `4.x-fix-exposed-category-filter` branch, renamed this PR to make it more general, and pushed my current work: https://github.com/farmOS/farmOS/compare/4.x...mstenta:farmOS:refs/heads/4.x-fix-exposed-category-filter

Log categories are working as expected, but now "Log type" has turned to an autocomplete. :-(

Screenshot from 2026-03-08 13-57-20

And... it seems that all the autocompletes are broken... they all result in 403 errors in the developer tools network tab when you type a letter into them. :-(

> 3. The filter plugin of taxonomy term reference bundle fields is not being set to `taxonomy_index_tid`.
> 
> This might just start working once we merge/refactor the code. But we should check that these dynamically-added exposed filters still work afterwards...

These are working now. For example, here is before/after of `/assets/plants`, which shows how the "Crop/variety" and "Season" filters change from autocompletes to multi-selects:

**Before**:

Screenshot from 2026-03-08 13-54-53

**After**:

Screenshot from 2026-03-08 13-56-26

I think this is an improvement, but it has performance considerations... those two taxonomies specifically may be very large.

So we may want to give some thought to refining this...

mstenta:
> Log categories are working as expected, but now "Log type" has turned to an autocomplete. :-(
> And... it seems that all the autocompletes are broken... they all result in 403 errors in the developer tools network tab when you type a letter into them. :-(

I fixed this by filtering out target entity types we don't care about (eg: `log_type`).

The 403 issue went away, so it may have been an artifact of my testing and cache rebuilds.

mstenta:
I merged my branch. This is a good cleanup and improvement to our code, as well as a fix for the original issue. :-)

PR #1061, opened 2026-03-07
Fix asset parent/child farm validation when neither asset is in a farm

### LINKED ISSUE #1059: Farm organization module does not allow assigning parent asset when parent/child are not assigned to farm
**Describe the bug**
When the Farm Organization module is enabled for an existing (populated) instance, the parent asset relationship becomes a validation blocker to editing assets.

For existing assets which already have parent/child relationship, but where neither is yet assigned to a farm org, one can not edit either parent or child.

Creating new assets also does not work, if also trying to assign a parent asset.  This is true even when neither parent nor child is assigned to any farm.

**To Reproduce**
1. Start with a 4.x farmOS instance (sans Farm Organization module), which has e.g. 2 land assets with parent/child relationship.
2. Enable the Farm Organization module.
3. Try to edit either land asset, just click Edit and then click Save.
4. See error message: "Assets must be in the same farm as their parents."
5. Alternatively, create a new asset and define its parent; clicking Save will give the same error message.
6. (Note all my examples do NOT contain any farm org assignments.)

**Expected behavior**
I understand the validation is meant to ensure the parent asset and child asset has the "same" farm org relationship.  So if parent was assigned to Farm A and child was assigned to Farm B, validation should fail and error message is expected.

But when both parent and child asset have a "null" farm org reference, I would expect validation to succeed and allow the edit.

Similarly when creating a new asset, with null farm, I would expect assigning a parent would work okay as long as parent also had no farm assigned.

**Desktop (please complete the following information):**
 - OS: Debian 12 bookworm
 - Browser: Firefox

**Additional context**
See original comments in #1058 - my workaround for now was to disable the Farm Organization module.  I re-enabled that in my test instance for sake of confirming behavior to create this issue.

FWIW the previous farm org data was partially preserved, so when I re-enabled the module I did have some farm org records defined, but the assets no longer referenced them.  This is just FYI and did not affect the actual behavior described AFAIK.

Perhaps a "warning" is appropriate at module install time, so admin knows what hurdles might come if enabling the module for an existing / populated instance which already makes use of parent/child relationships.  But that is kind of a separate issue, just a suggestion.

mstenta:
Fixes https://github.com/farmOS/farmOS/issues/1059

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - [Allow importing taxonomy term CSVs without parent column #1054](https://github.com/farmOS/farmOS/pull/1054)
 - [Fix exposed taxonomy term reference filters in Views #1060](https://github.com/farmOS/farmOS/pull/1060)
+- [Fix asset parent/child farm validation when neither asset is in a farm #1061](https://github.com/farmOS/farmOS/pull/1061)
 
 ### Changed
 
--- a/modules/organization/farm/src/Plugin/Validation/Constraint/AssetParentFarmValidator.php
+++ b/modules/organization/farm/src/Plugin/Validation/Constraint/AssetParentFarmValidator.php
@@ -69,8 +69,8 @@ class AssetParentFarmValidator extends ConstraintValidator implements ContainerI
         }
       }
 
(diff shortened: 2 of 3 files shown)


mstenta:
Merged (accidentally deleted the branch before I pushed the merged 4.x branch, so the PR was marked as "Closed" instead of "Merged").

PR #1063, opened 2026-03-09
Fix Fatal error: Trait FarmFormProtectionTrait not found during update from 3.x

### LINKED ISSUE #1062: PHP Fatal error:  Trait "Drupal\farm_form\Traits\FarmFormProtectionTrait" not found
During the upgrade from farmOS 3.5.1 to 4.0.0-beta3 @Fosten and I both (separately) ran into the following error (during `drush updb`):

`PHP Fatal error:  Trait "Drupal\farm_form\Traits\FarmFormProtectionTrait" not found in /opt/drupal/web/profiles/farm/modules/core/ui/theme/src/Form/GinContentFormBase.php on line 25`

I thought that I had fixed this in #1042 but apparently not... opening this bug report to investigate further...

mstenta:
Fixes https://github.com/farmOS/farmOS/issues/1062

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- [Fix Fatal error: Trait FarmFormProtectionTrait not found during update from 3.x #1063](https://github.com/farmOS/farmOS/pull/1063)
 - [Allow importing taxonomy term CSVs without parent column #1054](https://github.com/farmOS/farmOS/pull/1054)
 - [Fix exposed taxonomy term reference filters in Views #1060](https://github.com/farmOS/farmOS/pull/1060)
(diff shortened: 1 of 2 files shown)


RMattB1:
possibly helpful context for reviewing the PR: 
https://drupalsun.com/philipnorton42/2022/06/05/drupal-9-different-update-hooks-and-when-use-them

Fosten:
From the article @RMattB1 linked: "Do use hook_post_update_NAME() to: As a cache clear operation will be performed when this hook is run you can add an empty hook_post_update_NAME() hook to trigger a cache clear."

mstenta:
Thanks @RMattB1 and @Fosten! Yea... these particular updates are unique because they are adding a new module and moving a PHP trait to it at the same time, so we needed to be a little extra careful about how that gets deployed, to ensure that a cache clear does NOT happen before the new module is installed.

FWIW I am in the process of updating hundreds of farmOS instances to 4.x right now, and have not run into this issue again! So I think it's good!

PR #1064, opened 2026-03-10
Fix ContentEntityGeometryNormalizer support check

lgedgar:
This bug was triggered by configuring a "webhook" for asset create/update events.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- [Fix ContentEntityGeometryNormalizer support check #1064](https://github.com/farmOS/farmOS/pull/1064)
+
 ## [4.0.0-beta4] 2026-03-09
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


lgedgar:
I added more comments to the commit message.  Copying here:

The bug was triggered as follows:

I installed/enabled the [webhooks](https://www.drupal.org/project/webhooks) module.

I created a new outgoing webhook, defined for Asset
create/update/delete events.

I edited an Animal Asset, which did not have intrinsic geometry.
Clicking save caused a server error; message was:

    TypeError: strpos(): Argument #1 ($haystack) must be of type string, null given in strpos()
    (line 101 of /opt/drupal/web/profiles/farm/modules/core/geo/src/Normalizer/ContentEntityGeometryNormalizer.php).

Per @mstenta the logic did not expect `$format` to be NULL, but it was in this context.

This also replaces `strpos()` with `str_starts_with()` since we already require PHP 8.

symbioquine (approved):
LGTM

mstenta:
Merged! Thanks again @lgedgar!

I simplified the commit message and this PR title just a bit (along with the CHANGELOG.md line to match). I like to add the PR # at the end of the commit message so that it's easy to find again if we ever need to.

PR #1067, opened 2026-03-12
Fix doc header outline for OAuth section in hosting/install

lgedgar:
This is a very minor tweak, and feel free to ignore if I'm mistaken.  But it seemed to me the "OAuth2 keys" section was not behaving somehow.  In particular it did not appear in the table of contents (top right) at least on my screen..  I think this will fix, but didn't attempt to actually "build" the docs site, so not 100% sure.

Screenshot from 2026-03-11 19-43-37


symbioquine (approved):
I think this makes sense.

mstenta:
Ah good eye @lgedgar! Thanks!

Issue #82, opened 2018-04-19
mass repositioning in the tree


bobombolo:
so i have this view and i want to add an intermediate layer of areas in the hierarchy, i mass-select them and i only have the option to delete or flag. what i really want is the option to re-parent all these.

![screenshot from 2018-04-19 18-17-51](https://user-images.githubusercontent.com/25610408/39021353-6048bf66-43fe-11e8-8b70-976beb9490c6.png)

mstenta:
Up at the top of that page there is a "Hierarchy (change)" link, that gives you a form for clicking and dragging the hierarchy around.

http://farmos.org/guide/areas/#organizing-areas-hierarchically

bobombolo:
oh it's draggable!?

bobombolo:
draggable is nice but in the use-case that you add a new intermediate group that many areas would be subjugated to, then it would be nice to multiple select them and "assign new parent" button to appear just like delete or flag.
it's a question of data management and minimal number of clicks.

bobombolo:
again these are issues for people entering initial data, people that don't build a perfect schema first. people that discover the structure to their intuitive data for the first time as they are entering it into a relational database.

PR #1070, opened 2026-03-16
Validate that logs only reference assets in the same farm

mstenta:
This is a follow-up to the constraint validation logic that was added in #849 to ensure that the graphs of relationships between records do not cross farm organization boundaries.

This PR adds a more general constraint to log entities to ensure that all assets that they referenced by the log are assigned to the same farm (or no farm).

Without this, it's possible to create a log that references farms in more than one farm, which, in the context of access control (eg: as provided by the [Multitenant](https://www.drupal.org/project/farm_multitenant) module) raises questions about who has access to the log. So, in line with our decision to keep things simple in the first release of the Farm Organization module, this prevents that question from arising. We may re-assess this later, but it's easier to open things up in the future than lock things back down, so we are starting with this restriction in place.

This new constraint looks at all assets referenced in a log's `asset` and `location` fields, and checks to see if they are all assigned to the same farm (or no farm). If they are not in the same farm, then a violation is added.

Notably, this constraint essentially makes one of the other constraints that we added in #849 redundant, and allows us to simplify another...

The `LogMovementFarm` constraint checks to see if a log is a movement, and then basically performs the same check on the `asset` and `location` fields to see if there are assets from multiple farms. So this PR removes that constraint in favor of this more generic one. The tests for the movement constraint are maintained.

Likewise, the `LogGroupAssignmentFarm` constraint checks to see if a log is a group assignment, then performs a similar check. The difference is that it checks the `asset` and `group` reference fields (not `location`). So, this PR refactors this constraint to extend the new generic constraint, and only overrides the fields that it checks. It still checks to see if `is_group_assignment` is set to ensure that it doesn't get triggered alongside the generic constraint. The tests for the group assignment constraint are maintained.

So all in all, this PR expands the coverage while also simplifying the overall constraint logic.


modules/organization/farm/src/Plugin/Validation/Constraint/LogAssetFarmValidator.php
```diff
@@ -0,0 +1,97 @@
...
+class LogAssetFarmValidator extends ConstraintValidator {
+
+  /**
+   * Log asset reference fields.
+   *
+   * @var array|string[]
+   */
+  protected array $fieldNames = [
+    'asset',
+    'location',
```
symbioquine, line 25: 'equipment'?

modules/organization/farm/src/Hook/FieldHooks.php
symbioquine: Why can we remove the constraint for the `is_movement` field, but not the `is_group_assignment` field? Also, shouldn't we have a similar constraint added here for the `location` and `equipment` fields?

modules/organization/farm/src/Plugin/Validation/Constraint/LogAssetFarmValidator.php
```diff
@@ -0,0 +1,97 @@
...
+  ];
+
+  /**
+   * {@inheritdoc}
+   */
+  public function validate(mixed $value, Constraint $constraint) {
+    /** @var \Drupal\Core\Field\Plugin\Field\FieldType\BooleanItem $value */
```
symbioquine: This `@var` pragma seems a bit misleading because it's not always a `BooleanItem`? It can also be a `AssetReferenceField` (or something like that) too right?

modules/organization/farm/src/Plugin/Validation/Constraint/LogAssetFarmValidator.php
```diff
@@ -0,0 +1,97 @@
...
+    /** @var \Drupal\farm_farm\Plugin\Validation\Constraint\LogAssetFarm $constraint */
+
+    // Get the log entity.
+    /** @var \Drupal\log\Entity\LogInterface|null $log */
+    $log = $value->getParent()->getValue();
+
+    // Only continue if the log references assets.
+    if (!$this->logReferencesAssets($log)) {
+      return;
+    }
```
symbioquine: This seems unnecessary.

modules/organization/farm/src/Plugin/Validation/Constraint/LogAssetFarmValidator.php
```diff
@@ -0,0 +1,97 @@
...
+   */
+  protected function logReferencedAssetFarmIds(LogInterface $log) {
+    $farm_ids = [];
+    foreach ($this->fieldNames as $name) {
+      if ($log->get($name)->isEmpty()) {
+        continue;
+      }
+      foreach ($log->get($name)->referencedEntities() as $asset) {
+        if ($asset->get('farm')->isEmpty()) {
+          $farm_ids[] = 0;
```
symbioquine, line 83: How are we sure zero is never a valid `$farm->id()`? Is that definitely a characteristic of Drupal itself - or just the typical Postgres/MySQL database drivers?

modules/organization/farm/src/Plugin/Validation/Constraint/LogGroupAssignmentFarmValidator.php
```diff
@@ -5,12 +5,21 @@
...
+
+  /**
+   * Log asset reference fields.
+   *
+   * @var array|string[]
+   */
+  protected array $fieldNames = [
+    'asset',
+    'group',
+  ];
```
symbioquine: I'm a bit confused how we can have these (e.g. `group` and `location`) as separate validators. Doesn't that leave cross-farm references where there's only one for each of these untested?

For example, wouldn't both this validator and the location validator pass if a log referenced an asset from one farm as the location and separately referenced an asset from a different farm as the group assignment?

symbioquine (commented):
Also, shouldn't we test for inventory adjustments related to multiple farm assets?

modules/organization/farm/src/Plugin/Validation/Constraint/LogAssetFarmValidator.php
mstenta: Ah good catch! I copied that from another constraint and forgot to update it. This constraint should only be applied to `EntityReferenceItemInterface` values.

Fixup: https://github.com/farmOS/farmOS/pull/1070/commits/44be0cb6b104ce993cdb6a5d1760b92f7d1ae88f

modules/organization/farm/src/Plugin/Validation/Constraint/LogAssetFarmValidator.php
mstenta: You're right. My approach evolved as I wrote this PR and this ended up being vestigial and redundant.

Fixup: https://github.com/farmOS/farmOS/pull/1070/commits/7dd090ba3f2c3b80418022e1ad6fe7a67bcd5e92

modules/organization/farm/src/Plugin/Validation/Constraint/LogAssetFarmValidator.php
mstenta: Drupal entities always have an integer ID greater than zero in my experience. I'm not sure exactly at what level that's enforced, but I think we can safely rely on it.

I only used 0 as a placeholder, but we could use something else (like an arbitrary "`none`" string) if that would be better?

modules/organization/farm/src/Plugin/Validation/Constraint/LogAssetFarmValidator.php
mstenta: Ah yea, forgot about `equipment`. :thinking: 

That will only exist if the `farm_equipment` module is installed, which is not a hard requirement of the Farm Organization module. But I suppose we could add it to this list and add some logic to ensure each field exists before checking it.

modules/organization/farm/src/Plugin/Validation/Constraint/LogGroupAssignmentFarmValidator.php
mstenta: Hmm I think you're right. Maybe we should just combine them after all. I'm writing a more thorough comment in response to this and others below...

modules/organization/farm/src/Hook/FieldHooks.php
mstenta: Maybe we should just combine them all. I'm writing a more thorough comment in response to this and others below...

mstenta:
Thanks for the review @symbioquine - you bring up some very good questions... many are related so I'll respond here instead of in each thread above.

> Why can we remove the constraint for the `is_movement` field, but not the `is_group_assignment` field? Also, shouldn't we have a similar constraint added here for the `location` and `equipment` fields?
> ...
> I'm a bit confused how we can have these (e.g. `group` and `location`) as separate validators. Doesn't that leave cross-farm references where there's only one for each of these untested?
> 
> For example, wouldn't both this validator and the location validator pass if a log referenced an asset from one farm as the location and separately referenced an asset from a different farm as the group assignment?

Maybe it does make sense to just put everything together in one constraint. I was thinking it would be better to try to keep them separate, but realized that has the potential to causes duplicate violations (one for each field the constraint is applied to).

I was also trying to maintain separate violation messages, so they would be more specific to "movements" and "group assignments", but maybe that's not necessary either.

Ultimately, we just want to check ALL assets that are referenced from a log, and confirm they are all in the same farm (or no farm). It doesn't really matter if the log is a movement or group assignment.

> Also, shouldn't we test for inventory adjustments related to multiple farm assets?

Phew... you're right. That one might be a bit more complex, because it's a field on `quantity` entities... and we would want to compare it with all the other assets referenced directly from the log too... :thinking: I'll take a closer look at that...

mstenta:
I've pushed a set of fixup commits that effectively merge everything into a single entity-level constraint that checks assets referenced across the `asset`, `equipment`, `group`, and `location` fields on log entities. It checks to be sure the fields exist first, so it works when `farm_equipment` or `farm_group` modules are not installed.

Next thing to figure out is inventory assets...

mstenta:
I added logic to check the `inventory_asset` field on `quantity` entities attached to the log.

modules/organization/farm/src/Plugin/Validation/Constraint/LogAssetFarmValidator.php
symbioquine: I could go either way on this. Mostly just asking the question since I was curious and it seemed like it could be a potential issue.

symbioquine: It does look like there's some precedent (at least in earlier versions of Drupal) for zero being used as a placeholder for user ids: https://www.drupal.org/docs/7/understanding-drupal/users-permissions-and-roles

mstenta: 0 is used to represent the ID of the "anonymous user" (not logged in), but that's the only case I'm aware of.

modules/organization/farm/tests/src/Kernel/FieldConstraintsTest.php
```diff
@@ -155,6 +170,176 @@ public function testAssetParentFarmConstraint() {
...
+  public function testLogAssetConstraints() {
+    $entity_type_manager = $this->container->get('entity_type.manager');
+    $asset_storage = $entity_type_manager->getStorage('asset');
+    $log_storage = $this->container->get('entity_type.manager')->getStorage('log');
+    $organization_storage = $entity_type_manager->getStorage('organization');
+    $quantity_storage = $entity_type_manager->getStorage('quantity');
+
+    // Create two assets.
+    // Both are equipment and are locations so that they can be referenced in
+    // the asset, equipment, and location fields of a log.
```
symbioquine, line 185: nit:

```suggestion
    // the asset, equipment, and location fields logs.
```

modules/organization/farm/tests/src/Kernel/FieldConstraintsTest.php
```diff
@@ -155,6 +170,176 @@ public function testAssetParentFarmConstraint() {
...
+    $this->assertEquals(1, $violations->count());
+    $this->assertInstanceOf(LogAssetFarm::class, $violations->get(0)->getConstraint());
+
+    // Move the second asset to the first farm and confirm that the log
+    // validates.
+    $asset2->set('farm', [$farm1]);
+    $asset2->save();
+    $violations = $log->validate();
+    $this->assertEquals(0, $violations->count());
+  }
```
symbioquine, line 258: I don't know if it matters, but - despite testing an impressive amount of the permutations - I believe this test suite doesn't cover line `#82` of the validator.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- [Validate that logs reference assets in the same farm #1070](https://github.com/farmOS/farmOS/pull/1070)
```
symbioquine: nit:

```suggestion
- [Validate that logs only reference assets in the same farm #1070](https://github.com/farmOS/farmOS/pull/1070)
```

modules/organization/farm/src/Plugin/Validation/Constraint/LogAssetFarmValidator.php
```diff
@@ -0,0 +1,97 @@
...
+                $farm_ids[] = $farm->id();
+              }
+            }
+          }
+        }
+      }
+    }
+
+    return array_unique($farm_ids);
+  }
```
symbioquine, line 95: I'm not sure if this is better or not, but it is possible to flatten and deduplicate a tiny bit...

```suggestion
  protected function logReferencedAssetFarmIds(LogInterface $log) {
    $asset_refs = [];
    
    // Consider direct asset references
    foreach ($this->fieldNames as $name) {
      if ($log->hasField($name)) {
        $asset_refs[] = $log->get($name);
      }
    }
    
    // Consider inventory adjustment asset references
    if ($log->hasField('quantity') && !$log->get('quantity')->isEmpty()) {
      foreach ($log->get('quantity')->referencedEntities() as $quantity) {
        /** @var \Drupal\quantity\Entity\QuantityInterface $quantity */
        if ($quantity->hasField('inventory_asset')) {
          $asset_refs[] = $quantity->get('inventory_asset');
        }
      }
    }

    $farm_ids = [];

    // Accumulate the farm ids from all referenced assets
    foreach ($asset_refs as $asset_ref) {
      if ($asset_ref->isEmpty()) {
        continue;
      }
      foreach ($asset_ref->referencedEntities() as $asset) {
        if ($asset->get('farm')->isEmpty()) {
          $farm_ids[] = 0;
        }
        else {
          foreach ($asset->get('farm')->referencedEntities() as $farm) {
            $farm_ids[] = $farm->id();
          }
        }
      }
    }
    
    // Return true iff the set of ids has a single value
    return array_unique($farm_ids);
  }
```

symbioquine (approved):
LGTM!

mstenta:
Thanks for the thorough review @symbioquine! I will go through and incorporate your suggestions... glad we'll be able to merge this before the 4.0.0 release on Friday! :tada:

modules/organization/farm/tests/src/Kernel/FieldConstraintsTest.php
mstenta: Do you mean testing `$farm_ids[] = 0;` specifically? I think that is covered early in the test...

First it creates two assets that are not assigned to any farm and tests that a log referencing them validates:

https://github.com/mstenta/farmOS/blob/9823da3fbd6b804ecba459e566e2a8ed25cc929b/modules/organization/farm/tests/src/Kernel/FieldConstraintsTest.php#L183-L200

Then it assigns _one_ of the assets to a farm, but leaves the other unassigned, and confirms that the log does not validate:

https://github.com/mstenta/farmOS/blob/9823da3fbd6b804ecba459e566e2a8ed25cc929b/modules/organization/farm/tests/src/Kernel/FieldConstraintsTest.php#L214-L219

mstenta:
I squashed my fixup commits and force pushed, renaming the PR, commit message, and changelog line at the same time.

modules/organization/farm/tests/src/Kernel/FieldConstraintsTest.php
symbioquine: > Do you mean testing `$farm_ids[] = 0;` specifically? I think that is covered early in the test...

Yes. There are two spots where that `$farm_ids[] = 0;` statement happens. I believe you could comment out the second one (on line `#82` in the version I was reviewing) and all the tests would still pass.

Like I said, I don't know how much it matters, but since you went to the trouble of testing so many permutations I figured I'd mention it.

modules/organization/farm/src/Plugin/Validation/Constraint/LogAssetFarmValidator.php
symbioquine: For better or worse, this refactoring also means that the number of test cases needed to get full coverage (not necessarily the goal) goes down and that case I was referring to above stops being an issue.

symbioquine (approved):
LGTM!

modules/organization/farm/tests/src/Kernel/FieldConstraintsTest.php
mstenta: Ah I see! Reading your comments below about refactoring the `logReferencedAssetFarmIds()` as a potential solution...

modules/organization/farm/src/Plugin/Validation/Constraint/LogAssetFarmValidator.php
mstenta: Makes sense! Good suggestion!

I committed this refactor as a second commit and credited you as the author: https://github.com/farmOS/farmOS/pull/1070/commits/175555eb355e6632a4f49fcd0f9c986ddf57e52a

mstenta:
Great - just committed your refactor. If everything is still green after tests run I'll merge this.

PR #1071, opened 2026-03-17
Add a bulk action for abandoning logs

Fosten:
Adds a "MarkAsAbandoned” action to the Logs View to support the new “Abandoned” status.

### DIFF
--- a/modules/core/log/config/optional/system.action.log_mark_as_abandoned_action.yml
+++ b/modules/core/log/config/optional/system.action.log_mark_as_abandoned_action.yml
@@ -1,0 +1,11 @@
+langcode: en
+status: true
+dependencies:
+  module:
+    - farm_log
+    - log
+id: log_mark_as_abandoned_action
+label: 'Mark as abandoned'
+type: log
+plugin: log_mark_as_abandoned_action
+configuration: {  }
--- a/modules/core/log/farm_log.post_update.php
+++ b/modules/core/log/farm_log.post_update.php
@@ -35,3 +35,25 @@ function farm_log_post_update_move_asset_add_log_action(&$sandbox) {
     }
   }
 }
+
+/**
+ * Add 'Mark As Abandoned' option to actions.
+ */
+function farm_log_post_update_add_abandoned_status_action(&$sandbox) {
+
+  // Create action for assigning abandoned status to logs.
+  $action = Action::create([
+    'id' => 'log_mark_as_abandoned_action',
+    'label' => t('Mark as abandoned'),
+    'type' => 'log',
+    'plugin' => 'log_mark_as_abandoned_action',
+    'configuration' => [],
+    'dependencies' => [
+      'module' => [
+        'farm_log',
+        'log',
+      ],
+    ],
+  ]);
+  $action->save();
+}
--- a/modules/core/log/src/Plugin/Action/LogMarkAsAbandoned.php
+++ b/modules/core/log/src/Plugin/Action/LogMarkAsAbandoned.php
@@ -1,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Drupal\farm_log\Plugin\Action;
+
+use Drupal\Core\Action\Attribute\Action;
+use Drupal\Core\StringTranslation\TranslatableMarkup;
+use Drupal\log\Plugin\Action\LogStateChangeBase;
+
+/**
+ * Action that marks a log as abandoned.
+ */
+#[Action(
+  id: 'log_mark_as_abandoned_action',
+  label: new TranslatableMarkup('Sets a Log as abandoned'),
+  type: 'log',
+)]
+
+class LogMarkAsAbandoned extends LogStateChangeBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  protected $targetState = 'abandoned';
+
+}
--- a/modules/core/log/config/schema/farm_log.schema.yml
+++ b/modules/core/log/config/schema/farm_log.schema.yml
@@ -1,0 +1,3 @@
+action.configuration.log_mark_as_abandoned_action:
+  type: action_configuration_default
+  label: 'Configuration for the log mark as abandoned action'


modules/core/log/src/Plugin/Action/LogMarkAsAbandoned.php
mstenta, line 1: Can you move this to a `src/Plugin/Action` subdirectory? That's the standard place for action plugins.

Also, the `namespace` should be changed from `Drupal\log\Plugin\Action` to `Drupal\farm_log\Plugin\Action`.

modules/core/log/src/LogMarkAsAbandoned.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Drupal\log\Plugin\Action;
+
+/**
+ * Action that marks a log as abandoned.
+ *
+ * @Action(
```
mstenta: We've converted all our "annotations" (eg: `@Action(...)` in the docblock comment) to "attributes" (eg: `#[Action(...)]` below the docblock comment), so this should be converted as well.

See #908 

We did this in farmOS before doing it in the Log module, but I just recently updated them there too, so you can use the Log module's 4.x branch as an example. Here is the "mark as done" action's attribute in Log 4.x: https://git.drupalcode.org/project/log/-/blob/abc187b38bce86814b1229ce3d78024b878dcac7/src/Plugin/Action/LogMarkAsDone.php#L13-17

modules/core/log/farm_log.post_update.php
```diff
@@ -35,3 +35,18 @@ function farm_log_post_update_move_asset_add_log_action(&$sandbox) {
     }
   }
 }
+
+/**
+ * Add 'Mark As Abandoned' option to actions.
+ */
+function farm_log_post_update_add_abandoned_status_action(&$sandbox) {
+  // Create action for assigning abandoned status to logs.
```
mstenta, line 44: Minor nit: can you add a newline above this comment?

modules/core/log/farm_log.post_update.php
```diff
@@ -35,3 +35,18 @@ function farm_log_post_update_move_asset_add_log_action(&$sandbox) {
...
+  $action = Action::create([
+    'id' => 'log_mark_as_abandoned_action',
+    'label' => t('Mark As Abandoned'),
+    'type' => 'log',
+    'plugin' => 'log_mark_as_abandoned_action',
+    'configuration' => [],
+  ]);
```
mstenta, line 57: This needs `dependencies` too, otherwise the action config entity will appear as "overridden" in Config Update UI.

CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [4.0.0-beta4] 2026-03-09
 
+### Changed
+
+- [Add LogMarkAsAbandoned action #1071](https://github.com/farmOS/farmOS/pull/1071)
```
mstenta: Maybe we can rename this PR to make it more clear? Perhaps: "Add a bulk action for abandoning logs #1071"?

mstenta (changes requested):
Thanks @Fosten! This is looking great! I haven't tested yet, but I left a few comments and requested a few minor changes.

This will be a nice addition in farmOS 4.1.0 alongside some other "action" features I'm working on more generally. :-)

modules/core/log/src/LogMarkAsAbandoned.php
mstenta: (Doing this will fix the failing Rector test.)

modules/core/log/src/Plugin/Action/LogMarkAsAbandoned.php
mstenta: (I _think_ doing this will fix the failing PHPUnit tests.)

CHANGELOG.md
Fosten: PR renamed and CHANGELOG.md updated in ae52699345c6776acc2354e6c04d03bc1b1190d0

modules/core/log/farm_log.post_update.php
Fosten: Newline added in e5ad501f18d15bf363fd3b36cfd9bff16ab2cdd2

modules/core/log/farm_log.post_update.php
Fosten: Dependencies added in e5ad501f18d15bf363fd3b36cfd9bff16ab2cdd2
I copied verbatim from farm_log_post_update_move_asset_add_log_action.
I wasn’t sure if there should be alterations?

modules/core/log/src/LogMarkAsAbandoned.php
Fosten: Annotations converted to attributes in 5ce3796e307ff82b9b397173babde87413794225

modules/core/log/src/Plugin/Action/LogMarkAsAbandoned.php
Fosten: File moved in 849ba16e7372194cfbb1f103dd847c50a56963db
Namespace changed in 5ce3796e307ff82b9b397173babde87413794225

modules/core/log/src/LogMarkAsAbandoned.php
Fosten: The following PHPStan errors were fixed in a0a435bf84161a0889c1cab2479206c0553828bd

  10     Attribute class Drupal\farm_log\Plugin\Action\Action does not exist.  
         🪪  attribute.notFound                                                
  12     Instantiated class Drupal\farm_log\Plugin\Action\TranslatableMarkup   
         not found.                                                            
         🪪  class.notFound

modules/core/log/src/Plugin/Action/LogMarkAsAbandoned.php
Fosten: re: the following PHPStan error

10     Class Drupal\farm_log\Plugin\Action\LogMarkAsAbandoned extends        
         unknown class Drupal\farm_log\Plugin\Action\LogStateChangeBase.       
         🪪  class.notFound (non-ignorable)                                    
         💡  Learn more at https://phpstan.org/user-guide/discovering-symbols  

Should I also include a copy of https://git.drupalcode.org/project/log/-/blob/4.x/src/Plugin/Action/LogStateChangeBase.php in my PR?

modules/core/log/farm_log.post_update.php
Fosten: re: the following PHPStan error

  54     Call to function is_null() with Drupal\system\Entity\Action will       
         always evaluate to false.                                              
         🪪  function.impossibleType                                            
         💡  Because the type is coming from a PHPDoc, you can turn off this    
         check by setting treatPhpDocTypesAsCertain: false in your phpstan.neo  
         n.

Perhaps <code>if (!is_null($action))</code> is not needed for Action::create like it was in Action::load ?

modules/core/log/src/Plugin/Action/LogMarkAsAbandoned.php
mstenta: No you just need to add the following `use` line to your class:

`use Drupal\log\Plugin\Action\LogStateChangeBase;`

This tells PHP where to find the `LogStateChangeBase` class. This is necessary for your new `LogMarkAsAbandoned` class because it's in a different namespace than `LogStateChangeBase`. The `MarkAsDone` and `MarkAsPending` classes don't need that additional `use` statement because they are in the same namespace as it, so PHP infers the correct class.

modules/core/log/config/optional/system.action.log_mark_as_abandoned_action.yml
```diff
@@ -0,0 +1,10 @@
+langcode: en
+status: true
+dependencies:
+  module:
+    - log
```
mstenta, line 6: This needs `farm_log` as well, because that provides the `abandoned` status.

modules/core/log/farm_log.post_update.php
mstenta: > Dependencies added in https://github.com/farmOS/farmOS/commit/e5ad501f18d15bf363fd3b36cfd9bff16ab2cdd2
> I copied verbatim from farm_log_post_update_move_asset_add_log_action.
> I wasn’t sure if there should be alterations?

The dependencies should be declared explicitly in the `Action::create([...])` array, so that they match what's declared in `system.action.log_mark_as_abandoned_action.yml‎` exactly. `farm_log_post_update_move_asset_add_log_action` is an unusual case because it's changing an existing action, not creating a new one.

Like this:

```php
$action = Action::create([
    'id' => 'log_mark_as_abandoned_action',
    'plugin' => 'log_mark_as_abandoned_action',
    'configuration' => [],
    'dependencies' => [
      'module' => [
        'farm_log',
        'log',
      ],
    ],
  ]);
```

I realized that we need it to depend on `farm_log` in addition to `log`, because `farm_log` provides the `abandoned` status (I left another comment in `system.action.log_mark_as_abandoned_action.yml` too).

That should also resolve the PHPStan error because that chunk of code for setting dependencies is unnecessary.

mstenta (changes requested):

modules/core/log/farm_log.post_update.php
mstenta: Oh I just realized this is also missing the `label` and `type` properties.

The array that is passed to `Action::create()` essentially needs to match exactly what is in `system.action.log_mark_as_abandoned_action.yml` (because it is literally responsible for creating it on existing sites). You can omit `langcode` and `status`, but everything else should be the same.

modules/core/log/src/Plugin/Action/LogMarkAsAbandoned.php
Fosten: Import added in 25576366bc2376a04ffde27bcb30a7e3a6236ad8

modules/core/log/farm_log.post_update.php
Fosten: Dependencies added to Action::create array in f004531088eae2dd885189f6097a0cd43be6c60b

modules/core/log/config/optional/system.action.log_mark_as_abandoned_action.yml
Fosten: farm_log added to dependencies in f004531088eae2dd885189f6097a0cd43be6c60b

Fosten:
Thanks for the pointers @mstenta.  Tests are now passing.  I haven’t run it locally, but hope it will work.

mstenta:
Awesome thanks @Fosten! Looks like the linting tests are passing (PHP Codesniffer, PHPStan, and Rector), but the functional tests (PHPUnit) are failing with the following error:

```
Drupal\Core\Config\Schema\SchemaIncompleteException: Schema errors for system.action.log_mark_as_abandoned_action with the following errors: system.action.log_mark_as_abandoned_action:configuration missing schema
```

I think I've seen this before, but I forget what causes it... I can take a look at it myself. It's probably something simple.

mstenta:
Figured it out - we just need to add a `farm_log.schema.yml` to describe the configuration schema. Pushed two commits... one that adds that file, and one that moves the configuration to `config/optional` (that's the pattern we follow in other core modules, so just for consistency... doesn't make any difference).

Let's see if there are any other failures...

mstenta:
All green now!

mstenta:
I squashed all the commits and force-pushed. I also removed the `CHANGELOG.md` line for now, since this will go into 4.1.0 and its easier to just add those when I merge all the 4.1.0 PRs.

This is ready to merge! Thanks again @Fosten!

mstenta (approved):

mstenta:
We'll also be able to expose this new action on individual log entity pages once https://www.drupal.org/project/farm/issues/2513056 is done. That will just require implementing the new `hook_farm_exposed_entity_actions()` hook. I'll try to remember to do that when I'm merging everything ahead of 4.1.0. :-)

mstenta:
> We'll also be able to expose this new action on individual log entity pages

Ah actually I just refreshed on this, and I think I decided earlier that we don't need to expose this one on individual logs. We're only going to expose the "Mark as done" action, but not the "Mark as pending". Real estate for action buttons is limited, so I'm focusing only on the very commonly needed ones. Marking a log as pending/abandoned can be done in bulk, or by editing the log directly. It's not common enough to need this extra shortcut IMO.

PR #1072, opened 2026-03-20
Add support for Markdown in default text format

mstenta:
See forum topic for more information: https://farmos.discourse.group/t/support-for-markdown-in-assets-logs-etc/2494

### DIFF
--- a/modules/core/format/config/install/filter.format.default.yml
+++ b/modules/core/format/config/install/filter.format.default.yml
@@ -1,29 +1,56 @@
 langcode: en
 status: true
-dependencies: {  }
-name: 'Default'
+dependencies:
+  module:
+    - markdown
+name: Default
 format: default
 weight: 0
 filters:
-  filter_html:
-    id: filter_html
-    provider: filter
-    status: true
-    weight: -10
-    settings:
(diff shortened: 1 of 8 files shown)


modules/core/format/config/install/filter.format.default.yml
```diff
@@ -1,29 +1,56 @@
...
-dependencies: {  }
-name: 'Default'
+dependencies:
+  module:
+    - markdown
+name: Default
 format: default
 weight: 0
 filters:
-  filter_html:
```
RMattB1, line 8: Does this remove support for html tags in existing database fields, where people may have followed the help in this Notes field example screenshot?: 
https://global.discourse-cdn.com/free1/uploads/farm_os/original/2X/a/a3534d7876934df82c06da3713967992de6c16c7.png

mstenta: The Markdown filter has its own internal HTML filtering, which allows certain tags (or maybe denies certain tags, I'm not entirely sure). I tested most of the tags listed in the old help text, and they seem to work still. 

Could use more testing though!

See my test input and screenshot of the output, which tests some HTML tags:

https://farmos.discourse.group/t/support-for-markdown-in-assets-logs-etc/2494/15

mstenta:
One thing I noticed in testing: certain elements cause the text to overlap the field label. For example:

```md
## Test notes label overlap
```

Screenshot from 2026-03-23 09-01-29

```md
> Test notes label overlap
```

Screenshot from 2026-03-23 09-01-48

Currently, all text fields default to an "`inline`" label configuration. We could change that to show the label above the field content for `text_long` fields, and that solves the issue. So I'm going to include a simple commit that makes that change.

PR #1074, opened 2026-04-22
Remove Drupal core patch that adds EntityFieldManager::rebuildBundleFieldMap()

mstenta:
This was originally added to fix the following bug: https://www.drupal.org/project/farm/issues/3314741

It introduced a Drupal core patch from this issue: https://www.drupal.org/project/drupal/issues/3129179

This is no longer necessary because of this Drupal core change: https://www.drupal.org/project/drupal/issues/3045509

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Deprecated
+
+- [EntityFieldManager::rebuildBundleFieldMap() is deprecated](https://www.drupal.org/node/3591108)
+
 ### Fixed
 
 - [Fix dashboard block title access #1075](https://github.com/farmOS/farmOS/pull/1075)
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
     "require": {
         "cweagans/composer-patches": "^1.6",
         "drupal/admin_toolbar": "^3.5",
-        "drupal/core": "11.3.8",
+        "drupal/core": "11.3.9",
         "drupal/config_update": "^2.0@alpha",
         "drupal/consumers": "^1.22",
         "drupal/csv_serialization": "^4.0.1",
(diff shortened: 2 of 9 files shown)


mstenta:
I've been following the progress of https://www.drupal.org/project/drupal/issues/3045509, so I'm opening this PR to prepare for that, but this cannot be merged until that is released in Drupal core and we update to the version that includes it.

We should also consider whether or not this is a breaking change for farmOS modules... Right now we recommend using this method (provided by the patch) in module update hooks. This PR removes that recommendation from our docs, and removes our usage of the method. If we remove the core patch, however, and there are modules that have followed our recommendation, and someone runs their update hook, it will cause an error during update.

So we may actually need to keep the patch... at least until farmOS v5.

Ideally we would also provide some way of deprecating the patched method so that module maintainers can remove it.

One way to do this would be to replace the existing patch with one of our own, which adds the same method, but it can simply be a stub that throws a deprecation warning.

mstenta:
Discussed with @symbioquine and @paul121 and we agreed the best course of action is to maintain our own Drupal core patch that adds an empty `EntityFieldManager::rebuildBundleFieldMap()` which throws a deprecation warning - so that anyone who is using it downstream will see it.

mstenta:
OK I added two commits: one that updates to Drupal 11.3.9, where this change is introduced, and one that adds our own decorator class with an empty `rebuildBundleFieldMap()` method.

This class and method are deprecated in farmOS 4.0.2 (planned for release later today). I created a drupal.org issue and change notice for these:

- https://www.drupal.org/project/farm/issues/3591107
- https://www.drupal.org/node/3591108

If tests are green, and there are no objections, I'll merge this today.

mstenta:
Decorating the service was causing a weird test failure, so I changed it to override the service instead.

The error was:

```
    1) Drupal\Tests\farm_entity\Functional\FarmEntityBundleFieldTest::testAll
    TypeError: Drupal\Component\DependencyInjection\ReverseContainer::generateServiceIdHash(): Argument #1 ($object) must be of type object, null given
```


Also, due to the change in farmOS core, we needed to add some additional lines to one of our tests to reload the `entity_field.manager` service into the container so that we don't get stale state during the test. As far as @paul121 and I can tell, this is only a test issue, not a wider issue. The error this fixes is:

```
    1) Drupal\Tests\farm_entity\Functional\FarmEntityBundleFieldTest::testAll
    Failed asserting that an array has the key 'test_contrib_hook_bundle_field'.
```

PR #1076, opened 2026-04-23
Add an asset search block to the dashboard

mstenta:
This adds a simple "Search assets" box to the farmOS dashboard, which redirects to `/assets` with the `?name=...` query parameter set to the search text.

The goal is to make it easier to find assets with fewer clicks/navigation decisions. Currently, to achieve the same thing you must find your way to `/assets` manually (by going to Records > Assets in the main menu, or clicking "Assets" in the "Metrics" block), then you need to open the "Filter" box, type in your search terms, and click "Apply". This is too many steps for such a common need.

This is only a first step. Perhaps future PRs can extend the functionality to provide additional options, or allow searching other entities as well (eg: logs).

### DIFF
--- a/modules/core/ui/farm_ui.info.yml
+++ b/modules/core/ui/farm_ui.info.yml
@@ -12,6 +12,7 @@ dependencies:
   - farm:farm_ui_map
   - farm:farm_ui_menu
   - farm:farm_ui_metrics
+  - farm:farm_ui_search
   - farm:farm_ui_term
   - farm:farm_ui_theme
   - farm:farm_ui_user
--- a/modules/core/ui/farm_ui.post_update.php
+++ b/modules/core/ui/farm_ui.post_update.php
@@ -15,3 +15,12 @@ function farm_ui_post_update_install_farm_ui_term() {
     \Drupal::service('module_installer')->install(['farm_ui_term']);
   }
 }
+
+/**
+ * Install the farm_ui_search module.
+ */
+function farm_ui_post_update_install_farm_ui_search() {
+  if (!\Drupal::service('module_handler')->moduleExists('farm_ui_search')) {
+    \Drupal::service('module_installer')->install(['farm_ui_search']);
+  }
+}
--- a/modules/core/ui/search/farm_ui_search.info.yml
+++ b/modules/core/ui/search/farm_ui_search.info.yml
@@ -1,0 +1,8 @@
+name: farmOS UI Search
+description: Adds a dashboard pane with search features.
+type: module
+package: farmOS UI
+core_version_requirement: ^11
+dependencies:
+  - farm:farm_ui_dashboard
+  - farm:farm_ui_views
--- a/modules/core/ui/search/src/Form/AssetSearchForm.php
+++ b/modules/core/ui/search/src/Form/AssetSearchForm.php
@@ -29,6 +29,7 @@ class AssetSearchForm extends FormBase {
       '#type' => 'textfield',
       '#title' => $this->t('Asset search'),
       '#title_display' => 'invisible',
+      '#placeholder' => $this->t('Search assets'),
     ];
 
     // Search submit button.
--- a/modules/core/ui/search/src/Form/AssetSearchForm.php
+++ b/modules/core/ui/search/src/Form/AssetSearchForm.php
@@ -1,0 +1,57 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Drupal\farm_ui_search\Form;
+
+use Drupal\Core\Form\FormBase;
+use Drupal\Core\Form\FormStateInterface;
+
+/**
+ * Asset search form.
+ */
+class AssetSearchForm extends FormBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getFormId() {
+    return 'farm_ui_search_asset_form';
+  }
+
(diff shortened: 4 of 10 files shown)


mstenta:
Notes from the dev call:

- Would be great if it could use autocomplete, and if you select a specific asset it sends you directly to that asset (rather than /assets?name=...)
- Reduce vertical space
  - Move "Search" button inline
  - Move title into placeholder text
- Maybe move to top region?
- Or move to sidebar, when Drupal navigation is available?
- This only works on name search.

mstenta:
I pushed a few fixup commits for consideration, which makes changes discussed on dev call:

> - Reduce vertical space
>   - Move "Search" button inline
>   - Move title into placeholder text
> - Maybe move to top region?

This is what it currently looks like:

Screenshot from 2026-04-23 13-59-20

mstenta:
> Would be great if it could use autocomplete, and if you select a specific asset it sends you directly to that asset (rather than /assets?name=...)

I test this and it's a little more complicated than I anticipated...

The easiest way to make it into an autocomplete is to change the search text box `#type` from `textfield` to `entity_autocomplete`. This enables the autocomplete search functionality, so that matching assets are visible in a dropdown as you type.

The problem is the way the `entity_autocomplete` form widget processes the input during submission. I would like to be able to get an asset ID if a specific asset is selected, but I would also like to be able to get the raw text string if not. This would allow me to redirect to either `/asset/[id]` (specific asset) or `/assets` (pre-filtered list of assets). However, the `entity_autocomplete` widget is always trying to interpret it as a single asset, and it returns the asset ID in the `$form_state` values. If an asset is not found, it throws a validation error before it even gets to submission:

Screenshot from 2026-04-23 15-41-38

I'm sure we can still make it work, but it's not going to be as simple as I hoped, and will probably require some custom autocomplete logic...

mstenta:
> I'm sure we can still make it work, but it's not going to be as simple as I hoped, and will probably require some custom autocomplete logic...

I say we start with the simple approach, and explore this in a follow-up.

symbioquine:
I'm curious what the rational is for making this a dependency of `farm_ui` instead of a standalone module? Naively, it seems like it would be nice for folks to be able to disable it if they don't need search or want to substitute their own search UI implementation...

paul121:
> Naively, it seems like it would be nice for folks to be able to disable it if they don't need search or want to substitute their own search UI implementation...

Agreed - and with the new/improved setup functionality it would be easier for users to enable this if they want? I guess the challenge is that if we enable it by default, they can't disable later? :thinking: 

Here is a quick screenshot of the Rothamsted dashboard search we implemented. It does something similar except opens the results in a modal (not the standard view, just a simplified table with name and  status). Few suggestions:

- Change placeholder text to say "Search by asset name" to make it clear you aren't searching across all asset fields
- Consider centering search block content
- Consider adding background to the block. I forget if this is custom to Rothamsted or something we have on all/most dashboard blocks? :thinking: 

rothamsted-search-example

mstenta:
> I'm curious what the rational is for making this a dependency of `farm_ui` instead of a standalone module?

The `farm_ui` module is basically just a shortcut we use for installing/"locking" other "UI" (`farm_ui_*`) modules. It doesn't do anything itself - it simply depends on the other `farm_ui_*` sub-modules. farmOS installs `farm_ui` as part of the "base" modules, which then installs all the sub-modules.

> Naively, it seems like it would be nice for folks to be able to disable it if they don't need search or want to substitute their own search UI implementation...

Yep that's exactly what the `farm_ui` module strategy allows. If you want to customize, simply uninstall `farm_ui`, and then you have the ability to uninstall any of the `farm_ui_*` modules. It basically just serves as a "lock" - and if you want to take more control, you can "unlock" by uninstalling that one.

Maybe there are better ways to do this (eg: via the Setup Wizard, as @paul121 suggested), but that should be a separate reorganization task IMO. This PR follows the existing pattern, so I don't think it needs to wait for that. Would love to start brainstorming how that might work in a forum topic though!

symbioquine:
> Yep that's exactly what the `farm_ui` module strategy allows. If you want to customize, simply uninstall `farm_ui`, and then you have the ability to uninstall any of the `farm_ui_*` modules. It basically just serves as a "lock" - and if you want to take more control, you can "unlock" by uninstalling that one.

I guess that makes sense. My brain was filling in other stuff living in the base `farm_ui` module, but if it's just a [meta ~~package~~ [module]](https://help.ubuntu.com/community/MetaPackages) then an admin can just manually install a subset of its dependencies and get the same outcome minus the undesired features.

> Maybe there are better ways to do this (eg: via the Setup Wizard, as @paul121 suggested)

Assuming I'm understanding what you've just described, it seems like the current structure could just be leveraged by the setup wizard code... i.e. the top-level question could be "Do you need to customize which farmOS UI elements get installed?" If you leave it unchecked, it installs the `farm_ui` module. If you check it, then it gives you checkboxes for each of the dependencies of `farm_ui` and leaves `farm_ui` itself uninstalled.

mstenta:
Yea something like that would work!

mstenta:
On the community call yesterday we discussed potentially leap-frogging this PR and exploring https://www.drupal.org/project/search_api as a more general first-step solution...

mstenta:
Closing this in favor of #1085.

PR #1077, opened 2026-04-23
Build ARM Docker images separately

mstenta:
This PR changes the way that we build the AMD64 vs ARM Docker images. Previously, they were all built in parallel as a prerequisite to running tests. The AMD64 image takes ~2.5 minutes to build, but the ARM images take ~20 minutes, so this is a long time to wait before tests start running. This splits the AMD64 `build` step out, so it runs by itself, and renames the ARM step to `build-arm`. Most other steps depend on the AMD64 image, but not the ARM images, so they can continue depending on `build`. Then, down in the `publish-multi-arch-images` step, this adds an additional dependency on the new `build-arm` step, since it is responsible for pushing all of the images to Docker Hub.

### DIFF
--- a/.github/workflows/deliver.yml
+++ b/.github/workflows/deliver.yml
@@ -19,19 +19,63 @@ env:
 jobs:
 
   build:
-    name: Build Docker images
+    name: Build Docker AMD64 images
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout the repository
+        uses: actions/checkout@v6
+      - name: Set default FARMOS_REPO and FARMOS_VERSION
+        run: |
+          echo "FARMOS_REPO=${GITHUB_REPOSITORY}" >> $GITHUB_ENV
+          echo "FARMOS_VERSION=${FARMOS_DEV_BRANCH}" >> $GITHUB_ENV
+      - name: Set FARMOS_VERSION for branch push event
+        if: github.event_name == 'push' && github.ref_type == 'branch'
+        run: echo "FARMOS_VERSION=${GITHUB_REF:11}" >> $GITHUB_ENV
+      - name: Set FARMOS_VERSION for tag push event
+        if: github.event_name == 'push' && github.ref_type == 'tag'
+        run: echo "FARMOS_VERSION=${GITHUB_REF:10}" >> $GITHUB_ENV
+      - name: Set FARMOS_VERSION and FARMOS_REPO for pull request event
+        if: github.event_name == 'pull_request'
+        run: |
+          echo "FARMOS_VERSION=${GITHUB_HEAD_REF}" >> $GITHUB_ENV
+          echo "FARMOS_REPO=${{ github.event.pull_request.head.repo.full_name }}" >> $GITHUB_ENV
+      # This builds the production/dev Docker images using the specified FARMOS_VERSION,
+      # but notably it does NOT override the default PROJECT_VERSION, so the
+      # farmOS Composer project dev branch is always used.
+      - name: Build and save the farmOS Docker image
+        run: |
+          COMMON_BUILD_ARGS="--build-arg FARMOS_REPO=https://github.com/${FARMOS_REPO} --build-arg FARMOS_VERSION=${FARMOS_VERSION}"
+          PROD_IMG_NAME=${DOCKER_IMG_PREFIX}:${FARMOS_DEV_BRANCH}-amd64
+          DEV_IMG_NAME=${DOCKER_IMG_PREFIX}:${FARMOS_DEV_BRANCH}-dev-amd64
+          docker build ${COMMON_BUILD_ARGS} -t ${PROD_IMG_NAME} docker
+          docker build ${COMMON_BUILD_ARGS} -t ${DEV_IMG_NAME} --target dev docker
+          docker save ${PROD_IMG_NAME} > /tmp/farmos-amd64.tar
+          docker save ${DEV_IMG_NAME} > /tmp/farmos-dev-amd64.tar
+      - name: Cache the farmOS Docker image
+        uses: actions/cache@v5
+        with:
+          path: /tmp/farmos-amd64.tar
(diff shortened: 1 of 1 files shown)


wotnak:
For arm64 build there's also an option to run it on an arm runner to build it natively without the need for quemu emulation which should be much faster. `runs-on: ubuntu-24.04-arm` https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories

mstenta:
Oh interesting! Thanks for sharing that @wotnak! That would be easier to do now that the AMD/ARM builds are split...

I think we would still need `qemu` for `arm32v7`, though, right? If so, then we might need three separate build steps (and the `arm32v7` would take the longest to run).

I suppose we could also consider also dropping support for `arm32v7` in the future... which would bring down the overall build time. That would be a breaking change, though, so would need to wait for farmOS 5.x. :thinking:

mstenta:
> I think we would still need `qemu` for `arm32v7`, though, right? If so, then we might need three separate build steps (and the `arm32v7` would take the longest to run).

I pushed a commit that splits the `build-arm` step into `build-arm64` (on `ubuntu-22.04-arm`) and `build-arm32` (on `ubuntu-22.04` with `qemu`).

It works! The `build-arm64` step took ~4.5 minutes, as opposed to 23 minutes for `build-arm32`.

> I suppose we could also consider also dropping support for `arm32v7` in the future... which would bring down the overall build time. That would be a breaking change, though, so would need to wait for farmOS 5.x. 🤔

What do others think about this? Would it make sense to deprecate the `arm32v7` builds in this PR as well, so that we can plan to remove that step in 5.x?

wotnak:
I think deprecating arm32 should be fine. Maybe hub.docker.com provides some statistics when was the last time someone pulled arm32 version of images? It could give us an idea if anyone is actually using them.

mstenta:
> Maybe hub.docker.com provides some statistics when was the last time someone pulled arm32 version of images? It could give us an idea if anyone is actually using them.

Good thought. I checked and the only information I could find was "less than 1 day" for the "last pull" of the `arm32-v7` images... but that's the same for all of our images, so I'm not sure how reliable it is. I couldn't find any more detailed statistics about it. :thinking:

mstenta:
Update: I found a way to export raw data about pulls. I think it has the info we need, but might take some parsing...

symbioquine:
> I think we would still need `qemu` for `arm32v7`, though, right?

I'm not totally convinced of this. Even if it's true, I'd guess the emulation might be cheaper (doing arm32v7 on aarch64 instead of amd64 should be a simpler translation since the instructions are supposed to be a subset right?).

> I think deprecating arm32 should be fine. Maybe hub.docker.com provides some statistics when was the last time someone pulled arm32 version of images? It could give us an idea if anyone is actually using them.

Maybe it would also be worth a microblog post to the effect of "Hey, is anybody using farmOS via Docker on hardware like the Raspberry Pi Zero or Raspberry Pi 3b (and committed/stuck on the 32bit OS)?"

mstenta:
> Maybe it would also be worth a microblog post to the effect of "Hey, is anybody using farmOS via Docker on hardware like the Raspberry Pi Zero or Raspberry Pi 3b (and committed/stuck on the 32bit OS)?"

https://fosstodon.org/@farmOS/116476604509679551

mstenta:
> > I think we would still need `qemu` for `arm32v7`, though, right?
> 
> I'm not totally convinced of this. Even if it's true, I'd guess the emulation might be cheaper (doing arm32v7 on aarch64 instead of amd64 should be a simpler translation since the instructions are supposed to be a subset right?).

I tested both:

**a) building `arm32v7` and `arm64v8` together in the same build step without `qemu`...**

https://github.com/mstenta/farmOS/commit/fce25a743b8533d56cc608c56821921b0732c1d0

https://github.com/mstenta/farmOS/actions/runs/24995322754

Build times:

- `amd64`: 3m 40s
- `arm64v8`: 4m 44s
- `arm32v7`: 4m 49s

**b) building them separately, both on `ubuntu-22.04-arm`, using `qemu` for `arm32v7`...**

https://github.com/mstenta/farmOS/commit/fbdaa5f681149f21d238f913d16060473fab6081

https://github.com/mstenta/farmOS/actions/runs/24995236221

- `amd64`: 3m 40s
- `arm64v8`: 4m 18s
- `arm32v7`: 7m 16s

Both seem to have built successfully!!

The former (a) is a simpler change than the latter (b), because it doesn't require splitting the ARM images into separate steps at all. And it's faster than running with `qemu`!

So it seems like a win-win!

@symbioquine Do you want to take a quick peek at those to make sure I didn't mess anything up? I'm not actually testing these images. Is it possible that I accidentally built two `arm64v8`s in (a)?

symbioquine:
> @symbioquine Do you want to take a quick peek at those to make sure I didn't mess anything up? I'm not actually testing these images. Is it possible that I accidentally built two `arm64v8`s in (a)?

It looks pretty close... I'm seeing "arm" and "hf" references in the package repos and abi detection logs during build for the one that's supposed to be arm32v7

image

My only concern is that the non-qemu version shows some logs like 

> `#10 6.153 checking target system type... armv8l-unknown-linux-gnueabihf`

whereas the qemu version shows

> `#11 1.838 checking target system type... arm-unknown-linux-gnueabihf`

That makes me think we might be inadvertently targeting something between arm32v7 and aarch64.

If we're going to keep the "arm32" build, we should go for the lowest common denominator which I think is arm32v7. So maybe for now keep it in qemu because that's a known quantity instead of trying to figure out how to set our target architecture some other way?

mstenta:
Ah good eye @symbioquine - thanks for reviewing!

I agree, we may as well just keep `qemu` for the `amd32v7` image. A 7 minute build time is still a whole lot better than 20+!

I'll add that commit to this branch.

mstenta:
I think this is ready for final review!

symbioquine (approved):
LGTM!

PR #1081, opened 2026-05-24
Provide a trait for building inline form containers

mstenta:
See https://www.drupal.org/project/farm/issues/3591769

### DIFF
--- a/modules/core/form/src/Traits/FarmFormInlineContainerTrait.php
+++ b/modules/core/form/src/Traits/FarmFormInlineContainerTrait.php
@@ -1,0 +1,29 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Drupal\farm_form\Traits;
+
+/**
+ * Provides a standard method for creating inline containers.
+ */
+trait FarmFormInlineContainerTrait {
+
+  /**
+   * Build an inline container element.
+   *
+   * @return array
+   *   Returns a render array.
+   */
+  public function buildInlineContainer() {
+    return [
+      '#type' => 'container',
+      '#attributes' => [
+        'class' => [
+          'inline-container',
+        ],
+      ],
+    ];
+  }
+
+}
--- a/modules/core/quick/src/Traits/QuickFormElementsTrait.php
+++ b/modules/core/quick/src/Traits/QuickFormElementsTrait.php
@@ -4,26 +4,19 @@ declare(strict_types=1);
 
 namespace Drupal\farm_quick\Traits;
 
+use Drupal\farm_form\Traits\FarmFormInlineContainerTrait;
+
 /**
  * Provides methods for building common quick form elements.
+ *
+ * @deprecated in farm:4.1.0 and is removed from farm:5.0.0.
+ *   Use FarmFormInlineContainerTrait instead.
+ * @see https://www.drupal.org/node/3591770
+ *
+ * @phpstan-ignore trait.unused
  */
 trait QuickFormElementsTrait {
 
-  /**
-   * Build an inline container element.
-   *
-   * @return array
-   *   Returns a render array.
-   */
-  public function buildInlineContainer() {
-    return [
-      '#type' => 'container',
-      '#attributes' => [
-        'class' => [
-          'inline-container',
-        ],
-      ],
-    ];
-  }
+  use FarmFormInlineContainerTrait;
 
 }
--- a/modules/core/quick/tests/src/Kernel/QuickFormTestBase.php
+++ b/modules/core/quick/tests/src/Kernel/QuickFormTestBase.php
@@ -63,6 +63,7 @@ abstract class QuickFormTestBase extends KernelTestBase {
     'farm_entity_access',
     'farm_entity_fields',
     'farm_field',
+    'farm_form',
     'farm_format',
     'farm_geo',
     'farm_location',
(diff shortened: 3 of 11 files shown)


modules/core/form/src/Traits/FarmFormInlineContainerTrait.php
```diff
@@ -0,0 +1,29 @@
...
+   *   Returns a render array.
+   */
+  public function buildInlineContainer() {
+    return [
+      '#type' => 'container',
+      '#attributes' => [
+        'class' => [
+          'inline-container',
+        ],
+      ],
```
paul121, line 25: Not sure why we didn't do this before but should be able to attach the library here in this render array - that way we don't have to add the library to every form, even if they aren't using inline container?

paul121 (changes requested):
This seems like a fine change! I would be curious about making this available outside of forms and for any render array, too. I think the issue with doing is that there isn't a good module to host that functionality (for example, it would require everything to depend on `farm_ui_*`) which we wanted to avoid...

I guess that makes me wonder if there is any reason someone might not want to enable `farm_form` module. It looks like the behavior it provides is already configurable. Maybe something to consider?

modules/core/form/src/Traits/FarmFormInlineContainerTrait.php
mstenta: Ah I like this idea, but the library+CSS is not actually provided by the `farm_form` module. It is provided by `farm_ui_theme`, because that's where we try to isolate all theme-specific CSS.

It raises the question, though: how specific is this CSS to Gin? :thinking: 

Another question that comes to mind, of course, is: if [Gin is going to become part of Drupal core](https://www.drupal.org/about/core/blog/drupal-core-will-adopt-gin-admin-theme-to-replace-claro), should we reconsider `farm_ui_theme` entirely and create more of a hard-dependency on Gin? That would allow us to move a lot of stuff out of `farm_ui_theme` and into other modules... but it's a big change/decision, and would likely require waiting for 5.x. So maybe just something to think about, but probably not a short term option...

mstenta:

> I would be curious about making this available outside of forms and for any render array, too.

Good question! This hasn't come up as a request/need yet, but it's worth considering.

> I think the issue with doing is that there isn't a good module to host that functionality (for example, it would require everything to depend on `farm_ui_*`) which we wanted to avoid...

Yea - we just recently introduced the `farm_form` module, so this felt like a nice small next step (moving the inline container trait out of `farm_quick` to `farm_form` so it is more generally reusable). And yea, `farm_ui_*` modules generally should be considered "higher level" modules (not depended upon by "lower level" modules), so that they can be turned off if more customization is required.

This particular case (inline elements) is simple, but it does illustrate the fuzziness of some of these lines, and you're right to think about it as even lower-level than forms (it's about render array elements more generally).

> I guess that makes me wonder if there is any reason someone might not want to enable `farm_form` module. It looks like the behavior it provides is already configurable. Maybe something to consider?

I don't see any reason to uninstall the `farm_form` module, in its current state. It was added in order to provide a place for the "form protection" feature in #1025, and that also added an option to disable the form protection (without uninstalling the module). I have some ambitions in mind for additional form elements that the `farm_form` module can provide in the future... so I consider it a "lower level" module that can be depended on by others, and perhaps over time more core modules will depend on it. It is already depended on by `farm_quick` and `farm_ui_theme` for the form protection functionality.

mstenta:
> I think the issue with doing is that there isn't a good module to host that functionality

Trying to think if there's another place to put it... :thinking: 

I suppose in its current state nothing is preventing someone from using this outside of forms... it's just a bit odd that it's named `FarmFormInlineContainerTrait`.

Oh... actually no... because the CSS is for `form .inline-container`...

Hmm.

mstenta:
At the risk of this spiraling out and dying of analysis paralysis... if we are going to consider expanding the scope of this (to render arrays more generally), I wonder if the CSS is too simple? Will there be cases where it behaves oddly, depending on the size/complexity of the render arrays that are being "inlined"? :thinking:

Maybe there's an argument for constraining the scope, and not supporting wider use (yet)?

modules/core/form/src/Traits/FarmFormInlineContainerTrait.php
paul121: > Ah I like this idea, but the library+CSS is not actually provided by the farm_form module. It is provided by farm_ui_theme

Ah shoot, I didn't catch that.

I think I'm remembering this now. The fully correct way to do this would be to provide a new render element + template somewhere within farmOS core (in a module that can be expected to be enabled all the time). It can provide an initial, simple CSS implementation (likely what we already have). This way the element becomes first-class and essentially required within farmOS.

If someone is using a different theme, then they can implement/alter/override the rendering of this element as necessary. `farm_ui_*` might do this itself to workout and conflicts with Drupal core themes.

mstenta:
FYI one of the reasons I wanted this recently was for https://github.com/farmOS/farmOS/pull/1076 - which currently replicates the same CSS as the `farm_ui_theme/quick` library.

mstenta:
> FYI one of the reasons I wanted this recently was for #1076 - which currently replicates the same CSS as the `farm_ui_theme/quick` library.

#1076 has been closed in favor of #1085.

#1076 was the primary motivation for doing this (generalizing the inline form element styling). And the discussion here suggests that we might want to consider a broader solution...

Converting this back to a draft in the meantime...

PR #1082, opened 2026-05-24
Remove AutowireTrait usage from hook classes

paul121:
`AutowireTrait` only implements the `::create` method which is not needed in hook classes. There is some other magic in core that wires these dependencies just using the class constructor.


mstenta:
Nice! Good catch @paul121!

PR #1085, opened 2026-06-12
Add a unified asset+log search

mstenta:
This adds a unified asset+log search, which indexes the `name` and `notes` fields of all assets and logs, making it possible to quickly search records by keywords/phrases.

It accomplishes this with the [Search API](https://www.drupal.org/project/search_api) module, which provides a robust search framework for Drupal entities. Search API has a swappable backend plugin system, which means that it's possible to use Apace Solr, ElasticSearch/OpenSearch, etc as a server backend, or you can use the existing Drupal SQL database. This PR adds a `default` server and index using the existing SQL database.

Very little code is necessary to accomplish this, and this PR is mostly YML configuration for the `default` Search API server and index, and a View for searching assets and logs, with a page display at `/search`, and a block display that is added to the farmOS dashboard.

Screenshots:

Screenshot from 2026-06-11 20-03-13

Screenshot from 2026-06-11 20-03-28

Screenshot from 2026-06-11 20-03-35

This PR is a replacement for (and improvement over) #1076. I started with that one and modified/simplified it for this PR.

I am opening this as a draft / working proof of concept. There is more to do, including:

- [ ] Review Search API and Views configuration with community
- [x] Test/understand access control considerations
- [ ] Consider whether we should index other fields in this first pass, or wait and see...
- [ ] ...?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -39,6 +39,7 @@
         "drupal/migrate_source_csv": "^3.6",
         "drupal/migrate_tools": "^6.1.2",
         "drupal/role_delegation": "^1.3",
+        "drupal/search_api": "^1.41",
         "drupal/simple_oauth": "^6.1.1",
         "drupal/simple_oauth_password_grant": "^2.1",
         "drupal/state_machine": "^1.12",
@@ -72,6 +73,9 @@
             },
             "drupal/gin": {
                 "Issue #3342164: Remove implicit dependency on node module for gin content form.": "https://www.drupal.org/files/issues/2024-01-15/3342164-6.patch"
+            },
+            "drupal/search_api": {
+                "Issue #3598014: PluginNotFoundException when creating index for entity type with no bundles installed": "https://www.drupal.org/files/issues/2026-06-11/3598014-6.patch"
             },
             "drupal/subrequests": {
                 "Issue #3524429: PHP 8.4 deprecation: Implicitly marking parameter as nullable": "https://git.drupalcode.org/issue/subrequests-3524429/-/commit/784ccfe8af7d803f392263d2dbf7dcd89f9566d5.patch"
--- a/composer.json
+++ b/composer.json
@@ -39,7 +39,7 @@
         "drupal/migrate_source_csv": "^3.6",
         "drupal/migrate_tools": "^6.1.2",
         "drupal/role_delegation": "^1.3",
-        "drupal/search_api": "^1.41",
+        "drupal/search_api": "1.41",
         "drupal/simple_oauth": "^6.1.1",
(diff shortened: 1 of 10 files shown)


mstenta:
I force-pushed to remove the `farm_ui_theme/search` CSS that was copied from #1076, but is no longer necessary here.

mstenta:
FYI this includes a patch to fix a minor bug I discovered in Search API that was preventing me from installing a default search index before asset/log types are installed: https://www.drupal.org/project/search_api/issues/3598014

mstenta:
> - [ ] Test/understand access control considerations

I'm testing access to search results and it seems to be working pretty well out of the box with our "Entity Collection" Views access handler!

I wasn't sure what to expect, given the prominent warning on the Search API project page:

> ## Note on security
> 
> In general, the Search API doesn't provide any kind of access restrictions (as this cannot be done on a generic level). It is therefore your responsibility to take care that only accessible items are indexed or results displayed (e.g., by using appropriate filters).
> 
> There is, however, built-in support for node access checks via the [Node access](https://www.drupal.org/node/1254452#search_api_node_access) data alteration (q.v. for details). Also, the Search views module contains an option to do additional access checks on all entities in search views.

This is what I've tested:

As the admin user, I created one asset and one log (see the screenshots above).

I created a "test" user with a "test" role and only assigned the `access farm dashboard` permission to it. With that, the search block is not visible on the dashboard, and `/search` shows "Access denied".

The View is configured to use our "Entity Collection" access plugin, with both `asset` and `log` entity types selected. This access plugin was added in #965, and for each entity type selected it requires *both* the `access [entity-type] collection` *and* another permission for that type, such as `view [any/own] [entity-type]` or `view [any/own] [bundle] [entity-type]`. It then filters out results in the View's SQL query accordingly.

So I added the `access asset collection`, `access log collection`, `view own asset`, and `view own log` permissions to the test role, and this makes both the block and `/search` available. But searching for any keywords shows "No assets or logs found." because the test asset and log were created by the admin user. I also tested this with `view own land asset` and `view own activity log` and still the results are hidden. If I add `view any land asset` and `view any activity log` then both test entities show in the results (same as what the admin users sees in the screenshot above).

One thing I noticed, however, is that when the View includes a "Result summary" in the footer (which we include on other Views - it shows a count of results and pages), it shows a count of matching results that *the user does not have access to*. So I removed that from the Views footer configuration.

So this appears to be working pretty well!

mstenta:
I'm going to mark this as "Ready for review" since access control seems to be working as expected, and tests are green.

I would still like to review with the community to see if there are any additions/adjustments we should consider before merging (perhaps on next week's dev call).

PR #1088, opened 2026-06-15
Add the Diff module for comparing entity revisions

mstenta:
The Drupal [Diff](https://www.drupal.org/project/diff) module recently merged a long-awaited feature request: [Offer a revisions tab for all entities](https://www.drupal.org/project/diff/issues/2452523)

This adds UI for comparing the differences between entity revisions. Previously this was only available for `node` entities.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,6 +23,7 @@
         "drupal/consumers": "^1.22",
         "drupal/csv_serialization": "^4.0.1",
         "drupal/date_popup": "^2.0",
+        "drupal/diff": "^2.1.0",
         "drupal/entity": "1.6",
         "drupal/entity_browser": "^2.12",
         "drupal/entity_reference_integrity": "^2.0",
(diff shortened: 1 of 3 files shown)


symbioquine (approved):
Makes sense! Probably in the future we should document revisions (and this diff feature) better. I only see one reference to revisions in our docs - in the context of the log entity.

mstenta:
Thanks for reviewing @symbioquine! I'd like to take a look at this with you and others on the dev call too - and see if there are configuration settings we want to include. This is pretty barebones (just adding/installing the module) as is.

> Probably in the future we should document revisions (and this diff feature) better. I only see one reference to revisions in our docs - in the context of the log entity.

Agreed! I'll probably end up writing about it a bit in the 4.1.0 blog post too - so maybe we can copy some of that to docs if it makes sense.

Issue #1089, opened 2026-06-18
plan_record access discards the parent plan's access cacheability (PlanRecordAccess)


gbathree:
Hey Mike - working on including Plans in multi-farm access control.  Copying the way Miriam set it up already with assets.  But identified (I think) an issue that may need up-stream review with regard to catching.  It feels minor to me, but as a non-dev I'm doing a lot of guessing about this level of details.  Here's the fix it's offering:

--> claude --> 

### Problem

`PlanRecordAccess::checkAccess()` delegates a record's access to its parent plan, but discards the plan access result's cacheability:

https://github.com/farmOS/farmOS/blob/4.x/modules/core/plan/src/Access/PlanRecordAccess.php

```php
if ($entity instanceof PlanRecordInterface && $plan = $entity->getPlan()) {
  return AccessResult::allowedIf($plan->access($operation, $account));
}
```

`$plan->access($operation, $account)` returns a **bool** (the default `$return_as_object = FALSE`), so `AccessResult::allowedIf()` produces a fresh result with **no cache metadata** — none of the plan access handler's cache contexts, tags, or max-age. The plan_record access result is therefore cached without any dependency on what the plan's access actually depends on.

### Impact

Whenever a plan's access varies by cache context or is invalidated by a cache tag, the plan_record's access can go **stale**:

- An access handler that grants/denies a plan by group membership (or any per-user/per-context factor): the plan re-evaluates when membership changes (its result carries the right tags/contexts), but the plan_record keeps the stale grant.
- Any contrib/custom plan access handler that calls `cachePerUser()` / adds cache tags — all of it is lost at the plan_record boundary.

(Surfaced while adding membership-scoped plan access downstream: the plan correctly invalidated on a membership change but the plan_record did not. Related: the plan-access discussion in Discourse topic 2547.)

### Proposed fix

Preserve the plan access result's cacheability. Minimal and behavior-preserving:

```php
if ($entity instanceof PlanRecordInterface && $plan = $entity->getPlan()) {
  $plan_access = $plan->access($operation, $account, TRUE);
  return AccessResult::allowedIf($plan_access->isAllowed())->addCacheableDependency($plan_access);
}
```

This keeps the current allowed/neutral semantics exactly, but carries the plan access result's cache metadata.

Alternatively, return the plan's access result object directly:

```php
return $plan->access($operation, $account, TRUE);
```

[...]

PR #1093, opened 2026-06-24
Add a Farm reference field to plans

mstenta:
This adds a `farm` reference field to `plan` entities, for referencing `organization` entities of type `farm`. This is identical to the way that we add a `farm` reference field to `asset` entities. These reference fields are only added if the `farm_farm` module is installed (which provides the `farm` `organization` bundle).

The reason to do this is so that modules like `farm_multitenant` (https://www.drupal.org/project/farm_multitenant) can add access control to `plan` entities based on the `farm` they are associated with.

See this comment (and surrounding discussion) for more context and reasoning: https://farmos.discourse.group/t/question-about-moving-intake-process-into-multi-tenant-infrastructure/2547/6

Notably, this PR does NOT currently add any field constraint validation logic to this new reference field like we do already for `asset` entities. This is worth discussing and considering the ramifications in more depth before merging this. To summarize both sides of thinking: a) it is technically challenging to add constraint validation because we don't know what reference fields downstream modules will add to their `plan` and `plan_record` bundles, but b) if we don't, then it opens the possibility for data to cross the line between different farm organizations. Perhaps that should be the responsibility of the downstream modules to maintain that separation (or not). But if we allow it to start then there's no going back. This is why we decided to be conservative in #849 and add the constraints to asset `farm` reference fields, to prevent any cross-farm relationships.

### DIFF
--- a/modules/organization/farm/src/Hook/FieldHooks.php
+++ b/modules/organization/farm/src/Hook/FieldHooks.php
@@ -29,20 +29,22 @@ class FieldHooks {
   public function entityBaseFieldInfo(EntityTypeInterface $entity_type) {
     $fields = [];
 
+    // Define common Farm reference field options.
+    $options = [
+      'type' => 'entity_reference',
+      'label' => $this->t('Farm'),
+      'description' => $this->t('What farm is this associated with?'),
+      'target_type' => 'organization',
+      'target_bundle' => 'farm',
+      'multiple' => FALSE,
+      'weight' => [
+        'form' => 55,
+        'view' => -5,
+      ],
+    ];
+
     // Add a Farm reference field to assets.
     if ($entity_type->id() == 'asset') {
-      $options = [
-        'type' => 'entity_reference',
-        'label' => $this->t('Farm'),
-        'description' => $this->t('What farm is this associated with?'),
-        'target_type' => 'organization',
-        'target_bundle' => 'farm',
-        'multiple' => FALSE,
-        'weight' => [
-          'form' => 55,
-          'view' => -5,
-        ],
-      ];
       $fields['farm'] = $this->farmFieldFactory->baseFieldDefinition($options);
 
       // Add a constraint to ensure that assets are in the same farm as their
@@ -61,6 +63,12 @@ class FieldHooks {
         $fields['farm']->addConstraint('AssetGroupAssignmentFarm');
       }
     }
+
+    // Add a Farm reference field to plans.
+    elseif ($entity_type->id() == 'plan') {
+      $fields['farm'] = $this->farmFieldFactory->baseFieldDefinition($options);
+    }
+
     return $fields;
   }
 


gbathree:
Thanks Mike for making this.

I think the currently unconstrained nature of `plan` makes it logical that it is similarly unconstrained in how it handles organization-farm`.  I think it's worth highlighting anywhere in the code (and in the docs) that developers working with plans can break things, and even more so it's probably worth writing up something for the docs with a lot of SHOULD type language (like... "If you are creating a module using plans, you SHOULD ensure any entities you create or modify have or retain the correct connection to the farm organization", etc.).

If I have suggestions like this, especially going into core, for documentation, where should I put that?

Otherwise, I'll review this today and get back (probably via claude, they'll post here).

gbathree:
Reviewed from the downstream side — `quickform_plans` (farmOS soil-health) is one of the modules this is aimed at. A few notes, a take on the constraint question, and an answer to the docs question.

**This is exactly the primitive we needed.** We had to hand-roll a `plan.farm` config field plus a self-deactivating `PlanAccessControlHandler` to scope plans by organization, precisely because core had no plan→farm reference (the cross-org plan leak from [Discourse #2547](https://farmos.discourse.group/t/question-about-moving-intake-process-into-multi-tenant-infrastructure/2547)). With this field in core we can delete our local field, and once `farm_multitenant` scopes plans off it we can retire our interim handler entirely. So a strong +1 — and +1 specifically on gating it behind `farm_farm` (our local field wrongly depended on `farm_multitenant`; this puts it in the right place).

**On the constraint (the #849-style validation) — I'd land where Greg did (leave it unconstrained in core), but for a specific reason worth stating.** The cross-farm risk isn't really in the plan's own `farm` field; it's in the entities the plan's `plan_record`s reference (logs, assets, terms). A constraint on `plan.farm` alone wouldn't catch a `plan_record` that points at another farm's log, and you can't write that constraint in core because those reference fields are downstream-defined — which is your point (a). Where we found separation actually belongs is the **access layer, not field validation**: scope the plan by farm and let `plan_record` inherit the decision (this is also why we filed #1089 / #1095 — `PlanRecordAccess` delegates to the plan but drops its cacheability). That generalizes without core having to know downstream field shapes. The one guardrail that *is* core's to keep — and the PR already has it — is `target_bundle: farm`, so a plan can't reference a non-farm org. And the "once you allow it there's no going back" logic cuts both ways: a baked-in core constraint that assumes a data model downstream plans may not follow is itself hard to walk back.

**Docs (your question, Greg) — two homes:**
- a docblock right on the field in `FieldHooks::entityBaseFieldInfo`, noting it's intentionally unconstrained and that downstream modules own data separation;
- prose with the SHOULD-language in `docs/model/type/plan.md` + `plan_record.md` (alongside where the asset `farm` reference is documented), with a cross-cutting note in `docs/model/convention/`.

We already have that SHOULD-language sketched from the onboarding design work — happy to open a follow-up PR against the docs with a first draft so it rides alongside this.

— Greg / soil-health team (via Claude)

mstenta:
We looked at this briefly on the dev call and @symbioquine had a good idea: instead of always adding the `farm` field to plans as a "base field", maybe we should add a mechanism that let's downstream modules "opt-in" to adding it as a "bundle field" to their bundles specifically. And as part of that, in the API surface documentation for that mechanism, we can make it clear that it is the downstream module's responsibility to enforce farm relationship separation (or not).

Issue #875, opened 2024-09-18
plan_record API access not working


ktohalloran:
**Describe the bug**
I receive 403 errors when attempting to create new PlanRecords via the API, even for Plans which I have update access to.

**To Reproduce**
1. Try to POST to create a plan_record for a plan of any type

**Expected behavior**
I expect a 201 successful creation response, but instead receive a 403 error.

**Desktop (please complete the following information):**
 - Windows 11 + WSL2
 - Chrome, but I also see this issue in Postman
 - Chrome version 128.0.6613.138, farmOS version 3.2.3

**Additional context**
Here's the body of the request that's failing (this specific body is for a custom plan and plan record type but I get this error for plans and plan records of any type; the plan id is for an active plan that I can successfully PATCH to):
```
{
    "data": {
        "type": "plan_record--sample_site",
        "attributes": {
            "site_id": "LARA.24.RP.01",
            "indicators": [
                "SOC,Soil organic carbon (SOC)",
                "BD,Bulk density (BD)"
            ]
        },
        "relationships": {
            "plan": {
                "data": {
                    "type": "plan--ag_c_project",
                    "id": "cdda83eb-a759-42ed-b8db-bb773f27eab4"
                }
            },
            "site": {
                "data": {
                    "type": "asset--land",
                    "id": "48108a58-008d-4927-bfc3-25fcb27fea12"
                }
            }
        }
    }
}
```

paul121:
I've started some investigation into this... WIP in the last commit https://github.com/paul121/farmOS/commit/1375cfca335375434c6a07abafb8b7f5cb8a1916 on my 3.x-plan-record-access branch: https://github.com/paul121/farmOS/tree/3.x-plan-record-access (this includes all of 3.3.x changes so not making a PR yet)

I think we can solve this by implementing the `checkCreateAccess()` and `checkFieldAccess()` methods of `PlanRecordAccess`. The combination of these two checks will 1) first always allow creation access but 2) restrict creation/updating of `plan_record` entities by requiring that the user has `update` access to the `plan` entity the record is associated with. Some initial testing via the API using curl and simple oauth I'm able to demonstrate that the `checkFieldAccess()` method blocks creation of the entire entity if access is denied for any reason.

As a side note, I think we should make the `plan` record required? Not sure why this isn't already the case, but maybe there is something we thought of?

mstenta:
[...]

PR #1094, opened 2026-06-24
Require plan records to reference a plan

mstenta:
Currently, `plan_record` entities have a `plan` reference field for linking them to a `plan`. This field is currently optional, which was not intentional. It should have been required, and we describe it as such in the data model docs:

> Every Plan Record must specify a single [Plan](https://farmos.org/model/type/plan) that it is associated with.

https://farmos.org/model/type/plan_record/

This PR makes it required.

Note: this is probably also a prerequisite for fixing: https://github.com/farmOS/farmOS/issues/875

See @paul121's comment about this specifically:

> As a side note, I think we should make the plan record required? Not sure why this isn't already the case, but maybe there is something we thought of?

https://github.com/farmOS/farmOS/issues/875#issuecomment-2362416235

And my response:

> I don't recall any reason that it was not required. I agree that it should be.

https://github.com/farmOS/farmOS/issues/875#issuecomment-2363671763

One could argue that making it required in 4.x is a breaking change, but I think we should do it. I'm not aware of any modules that are using this field optionally, and the intention for these entities is that they are always tied to a plan.

### DIFF
--- a/modules/core/plan/src/Entity/PlanRecord.php
+++ b/modules/core/plan/src/Entity/PlanRecord.php
@@ -76,6 +76,7 @@ class PlanRecord extends ContentEntityBase implements PlanRecordInterface {
       ->setDescription(t('Associate this plan record relationship with a plan entity.'))
       ->setTranslatable(FALSE)
       ->setCardinality(1)
+      ->setRequired(TRUE)
       ->setSetting('target_type', 'plan')
       ->setDisplayOptions('form', [
         'type' => 'entity_reference',


modules/core/plan/src/Entity/PlanRecord.php
```diff
@@ -76,6 +76,7 @@ public static function baseFieldDefinitions(EntityTypeInterface $entity_type) {
       ->setDescription(t('Associate this plan record relationship with a plan entity.'))
       ->setTranslatable(FALSE)
       ->setCardinality(1)
+      ->setRequired(TRUE)
```
symbioquine, line 79: Is this enforced at the DB constraint level? Do we need update hooks for this change?

mstenta: No it is enforced at the Entity/Field API level and is only checked when `$entity->validate()` is run.

symbioquine (approved):

paul121 (approved):

PR #1098, opened 2026-08-18
Pin rector/rector to 2.5.9

mstenta:
Something changed in `rector/rector` 2.6+ that breaks `palantirnet/drupal-rector`, which causes our automated tests to fail with:

`[ERROR] Could not detect twig set.`

I opened an upstream issue: https://git.drupalcode.org/project/rector/-/work_items/3600967

This PR simply pins farmOS to `rector/rector` `2.5.9` until that is fixed.

### DIFF
--- a/composer.project.json
+++ b/composer.project.json
@@ -18,6 +18,7 @@
         "phpstan/phpstan-deprecation-rules": "^2.0",
         "phpunit/phpunit": "^11.5.3",
         "phpspec/prophecy-phpunit": "^2",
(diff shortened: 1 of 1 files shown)


mstenta:
I am not including a CHANGELOG.md entry for this because it is very internal.

I will open a draft PR to remove this pin when the upstream issue is resolved so we don't forget about it.

mstenta:
Follow-up: https://github.com/farmOS/farmOS/pull/1099

Issue #865, opened 2024-08-05
SW MAPS export to KML and subsequent imports has an issue. 


Middyeasy:
**Describe the bug**
When you attempt import a KML file that is directly exported from SW MAPS a well know free and hardware supported GNSS Android survey tool by any number of most major hardware makers you will get an error importing that is very to easy to fix so if there is a way to address this issue then you can recommend that people can use this tool to get data into FarmOS.

**To Reproduce**
copy this to notepad and save as say test.kml
<?xml version="1.0" encoding="UTF-8"?>
<kml xmlns="http://www.opengis.net/kml/2.2">
<Document>
<name>rtk setup</name>
<open>0</open>
<description>Exported using SW Maps</description>
<Folder>
<name>Layer2</name>
<visibility>1</visibility>
<open>0</open>
<Style id="Layer2-PointStyle">
<IconStyle>
<color>FF1DA75F</color>
<scale>0.5</scale>
<Icon>
<href>http://maps.google.com/mapfiles/kml/shapes/donut.png</href>
</Icon>
<hotSpot x="0.5" y="0.5" xunits="fraction" yunits="fraction"/>
</IconStyle>
<LabelStyle>
<scale>0</scale>
</LabelStyle>
</Style>
<Style id="Layer2-Style">
<LineStyle>
<color>FF1DA75F</color>
<width>3</width>
</LineStyle>
<PolyStyle>
<color>501DA75F</color>
</PolyStyle>
</Style>
<Placemark>
<name>Layer2 2</name>
<description> <![CDATA[<h4>Layer: Layer2</h4><h4>Feature ID: 2</h4><p>Remarks: </p><p>Fix Quality: DGPS</p>]]></description>
<styleUrl>#Layer2-PointStyle</styleUrl>
<Point>
<coordinates>-97.64455901341667,30.530107166625,179.72090000000003</coordinates>
</Point>
</Placemark>
</Folder>
</Document>
</kml>

note the failure. 

Copy this in where the following lines are removed after <placemark> and it imports perfectly.

<name>Layer2 2</name>
<description> <![CDATA[<h4>Layer: Layer2</h4><h4>Feature ID: 2</h4><p>Remarks: </p><p>Fix Quality: DGPS</p>]]></description>
<styleUrl>#Layer2-PointStyle</styleUrl>

<?xml version="1.0" encoding="UTF-8"?>
<kml xmlns="http://www.opengis.net/kml/2.2">
<Document>
<name>rtk setup</name>
[...]

PR #1100, opened 2026-08-29
Fix KML import crash on placemarks containing CDATA

AbhinavMir:
The bulk KML importer crashes when a placemark's name or description is wrapped in a CDATA section.

## Why

`KmlNormalizer::denormalize()` reads placemark properties straight out of the array that `simplexml_load_string()` produced. For a plain text element that value is a string. For an element containing a CDATA section, or an element with no text at all, PHP hands back a `SimpleXMLElement` object instead.

That object is then stored in form state. `SimpleXMLElement` cannot be serialized, so the import fails.

## How

Cast the value to a string when it is a `SimpleXMLElement`. CDATA text and empty elements then arrive as ordinary strings, which is what every consumer already expects. Plain text elements are unaffected.

## Tests

Adds `KmlTest`, a kernel test covering four cases: plain text, CDATA in the name, CDATA in the description, and an empty element. Each asserts the resulting property is a string and survives serialization.

### DIFF
--- a/modules/core/kml/tests/src/Kernel/KmlTest.php
+++ b/modules/core/kml/tests/src/Kernel/KmlTest.php
@@ -110,8 +110,7 @@ class KmlTest extends KernelTestBase {
       $this->assertEquals($placemark['name'], $geometries[0]->properties['name'], $label);
       $this->assertEquals($placemark['description'], $geometries[0]->properties['description'], $label);
 
-      // Confirm that the properties can be serialized. The KML importer form
-      // puts them into form state, which Drupal serializes.
+      // Confirm that the properties can be serialized.
       // @see \Drupal\farm_import_kml\Form\KmlImporter
       $serialized = PhpSerialize::encode($geometries[0]->properties);
       $this->assertEquals($geometries[0]->properties, PhpSerialize::decode($serialized), $label);
--- a/modules/core/kml/tests/src/Kernel/KmlTest.php
+++ b/modules/core/kml/tests/src/Kernel/KmlTest.php
@@ -87,6 +87,26 @@ class KmlTest extends KernelTestBase {
         'name' => 'Empty description',
         'description' => '',
       ],
+      // An element only keeps the text that is directly inside it. Text that
+      // is inside child elements is dropped. This is a known limitation. All
+      // supported properties are plain scalar values in practice.
+      'nested child elements' => [
+        'kml' => '<name>Nested</name><description>Hello <b>world</b> and more</description>',
+        'name' => 'Nested',
+        // The text of the <b> element is dropped. Two spaces remain.
+        'description' => 'Hello  and more',
+      ],
+      'nested child element only' => [
+        'kml' => '<name>Nested only</name><description><b>world</b></description>',
+        'name' => 'Nested only',
+        // The element holds no direct text, so the result is an empty string.
+        'description' => '',
+      ],
+      'nested child elements in name' => [
+        'kml' => '<name>Field <b>A</b> north</name><description>Nested name</description>',
+        'name' => 'Field  north',
+        'description' => 'Nested name',
+      ],
     ];
 
     foreach ($placemarks as $label => $placemark) {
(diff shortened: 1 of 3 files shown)


mstenta:
Thank you for the PR @AbhinavMir. Can you please fill in the PR description with some more detail? I see that this aims to fix #865, but I'd like to document a bit more of the *how* and *why* it works to fix that. That will make it easier to review, and provide context to refer back to in the future.

Thanks again!

AbhinavMir:
Hi @mstenta - flipped it for review by mistake! Was still working on it

mstenta:
@AbhinavMir Ah gotcha! I'll flip it to a draft. Thanks for investigating!

For what it's worth, I have wondered about unifying the two different places we handle KML parsing.

I noted in #865 that it only affects the bulk KML importer, which is a more custom implementation. IIRC GeoPHP is used in other contexts and does not have this issue.

So I wonder if a more holistic refactor is a better long-term strategy than playing whack-a-mole with bugs in our custom KML parsing.

That said, your PR looks pretty simple, and it includes test coverage that might be nice to have either way. I'll review in more detail when you're ready. :+1:

mstenta:
Thanks @AbhinavMir! We spent a little bit of time reviewing this today on the weekly dev call. We ran out of time so didn't finish completely, but here is a quick summary so we can pick up again later...

One thing we looked into was how the `(string)` casting works in `SimpleXML` objects, and found this (emphasis mine):

> Returns text content that is directly in this element. _Does not return text content that is inside this element's children._

Source: https://www.php.net/manual/en/simplexmlelement.tostring.php

Currently, the `KmlNormalizer` class only handles properties defined in its `supportedProperties()` method (`['name', 'entity_type', 'bundle', 'internal_id', 'description']`), and those *should* always be plain scalar values in practice, so this makes sense as-is.

If we ever add a "supported property" that includes child elements (non-scalar), then all of that will be lost when it is cast to a string. This is only a hypothetical, and I can't think of any cases where we would have to worry about it right now. So IMO it doesn't block this from being merged.

I have a few small nitpicks that I will leave comments for inline...

Also, worth noting, regarding my previous comment:

> I noted in https://github.com/farmOS/farmOS/issues/865 that it only affects the bulk KML importer, which is a more custom implementation. IIRC GeoPHP is used in other contexts and does not have this issue.
> 
> So I wonder if a more holistic refactor is a better long-term strategy than playing whack-a-mole with bugs in our custom KML parsing.

We talked about this in the dev call, and upon reviewing we were reminded why we did it this way: the GeoPHP library itself only cares about the geometries in the KML, and drops everything else. We need to extract name, description, etc from the `Placemarks`, so we need to do our own XML parsing for that.

modules/core/kml/tests/src/Kernel/KmlTest.php
```diff
@@ -0,0 +1,121 @@
...
+
+      // Confirm that the name and description are strings with the expected
+      // values. CDATA content must be preserved, not dropped.
+      $this->assertIsString($geometries[0]->properties['name'], $label);
+      $this->assertIsString($geometries[0]->properties['description'], $label);
+      $this->assertEquals($placemark['name'], $geometries[0]->properties['name'], $label);
+      $this->assertEquals($placemark['description'], $geometries[0]->properties['description'], $label);
+
+      // Confirm that the properties can be serialized. The KML importer form
+      // puts them into form state, which Drupal serializes.
```
mstenta: This module is lower-level and has no knowledge of the `farm_import_kml` module, so let's simplify this comment by removing the second sentence.

mstenta (changes requested):
> Returns text content that is directly in this element. _Does not return text content that is inside this element's children._

@symbioquine suggested that we might want to add a test that confirms this works when there are nested child elements within the element that is being cast to a string.

mstenta (approved):

symbioquine (approved):
LGTM!

mstenta:
Merged - thanks again @AbhinavMir!

PR #1101, opened 2026-09-10
Allow referencing multiple genetic fathers in birth quick form

mstenta:
Context: https://farmos.discourse.group/t/cattle-admin-workflow/2598/6

Note: this PR contains two commits. The first moves/combines the two existing identical copies of `loadEntityAutocompleteAsset()` from the `Group` and `Movement` quick forms up to the `QuickFormElementsTrait`, which both use. We need this same method in the `Birth` quick form for processing the multiple-value "Genetic father" field input. The second enables multiple fathers to be selected.


symbioquine (approved):
Seems reasonable

PR #1103, opened 2026-09-21
Support building/testing 4.*.x branches

mstenta:
Per discussion w/ @symbioquine and @paul121 in chat, we are going to start moving ahead with merging pull requests that add new features directly into `4.x` (treating it like `main`), rather than creating a `4.1.x` branch. If we need to tag another `4.0.x` release before `4.1.0` is tagged (eg: for security updates or CI fixes), we can create commits on `4.x` first and then create a `4.0.x` branch to backport them to.

In order to support this, our `deliver.yml` GitHub Actions workflow and our `build-farmOS.sh` script need to be able to handle `4.*.x` branches. Currently they only support the "`4.x`" branch explicitly, and tags (eg: `4.0.0`).


mstenta:
Test failures seem to be due to a transient issue with a `curl` command returning a `504` trying to download a ZIP file from GitHub - so it seems to be due to a GitHub issue, not an issue with this change.

symbioquine (approved):
LGTM!

mstenta:
Re-ran tests... they are passing on this PR branch now (`4.x-4.x.y`) which proves that the changes to `build-farmOS.sh` still work on `4.x-*` branches.

I also pushed a `4.0.x` branch to my fork and tests passed there as well, indicating that both changes are working as expected: https://github.com/mstenta/farmOS/actions/runs/35613876198

PR #1104, opened 2026-09-23
Fix JSON:API entity filter access

mstenta:
The recent update to [Entity API 1.8](https://www.drupal.org/project/entity/releases/8.x-1.8) (in farmOS 4.0.5) broke `?filter` functionality in farmOS's JSON:API endpoints.

Originally reported by @agnesfa: https://farmos.discourse.group/t/json-api-collection-filters-return-empty-for-non-administrator-users-since-farmos-4-0-5-entity-api-1-8/2608

> Since farmOS 4.0.5, every `filter[...]` on `asset`, `log` and `quantity` collections through JSON:API returns HTTP 200 with an empty `data` array for an OAuth2 user holding the Farm Manager role, whatever the filter says. Unfiltered reads, `sort` and `page[offset]` work; an unknown field still returns a 400. Taxonomy term and file collections filter normally. Reproduced on 2026-09-22 against [our instance] (4.0.5) with an exact-name filter on a land asset that an unfiltered read returns one request earlier, and with a not-equal filter that should match every row.
> 
> Cause, as far as we can trace it: farmOS 4.0.5 updates the Entity API module to 1.8 (SA-CONTRIB-2026-113). Its commit c07395aa (“fix: Update JSONAPI filtering”) removes the branch in `entity_jsonapi_entity_filter_access()` that granted `JSONAPI_FILTER_AMONG_ALL` when the query-access handler returned no conditions. Assets, logs and quantities use Entity API’s query access handler; for a manager the conditions are empty (the account has `view any asset`), so the hook now returns nothing, core’s generic rule then requires the entity type’s admin permission (`administer assets`, `administer log`, `administer quantity`), the managed roles do not hold it, and `TemporaryQueryGuard` appends `alwaysFalse()`.
> 
> Effect: any integration that filters assets or logs through JSON:API with a manager, worker or viewer token stops seeing data, silently. Our MCP server, and any farmOS client using `filter[name]` to resolve an asset before a write, is affected.
> 
> Possible fixes upstream: implement `hook_jsonapi_entity_filter_access()` for `asset`, `log` and `quantity` in farmOS, granting `JSONAPI_FILTER_AMONG_ALL` to accounts with the `view any <type>`permission (mirroring what Entity API 1.7 did), or add the admin permissions to the managed roles’ “view all” setting.

This was hinted at in the [Entity API 1.8 release notes](https://www.drupal.org/project/entity/releases/8.x-1.8):

> This release changes JSONAPI entity collection access control. In certain circumstances this may mean that JSONAPI clients can access less data than expected. Sites should carefully review their JSONAPI integrations and implement their own json api filter access rules if necessary. See examples in jsonapi module in core on how to do so for your own entity types: [https://git.drupalcode.org/project/drupal/-/blob/main/core/modules/jsona...](https://git.drupalcode.org/project/drupal/-/blob/main/core/modules/jsonapi/src/Hook/JsonapiHooks.php?ref_type=heads)

farmOS tests did not catch this, so the first step is to add tests to reproduce it.


mstenta:
First commit adds very simple tests that reproduce the issue. Right now they only test asset and log entities, but other entities are also affected. Working on a quick fix, and then we can refine...

mstenta:
Tests failed as expected.

```
There were 3 failures:

1) Drupal\Tests\farm_role\Kernel\FarmApiOauthRoleTest::testJsonApiEntityFilterAccess
Failed asserting that an array is not empty.

/opt/drupal/web/profiles/farm/modules/core/api/tests/src/Kernel/FarmApiTest.php:317

2) Drupal\Tests\farm_api\Kernel\FarmApiTest::testJsonApiEntityFilterAccess
Failed asserting that an array is not empty.

/opt/drupal/web/profiles/farm/modules/core/api/tests/src/Kernel/FarmApiTest.php:317

3) Drupal\Tests\farm_l10n\Kernel\FarmLocalizationApiTest::testJsonApiEntityFilterAccess
Failed asserting that an array is not empty.

/opt/drupal/web/profiles/farm/modules/core/api/tests/src/Kernel/FarmApiTest.php:317
```

Now pushing a commit that fixes the issue (for assets and logs only). The tests should pass.

mstenta:
Tests passed (need to clean up linting a bit)...

Now to expand this to other entity types...

mstenta:
Force-pushed to fix linting and include a `CHANGELOG.md` line, so tests should be all green.

I also added a restriction so that filtering can only be done by authenticated users.

mstenta:
Added support and testing for other farmOS entity types.

mstenta:
I split `plan_record` support out to its own branch, because it has some additional considerations. This PR will not add support for `plan_record` filtering in the API. We can tackle that in a follow-up. I'll open a second draft PR with the commit that I removed from this one, once this one is merged.

